PR #1, opened 2015-10-26
Fix direct modex logic when local rank information is requested.

artpol84:
@rhc54 
This PR address the following problem seen in OMPI integration:
**Configuration**: 2 processes on the same node.
**Changes in OMPI**: changing the OPAL_MODEX macro in opal/mca/pmix/pmix.h to use the fence_nb call with a “0” for collect_data.
**What is supposed to happen** is that the PMIx server library sees that these are both local procs, and so it “holds” the request inside itself until the requested data arrives. Then it delivers the data to the requesting process.
**What is happening is** that the PMIx server library immediately fires a direct_modex request to the host. Somehow, the “lcd” portion of the request is getting trashed, and chaos ensues.

### DIFF
--- a/src/server/pmix_server_ops.c
+++ b/src/server/pmix_server_ops.c
@@ -123,6 +123,7 @@ pmix_status_t pmix_pending_request(pmix_nspace_t *nptr, int rank,
     pmix_rank_info_t *iptr;
     pmix_hash_table_t *ht;
     pmix_status_t rc;
+    bool is_local = false;
 
     /* 1. Try to satisfy the request right now */
 
@@ -132,6 +133,7 @@ pmix_status_t pmix_pending_request(pmix_nspace_t *nptr, int rank,
         if (iptr->rank == rank) {
             /* in case it is known local rank - check local table */
             ht = &nptr->server->mylocal;
+            is_local = true;
             break;
         }
     }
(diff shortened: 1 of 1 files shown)


rhc54:
laugh - yes, I found that same one during my code revamp. I'm working on the rest now, and hope to push something later this evening.

jladd-mlnx:
@annu13 just FYI, if you attach to the running processes with gdb (after inserting a delay) and step through the code, it's much easier to understand the flow. We found this in a couple of minutes using this tried and true technique.

jladd-mlnx:
@rhc54 "revamp"? I would suggest we stabilize what we have before we start running down the refactoring rabbit hole.

jladd-mlnx:
A more productive use of our precious little time would be to do the following:

1.  @annu13 and @rhc54 to run as many combinations and permutations as possible and to report failures by opening issues and providing all of the relevant info to reproduce: command line, number of processes, process per node, btl, etc.  

2. If we identify issues in the OMPI integration, then we'll be happy to hand those over to you @rhc54. 

We really only care about the SLURM plugin which, obviously, necessitates a functional standalone client/server.

jladd-mlnx:
@rhc54 given that time is of the essence here, I would ask that you not commit anything directly to master. Please follow standard operating protocol and open a pull request so that @artpol84 and I may review your code.

rhc54:
Fixed in #3

PR #4, opened 2015-10-30
No real need to generate the pmix_common.h header, so simplify by mov…

rhc54:
…ing the version strings to config.h. Add the requested info keys to server_init for passing uid and grpid. Still need to test to see if that is going to cause problems for systems where it isn't allowed


rhc54:
@artpol84 This contains the requested change to server_init, but I'll need to check it on OMPI to ensure it doesn't cause problems.

rhc54:
will move to proper location

PR #5, opened 2015-10-30
No real need to generate the pmix_common.h header, so simplify by mov…

rhc54:
…ing the version strings to config.h. Add the requested info keys to server_init for passing uid and grpid. Still need to test to see if that is going to cause problems for systems where it isn't allowed

Move the chmod and chown calls to their proper order so the rendezvous file will have been created.


rhc54:
@artpol84 As I said, contains the uid/gid changes you requested.

artpol84:
@rhc54 thank you! Will check asap

PR #10, opened 2015-11-19
Fix double-free error in pmix_server_commit.

artpol84:
This error was noticed when running with
```export OMPI_MCA_opal_pmix_base_async_modex=1```
In this case it is possible that dmdx requests will arrive before the data is available. This means that the following codepath was enabled:
https://github.com/pmix/releases/blob/v1.1/src/server/pmix_server_ops.c#L187:L213

The error occurs because data was free()'d twice, here:
https://github.com/pmix/releases/blob/v1.1/src/server/pmix_server_ops.c#L200
and here:
https://github.com/pmix/releases/blob/v1.1/src/server/pmix_server_ops.c#L207.


rhc54:
Nice, much cleaner -- thanks for tracking it down!

:+1:

jladd-mlnx:
Nice job, Tema!

Issue #13, opened 2015-11-24
fi_{rma,atomic} in pmix_constants.7


PHHargrove:
The current man page for pmix_constants.7 contains descriptions of two OFI constants. :wink:

rhc54:
(laugh) oops. I'll take a gander - thanks!

Issue #12, opened 2015-11-24
PMIx_Init(NULL) supported?


PHHargrove:
The man page for pmix_init.3 is clear that a `NULL` argument is legal:
```
Passing a parameter of NULL for this parameter is allowed if
the user wishes solely to initialize the PMIx system and does not
require return of the identifier at that time.
```

However, PMIx_Init() in src/client/pmix_client.c has the following lines as its first executable statements:
```
    if (NULL == proc) {
        return PMIX_ERR_BAD_PARAM;
    }
```

A simple test with the pmix-1.1.1 tarball confirms a return of -19 for `PMIx_Init(NULL)`

rhc54:
Hmmm....obviously, something got out-of-phase. Thx!

Issue #11, opened 2015-11-23
Missing prototype for pmix_value_load()


PHHargrove:
This report is against pmix-1.1.1 from the tarball posted on the Open MPI site.

The macro PMIX_INFO_LOAD() in pmix_common.h expands to calls to pmix_value_load(), for which there is no prototype.  This results in compiler warning about the implicit declaration.

PR #9, opened 2015-11-10
Put pmix usock file chown to the right place (after usock was created).

artpol84:

PR #16, opened 2015-11-25
Fix a few places where thread safety wasn't provided

rhc54:


rhc54:
@artpol84 Do you have time to give this a once-over? I'd like to see if this solves some of the things @ggouaillardet is seeing on the OMPI master.

@ggouaillardet if you want to try it over there, I'd consider that an adequate proof of it being okay.

src/client/pmix_client.c
```diff
@@ -570,56 +561,81 @@ pmix_status_t PMIx_Put(pmix_scope_t scope, const char key[], pmix_value_t *val)
...
+
+    if (pmix_globals.init_cntr <= 0) {
+        return PMIX_ERR_INIT;
+    }
+
+    /* create a callback object */
+    cb = PMIX_NEW(pmix_cb_t);
+    cb->active = true;
+    cb->scope = scope;
+    cb->key = (char*)key;
```
artpol84: You need to strdup here. I'm gettin segfaults at cb with following backtrace:
```
(gdb) bt
#0  0x00007ffff770aef2 in opal_memory_ptmalloc2_int_free (av=0x7ffff79672e0, mem=0x7ffff7a2fbac) at malloc.c:4387
#1  0x00007ffff770970b in opal_memory_ptmalloc2_free (mem=0x7ffff7a2fbac) at malloc.c:3513
#2  0x00007ffff7708445 in opal_memory_linux_free_hook (__ptr=0x7ffff7a2fbac, caller=0x7ffff76ab398) at hooks.c:708
#3  0x00007ffff76ab398 in opal_free (addr=0x7ffff7a2fbac, file=0x7ffff4fd964c "src/usock/usock.c", line=284) at malloc.c:184
#4  0x00007ffff4fb4d58 in cbdes (p=0xabb8a0) at src/usock/usock.c:284
#5  0x00007ffff4fb6939 in pmix_obj_run_destructors (object=0xabb8a0) at ./src/class/pmix_object.h:449
#6  0x00007ffff4fb902f in OPAL_PMIX_PMIX1XX_PMIx_Put (scope=PMIX_GLOBAL, key=0x7ffff7a2fbac "pmix.cpuset", val=0x7fffffffc230)
    at src/client/pmix_client.c:620
#7  0x00007ffff51ec50e in pmix1_put (scope=OPAL_PMIX_GLOBAL, val=0x7fffffffca70) at pmix1_client.c:395
#8  0x00007ffff79cf064 in orte_ess_base_proc_binding () at base/ess_base_fns.c:297
#9  0x00007ffff63f0241 in rte_init () at ess_pmi_module.c:344
#10 0x00007ffff7982406 in orte_init (pargc=0x0, pargv=0x0, flags=32) at runtime/orte_init.c:234
#11 0x00007ffff7cd0044 in ompi_mpi_init (argc=1, argv=0x7fffffffd0c8, requested=0, provided=0x7fffffffcf48) at runtime/ompi_mpi_init.c:488
#12 0x00007ffff7d069d5 in PMPI_Init (argc=0x7fffffffcfac, argv=0x7fffffffcfa0) at pinit.c:66
#13 0x000000000040086e in main (argc=1, argv=0x7fffffffd0c8) at ring_c.c:19
(gdb) frame 6
#6  0x00007ffff4fb902f in OPAL_PMIX_PMIX1XX_PMIx_Put (scope=PMIX_GLOBAL, key=0x7ffff7a2fbac "pmix.cpuset", val=0x7fffffffc230)
    at src/client/pmix_client.c:620
620         PMIX_RELEASE(cb);
(gdb) frame 4
#4  0x00007ffff4fb4d58 in cbdes (p=0xabb8a0) at src/usock/usock.c:284
284             free(p->key);
(gdb) p p->key
$1 = 0x7ffff7a2fbac "pmix.cpuset"
(gdb) frame 6
#6  0x00007ffff4fb902f in OPAL_PMIX_PMIX1XX_PMIx_Put (scope=PMIX_GLOBAL, key=0x7ffff7a2fbac "pmix.cpuset", val=0x7fffffffc230)
    at src/client/pmix_client.c:620
```

src/client/pmix_client.c
```diff
@@ -532,37 +532,28 @@ int PMIx_Abort(int flag, const char msg[],
...
-                        key, val->type);
-
-    if (pmix_globals.init_cntr <= 0) {
-        return PMIX_ERR_INIT;
-    }
-
     /* setup to xfer the data */
     kv = PMIX_NEW(pmix_kval_t);
-    kv->key = strdup((char*)key);
+    kv->key = strdup(cb->key);
```
artpol84: And to avoid extra strdup I'd suggest to add:
`cb->key = NULL` here

artpol84:
With mentioned fix it one-run test is OK. I'm starting multiple-run test to see if any hangs/errors remains.

ggouaillardet:
i am still seeing some hangs on ompi master (with the latest pmix master) and this patch and the strdup fix

artpol84:
What is your testing suite?

artpol84:
@rhc54 I see hangs too. With mpirun launch and "sm" BTL instead of vader.
```
(gdb) bt
#0  0x00000037694ac9fd in nanosleep () from /lib64/libc.so.6
#1  0x00000037694e1b54 in usleep () from /lib64/libc.so.6
#2  0x00007ffff65b48fc in OPAL_PMIX_PMIX1XX_PMIx_Fence (procs=0x0, nprocs=0, info=0x0, ninfo=0) at src/client/pmix_client_fence.c:100
#3  0x00007ffff67e4f1c in pmix1_fence (procs=0x0, collect_data=0) at pmix1_client.c:306
#4  0x00007ffff7cd0766 in ompi_mpi_init (argc=1, argv=0x7fffffffd058, requested=0, provided=0x7fffffffced8) at runtime/ompi_mpi_init.c:812
#5  0x00007ffff7d069d5 in PMPI_Init (argc=0x7fffffffcf3c, argv=0x7fffffffcf30) at pinit.c:66
#6  0x000000000040086e in main (argc=1, argv=0x7fffffffd058) at ring_c.c:19
```

ggouaillardet:
```mpirun  --oversubscribe -np 4 --mca orte_startup_timeout 10000 --mca coll ^ml --mca btl tcp,self --mca mpi_add_procs_cutoff 0 --mca pmix_base_async_modex 1 -host n0,n1 --mca pmix_base_verbose 19 ./intercomm/allgather_inter```

from the ```ibm/collective``` test suite

two nodes are required, and the test must be ran within a loop, and once in a while, it hangs in pmix_get("btl.tcp")

note you are much more likely of hanging with ```--mca pmix_base_verbose 0```

```
[gilles@motomachi collective]$ pstack 18793
Thread 3 (Thread 0x7f21ed861700 (LWP 18795)):
#0  0x0000003c61ce5d03 in epoll_wait () from /lib64/libc.so.6
#1  0x00007f21ee5909da in epoll_dispatch () from /home/RIST/gilles/local/ompi-master/lib/libopen-pal.so.0
#2  0x00007f21ee5945a6 in opal_libevent2022_event_base_loop () from /home/RIST/gilles/local/ompi-master/lib/libopen-pal.so.0
#3  0x00007f21ed88375d in progress_engine () from /home/RIST/gilles/local/ompi-master/lib/libpmix.so.0
#4  0x0000003c620077f1 in start_thread () from /lib64/libpthread.so.0
#5  0x0000003c61ce570d in clone () from /lib64/libc.so.6
Thread 2 (Thread 0x7f21ece60700 (LWP 18797)):
#0  0x0000003c61cdc053 in poll () from /lib64/libc.so.6
#1  0x00007f21ee59fb91 in poll_dispatch () from /home/RIST/gilles/local/ompi-master/lib/libopen-pal.so.0
#2  0x00007f21ee5945a6 in opal_libevent2022_event_base_loop () from /home/RIST/gilles/local/ompi-master/lib/libopen-pal.so.0
#3  0x00007f21ee53ab8c in progress_engine () from /home/RIST/gilles/local/ompi-master/lib/libopen-pal.so.0
#4  0x0000003c620077f1 in start_thread () from /lib64/libpthread.so.0
#5  0x0000003c61ce570d in clone () from /lib64/libc.so.6
Thread 1 (Thread 0x7f21ee2ec700 (LWP 18793)):
#0  0x0000003c61caab9d in nanosleep () from /lib64/libc.so.6
#1  0x0000003c61cdeb64 in usleep () from /lib64/libc.so.6
#2  0x00007f21ed890c36 in OPAL_PMIX_PMIX111_PMIx_Get () from /home/RIST/gilles/local/ompi-master/lib/libpmix.so.0
#3  0x00007f21edad1792 in pmix1_get () from /home/RIST/gilles/local/ompi-master/lib/openmpi/mca_pmix_pmix111.so
#4  0x00007f21eadb19ae in mca_btl_tcp_proc_create () from /home/RIST/gilles/local/ompi-master/lib/openmpi/mca_btl_tcp.so
#5  0x00007f21eada9e69 in mca_btl_tcp_add_procs () from /home/RIST/gilles/local/ompi-master/lib/openmpi/mca_btl_tcp.so
#6  0x00007f21eb23f652 in mca_bml_r2_add_proc () from /home/RIST/gilles/local/ompi-master/lib/openmpi/mca_bml_r2.so
#7  0x00007f21ea777091 in mca_bml_base_get_endpoint () from /home/RIST/gilles/local/ompi-master/lib/openmpi/mca_pml_ob1.so
#8  0x00007f21ea7780ed in mca_pml_ob1_isend () from /home/RIST/gilles/local/ompi-master/lib/openmpi/mca_pml_ob1.so
#9  0x00007f21eec2f5a4 in ompi_coll_base_sendrecv_nonzero_actual () from /home/RIST/gilles/local/ompi-master/lib/libmpi.so.0
#10 0x00007f21eec2caa1 in ompi_coll_base_sendrecv () from /home/RIST/gilles/local/ompi-master/lib/libmpi.so.0
#11 0x00007f21eec2d2ef in ompi_coll_base_allgather_intra_recursivedoubling () from /home/RIST/gilles/local/ompi-master/lib/libmpi.so.0
#12 0x00007f21e9b01b4e in ompi_coll_tuned_allgather_intra_dec_fixed () from /home/RIST/gilles/local/ompi-master/lib/openmpi/mca_coll_tuned.so
#13 0x00007f21eeb7eceb in ompi_comm_split () from /home/RIST/gilles/local/ompi-master/lib/libmpi.so.0
#14 0x00007f21eebd16cb in PMPI_Comm_split () from /home/RIST/gilles/local/ompi-master/lib/libmpi.so.0
```

**Comment by artpol84** [comment]
Ok, so you intensively use direct modex capabilities.

**Comment by artpol84** [comment]
Update: I've done some advanced analysis of the hange'd job. Using pdsh I've collected pstack output for all 128x28 processes. Then I filtered "Thread 1" backtraces that looks like this:
```
cnXXX: Thread 1 (Thread 0x7ffff740b700 (LWP 20107)):                                                                                          
cnXXX: #0  0x0000003caecac9fd in nanosleep () from /lib64/libc.so.6                                                                           
cnXXX: #1  0x0000003caece1b54 in usleep () from /lib64/libc.so.6                                                                              
cnXXX: #2  0x00007ffff65b48fc in OPAL_PMIX_PMIX1XX_PMIx_Fence () from /tmp/artemp-ompi/lib/libpmix.so.0                                       
cnXXX: #3  0x00007ffff67e4f1c in pmix1_fence () from /tmp/artemp-ompi/lib/openmpi/mca_pmix_pmix1xx.so                                         
cnXXX: #4  0x00007ffff7cd03e7 in ompi_mpi_init () from /tmp/artemp-ompi/lib/libmpi.so.0                                                       
cnXXX: #5  0x00007ffff7d069d5 in PMPI_Init () from /tmp/artemp-ompi/lib/libmpi.so.0
```

Then I split traces into different files by node-name prefix. And afterwards I've counted number of lines containing `OPAL_PMIX_PMIX1XX_PMIx_Fence` in each of them. The result shows that **all** processes entered Fence and the problem most probably sits on the server side. 

I will check orted's next time I'll hit this problem.

**Comment by artpol84** [comment]
Update #2:
Here is the data from the orted (from node #1):
```
(gdb) p *(pmix_server_trkr_t*)pmix_server_globals.collectives.pmix_list_sentinel.pmix_list_next
$3 = {
super = {...},
type = PMIX_FENCENB_CMD, pcs = 0x7fffec216bf0, npcs = 1, active = true, def_complete = true, ranks = {
      super = {...}, pmix_list_sentinel = {super = {...}, 
          pmix_list_next = 0x7fffec0174d0, pmix_list_prev = 0x7fffec217660, item_free = 0}, 
          pmix_list_length = 28
      }, 
local_cbs = {
    super = {...},
    pmix_list_sentinel = {super = {...}, 
        pmix_list_next = 0x7fffec017410, pmix_list_prev = 0x7fffec1185b0, item_free = 0
    }, 
    pmix_list_length = 28
},
nlocal = 28, local_cnt = 0, 
info = 0x7fffec2168d0, ninfo = 1, 
collect_type = PMIX_COLLECT_YES, 
modexcbfunc = 0x7ffff3aa350c <modex_cbfunc>, op_cbfunc = 0
}
```
PMIx lib part of the fence seems done: `local_cbs.pmix_list_length == 28 == nlocal`

**Comment by artpol84** [comment]
Using the same technique as for MPI processes I checked that **all** daemons and mpirun have `local_cbs.pmix_list_length == 28` so I the problem resides somewhere in exchange logic.

**Comment by ggouaillardet** [comment]
Here is a gist with the logs of a hang
https://gist.github.com/ggouaillardet/6c6b371c228b194df4d3
Task 3 hangs in pmix_get
I noticed tasks 2 and 3 issued requests at the same time to orted of n1 for Info about tasks 0 and 1.
Then orted 1 asked mpirun about tasks 0 and then 1
Task 2 got a reply but not task 3
First naive question ...
Is task 2 supposed to receive info for task 0 or 0-1 ? Aka all tasks on node 0 ?
If yes, is orted n1 supposed to ask again info about 0-1 for task 3 ?
I will resume debugging tomorrow if needed

**Comment by artpol84** [comment]
2015-11-26 16:29 GMT+06:00 Gilles Gouaillardet <notifications@github.com>:

> Here is a gist with the logs of a hang
> https://gist.github.com/ggouaillardet/6c6b371c228b194df4d3
> Task 3 hangs in pmix_get
> I noticed tasks 2 and 3 issued requests at the same time to orted of n1
> for Info about tasks 0 and 1.
> Then orted 1 asked mpirun about tasks 0 and then 1
> Task 2 got a reply but not task 3
> First naive question ...
> Is task 2 supposed to receive info for task 0 or 0-1 ? Aka all tasks on
> node 0 ?
>
There will be 2 separate requests to n1 for task 0 and task 1.

> If yes, is orted n1 supposed to ask again info about 0-1 for task 3 ?
>
if task 3 will request data after response for task 2 will be received
orted n1 will cache data for tasks 0 and 1 and will respond to task 3
locally.
If both requests will appear before response is received they will be
queued in the list that will be subsequently scanned and all requests
waitin for the certain task will be satisfied.

> I will resume debugging tomorrow if needed
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pmix/master/pull/16#issuecomment-159870972>.
>



-- 
С Уважением, Поляков Артем Юрьевич
Best regards, Artem Y. Polyakov

**Comment by artpol84** [comment]
@ggouaillardet One question: you use 2 nodes and 4 tasks and enable oversubscrition. Does that mean that your nodes have one core only?

**Comment by ggouaillardet** [comment]
This is just a stupid copy/paste from mtt
Both nodes have several cores

**Comment by artpol84** [comment]
BTW if I run this test with more than 7 processes it segfaults. Do you have the same behavior? Is it expected? 
And am I right that if all is correct empty output is expected.
Here is what I get from credump:
```
(gdb) bt
#0  0x00007f08286985d0 in ?? ()
#1  <signal handler called>
#2  0x00007f082d99452a in orte_util_convert_process_name_to_string (name_string=0x7fffaa2c71d8, name=0x2bfa8004b) at util/name_fns.c:451
#3  0x00007f082d97d10b in _convert_process_name_to_string (name_string=0x7fffaa2c71d8, name=0x2bfa8004b) at runtime/orte_init.c:97
#4  0x00007f082dc9eca2 in ompi_dpm_connect_accept (comm=0x19c2800, root=0, port_string=0x0, send_first=true, newcomm=0x7fffaa2c74a8) at dpm/dpm.c:170
#5  0x00007f082dce17f9 in PMPI_Comm_connect (port_name=0x0, info=0x6035e0, root=0, comm=0x19c2800, newcomm=0x7fffaa2c7508) at pcomm_connect.c:113
#6  0x00000000004013b2 in main (argc=1, argv=0x7fffaa2c7638) at allgather_inter.c:71
(gdb) frame 3
#3  0x00007f082d97d10b in _convert_process_name_to_string (name_string=0x7fffaa2c71d8, name=0x2bfa8004b) at runtime/orte_init.c:97
97          return orte_util_convert_process_name_to_string(name_string, name);
(gdb) p *name
Cannot access memory at address 0x2bfa8004b
#4  0x00007f082dc9eca2 in ompi_dpm_connect_accept (comm=0x19c2800, root=0, port_string=0x0, send_first=true, newcomm=0x7fffaa2c74a8) at dpm/dpm.c:170
170                 rc = opal_convert_process_name_to_string(&nstring, &(proc_list[i]->super.proc_name));
(gdb) p *proc_list[0]
Cannot access memory at address 0x2bfa80003
```
Looks like proc_list gets corrupted. I'll try to see what happens.

**Comment by artpol84** [comment]
I see the same SEGFAULT with both PMIx plugin and mpirun. The backtrace that I've previously provided is a consequence not a cause. The root cause of error occurs in MPI_Comm_split, here:
```
(gdb) bt
#0  ompi_proc_lookup (proc_name=...) at proc/proc.c:198
#1  0x00007f10264d6dab in ompi_group_dense_lookup_raw (group=0x1b742b0, peer_id=6) at group/group_plist.c:64
#2  0x00007f10264d6f10 in ompi_group_get_proc_ptr_raw (group=0x1b742b0, rank=6) at group/group_plist.c:88
#3  0x00007f10264d707f in ompi_group_incl_plist (group=0x1b742b0, n=4, ranks=0x1b929e0, new_group=0x1ba24f8) at group/group_plist.c:120
#4  0x00007f10264d480b in ompi_group_incl (group=0x1b742b0, n=4, ranks=0x1b929e0, new_group=0x1ba24f8) at group/group.c:213
#5  0x00007f10264b72d3 in ompi_comm_set_nb (ncomm=0x7fff643f4e88, oldcomm=0x603960, local_size=4, local_ranks=0x1b929e0, remote_size=0, 
    remote_ranks=0x0, attr=0x0, errh=0x7f10267fd920, copy_topocomponent=false, local_group=0x0, remote_group=0x0, req=0x7fff643f4e00)
    at communicator/comm.c:167
#6  0x00007f10264b7191 in ompi_comm_set (ncomm=0x7fff643f4e88, oldcomm=0x603960, local_size=4, local_ranks=0x1b929e0, remote_size=0, 
    remote_ranks=0x0, attr=0x0, errh=0x7f10267fd920, copy_topocomponent=false, local_group=0x0, remote_group=0x0) at communicator/comm.c:114
#7  0x00007f10264b838b in ompi_comm_split (comm=0x603960, color=1, key=0, newcomm=0x7fff643f4fb0, pass_on_topo=false) at communicator/comm.c:585
#8  0x00007f102650b9a3 in PMPI_Comm_split (comm=0x603960, color=1, key=0, newcomm=0x7fff643f4fb0) at pcomm_split.c:69
#9  0x0000000000401255 in main (argc=1, argv=0x7fff643f50d8) at allgather_inter.c:41
(gdb) l 192
187     {
188         ompi_proc_t *proc = NULL;
189         int ret;
190
191         /* try to lookup the value in the hash table */
192         ret = opal_hash_table_get_value_ptr (&ompi_proc_hash, &proc_name, sizeof (proc_name), (void **) &proc);
193
194         if (OPAL_SUCCESS == ret) {
195             return &proc->super;
196         }
(gdb) p ret
$10 = -13
(gdb) n
199     }
(gdb) 
ompi_group_dense_lookup_raw (group=0x1b742b0, peer_id=6) at group/group_plist.c:65
65              if (NULL != proc) {
(gdb) 
73          return group->grp_proc_pointers[peer_id];
```

If `ompi_proc_lookup` is unable to find the process that appears to be a sentinel it will fallback to the regular return, see here:
https://github.com/open-mpi/ompi/blob/master/ompi/group/group_plist.c#L65
In our case `ompi_proc_lookup` is unable to find data for rank = 6 and returns NULL, however `ompi_group_dense_lookup_raw` treats this as OK and falls back to `group->grp_proc_pointers[peer_id]`:
https://github.com/open-mpi/ompi/blob/master/ompi/group/group_plist.c#L73
 However this array is not completely filled:
```
(gdb) p *group->grp_proc_pointers[0]
$11 = {super = {super = {super = {obj_magic_id = 16046253926196952813, obj_class = 0x7f10267ed320, obj_reference_count = 6, 
        cls_init_file_name = 0x7f1025f3655b "class/opal_list.c", cls_init_lineno = 254}, opal_list_next = 0x1b842d0, 
      opal_list_prev = 0x7f102680d8a8, item_free = 1, opal_list_item_refcount = 1, opal_list_item_belong_to = 0x7f102680d880}, proc_name = {
      jobid = 1364787201, vpid = 0}, proc_arch = 4291367424, proc_flags = 4095, proc_convertor = 0x1af95d0, proc_hostname = 0x1af9870 "sputnik1"}, 
  proc_endpoints = {0x1b80540, 0x0}, padding = '\000' <repeats 15 times>}
(gdb) p *group->grp_proc_pointers[1]
$12 = {super = {super = {super = {obj_magic_id = 16046253926196952813, obj_class = 0x7f10267ed320, obj_reference_count = 4, 
        cls_init_file_name = 0x7f10265a5782 "proc/proc.c", cls_init_lineno = 106}, opal_list_next = 0x1b84730, opal_list_prev = 0x1b31580, 
      item_free = 1, opal_list_item_refcount = 1, opal_list_item_belong_to = 0x7f102680d880}, proc_name = {jobid = 1364787201, vpid = 1}, 
    proc_arch = 4291367424, proc_flags = 32768, proc_convertor = 0x1af95d0, proc_hostname = 0x1b848c0 "sputnik2"}, proc_endpoints = {0x1b848e0, 
    0x0}, padding = '\000' <repeats 15 times>}
(gdb) p *group->grp_proc_pointers[2]
$13 = {super = {super = {super = {obj_magic_id = 16046253926196952813, obj_class = 0x7f10267ed320, obj_reference_count = 4, 
        cls_init_file_name = 0x7f10265a5782 "proc/proc.c", cls_init_lineno = 106}, opal_list_next = 0x7f102680d8a8, opal_list_prev = 0x1b84730, 
      item_free = 1, opal_list_item_refcount = 1, opal_list_item_belong_to = 0x7f102680d880}, proc_name = {jobid = 1364787201, vpid = 2}, 
    proc_arch = 4291367424, proc_flags = 32768, proc_convertor = 0x1af95d0, proc_hostname = 0x1b91ee0 "sputnik3"}, proc_endpoints = {0x1ba2150, 
    0x0}, padding = "\000\000\000\000\000\000\000\000\220\000\000\000\000\000\000"}
(gdb) p *group->grp_proc_pointers[3]
Cannot access memory at address 0x6a2b20003
(gdb) p *group->grp_proc_pointers[4]
$14 = {super = {super = {super = {obj_magic_id = 16046253926196952813, obj_class = 0x7f10267ed320, obj_reference_count = 4, 
        cls_init_file_name = 0x7f10265a5782 "proc/proc.c", cls_init_lineno = 106}, opal_list_next = 0x1b920a0, opal_list_prev = 0x1b842d0, 
      item_free = 1, opal_list_item_refcount = 1, opal_list_item_belong_to = 0x7f102680d880}, proc_name = {jobid = 1364787201, vpid = 4}, 
    proc_arch = 4291367424, proc_flags = 32768, proc_convertor = 0x1af95d0, proc_hostname = 0x1b84860 "sputnik5"}, proc_endpoints = {0x1ba22b0, 
    0x0}, padding = "\025\000\000\000\000\000\000\000\300y\252\001\000\000\000"}
(gdb) p *group->grp_proc_pointers[5]
Cannot access memory at address 0xaa2b20003
(gdb) p *group->grp_proc_pointers[6]
Cannot access memory at address 0xca2b20003
(gdb) p *group->grp_proc_pointers[7]
Cannot access memory at address 0xea2b20003
```
I think that `grp_proc_pointers` is incomplete because we use direct modex. If I remove cutoff=0 and async modex - the array is correct. 
I think that this may be related to problems that Gilles sees.

**Comment by artpol84** [comment]
@ggouaillardet I tried to run your example 500 times without any hangs. How many times do you usually need to reproduce?

**Comment by ggouaillardet** [comment]
@artpol84 when i run this in a loop, it usually takes less than one minute with ```--mca pmix_base_verbose 19``` and up to 5 minutes with ```--mca pmix_base_verbose 0```

note i am running the latest ompi master, in which i rebased open-mpi/ompi#1114 and i back ported all recent commits from pmix master, plus this PR and your strdup fix

stock ompi has some hangs (and maybe some crashes too) but since latest pmix is the way we will go anyway, i chose to debug the latest pmix.

with 4 procs and the latest pmix, i do not see any crash.
note the hang occurs at the very beginning of the MPI app (e.g. the first MPI_Comm_split) which occurs before MPI_{connect,accept}

**Comment by rhc54** [comment]
You might try again with this updated PR just for baseline - I fixed the strdup and a couple of other places. I see the problem in the direct modex code - there are a couple of places where we don't thread shift and need to do so. I'll add those fixes here later tonight.

**Comment by ggouaillardet** [comment]
@rhc54 with this update, i got some crashes
for now, i fixed that in ompi
```patch
diff --git a/opal/mca/pmix/pmix.h b/opal/mca/pmix/pmix.h
index 64c7895..ff1ac91 100644
--- a/opal/mca/pmix/pmix.h
+++ b/opal/mca/pmix/pmix.h
@@ -110,7 +110,6 @@ extern int opal_pmix_base_exchange(opal_value_t *info,
         char *_key;                                             \
         _key = mca_base_component_to_string((s));               \
         OPAL_MODEX_SEND_STRING((r), (sc), _key, (d), (sz));     \
-        free(_key);                                             \
     } while(0);
 
 /**
```

ggouaillardet:
i did a bit more of debugging
so far, the hang seems to always occur when data is on its way (e.g. the return statement is executed)

from ```dmodex_req``` from ```orte/orted/pmix/pmix_server_fence.c```

```c
    /* has anyone already requested data for this job from this node? If so,
     * then the data is already on its way */
    for (rnum=0; rnum < orte_pmix_server_globals.reqs.num_rooms; rnum++) {
        opal_hotel_knock(&orte_pmix_server_globals.reqs, rnum, (void**)&r);
        if (NULL == r) {
            continue;
        }
        if (r->target.jobid == req->target.jobid &&
            r->proxy.vpid == dmn->name.vpid) {
            /* we have this request, so just wait for the return */
                return;
        }
    }
```

could this be an ompi problem and not a pmix one ?

rhc54:
Certainly could be on the upstream server side - I need to check the threads up there to ensure things get shifted correctly

artpol84:
@rhc54 Can you point to the places in PMIx direct modex where thread
shifting is needed?

пятница, 27 ноября 2015 г. пользователь rhc54 написал:

> Certainly could be on the upstream server side - I need to check the
> threads up there to ensure things get shifted correctly
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pmix/master/pull/16#issuecomment-160042994>.
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

rhc54:
I pushed an updated branch to this PR that contains the rest of the client-side changes that were required. I took a quick look on the orte side, and it looked okay thread-wise, so I'm now digging in a bit to see why we think we already have a request when we don't. Probably have to look deeper tomorrow as it's getting a little late here

rhc54:
I believe I see one problem, though it may not be the cause here. If you look at the dmodex_req code in orte/orted/pmix/pmix_server_fence.c, you'll see that we never set the proxy.vpid field in the request, and the constructor for that object doesn't initialize that field. So I'm guessing that the proxy is probably defaulting to zero, which is okay for this test case but certainly not in general.

rhc54:
Okay, I'm zeroing in on the problem. Apparently, the pmix server library is passing up a "wildcard" rank to the ORTE server. This causes the ORTE server to return "not found", and things go downhill from there.

I'll dig deeper to find the cause of the problem tomorrow. I vaguely remember something about an optimization where we would ask for data from -all- clients on the target daemon as we assumed that if a proc asked for one, it or someone else on the node would ask for other from that location. The ORTE server isn't expecting that, so if we did it, that's the source of the problem.

artpol84:
пятница, 27 ноября 2015 г. пользователь rhc54 написал:

> Okay, I'm zeroing in on the problem. Apparently, the pmix server library
> is passing up a "wildcard" rank to the ORTE server. This causes the ORTE
> server to return "not found", and things go downhill from there.
>
> I'll dig deeper to find the cause of the problem tomorrow. I vaguely
> remember something about an optimization where we would ask for data from
> -all- clients on the target daemon as we assumed that if a proc asked for
> one, it or someone else on the node would ask for other from that location.
> The ORTE server isn't expecting that, so if we did it, that's the source of
> the problem.
>
According to my observations, at least plugin gets separate direct modex
requests for each task.

> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pmix/master/pull/16#issuecomment-160047894>.
>


-- 
-----
Best regards, Artem Polyakov
(Mobile mail)

artpol84:
@ggouaillardet sorry for asking again, but what do you see if you run with 8 processes and 8 nodes?

artpol84:
@rhc54 @ggouaillardet 
I've run one more experiment: LU test from NPB 3.x; with enabled direct modex-related debug in PMIx plugin. I was using 8 nodes and 32 processes. After I filtered slurmd.log's I realized that not all node daemons was queried with dmdx requests. Moreover, I saw the following picture:
```
$ ls -l
-rw-r--r-- 1 user group 5670 Ноя 27 12:14 slurmd.log.cn1
-rw-r--r-- 1 user group 1884 Ноя 27 12:15 slurmd.log.cn2
-rw-r--r-- 1 user group 1888 Ноя 27 12:15 slurmd.log.cn3
-rw-r--r-- 1 user group    0 Ноя 27 12:12 slurmd.log.cn4
-rw-r--r-- 1 user group 1898 Ноя 27 12:16 slurmd.log.cn5
-rw-r--r-- 1 user group    0 Ноя 27 12:12 slurmd.log.cn6
-rw-r--r-- 1 user group    0 Ноя 27 12:12 slurmd.log.cn7
-rw-r--r-- 1 user group    0 Ноя 27 12:12 slurmd.log.cn8
```
So only nodes 1,2,3 and 5 was communicating. If you will compare this with gdb debug that I posted before with `group->grp_proc_pointers[x]` values you'll see that this strongly correlates.

So my strong believe is that for some reasons OMPI incorrectly sets `group->grp_proc_pointers[x]`. 
Also @ggouaillardet  if I remove cutoff=0 option leaving async=1 enabled I see that it works.
Can you verify the following scenarios:
1. Non of cutoff and async options are enabled
2. only async option is enabled.
Will your hangs go away then?

src/client/pmix_client.c
artpol84: I think that for better readability and safety strdup has to be done here. And in `_putfn` you don't do strdup and just copy already strdup'ed pointer and set `cb->key` to `NULL`

ggouaillardet:
i will test that on monday.
with an even number of tasks, the program should success without any error.
just to be clear, I am not stuck whatsoever.
I saw Cisco mtt now uses async modex and cutoff, and there are some hangs/failure, so I tried to debug them.

artpol84:
Here is the simple reproducer for my NPB hang:
```C
#include <stdio.h>                                                                                                                                    
#include <mpi.h>                                                                                                                                      
                                                                                                                                                     
{                                                                                                                                                     
    int val = 10;                                                                                                                                     
    int rank, color, size;                                                                                                                            
    MPI_Comm comm;                                                                                                                                    
    MPI_Init(&argc, &argv);                                                                                                                           
    MPI_Comm_rank(MPI_COMM_WORLD,&rank);                                                                                                              
    MPI_Comm_size(MPI_COMM_WORLD,&size);                                                                                                              
                                                                                                                                 
    MPI_Bcast(&val, 1, MPI_INT, 0, MPI_COMM_WORLD);                                                                                                   
    if( rank == 0 ){                                                                                                                                  
        printf("Bcas#1 - OK\n");                                                                                                                      
    }                                                                                                                                                 
    MPI_Finalize();                                                                                                                                   
    return 0;                                                                                                                                         
}
```
If I take 2 nodes and play with ppn here is what I get:

ppn  | result
------------- | -------------
1  | OK
2  | OK
4  | OK
5  | **hang**

artpol84:
**Update:** Checking plugin dmdx logs I see that node2 didn't request data about rank=4, information about all other ranks was exchanged.

ggouaillardet:
@artpol84 does the hang only occurs with async modex and cutoff=0 ?

artpol84:
@ggouaillardet yes.

artpol84:
@ggouaillardet @rhc54 And I'm using openib. will try TCP later.

artpol84:
The hang appears with TCP as well

artpol84:
Not true: TCP works, it was forgotten debug statement.

artpol84:
And LU test works well if I switch to tcp BTL on filly packed 8-node allocation (128 processes). Once I switch to openib it hangs.
So this problem differs from the one than Gilles observes.

artpol84:
Comparison of dmdx requests for working tcp btl and hanging openib btl shows that they are identical. In both cases data for rank=4 wasn't requested.

src/client/pmix_client_get.c
```diff
@@ -149,172 +134,18 @@ pmix_status_t PMIx_Get_nb(const pmix_proc_t *proc, const char *key,
...
-
-    /* Now we need to thread shift to the service
-     * thread for request processing */
+    /* thread-shift so we can check global objects */
     cb = PMIX_NEW(pmix_cb_t);
-    (void)strncpy(cb->nspace, nm, PMIX_MAX_NSLEN);
-    cb->rank = proc->rank;
-    cb->key = strdup(key);
+    cb->active = true;
+    cb->procs = (pmix_proc_t*)proc;
```
artpol84: @rhc54 I'm seing failures here. This is what @ggouaillardet fixed at OMPI site. In pmix_cb_t we have both nspace/rank fields and procs field.
Here I'd suggest either allocate and copy proc data or use nspace/rank.

src/client/pmix_client_get.c
```diff
@@ -497,64 +335,192 @@ static void getnb_cbfunc(struct pmix_peer_t *pr, pmix_usock_hdr_t *hdr,
...
+            PMIX_VALUE_RELEASE(val);
+        }
+        /* protect the data */
+        cb->procs = NULL;
+        cb->key = NULL;
+        cb->info = NULL;
+        PMIX_RELEASE(cb);
+        return;
+    }
+    if (PMIX_RANK_WILDCARD == cb->procs->rank) {
```
artpol84: @rhc54 here cb->proc->rank is used (and we don't actually know if proc still exists).

src/client/pmix_client_get.c
```diff
@@ -497,64 +335,192 @@ static void getnb_cbfunc(struct pmix_peer_t *pr, pmix_usock_hdr_t *hdr,
...
-            goto exit;
+            pmix_list_append(&pmix_client_globals.pending_requests, &cb->super);
+            return;
         }
     }
 
     /* we don't have a pending request, so let's create one - don't worry
      * about packing the key as we return everything from that proc */
-    msg = pack_get(data->cb->nspace, data->cb->rank, data->info, data->ninfo, PMIX_GETNB_CMD);
+    msg = pack_get(nm, cb->rank, cb->info, cb->ninfo, PMIX_GETNB_CMD);
```
artpol84: @rhc54  and here cb->rank is used, which we haven't initialized because was relying on cb->procs.

src/usock/usock.c
```diff
@@ -280,6 +285,15 @@ static void cbdes(pmix_cb_t *p)
     if (NULL != p->key) {
         free(p->key);
     }
+    if (NULL != p->nsp) {
+        free(p->nsp);
+    }
+    if (NULL != p->procs) {
```
artpol84: @rhc54 What if procs was a pointer to a local variable?

src/client/pmix_client.c
artpol84: @rhc54 Still no strdup here!

rhc54: @artpol84 I have no issue with doing extra strdup's - I didn't include it because you had later flagged lines for having strdup's that weren't needed because we could NULL the variable prior to release. At some point, we need to decide which way we are going to go and be consistent.

Adding strdup's is indeed safer and protects against any future case where the data isn't statically passed to us. There is a performance penalty to all these tiny malloc's since we don't have a memory manager in the code like OMPI does, but it probably isn't significant - modern Linux implementations have gotten fairly efficient at malloc.

The only issue would be that we definitely don't want to copy all those info arrays and proc arrays! So we will be treating the string values different from any other passed value. This is a little troubling.

Other option is to remove the free's from the pmix_cb_t destructor, and require that people take care of it directly in the code. It might result in leaks, but those can be stomped out. Frankly, I'm thinking this is the correct path.

Thoughts?

src/usock/usock.c
rhc54: @artpol84 See my other response

src/client/pmix_client_get.c
rhc54: @artpol84 Sure, we can copy the data across - it's a minor performance hit, but probably not noticeable. However, you made one incorrect statement here - the caller is absolutely required to maintain all passed data structures until Get_nb executes the provided callback function. This is clearly stipulated in the documentation.

src/client/pmix_client_get.c
rhc54: @artpol84 Again, yes we do - however, I have no issue with copying the data.

src/client/pmix_client_get.c
rhc54: @artpol84 fair enough - addressing as per above.

src/client/pmix_client.c
artpol84: @rhc54 the current version segfaults. So we need to solve it somehow. 
My suggestion is:
1. strdup key in PMIx_Put.
2. in _putfn pass it further to the kv->key data structure:
```
kv->key = cb->key;
cb->key = NULL;  // protect the data
```
you will need to `NULL` `cb->key` because `pmix_kval_t` destructor also free()'s the key if it is not zero. So we **must** have key malloc'ed.

this was my original suggestion. Sorry if it wasn't clear.

src/client/pmix_client_get.c
artpol84: @rhc54 Ok, I agree about demand to keep proc until Get_nb finishes. 
If we want to rely on user we probably need to provide him with the macro to create proper proc as a "blackbox" and take care about it's free'ing ourself.
like this:
```
pmix_proc_t *proc = PMIX_CREATE_PROC(nspace, rank);
PMIx_Get_nb(proc, ....);
```
and he forgot's about this proc.
But still this approach will rise doubts and intentions to free this pointer. Copying would make things clear from user perspective. AND user will be able to use local variables for proc. I personally would consider proc as input data and (without going to the documentation) would assume that I can do what I want after Get_nb locally finished.

src/client/pmix_client.c
rhc54: @artpol84 Okay, so let's try following the general policy of not having pmix_cb_t free anything. The strdup then moves to the kv->key line, which is probably where it belongs. We then don't need to add all these lines where we set cb fields to NULL prior to release.

I'll update accordingly.

src/client/pmix_client_get.c
rhc54: @artpol84 That would definitely be contrary to standard practice for non-blocking functions. I've never seen any library that copied all input data on a non-blocking call - what we are doing here is the normal and expected.

I'm not sure we really want to protect against user errors - this could get really cumbersome if we do.

artpol84:
Here is the summary for the problem I was working on today:
* **tcp** BTL works well with both **NPB.LU** and the **reproducer** I've posted previously in all combinations of **cutoff**/**async** values.
* **openib** BTL hangs if the **cutoff** parameter is set to 0, modex can be both **full** or **async**.
* the hang with **openib** appears with **mpirun**, **pmix** AND! **pmi2** which makes me think that this is not PMIx-related problem since PMI2 plugin goes through the s2 component and doesn't involve **libpmix**.
* the hang of **openib** occurs not always, see the table I've provided earlier.
* for some reasons setting `OMPI_MCA_btl_openib_if_include` helps to increase ppn and number of nodes, i.e. if I set this variable I can run my reproducer on **2x16**, **4x16** configurations, but on **5x16** and **8x16** it hangs.
* for all cases when **reproducer** works fine **NPB.LU** works fine as well.
* I made some analysis for the reproducer and without fixed interface on 2x4 configuration. Bcast algorithm has the following structure:
```
rank 0 sends to 1, 2, 4, 8;
rank 1 sends to 3, 5, 9
...
```
When hang occurs I see that all ranks except for 1, 5 and 9 reached MPI_Finalize(). Ranks 1, 5 and 9 are forever-waiting for completion of their send (1) or receive (5, 9) requests.

rhc54:
@artpol84 Thanks! I believe that should move over to the OMPI issues list as it relates to the on-demand add_procs code, and not PMIx. Do you want to enter it there, or would you prefer I do so?

artpol84:
@rhc54 I'll do

rhc54:
This has now been thoroughly tested in the OMPI repo, and all review comments have been addressed, so I'm going to bring it in here.

src/client/pmix_client_get.c
```diff
@@ -149,172 +134,19 @@ pmix_status_t PMIx_Get_nb(const pmix_proc_t *proc, const char *key,
...
-    /* Now we need to thread shift to the service
-     * thread for request processing */
+    /* thread-shift so we can check global objects */
     cb = PMIX_NEW(pmix_cb_t);
-    (void)strncpy(cb->nspace, nm, PMIX_MAX_NSLEN);
+    cb->active = true;
+    (void)strncpy(cb->nspace, proc->nspace, PMIX_MAX_NSLEN);
     cb->rank = proc->rank;
-    cb->key = strdup(key);
+    cb->key = (char*)key;
```
artpol84: @rhc54 This particular case differs from all other `key` usages. I would suggest to strdup the `key` here and free it in `_getnbfn`.
In `PMIx_Put` we are explicitly waiting until corresponding callback in the thread shifting finishes. So we can be sure that the key that was passed by user won't used anymore.
In this case user have no way to know whether  `_getnbfn` was finished or not. And unlike other functions we need to force user to keep the allocated key until Get_nb finished which is not confinient in my understanding.

rhc54: @artpol84 I'm afraid I don't understand your comment about the user not knowing when _getnbfn is done. In every code path, we ensure that their callback function (with cbdata) is executed. If there is an error in _getnbfn, then it occurs immediately from that point. Otherwise, it is executed once the data is returned.

So the user will always know exactly when the key, info array, and all other provided values are no longer required. Why single out the key as opposed to the info array and treat it differently?

artpol84: @rhc54 
Here https://github.com/pmix/master/blob/master/src/client/pmix_client.c#L618 we do explicitly wait for `_putfn` to finish.
While here:
https://github.com/pmix/master/blob/master/src/client/pmix_client_get.c#L147
we do a threadshift and return immidiately.

artpol84: @rhc54 Ok, i see what you are saying. You mean that user has to keep `key` `info` etc. till the callback from the PMIx_Get_nb will be executed.
Ok, that makes sense.

PR #19, opened 2015-12-01
Remove early development assumptions that force debug.

artpol84:
This brakes server-side with client-side communications when:
a. Server-side was compiled separately with "--disable-debug" which
should be the same as default.
b. Client-side is compiled inside OMPI, OMPI assumes that if no
"--enable-debug" was passed to OMPI ./configure than all it needs
is not to specify "--enable-debug" in integrated pmix ./configure.
However in fact it will be implicitly enabled by the code that this
commit removes.
**Upd**: this is related to master in git only.

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -668,12 +668,6 @@ else
     AC_MSG_RESULT([no])
     WANT_DEBUG=0
 fi
-#################### Early development override ####################
-if test "$WANT_DEBUG" = "0" && test -z "$enable_debug" && test "$PMIX_DEVEL" = "1"; then
-    WANT_DEBUG=1
-    echo "--> developer override: enable debugging code by default"
-fi
-#################### Early development override ####################
 if test "$WANT_DEBUG" = "0"; then
     CFLAGS="-DNDEBUG $CFLAGS"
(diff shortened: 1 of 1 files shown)


rhc54:
@artpol84 Yes, OMPI will always default to enabling debug when compiled from a git clone. If the server was compiled against a PMIx library built against a tarball, then the default would be to have debug disabled on that side. This is why we stated that you have to compile your app against the same library used by the server.

There is a PR under development for OMPI to allow use of an external PMIx library for just this reason. After all, there is no reason to believe that the RM is using the same PMIx version as is embedded in OMPI.

Leaving the default git clone build to be with debug enabled has saved our hide on more than one occasion, and so I'd prefer to leave it. However, I don't think it would be a big deal to make the buffer operations work even in the case of a mismatch, and I can see where that might be useful.

So would it be okay if we replaced this with a PR to handle buffer type mismatch? If so, I can try to work that up in the next day or two.

artpol84:
@rhc54 
Ok, sounds good. We can help with that. Lets talk about this on the call.

rhc54:
Closing this per telecon agreement that we'll provide the external PMIx component for OMPI.

PR #24, opened 2016-01-07
dist: do not include include/private/autogen/config.h in dist tarball

ggouaillardet:
this file is  automatically generated from include/private/autogen/config.h.in

### DIFF
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -1,5 +1,7 @@
 #
 # Copyright 2015-2016 Intel, Inc. All rights reserved
+# Copyright 2016      Research Organization for Information Science
+#                     and Technology (RIST). All rights reserved.
 #
 # $COPYRIGHT$
 #


ggouaillardet:
@rhc54 could you please review this ?

that solves one issue initially reported at http://www.open-mpi.org/community/lists/users/2016/01/28215.php
sun compiler uses config.h from the tarball (and that is good for gcc compiler) instead of config.h from the build directory (and that has been generated with sun compiler)

ggouaillardet:

since ```top_srcdir``` is before ```top_builddir```, the wrong config.h (e.g. the one created when building the tarball) is used when building from a tarball __and__ when using VPATH.

that should be fine most of the time since the prominent platform is x86 / gcc, but we already got a report on the ompi ml (building v2.x from the tarball with sun compiler)

rhc54:
@ggouaillardet All I see here is a copyright change - did you forget to push something?

ggouaillardet:
i missed you already made the changes in pmix/master@14aa4fde8209b2e6463cd9cabedbab5354177af2

let's forget about the copyright and close this PR

PR #48, opened 2016-02-10
Replace the "required" boolean in the pmix_info_t struct with a flag-based approach

rhc54:
Take 24-bits from the pmix_data_type_t definition for bit-flags to indicate different behaviors for pmix_info_t, leaving 8-bits for integer values of data types. This limits us to 256 data types, but that should be sufficient.

Because we have to deal with endianness, provide macros for setting and getting the data type when flags may be present. Ditto for setting and getting of flags


rhc54:
Closed in favor of plugin-based approach to compatibility

PR #62, opened 2016-03-22
Correct use of do/while loops to consume semicolon.

ashleypittman:
Remove extra semicolon from the end of do/while loops throughout
code, fix whitespace before and after the while keyword.

Fix an example which was not correctly providing a semicolon.

Signed-off-by: Ashley Pittman <ashley.m.pittman@intel.com>


rhc54:
:+1: 

Thanks!!

PR #65, opened 2016-03-28
usock_connect() revamp

ggouaillardet:


ggouaillardet:
then i will bring it into OpenMPI master

rhc54:
The correct fix would be to just change the return value of the function to pmix_status_t - can we do that instead?

ggouaillardet:
strictly speaking, we should change the prototype of this function.
when successful, it returns an int which is the file descriptor of the socket.
on error, it returns a pmix_status_t

I will fix that tomorrow

ggouaillardet:
@rhc54 done !

rhc54:
much better - thanks!

:+1:

PR #70, opened 2016-03-31
Fix C++ usage of PMIx

rhc54:
C++ mandates that we provide a name to each struct - i.e., it doesn't like:

typedef struct {
...
} foo_t;

Instead, it requires:

typedef struct foo {
...
} foo_t;

So make it happy and let C++ apps use PMIx

### DIFF
--- a/include/pmix.h
+++ b/include/pmix.h
@@ -55,7 +55,7 @@
 #include <pmix/pmix_common.h>
 
 
- BEGIN_C_DECLS
+BEGIN_C_DECLS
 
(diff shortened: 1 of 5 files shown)


rhc54:
@artpol84 @igor-ivanov This look okay to you guys? Our internal folks hit it when trying to build a C++ app that calls PMIx functions

jolivier23:
This fixes our issues

rhc54:
Very good - thanks!

PR #71, opened 2016-04-06
OSX tempdir too long for sun_path

hpcraink:
On Mac OSX, the temporary directory can get too long for the socket's sun_path (which is on LINUX 108 chars, on Mac OSX 103 chars), making Open MPI fail with cryptic: "bind() failed"...

If this happens (due to long tmp-dir and machine name), bail to the shortest possible version. Yes, this overrides any previously set environment-variables, but then again, if they're too long, they're wrong anyhow...

Before:
user@MYMACHINE-MacBook-Pro:~/C/MPI> mpirun -np 2 mpi_wtick
myaddress.sun_path:/var/folders/30/c1t3_6m546n62qrckq5fs_wc0000gp/T/openmpi-sessions-502@MYMACHINE-MacBook-Pro_0/48892/0/0/ len(tdir):101 MAX:103

After:
user@MYMACHINE-MacBook-Pro:~/C/MPI> mpirun -np 2 mpi_wtick
myaddress.sun_path:/tmp/pmix-51040 len(tdir):4 MAX:103

### DIFF
--- a/src/server/pmix_server.c
+++ b/src/server/pmix_server.c
@@ -147,6 +147,7 @@ static pmix_status_t initialize_server_base(pmix_server_module_t *module)
 {
     int debug_level;
     char *tdir, *evar;
+    char *pmix_pid;
     pid_t pid;
 
     /* initialize the output system */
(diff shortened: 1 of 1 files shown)


rhc54:
What does OMPI do for its session directory? I should think it has the same problem, yes?

rhc54:
Per my email, I think that failing over to some unexpected location is a tad risky. Instead, why don't we return an error indicating that the path length was too long, and then let OMPI error out with a nice message suggesting the user set their TMPDIR envar to something shorter?

hpcraink:
I agree with Your comment "setting to some unexpected location", even though this is just overriding to the default location, if the "provided" tdir is way to long (and OSX tends to set TMPDIR by default to something 50 characters long).

So, instead of resetting `tmpdir`, just return an error code like `PMIX_ERR_INVALID_LENGTH` and in any user in OMPI, augment the warning about resetting their tmpdir:
error = "Try resetting shorter TMPDIR var. or change your computer's name (see uname -n), since pmix server init";
Works for me.

rhc54:
Replaced by https://github.com/pmix/master/pull/72

@hpcraink Can you please take a look at that PR and ensure I captured everything?

PR #72, opened 2016-04-08
Backport the relevant changes from OMPI

rhc54:
If the tmpdir name is too long, then return an error so the host RM can know we couldn't setup the rendezvous point. Also, minor cleanup of some debug output

@artpol84 Can you please take a look? I think I captured everything from Rainer's original PR

### DIFF
--- a/src/server/pmix_server.c
+++ b/src/server/pmix_server.c
@@ -148,6 +148,7 @@ static pmix_status_t initialize_server_base(pmix_server_module_t *module)
     int debug_level;
     char *tdir, *evar;
     pid_t pid;
+    char * pmix_pid;
 
     /* initialize the output system */
     if (!pmix_output_init()) {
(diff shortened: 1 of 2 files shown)


hpcraink:
Looks good to me

src/server/pmix_server.c
```diff
@@ -223,7 +224,16 @@ static pmix_status_t initialize_server_base(pmix_server_module_t *module)
     /* now set the address - we use the pid here to reduce collisions */
     memset(&myaddress, 0, sizeof(struct sockaddr_un));
     myaddress.sun_family = AF_UNIX;
-    snprintf(myaddress.sun_path, sizeof(myaddress.sun_path)-1, "%s/pmix-%d", tdir, pid);
+    asprintf(&pmix_pid, "pmix-%d", pid);
+    // If the above set temporary directory name plus the pmix-PID string
+    // plus the '/' separator are too long, just fail, so the caller
+    // may provide the user with a proper help... *Cough*, *Cough* OSX...
+    if ((strlen(tdir) + strlen(pmix_pid) + 1) > sizeof(myaddress.sun_path)-1) {
```
artpol84: I think you need +2 here instead of +1:
'\0' and '/'

artpol84: Ok, you have it on the right side. So please ignore the prev comment

artpol84:
:+1:

PR #79, opened 2016-04-29
Fix typo in packing of pmix_app_t - thanks to Gilles for pointing it out

rhc54:
Needs to come across to 1.1.4


rhc54:
Reviewed by @ggouaillardet on OMPI (same patch here)

Issue #83, opened 2016-05-06
pmi_client (PMIv1) test fails and is getting incorrect size from PMI_Get_size()


garlick:
On current master e6747971774dad02f99de99e2783abd7ca8ba65e the PMI1 test appears to be failing.  Note the incorrect size=12 on the INFO line below.

```
$ ./configure --disable-visibility; make
[snip]
$ cd test
$ VERBOSE=3 ./pmix_test -v -n 2 -e ./pmi_client
pmix_test.c:main: Start PMIx_lite smoke test (timeout is 10)
pmix_test.c:main: Executing test: ./pmi_client -n 2 -v
utils.c:launch_clients: Setting job info
cli_stages.c:errhandler_reg_callbk: ERRHANDLER REGISTRATION CALLBACK CALLED WITH STATUS 0, ref=0
INFO  spawned=0 size=12 rank=0 appnum=11
FATAL  at test_item1:175
ERROR [cli_stages.c:252:errhandler]:  PMIX server Error handler with status = -27
cli_stages.c:cli_wait_all: OP CALLBACK CALLED WITH STATUS 0
cli_stages.c:op_callbk: the child with pid = 45294 has rank = 0
		exited = 1, signalled = 0
cli_stages.c:cli_wait_all: the child with pid = 45294 has rank = 0
		exited = 1, signalled = 0
ERROR [cli_stages.c:148:cli_cleanup]: rank 0 with state 1 unexpectedly terminated.
cli_stages.c:cli_terminate: Client rank = 0 terminated
INFO  spawned=0 size=12 rank=1 appnum=11
FATAL  at test_item1:175
ERROR [pmix_test.c:167:main]: Test was aborted!
cli_stages.c:cli_kill_all: Skip rank 0 as it was already terminated.
cli_stages.c:cli_kill_all: Kill rank 1 (pid = 45295).
cli_stages.c:cli_terminate: Client rank = 1 terminated
cli_stages.c:op_callbk: OP CALLBACK CALLED WITH STATUS 0
```

The PMI2 test, by contrast, runs OK:
```
./pmix_test -n 2 -e ./pmi2_client
pmix_test.c:main: Test finished OK!
```

PR #84, opened 2016-05-06
Return sensible errors from PMI1 when RM registers incomplete namespace

garlick:
Several PMI1 functions returned success without setting their parameter when `PMIx_Get()` failed.  Fix the logic in those functions to return the (PMI1-converted) error code of `PMIx_Get()` when it fails.  Also define PMIX_JOB_SIZE and PMIX_APPNUM in the `pmix_test` namespace so the `pmi_client` test no longer fails as reported in #83.

Note: `PMI2_Init()` returns PMIX_UNIV_SIZE for its `size` parameter.  Perhaps `PMI1_Get_size()` should use that instead of PMIX_JOB_SIZE?  I wasn't too sure about that one so I left it as is.


rhc54:
I'll have to think about PMI2_Init - you make a good point, but I need to check definitions.

👍

rhc54:
Thanks

Issue #87, opened 2016-05-10
Building C++ programs fails due to stdint.h


jjhursey:
A little C/C++ program:
```C
#include <stdio.h>
#include "pmix/pmix_common.h"

int main() {
    int rank = PMIX_RANK_WILDCARD;
    return 0;
}
```

Compiles fine with `gcc`, but fails with `g++`
```shell
shell$ gcc cpp_test.c -I$PMIX/include -L$PMIX/lib -lpmix
shell$ g++ cpp_test.c -I$PMIX/include -L$PMIX/lib -lpmix
In file included from cpp_test.c:5:0:
cpp_test.c: In function ‘int main()’:
/u/jjhursey/pmix-install/include/pmix/pmix_common.h:79:29: error: ‘INT32_MAX’ was not declared in this scope
 #define PMIX_RANK_WILDCARD  INT32_MAX-1
                             ^
cpp_test.c:9:16: note: in expansion of macro ‘PMIX_RANK_WILDCARD’
     int rank = PMIX_RANK_WILDCARD;
                ^
```

It looks like there is code to handle this in:
 * [include/pmix/autogen/pmix_config_bottom.h](https://github.com/pmix/master/blob/master/include/pmix/autogen/pmix_config_bottom.h)

But it is not included in `config.h.in`. Doing so results in more compilation problems.

I'll keep investigating.

jjhursey:
This was fixed by PR #88 in the `master` branch. After a soak there for a few days we'll try to get it into the v1.1.4 release branch.

PR #88, opened 2016-05-11
config: Consolidate private and public config.h to resolve C++ stdint.h issue

jjhursey:
Related to Issue #87 

This branch works, but I think the *true* fix will require a discussion about the configure logic and contents of the internal `config.h.in` and the external `config.h.in`.


jjhursey:
@rhc54 Is there a reason why we have two different `config.h.in` files? The one that is installed does not include the `include/pmix/autogen/pmix_config_bottom.h` and `include/pmix/autogen/pmix_config_top.h` as I would expect it to, but the internal one does.

The `include/pmix/autogen/pmix_config_bottom.h` contained the STL fix for `stdint.h`, but when I included it it also needed all of the `HAVE_` macros that only seem to be defined in the internal `config.h.in` in `src/include/private/autogen/config.h.in`.

rhc54:
It's purely historical, and the reason is no longer valid. At one time, we had a config.h in pmix_common to support some config-set values. We have since eliminated those (hooray!).

So I'd suggest we consolidate to one and avoid the problems.

jjhursey:
I should also note that this patch will not always work since `pmix_stdint.h` is not where the installed `pmix_config_bottom.h` expects to find it: `<pmix/autogen/pmix_stdint.h>`

I can work on consolidating the header file as part of this ticket. I would want this to be in PMIx 1.1.4 - hopefully that won't be too bad to sort out later.

jjhursey:
@rhc54 I did the config.h header consolidation. I can confirm that this resolves the C++ issue and builds on the few platforms I have access too. Can you review this PR before we merge it into the master?

Once it is merged into the master branch, it should probably soak there for a few days before pulling it over to the v1.1.4 rc branch. At which point I think it would be good to roll a new rc to get some broader testing.

rhc54:
I think this look okay. IIRC, we were going to get rid of PMIX_DECLSPEC because of confusion it created. Instead, we use PMIX_EXPORT for symbols that should be externally visible, and don't use anything else. Should we clean that up here? Or do it separately?

jjhursey:
I'm fine either way, as long as it comes in as a separate commit. Tomorrow, I can convert all of these over to `PMIX_EXPORT` and remove the `PMIX_DECLSPEC` symbol.

I just did a quick grep and it looks like both are used quite a bit:
```shell
shell$ grep -rnI "PMIX_EXPORT" * | wc -l
99
shell$ grep -rnI "PMIX_DECLSPEC" * | wc -l
103
```

rhc54:
Let's do it separately, then.

👍

Issue #91, opened 2016-05-18
"native" security mode does not check socket creds


garlick:
As noted in #90, the "native" security mode accepts user credentials without validation.

If I understand correctly, all that prevents a user from hijacking another user's pmix session, and thus gaining the ability to run processes as that user via `PMIx_Spawn()`, are the file system permissions on the UNIX domain socket.  Since not all UNIXes enforce those permissions, and since socket bind, chmod, chown operations are not applied together atomically, this is fairly weak security.

An improvement would be to modify the "native" plugin to call `getsockopt(SO_PEERCRED)` to retrieve the `struct ucred` for `peer->sfd` and validate it against the `char *cred` argument.

SO_PEERCRED also has some portability challenges but there are equivalent options on most UNIXes should support be required for non-Linux OS's later on, and a compilation failure is preferable to silently allowing anyone in.

garlick:
Also noted: if the client sends an empty "cred" portion of the initial handshake, the server validation check is bypassed.

Will have PR for both issues shortly.

PR #92, opened 2016-05-18
security: validate connections with SO_PEERCRED + misc fixes

garlick:
This PR changes the way that the "native" security module works, as proposed in #91.  Instead of checking a client-supplied credential against the uid/gid registered by the server, send a NULL credential and validate a `struct ucred` from SO_PEERCRED against the uid/gid registered by the server.

This also closes a vulnerability where a client could bypass a security module's `validate_cred` method by sending a NULL cred.

Finally,  munge is removed from the list of security modules, since only UNIX domain socket connections are supported by PMIx presently, and it adds latency to the synchronous connect handshake of clients without improving security over SO_PEERCRED.

Testing:
* The fact that the "native" module no longer has a `create_cred` method means the client sends an empty cred; so the fact debug logs show its `validate_cred` method being called (where before it wasn't) indicate that NULL cred vulnerability is fixed.
* debug logs show ucred values are as expected, and that validation is successful
* I've run my RM with its PMIx server and PMI1 client (linked against libpmix.so) many times without issue
* Briefly tried `PMIX_DEBUG=5 ./pmix_test -n 16 -e ./pmix_client` and all looks OK.

### DIFF
--- a/include/pmix/pmix_common.h
+++ b/include/pmix/pmix_common.h
@@ -97,6 +97,9 @@ BEGIN_C_DECLS
 #define PMIX_USERID                "pmix.euid"              // (uint32_t) effective user id
 #define PMIX_GRPID                 "pmix.egid"              // (uint32_t) effective group id
 
+/* attributes for the rendezvous socket  */
+#define PMIX_SOCKET_MODE           "pmix.sockmode"          // (uint32_t) POSIX mode_t (9 bits valid)
+
 /* general proc-level attributes */
 #define PMIX_CPUSET                "pmix.cpuset"            // (char*) hwloc bitmap applied to proc upon launch
 #define PMIX_CREDENTIAL            "pmix.cred"              // (char*) security credential assigned to proc
--- a/src/sec/pmix_munge.c
+++ b/src/sec/pmix_munge.c
@@ -118,7 +118,7 @@ static int validate_cred(pmix_peer_t *peer, char *cred)
     munge_err_t rc;
 
     pmix_output_verbose(2, pmix_globals.debug_output,
-                        "sec: munge validate_cred %s", cred);
+                        "sec: munge validate_cred %s", cred ? cred : "NULL");
 
     /* parse the inbound string */
     if (EMUNGE_SUCCESS != (rc = munge_decode(cred, NULL, NULL, NULL, &uid, &gid))) {
--- a/src/sec/pmix_native.c
+++ b/src/sec/pmix_native.c
@@ -12,6 +12,7 @@
 
 #include <pmix/pmix_common.h>
 
+#include "src/include/pmix_socket_errno.h"
 #include "src/include/pmix_globals.h"
 #include "src/util/argv.h"
 #include "src/util/output.h"
@@ -27,14 +28,13 @@
 
 static int native_init(void);
 static void native_finalize(void);
-static char* create_cred(void);
 static pmix_status_t validate_cred(pmix_peer_t *peer, char *cred);
 
 pmix_sec_base_module_t pmix_native_module = {
     "native",
     native_init,
     native_finalize,
-    create_cred,
+    NULL,
     NULL,
     validate_cred,
     NULL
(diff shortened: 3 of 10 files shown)


rhc54:
I'd like to get some broader checks on this - @jjhursey @artpol84  can you also verify this works for you?

garlick:
I should have added to my description that  I'm new to this code base and there is a lot I don't understand, so please review with extra skepticism.

rhc54:
Checked okay on my system

artpol84:
Looks good, will check early next week.

rhc54:
Fixes #91

jjhursey:
I did a simple build test on this, and it was fine. Is there a test case I can run that would exercise this code correctly?

rhc54:
Given the way we execute (i.e., a daemon fork/exec'ing a child of the same user), I don't see how we can do it. Perhaps we could create a test that runs as root and then fork/exec's a child that sets uid to some other user and then tries to connect?

garlick:
Maybe something like
* add option to `pmix_test` to set permissive mode on socket (otherwise open will fail before SO_PEERCRED is tested
* run test prefixed with sudo so client is a different user e.g. `pmix_test -e sudo -u testuser pmix_test`

The test would require setting up a user and sudo but maybe that's not unreasonable in this case?

garlick:
What about adding some new keys for the `PMIx_server_init()` info for controlling the socket creation:
* PMIX_SOCKET_MODE
* PMIX_SOCKET_PATH

This is in addition to the existing keys that control uid:gid of the socket
* PMIX_USERID
* PMIX_GRPID

Then we could add options to pmix_test to set these.

If that's an agreeable approach I could tack a couple of commits to do this onto this PR.  Let me know what you think.

rhc54:
No objections - what would those do?

garlick:
I was thinking the test could grow an option that would set PMIX_SOCKET_MODE to 0666 so that launching the client under sudo could actually fail the SO_PEERCRED check not `open(2)`.
Never mind on PMIX_SOCKET_PATH - probably not needed here.

rhc54:
Sounds fine with me

garlick:
Hang on I've screwed something up in the last push, will update shortly.

garlick:
Ok, sorry about that.  Just forced a push with the following changes
* fixed couple more buffer overrun vulnerabilities in connect-ack parsing (brought over from pr #93 to replace the one case that was previously fixed here)
* added PMIX_SOCKET_MODE attr as discussed
* pmix_server sets socket to mode 666
* provide a script so native auth can be tested

The test is documented at the top of `pmix_client_otheruser.sh`:
```
# To test sec:native we run client as a different user e.g.
#
#  ./pmix_test -e ./pmix_client_otheruser.sh
#
# Prerequisites:
# - ensure directories leading up to source tree are o+x
# - add dummy user and set OTHERUSER to its name
# - give yourself passwordless sudo capability to that user
#
# The test should fail with message similar to
#   PMIX ERROR: INVALID-CREDENTIAL in file 
#       ../src/server/pmix_server_listener.c at line 524
#
```

Note: if you use vpath builds, change path to script to be relative to srcdir.  Also run pmix_client once first so libtool can `.libs/lt-pmix_client`.  E.g.
```
pmix/build/test > ./pmix_client
ERROR [pmix_client.c:86:main]: Client ns (null) rank -1: PMIx_Init failed: -46
pmix/build/test > ./pmix_test -e ../../test/pmix_client_otheruser.sh
[jimbo:53038] PMIX ERROR: INVALID-CREDENTIAL in file ../src/server/pmix_server_listener.c at line 528
[jimbo:53056] PMIX ERROR: UNREACHABLE in file ../src/client/pmix_client.c at line 1044
[jimbo:53056] PMIX ERROR: UNREACHABLE in file ../src/client/pmix_client.c at line 203
ERROR [pmix_client.c:86:main]: Client ns smoky_nspace-0 rank 0: PMIx_Init failed: -27
ERROR [cli_stages.c:148:cli_cleanup]: rank 0 with state 1 unexpectedly terminated.
ERROR [pmix_test.c:172:main]: Test was aborted!
```

rhc54:
I do have a concern here, but perhaps I'm just misreading the code. It seems to me that you are restricting the server to handling a single user - i.e., the PMIx server creates a socket that only a specified user can utilize. Thus, all procs that will connect to that server must belong to that user.

It strikes me that this may be too restrictive. Some systems may well choose to have a single PMIx server supporting connections from multiple users (e.g., someone wanting to connect with a tool that isn't directly spawned by the server). Will these changes preclude that use-case?

garlick:
I could be confused but I thought I was matching the connecting uid,gid
against the ones provided to pmix_ server_register_client()?  I am fuzzy on
how that gets looked up for auth though.  Will have a closer look tomorrow.
On May 23, 2016 9:28 PM, "rhc54" <notifications@github.com> wrote:

I do have a concern here, but perhaps I'm just misreading the code. It
seems to me that you are restricting the server to handling a single user -
i.e., the PMIx server creates a socket that only a specified user can
utilize. Thus, all procs that will connect to that server must belong to
that user.

It strikes me that this may be too restrictive. Some systems may well
choose to have a single PMIx server supporting connections from multiple
users (e.g., someone wanting to connect with a tool that isn't directly
spawned by the server). Will these changes preclude that use-case?

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/pmix/master/pull/92#issuecomment-221163535>

garlick:
It seems like other changes are needed to support that use case.  For example, the RM needs a way to register access control lists to limit which users are allowed in and what they can do.

Maybe the security plugins should grow a new "authorization" method that allows a particular capability (starting with "connect") to be matched against an acl?

Anyway for now it seems like there is a security problem here, given the potential for just any user to get connected and call "spawn" or something.  Maybe we could fix that and open an issue for the other use case?

rhc54:
Remember, PMIx doesn't actually _do_ anything. Thus, the fact that you have connected to a PMIx server grants you no capabilities.

I'd rather not inject ACL responsibilities into PMIx, nor do I see a clear reason for doing so. So perhaps the best solution for the future would be to simply pass a connection request back to the RM and let it tell us if that connection is to be allowed?

Regardless, folks at the telecon had no issue with this constraint for 1.1.4, and we will look further at it for the future. I've asked that they give it a quick review just to ensure we are good.

garlick:
> simply pass a connection request back to the RM and let it tell us if that connection is to be allowed?

That sounds reasonable.  Then the security plugin could just establish the validity of the user's cred rather than authorize the connection.

garlick:
> I could be confused but I thought I was matching the connecting uid,gid
against the ones provided to pmix_ server_register_client()?  I am fuzzy on
how that gets looked up for auth though.  Will have a closer look tomorrow.

Looks like
* client provides (nspace,rank) in init-auth handshake
* server looks up (nspace,rank) in pmix_globals.nspaces to find pmix_rank_info_t which contains (uid,gid)
* server calls `sec->validate_cred()` compares pmix_rank_info_t (uid,gid) against credential (uid,gid)
* connect is denied or allowed depending on result of above

The pmix_globals.nspaces data is populated by `PMIx_server_register_client()`.  So yes we are comparing the (uid,gid) provided to register_client() with those in the cred.

The upshot is that the RM can,  within the current PMIx API, allow more than one user to connect e.g.
* each nspace can have a different (uid,gid)
* each rank within a nspace can have a different (uid,gid)
* but only one (uid,gid) can connect to one (nspace,rank)

Hopefully I read the code right.

rhc54:
Yes, you read it correctly! I think we are okay with things in this PR and can address the multi-user issue in 2.0.

What we would like to do is not hardcode the munge support to "off". Instead, we will make it not build by default - i.e., instead of building if we find the munge header, we will only build the munge support if explicitly told to do so.

I'll add that change after this is committed, and then PR both changes to 1.1.4

garlick:
Thanks!

Issue #90, opened 2016-05-17
consider file descriptor passing from RM to client instead of accept()


garlick:
PMIx currently creates a UNIX domain socket and accepts connections from clients on it.

Is there a reason there shouldn't _also_ be an optional mode where the RM calls `socketpair()` to create both client and server fds in one go, and passes them to PMIx RM interfaces, then ensures the client fd remains open across fork when the client is started?

PMIx would construct `PMIX_SERVER_URI` to contain the client fd number, so that the client can get hold of the server socket.  The server would use the other fd instead of one returned from `accept()`.

This way PMIx doesn't need to manage the socket in the file system namespace, worry about listen queues and acceptor threads, authenticate connections, etc., and thus be a bit more efficient.

rhc54:
I believe this is doable as an optional mode - for simplicity, we could just pass a ```PMIX_SERVER_FD``` in this case so the URI isn't confusing. The only limitation you should be aware of is that a client process won't be able to fork/exec and have its child connect to the PMIx server. We'll also need to protect the server against such connection attempts so we don't lose the original connection or other bad things.

Please feel free to provide a patch for it 😄

garlick:
> The only limitation you should be aware of is that a client process won't be able to fork/exec and have its child connect to the PMIx server.

Hmm, I hadn't thought about that one.  Did you have a specific use case in mind for that?

rhc54:
We ran into it with some applications running under SLURM, but it could become more prevalent given the broader use-cases from the extended PMIx APIs. For example, a client might fork/exec a child that wants to receive event notifications.

garlick:
I did get this working, nominally  - here are a few notes.

There are some client mods to skip the socket connect if `PMIX_SERVER_FD` is set in `PMIx_Init()` on this [socketpair](https://github.com/garlick/pmix/tree/socketpair) branch, in commit 625ccac97017a3e715d6f6824f89a31aa1b98c51.  This included a little refactoring to avoid code duplication.

I didn't have change anything in the PMIx server code to get this working;  instead I registered a listener callback to capture the `pmix_connection_cbfunc_t`.  Then in my server, I manually set `PMIX_SERVER_FD` in the client environment, and hand the server fd to the connection cb.  This could be made easier on the RM implementer if 
* pmix_connection_cbfunc_t was exposed directly
* environment setting was incorporated into `PMIx_server_setup_fork()` (somehow),
* there were a way to disable creation of the UNIX domain socket. 

[...]

PR #93, opened 2016-05-20
add support for "preconnected" clients

garlick:
This PR adds support for a slightly more efficient mode of wiring up clients, in which the resource manager calls `socketpair(2)`, tells PMIx about one fd, and passes the other fd to the client via a PMIX_SERVER_FD environment varaible, as discussed in #90.

One API function was added to support this, which can be called right after `PMIx_server_register_client()`:

```C
/* Register a "pre-connected" socket for the specified client.
 * The resource manager must arrange for this fd to remain open
 * across fork/exec when the client is started, and for
 * PMIX_SERVER_FD=fd to be set in the client's environment. */
pmix_status_t PMIx_server_register_client_fd(const pmix_proc_t *proc, int fd);
```

For clients connected in this way, the `init-auth` handshake is disabled.  Since the connection was established by the server, it is "pre-authenticated".

There are still some unresolved issues with this PR:
- [x] find some way to ask `PMIx_server_init()` not to create the UNIX domain socket if it is not to be used (one less thing to cleanup/protect)
- [ ] client `PMIx_Init()` should ignore PMIX_SERVER_URI if PMIX_SERVER_FD is set
- [ ] find some way for  `PMIx_server_setup_fork()` to set PMIX_SERVER_FD
- [ ] add option to `pmix_test` to run this way for testing

Despite those gaps, it is working well with my test RM integration and seems to speed things up a bit when launching a large number of clients.

Any feedback appreciated!

### DIFF
--- a/include/pmix/pmix_common.h
+++ b/include/pmix/pmix_common.h
@@ -99,6 +99,8 @@ BEGIN_C_DECLS
 
 /* attributes for the rendezvous socket  */
 #define PMIX_SOCKET_MODE           "pmix.sockmode"          // (uint32_t) POSIX mode_t (9 bits valid)
+#define PMIX_SOCKET_FILENAME       "pmix.sockname"          // (char*) filename to use for socket
+#define PMIX_SOCKET_SUPPRESS       "pmix.nosock"            // (bool) suppress creation/listening on socket
 
 /* general proc-level attributes */
 #define PMIX_CPUSET                "pmix.cpuset"            // (char*) hwloc bitmap applied to proc upon launch
--- a/include/pmix_server.h
+++ b/include/pmix_server.h
@@ -387,6 +387,12 @@ pmix_status_t PMIx_server_register_client(const pmix_proc_t *proc,
                                           void *server_object,
                                           pmix_op_cbfunc_t cbfunc, void *cbdata);
 
+/* Register a "pre-connected" socket for the specified client.
+ * The resource manager must arrange for this fd to remain open
+ * across fork/exec when the client is started, and for
+ * PMIX_SERVER_FD=fd to be set in the client's environment. */
+pmix_status_t PMIx_server_register_client_fd(const pmix_proc_t *proc, int fd);
+
 /* Deregister a client and purge all data relating to it. The
  * deregister_nspace API will automatically delete all client
  * info for that nspace - this API is therefore intended solely
--- a/src/client/pmix_client.c
+++ b/src/client/pmix_client.c
@@ -192,18 +192,16 @@ static void job_data(struct pmix_peer_t *pr, pmix_usock_hdr_t *hdr,
     cb->active = false;
 }
 
-static pmix_status_t connect_to_server(struct sockaddr_un *address, void *cbdata)
+static pmix_status_t setup_server(int sd, void *cbdata)
 {
-    int sd;
     pmix_status_t ret;
     pmix_cmd_t cmd = PMIX_REQ_CMD;
     pmix_buffer_t *req;
 
-    if (PMIX_SUCCESS != (ret=usock_connect((struct sockaddr *)address, &sd))) {
-        PMIX_ERROR_LOG(ret);
-        return ret;
-    }
+    /* mark the connection as made */
+    pmix_globals.connected = true;
     pmix_client_globals.myserver.sd = sd;
+
     /* setup recv event */
     event_assign(&pmix_client_globals.myserver.recv_event,
                  pmix_globals.evbase,
@@ -244,7 +242,7 @@ PMIX_EXPORT pmix_status_t PMIx_Init(pmix_proc_t *proc,
                           pmix_info_t info[], size_t ninfo)
 {
     char **uri, *evar;
-    int rc, debug_level;
+    int rc, debug_level, sd;
     struct sockaddr_un address;
     pmix_nspace_t *nsptr;
     pmix_cb_t cb;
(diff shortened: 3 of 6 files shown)


rhc54:
Fixes #90

dsolt:
I like the idea and the code actually a bit easier to read even though it is complicated by having two code paths.   If the user's command is actually a wrapper script or other complicated series of fork/exec's that finally results in the process that actually calls PMIx_Init() and if at some point in that chain some process closes the fd to the server, then PMIx_Init() will fail and the server will also detect that the client has closed its fd.   Not sure how common it would be for code to close their fd before calling exec?   I remember our (Platform) MPI daemon used to do that though.   Is it completely up to the server  whether to call PMIx_server_register_client_fd or PMIx_server_register_client?   Should there be an info that pmix_server_spawn_fn_t should take influence the choice?  Overall, I like it.

rhc54:
FWIW: OpenMPI always closes its fd's on fork before calling exec. It's a very common practice. So this mechanism will definitely not work with OMPI.

Cray used this approach and it caused considerable difficulties. Once a Cray person joined OMPI, they went in and hardcoded the file descriptors to be avoided (apparently there was some unpublished range they used as they have a strict limit on the number of supported clients).

So I have no issue with it as an option - just be aware that OMPI and perhaps other libraries will not work with it.

garlick:
>  Is it completely up to the server whether to call PMIx_server_register_client_fd or PMIx_server_register_client?

One would always call `PMIX_server_register_client()`, and then optionally, `PMIx_server_register_client_fd()`.

I was sort of wondering if it would be better API planning to add a `PMIx_server_register_client_info()` that takes a pmix_info_t array instead, and make the client fd one of the optional attributes?  This could be a replacement for `PMIx_server_register_client()` rather than an additional call, with uid, gid, and server_object arguments encoded as optional info attributes?  Then the original could be gently deprecated?

> OpenMPI always closes its fd's on fork before calling exec

In what situation would OMPI call exec?  Like when RM starts orte and orte starts MPI program?   But then isn't orte the PMIx provider?

I have a rebase on current master of this coming shortly, with some improved comments and a little less heavy handed change in `PMIx_Init()`.

garlick:
One question while working on `PMIx_Init()`:

By parsing PMIX_SERVER_URI we get the "server rank" and "server nspace", which are set in
```
pmix_client_globals.myserver.info->nptr->nspace
pmix_client_globals.myserver.info->rank
```

What's the distinction between these values and the ones from PMIX_NAMESPACE and PMIX_RANK, which are set in
```
pmix_globals.myid.nspace
pmix_globals.myid.rank
```

I'm wondering if I can use PMIX_NAMESPACE and PMIX_RANK for both if PMIX_SERVER_URI is not set.

dsolt:
Yes I don't think I can see a case where a pmix server would launch an orted or other MPI's daemon, so the discussion about orted or Platform's mpid closing all fd isn't completely relevant.  I'm just not sure what other tools might do that.   

I think 

pmix_client_globals.myserver.info->nptr->nspace
pmix_client_globals.myserver.info->rank

are some information about the server and not about the client.  PMIX_NAMESPACE/PMIX_RANK have the nspace/rank of the client while the structures above store some information about who its server is.  Not sure what its used for.  @rhc54 will know.   I notice that these nspace/rank seem to be "pmix-server" and <pid of pmix-server> respectively so they seem completely unrelated to a client nspace/rank and I'm not sure why they are stored in variables that imply they are a nspace/rank.

garlick:
Thanks @dsolt - will look for guidance from @rhc54 on those server values then.

It seems like a valid concern, clients closing the fd before `PMIx_Init()` can use it.   This might not be a good way to launch debugging tools for example.

As a data point, MPICH hydra does work like this, for the "simple PMI v1 wire protocol" that we were encouraged to implement in Flux to support launching MPICH.  I like the fact that file system namespaces and authentication become non-issues in this approach.

rhc54:
I think @dsolt misunderstood my comment. My point was that we need to keep this as an option, and not mandate that this is how everything will work. OMPI, when launching via mpirun, cannot use this model. There are other programming libraries that will have similar issues with it, as will the debuggers (as @garlick mentioned).

I suspect we will always have authentication concerns to deal with since we have applications interacting with the resource manager, which at some level has to reach into a privileged daemon. So whether the authentication occurs between the client and PMIx server library, or between the daemon hosting the PMIx server and the root-level daemon, there will be some concern somewhere along that chain. Might fall outside of our scope, though, which is fine.

The nspace/rank of the process is passed into the client for identification purposes. The nspace/rank of the server is required because we use it internally (e.g., for storing connection info and the peer arrays). So you'll always need something that represents the ID of the local server, and it cannot be the same as the process itself without making major changes to the code base.

garlick:
> I suspect we will always have authentication concerns to deal with since we have applications interacting with the resource manager, which at some level has to reach into a privileged daemon. So whether the authentication occurs between the client and PMIx server library, or between the daemon hosting the PMIx server and the root-level daemon, there will be some concern somewhere along that chain. Might fall outside of our scope, though, which is fine.

Are you speaking generally, or disagreeing with disabling the init-auth handshake for pre-connected clients in this PR?

rhc54:
I have no issue with disabling it for this optional connection method - just don't think we should remove it entirely.

garlick:
Just forced a push that is a rebase on current master, some cleanup, and two additional options that can be set in the info array of PMIx_server_init():
* PMIX_SOCKET_FILENAME: override the generated path for the socket, useful if the RM wants the PMIx socket created in its own controlled directory
* PMIX_SOCKET_SUPPRESS: suppress socket bind/listen thread, useful if RM is only launching pre-connected clients

garlick:
I think this PR has fallen behind upstream a bit, and I don't have time to work on it right now, so I'll withdraw it for now and resubmit when I have time to catch it up.

PR #100, opened 2016-05-31
Change how the PMIx constants are defined

rhc54:
Change the PMIx constants from an enum type to a list of #defines so that values can be added by users. Separate out some values that shouldn't be exposed outside the library as they won't have external meaning. Surround the #defines with #ifndef's to allow an embedding code to automatically make our values align with theirs - this eliminates one set of conversions


garlick:
> Surround the #defines with #ifndef's to allow an embedding code to automatically make our values align with theirs

This might be a bit dangerous from an ABI perspective.  E.g. if pmix is packaged for a distro, and other packages (RM, MPI, tools) build against it, they all have to be using the same values for constants, so it would be unsafe to allow any of those other packages to redefine them.

rhc54:
Hmmm...you are correct. It works fine within OMPI, but breaks when I direct launch against the SLURM plugin.

Oh well - we still need to remove the enum, but will leave the values #defined to hard numbers

garlick:
Apologies if this is off topic here, but what's the motivation to embed libpmix in OMPI?  Is it to integrate the server code with a libevent loop in ORTE?

One of the things I'm liking about libpmix is that it's factored out of the projects that use it and could evolve somewhat independently (within ABI constraints of course).  If projects start to embed it, then that advantage is diminished.  I guess in the case of ORTE/OPMI that's not so much a problem as they are components in the same project.

I also have the need to integrate pmix server bits with an external event loop (not libevent), so if that's the primary motivation to embed, maybe we could put our collective heads together and try to figure out how to accomplish that through exported interfaces, without giving up the shared library?

rhc54:
The only reason PMIx is embedded in OMPI is that we cannot be sure of its availability upstream when we are ready to release OMPI 2.0 (already past due), which depends on it. As PMIx becomes generally available, we will remove it from OMPI. So I would encourage anyone to not embed PMIx directly.

As for the event library, there is a way to tell PMIx to use your own event progress thread (as opposed to having it start its own) in 2.0, but not in 1.1.4. In 2.0, you pass the pointer to the event base with the PMIX\_EVENT\_BASE key when calling PMIx_Init or PMIx_server_init (I think this latter case is still sitting on my branch).

rhc54:
I should have noted: that external event progress thread has to be libevent. I don't know how to avoid that connection - we do require libevent during configure for a reason

garlick:
Right I was thinking there might be some way to expose the right interfaces to integrate it with another event loop type.   Other libraries have evolved this sort of interface.  For example see [libadns](http://www.chiark.greenend.org.uk/~ian/adns/docs/adns.h.txt) - specifically `adns_beforepoll()` and `adns_afterpoll()`.  The idea is to find out what fds and timeouts PMIx needs to poll on in the first call, and then when one of those is ready, give PMIx control in the second call, until it would block, e.g. run the libevent loop with EVLOOP_NONBLOCK.

garlick:
Getting a bit off-topic for this PR so I opened issue #102

PR #105, opened 2016-06-09
server: use memcpy in parse_connect_ack() to make sparc happy with un…

ggouaillardet:
…aligned data


ggouaillardet:
@rhc54 this should be back-ported in OpenMPI master (pmix114)
v2.x (pmix112) is safe

rhc54:
@ggouaillardet Good catch - thanks!

PR #106, opened 2016-06-10
Topic/tools

austenlauria:
This adds support on the convenience library side for a tool to connect to the PMIx server. A tool can be anything not forked by the resource manager.

### DIFF
--- a/src/server/pmix_server.c
+++ b/src/server/pmix_server.c
@@ -285,7 +285,8 @@ PMIX_EXPORT pmix_status_t PMIx_server_init(pmix_server_module_t *module,
     }
 
     /* check the info keys for a directive about the uid/gid
-     * to be set for the rendezvous file */
+     * to be set for the rendezvous file, and for indication
+     * of willingness to support tool connections */
     if (NULL != info) {
         for (n=0; n < ninfo; n++) {
             if (0 == strcmp(info[n].key, PMIX_USERID)) {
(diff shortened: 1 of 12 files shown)


src/server/pmix_server_listener.c
```diff
@@ -163,11 +164,15 @@ pmix_status_t pmix_start_listening(pmix_listener_t *lt)
...
-        }
+	if(!pmix_server_globals.listen_thread_active) {
+            if (0 > pthread_create(&engine, NULL, listen_thread, NULL)) {
+                pmix_server_globals.listen_thread_active = false;
+                return PMIX_ERROR;
+            }
+	    else {
+                pmix_server_globals.listen_thread_active = true;
+	    }
+	}
```
rhc54: I don't understand this code block. We just checked above (line 151) that the listen thread is not active, so why are we checking it again here?

src/tool/pmix_tool.c
```diff
@@ -0,0 +1,691 @@
...
+    /* setup the path to the daemon rendezvous point */
+    memset(&address, 0, sizeof(struct sockaddr_un));
+    address.sun_family = AF_UNIX;
+    
+    /* Get first 10 char's of hostname to match what the server is doing */
+    int hostnamelen = 10;
+    char hostname[hostnamelen];
+    gethostname(hostname, hostnamelen);
+
+    snprintf(address.sun_path, sizeof(address.sun_path)-1, "/tmp/pmix.%s.%d", hostname, server_pid);
```
rhc54: I must just be missing it - where did this file get created? I don't see that code.

src/server/pmix_server.c
```diff
@@ -294,17 +298,39 @@ PMIX_EXPORT pmix_status_t PMIx_server_init(pmix_server_module_t *module,
...
+                pmix_listener_t *tl = PMIX_NEW(pmix_listener_t);
+                memset(tl, 0, sizeof(pmix_listener_t));
+		
+		memset(&tl->address, 0, sizeof(struct sockaddr_un));
+		tl -> address.sun_family = AF_UNIX;
+		tl -> mode = ACCESSPERMS;
+
+		/* Get up to 10 chars of hostname.*/
+		gethostname(myhostname, myhostnamelen);
+                snprintf(tl->address.sun_path, sizeof(tl->address.sun_path) - 1, "/tmp/pmix.%s.%d", myhostname, getpid());
```
rhc54: Ah, I see - you made the file be an actual Unix domain socket itself, as opposed to containing info. Okay.

rhc54:
We'll use the official RFC PR

PR #111, opened 2016-07-04
add missing include files

ggouaillardet:
pmix cannot be built on alpine linux because of some missing includes.
uid_t and gid_t are defined in unistd.h or sys/types.h, and unistd.h
is not indirectly pulled under alpine linux, so do it manually.

Thanks N.L.K Nguyen for the report


ggouaillardet:
this issue was initially reported at open-mpi/ompi#1838

PR #112, opened 2016-07-05
server: fix a race condition in PMIx_server_init

ggouaillardet:
split pmix_start_listening() into two phases
 - pmix_prepare_listening() creates the socket without starting the listen_thread, this is invoked in a loop.
 - pmix_start_listening() only starts the listen_thread. this is only invoked once


rhc54:
@ggouaillardet this looks okay, but could you please tell me what the race condition was? I don't see any real change in logic here as the listen_thread_active variable would have been set by the first time thru the loop, and the second loop couldn't start in advance of that happening.

ggouaillardet:
here is what happened
first iteration
- create the pmix socket
- bind and listen on the pmix socket
- start the listen_thread 
second iteration
- create the tool socket

and the race occurs if the listener thread select on both socket.
it will fail on the tool socket, since it bind/listen was not yet invoked on it
/* select returns with an event on the tool socket, and accept fails with EINVAL,
this is interpreted as a race condition in PMIx_Finalize, and the listen_thread silently exits */


the logic was updated so all sockets are created, bound and listened on.
and then the listen_thread can be started

makes sense ?

rhc54:
I see - yes, thanks!

ggouaillardet:
btw, it is very likely I can simplify this pr a bit by removing the test on the listen_thread_active global

rhc54:
i already merged it, but you are welcome to submit another one with the optimization

PR #118, opened 2016-07-29
Provide a mechanism for the host to tell the PMIx server the system-l…

rhc54:
…evel tmpdir to use, and a server-specific tmpdir to use.

Cleanup some permissions


rhc54:
@artpol84 Does this look okay to you?

artpol84:
What is the usecase for this?

rhc54:
OMPI requires it, for one. The use case is where the system temporary directory is not specified by any of the standard locations, but rather by the host RM. So you need some way to tell the PMIx server library where to put the tool attachment point. You also need someway other than an envar to tell the server library where to put the session-level attachment point (the client one).

src/server/pmix_server.c
```diff
@@ -181,7 +183,9 @@ static pmix_status_t initialize_server_base(pmix_server_module_t *module)
     security_mode = strdup(pmix_sec.name);
 
     /* find the temp dir */
-    if (NULL == (tdir = getenv("PMIX_SERVER_TMPDIR"))) {
+    if (NULL != mytmpdir) {
+        tdir = mytmpdir;
```
artpol84: In which case `mytmpdir` can be non-NULL?

src/server/pmix_server.c
```diff
@@ -181,7 +183,9 @@ static pmix_status_t initialize_server_base(pmix_server_module_t *module)
...
+    } else if (NULL == (tdir = getenv("PMIX_SERVER_TMPDIR"))) {
```
artpol84: I think that we need to use MACRO's as env names instead of strings.

src/server/pmix_server.c
```diff
@@ -289,40 +294,51 @@ PMIX_EXPORT pmix_status_t PMIx_server_init(pmix_server_module_t *module,
...
-                if (0 > asprintf(&pmix_pid, "%s/pmix.%s.tool.%d", tdir, myhostname, mypid)) {
-                    return PMIX_ERR_NOMEM;
-                }
-                if ((strlen(pmix_pid) + 1) > sizeof(tl->address.sun_path)-1) {
-                    free(pmix_pid);
-                    return PMIX_ERR_INVALID_LENGTH;
+                /* defer processing to ensure we pickup any tmpdir
+                 * directives before setting location */
+                tool_support = true;
+            } else if (0 == strcmp(info[n].key, PMIX_SERVER_TMPDIR)) {
```
artpol84: **Minor comment**: Key macro has the same name as env variable. May lead to confusion at some point.

artpol84:
So in general what this PR adds is a new set of info keys:
* one of them duplicates existing env variable `PMIX_SERVER_TMPDIR`
* another introduces other tmpdir that is needed for the tool connection.

Question: why can't we handle this consistently with ENV vars?

rhc54:
Per the telecon, we will standardize on passing such values via info keys and not envars. I will remove the envar check so it solely uses info keys, and then we will bring that across to the 1.1 series and release an updated 1.1.5 so it can go into OpenMPI 2.0.1 release (this allows us to avoid having to carry the envar method forward for backward compatibility)

PR #121, opened 2016-08-02
Provide a key by which users can register a callback object to be returned whenever the registered event handler is invoked

### LINKED ISSUE #120: User object for notifications
A user would like to include a void* object when registering for notifications that will be returned to them whenever that specific notification callback function is executed. I believe this can be accomplished without modifying the current notification registration API by simply adding a new key (PMIX_EVENT_RETURN_OBJECT) that contains a pointer object as its value.

Only change required is to the event notification code so that it (a) retains an object provided via that key, and then (b) returns it in the cbfunc array of info.

rhc54:
The object will be returned in one of the info structures (in fact, it is the last info struct in the array, though that is not a requirement).

Fixes #120


rhc54:
@bosilca Would you mind taking a peek and letting me know what you think?

PR #123, opened 2016-08-06
Roll in the MCA infrastructure.

rhc54:
Includes:

* cleanup of the public headers to remove reference to configure-generated files

* addition of show_help and other utilities used by MCA

* addition of the MCA files themselves, without any components

Note that the MCA system is built, but is not referenced or initialized at this time


artpol84:
bot:retest

PR #124, opened 2016-08-09
Fetch some info from the info array before initializing server base.

karasevb:
In particular, this fixes initialization of the 'mytmpdir'.

### DIFF
--- a/src/server/pmix_server.c
+++ b/src/server/pmix_server.c
@@ -245,6 +245,23 @@ PMIX_EXPORT pmix_status_t PMIx_server_init(pmix_server_module_t *module,
     pmix_output_verbose(2, pmix_globals.debug_output,
                         "pmix:server init called");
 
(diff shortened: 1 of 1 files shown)


karasevb:
@artpol84 please review.

artpol84:
Ok, so the problem here is that `mytmpdir` is expected to be initialized in `initialize_server_base`:
https://github.com/pmix/master/blob/master/src/server/pmix_server.c#L186

But actual initialization of this variable is performed after  `initialize_server_base` in `PMIx_Server_init`:
https://github.com/pmix/master/blob/master/src/server/pmix_server.c#L299

artpol84:
:+1:

artpol84:
@rhc54 I think this is a minor fix so I'm merging it now. We need this fix for our dstor PR that we will merge later today.

artpol84:
@karasevb please go ahead with the DSTOR merge

PR #135, opened 2016-08-24
fix sec/native module under Solaris and other misc issues

ggouaillardet:


ggouaillardet:
@rhc54 can you please have a look ?
i will then update Open MPI master and PR for v2.x

config/pmix.m4
```diff
@@ -301,7 +301,8 @@ AC_DEFUN([PMIX_SETUP_CORE],[
                       time.h unistd.h dirent.h \
                       crt_externs.h signal.h \
                       ioLib.h sockLib.h hostLib.h limits.h \
-                      sys/statfs.h sys/statvfs.h])
+                      sys/statfs.h sys/statvfs.h \
+                      netdb.h ucred.h])
```
rhc54: I don't actually mind checking for netdb.h as the check is quick - just wondering why you added it since I don't see the result of that test being used?

ggouaillardet: from ```src/mca/base/pmix_mca_base_component_find.c```

```c
#ifdef HAVE_NETDB_H
#include <netdb.h>
#endif
```

rhc54: Ah, okay - I didn't see you add something and didn't realize it was missing. Thanks

PR #139, opened 2016-09-01
Support PMIx_Get calls on a namespace that the client is not a member and may not exist locally

dsolt:
Support a client (likely a tool) to call PMIx_Get on a namespace other than
    its own and in particular:
    
    1) call PMIx_Get on a namespace that is not located locally
    2) call PMIx_Get on either a specific rank or PMIX_RANK_WILDCARD
       In the later case, we expect job_info data to be collected
    
    To fully support this, the RM code for dmodex get has to be able
    to handle the case where we request a rank PMIX_RANK_WILDCARD and
    has to know to figure out which host to send that request to to
    get the job_info data.

Also piggybacked 2 minor bug fixes


src/buffer_ops/open_close.c
```diff
@@ -46,6 +46,8 @@ pmix_pointer_array_t pmix_bfrop_types = {{0}};
 pmix_data_type_t pmix_bfrop_num_reg_types = PMIX_UNDEF;
 static pmix_bfrop_buffer_type_t pmix_default_buf_type = PMIX_BFROP_BUFFER_NON_DESC;
 
+
+PMIX_EXPORT
```
rhc54: I don't believe we want to expose the bfrop code outside of the PMIx library - did you intend to do so?

src/server/pmix_server_get.c
```diff
@@ -457,7 +466,8 @@ static pmix_status_t _satisfy_request(pmix_nspace_t *nptr, pmix_rank_t rank,
             PMIX_DESTRUCT(&pbkt);
             cbfunc(rc, NULL, 0, cbdata, NULL, NULL);
             return rc;
-        }
+        }    
+	if (rank == PMIX_RANK_WILDCARD) found++;
```
rhc54: Minor nit: coding standards require use of braces around every if clause as it helps when running a debugger

src/tool/pmix_tool.c
```diff
@@ -664,7 +664,14 @@ PMIX_EXPORT pmix_status_t PMIx_tool_finalize(void)
     if (0 <= pmix_client_globals.myserver.sd) {
         CLOSE_THE_SOCKET(pmix_client_globals.myserver.sd);
     }
-    event_base_free(pmix_globals.evbase);
+/*
```
rhc54: Agreed - you can just delete this code (I may have already done this in a recent commit)

src/server/pmix_server_get.c
```diff
@@ -199,7 +199,14 @@ pmix_status_t pmix_server_get(pmix_buffer_t *buf,
          * give the host server a chance to tell us about it */
         rc = create_local_tracker(nspace, rank, info, ninfo,
                                   cbfunc, cbdata, &lcd);
-        return rc;
+/*
+ * Its possible there are no local processes on this host, so lets ask 
+ * for this explicitly.   
+ */
+        if( NULL != pmix_host_server.direct_modex ){
+                pmix_host_server.direct_modex(&lcd->proc, info, ninfo, dmdx_cbfunc, lcd);
```
rhc54: Indentation looks odd here

src/buffer_ops/open_close.c
dsolt: Yes, we like to use it... so we always make this change every time we pull down new bits.   Its just very convenience for the RM to be able to use this code to move PMIx-related structures around.

src/server/pmix_server_get.c
dsolt: I think when the namespace registration happens we are safe:  

    /* see if we already have this nspace */
    nptr = NULL;
    PMIX_LIST_FOREACH(tmp, &pmix_globals.nspaces, pmix_nspace_t) {
        if (0 == strcmp(tmp->nspace, cd->proc.nspace)) {
            nptr = tmp;
            /* release any existing packed data - we will replace it */
            if (0 < nptr->server->job_info.bytes_used) {
                PMIX_DESTRUCT(&nptr->server->job_info);
                PMIX_CONSTRUCT(&nptr->server->job_info, pmix_buffer_t);
            }
            break;
        }
    }
But the checking of trackers I need to add.

dsolt: Actually, I'm not sure we need to check for trackers either.  Won't that already happen once the last local rank registers itself?  At that point we call pmix_pending_nspace_requests(nptr) when the last _register_client call happens?

rhc54: Maybe? I don't know - my point was only that we should check to ensure that if we get the desired info via registration prior to receiving the direct modex response, then we need to ensure we don't pass the info down to the requesting client twice

rhc54:
@dsolt @jjhursey Looks like this now needs rebasing - is it otherwise ready to go?

jjhursey:
Replaced by PR #200

PR #145, opened 2016-09-08
Remove stale usock files if bind() failed.

artpol84:
We see sporadic MTT failures with the following error
```
[cn1:10888] [[44817,0],0] ORTE_ERROR_LOG: Error in file orted/pmix/pmix_server.c at line 254
[cn1:10888] [[44817,0],0] ORTE_ERROR_LOG: Error in file ess_hnp_module.c at line 656--------------------------------------------------------------------------
It looks like orte_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during orte_init; some of which are due to configuration or
environment problems.  This failure appears to be an internal failure;
here's some additional information (which may only be relevant to an
Open MPI developer):

  pmix server init failed
  --> Returned value Error (-1) instead of ORTE_SUCCESS
--------------------------------------------------------------------------
src/server/pmix_server_listener.c:92 bind() failed+ rc=213
+ exit 213
```

Looking into the /tmp directory I see lot's of pmix usock files and given that PID conflict can appear sometimes this may explain sporadic nature of MTT failures.

Also this PR makes bind failures more verbosive - errno is printed out helping to locate the reason if my original guess is incorrect.

I'm also investigating possible reasons of usock leftower's.

### DIFF
--- a/src/server/pmix_server.c
+++ b/src/server/pmix_server.c
@@ -213,8 +213,11 @@ static pmix_status_t initialize_server_base(pmix_server_module_t *module)
     }
 
     /* for now, just setup the v1.1 series rendezvous point
-     * we use the pid to reduce collisions */
-    if (0 > asprintf(&pmix_pid, "%s/pmix-%d", tdir, mypid)) {
+     * - use userid to avoid situation where user with other UID but same PID 
+     *   has created the usock and terminate abnormally and we have no way to
+     *   remove the file
+     * - use the pid to reduce collisions */
+    if (0 > asprintf(&pmix_pid, "%s/pmix-%d.%d", tdir, getuid(), mypid)) {
         return PMIX_ERR_NOMEM;
     }
 
--- a/src/server/pmix_server_listener.c
+++ b/src/server/pmix_server_listener.c
@@ -70,6 +70,37 @@ static void tool_handler(int incoming_sd, short flags, void* cbdata);
 static char *myversion = NULL;
 static pthread_t engine;
 
+static int _try_cleanup_usock(char *fname)
+{
+    struct stat sbuf;
+    
+    /* if socket file not exist we don't have the workaround */
+    if( stat(fname, &sbuf) ){
+        return PMIX_ERROR;
+    }
+
+    /* If usock filename corresponds to something not a socket - no workaround */
+#ifdef S_ISSOCK
+    /* if this is POSIX version > POSIX.1-1996 we can check for unix socket */
+    if( !S_ISSOCK(sbuf.st_mode) ){
+        return PMIX_ERROR;
+    }
+#else
+    /* otherwise - use negation */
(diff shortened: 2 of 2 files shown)


artpol84:
@rhc54, please review

artpol84:
P.S. We need this in OMPI-2.0.2 ASAP. I'll PR to release later today.

src/server/pmix_server.c
```diff
@@ -213,8 +213,11 @@ static pmix_status_t initialize_server_base(pmix_server_module_t *module)
     }
 
     /* for now, just setup the v1.1 series rendezvous point
-     * we use the pid to reduce collisions */
-    if (0 > asprintf(&pmix_pid, "%s/pmix-%d", tdir, mypid)) {
+     * - use userid to avoid situation where user with other UID but same PID 
+     *   has created the usock and terminate abnormally and we have no way to
+     *   remove the file
+     * - use the pid to reduce collisions */
+    if (0 > asprintf(&pmix_pid, "%s/pmix-%d.%d", tdir, getuid(), mypid)) {
```
rhc54: This is going to create a problem - we are already encountering issues with the length of the usock pathname on recent Linux kernels as the temp directory base has moved down to the var directory under a name created by a hash. So adding another field to it is going to add to the problem.

rhc54:
This honestly looks like a band-aid to me. I considered and rejected the same approach awhile ago as it really isn't addressing the problem.

What I think we really need to do is move away from usocks as they require "physical" rendezvous points that the OS doesn't clean up for us. If the PMIx server fails, then the litter is left behind to cause additional problems at a later time.

As for OMPI: we already have a solution to this problem in the master (we move the contact file down inside the session directory to eliminate the later conflicts). What remains unclear is if that solution should be moved to 2.0.2, or brought in as part of 2.1.0.

src/server/pmix_server.c
artpol84: Let's add UID only if RM haven't specified server tempdir. In this case probability of conflict is low, we may even don't need to have PID there.
Otherwise we may hit access right problems.

artpol84:
I guess that OMPI approach doesn't conflict with this one.
I was about to add more verbosity to try_cleanup function so that in case wi trigger this problem in future we will get more detailed description of the problem. 
I'm really missing this output now to diagnose.

rhc54:
We already know one reason - if the orted dies for some reason, then the PMIx server doesn't get called to cleanup and the litter is left behind. In all other cases, the cleanup in ```pmix_server_listener.c``` should take care of it.

artpol84:
it would great if this is the reason of our faults. I'm 99% sure it is, but still it is not 100% confirmed.
More detailed output would help to clarify.

artpol84:
Also I'd say we need this for our test suite.

artpol84:
Should be solved by external entity

PR #149, opened 2016-09-15
Have configure create a rename file for us to hide symbols when embedded.

rhc54:
Include pmix_globals in the exported symbols so the plugins can find it

Must export the component structure for dynamic plugins so the linker can find them as public data

Must include the pmix_output and pmix_output_verbose functions (and hide them) since plugins reference those


rhc54:
@dsolt @artpol84

jolivier23:
I can't compile a simple program that checks for pmix header.   Here is the program:
    #include <pmix.h>

It gives the following compiler error:
jvolivie-desk1:[~/scons_local]$ gcc -o .sconf_temp/conftest_4.o -c -Iinstall/include .sconf_temp/conftest_4.c
In file included from .sconf_temp/conftest_4.c:2:0:
install/include/pmix.h:51:25: fatal error: pmix_rename.h: No such file or directory
 #include <pmix_rename.h>
                         ^
compilation terminated.

rhc54:
@jolivier23 Try now - the generated header wasn't being installed.

jolivier23:
appears to work now.  Thanks

Issue #136, opened 2016-08-24
flock is not available on Solaris


ggouaillardet:
```src/dstore/pmix_esh.c``` uses the ```flock``` function in order to lock (shared or exclusive) or unlock a file.

unfortunatly, this is not available on Solaris (and maybe some other OSes as well).
sadly, i have no alternative to suggest yet.

could someone please find a way to fix this ?

rhc54:
This won't affect the OMPI v2.0.1 release, but will block us from the 2.1 release. OMPI has methods for dealing with this problem - perhaps we can borrow from the opal/shmem support over there?

artpol84:
@rhc54 I tried to assign @karasevb but failed. Do we need to do something special to enable this?

rhc54:
He has to join the organization - I just sent him an invite.

artpol84:
@rhc54 thank you!

artpol84:
@ggouaillardet do you know if fcntl locking is supported on Solaris?
If it is we will provide configuration-driven way to choose between flock and fcntl.

artpol84:
I'm talking about Advisory Record Locking:
http://man7.org/linux/man-pages/man2/fcntl.2.html

We may even improve concurrency of DSTOR by locking only portions of the shmem files (if this is possible)

ggouaillardet:
that is supported, at least per the man page.
and since ROMIO uses that, i do believe this is supported anywhere (except maybe under windows, but we do not have to care any more)

artpol84:
Great, thank you!

PHHargrove:
For more info on replacing flock() w/ fcntl() you may wish to have a look at [this page](https://www.perkin.org.uk/posts/solaris-portability-flock.html)

artpol84:
Closing through https://github.com/pmix/master/pull/157

PR #150, opened 2016-09-19
dstore/pmix: flock portability

karasevb:
Fixed the problem introduced in #136.

@ggouaillardet, @artpol84 please review.


config/pmix.m4
```diff
@@ -659,6 +659,14 @@ AC_DEFUN([PMIX_SETUP_CORE],[
     AC_SUBST(pmixlibdir)
     AC_SUBST(pmixincludedir)
 
+    ##################################
+    # Flock
+    ##################################
+    pmix_show_subtitle "Configuring flock"
+
+    AC_CHECK_FUNC([flock], [AC_DEFINE([PMIX_ENABLE_FLOCK], [ ],
```
ggouaillardet: or you can add ```flock``` to the arguments of ```AC_CHECK_FUNCS``` and test with
```c
#ifdef HAVE_FLOCK
...
#endif
```

artpol84:
Thank you.
Can you verify that it works ok on Solaris as is?
And then we can address the comment.

ggouaillardet:
Will do tomorrow

artpol84:
Here we obviously will not sync procs from several nodes using it. But I think we still might need NFS support to sync procs on the same node.

ggouaillardet:
i was able to build pmix on Solaris, i needed #151 though (not critical, since only test and example are affected)

artpol84:
@ggouaillardet thank you.

To close the question on NFS locking here is the excerpt from the `flock` man page:
```
       In Linux kernels up to 2.6.11, flock() does not lock files over NFS (i.e., the scope of locks was limited to the local system).  Instead, one could use fcntl(2) byte-range locking, which does work over NFS, given a sufficiently recent version of Linux and a server which supports locking.  Since Linux 2.6.12, NFS clients support flock() locks by emulating them as byte-range locks on the entire file.  This means that fcntl(2) and flock() locks do interact with one another over NFS.  Since Linux 2.6.37, the kernel supports a compatibility mode that allows flock() locks (and also fcntl(2) byte region locks) to be treated as local; see the discussion of the local_lock option in nfs(5).
```
so `flock` is fine for NFS in recent kernels. However I agree that since we can achieve the same effect with both ways. But `fcntl` is known to be more portable and have wider coverage in FS so let's go ahead with it.

artpol84:
All we need from NFS (if the dstor session dir will be located there) is `local_lock` to allow the locking mechanism that we are using.

karasevb:
@artpol84 please review

artpol84:
Looks OK.
@ggouaillardet could you double-check before we merge?

ggouaillardet:
that looks good to me too !

artpol84:
@ggouaillardet thank you!

artpol84:
@karasevb please, PR to the v2.x

PR #151, opened 2016-09-20
misc fixes to get pmix build on Solaris

ggouaillardet:
- MAXHOSTNAMELEN is defined in <netdb.h> on Solaris
- stdout cannot be assigned, so use freopen instead

### DIFF
--- a/examples/dynamic.c
+++ b/examples/dynamic.c
@@ -15,6 +15,8 @@
  * Copyright (c) 2011      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2013-2016 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Mellanox Technologies, Inc.  All rights reserved.
+ * Copyright (c) 2016      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


examples/dynamic.c
```diff
@@ -31,6 +33,10 @@
 #include <unistd.h>
 #include <time.h>
 #include <sys/param.h>
+#include <pmix_config.h>
+#ifdef HAVE_NETDB_H
+#include <netdb.h>
+#endif
```
rhc54: Can you pinpoint what in this example required netdb.h? I ask because this is supposed to be an example user program, and including pmix_config.h would not generally be acceptable. Instead, we should remove from the example whatever is forcing that include file to be required.

ggouaillardet: MAXHOSTNAMELEN is defined in netdb.h under Solaris (my commit message was truncated, I will update it tomorrow)
pmix_config.h is where HAVE_NETDB_H is #define'd
netdb.h is likely present under Linux, so removing the #ifdef could be acceptable too

rhc54: I see - let's just replace that MAXHOSTNAMELEN with some arbitrary number (e.g., 1024) so we remove the issue. It's unimportant in terms of the example we are trying to provide.

ggouaillardet:
Ok, will do

Issue #146, opened 2016-09-08
Restore smoky test functionality


artpol84:
Running 
```
./test/pmix_test -n 1 --job-fence 
```
I'm retting the error:
```
ERROR [test_fence.c:370:get_local_peers]: smoky_nspace-0:0: PMIx_Get local peer # failed: -107
ERROR [pmix_client.c:147:main]: smoky_nspace-0:0 Job fence test failed: -1
ERROR [cli_stages.c:148:cli_cleanup]: rank 0 with state 1 unexpectedly terminated.
ERROR [pmix_test.c:173:main]: Test was aborted!
```

Need to fix this. This should work for any sane `-n`.

artpol84:
Closing throug https://github.com/pmix/master/pull/174

PR #152, opened 2016-09-20
test: fixed fence test

karasevb:
Fixed problem introduced in #146


rhc54:
👍

artpol84:
@karasevb please check if we need this in v2.x and PR if we do.

artpol84:
please, note that v2.x now is in this (master) repository and not in the release repo.

karasevb:
@artpol84 I checked it in master/v2.x, and it is does not required for fix there.

PR #153, opened 2016-09-21
configury: correctly handle --disable-dlopen

ggouaillardet:
the LT_* macros do overwrite the enable_dlopen variable,
so it must be tested and saved before invoking LT_INIT.
delay the invokation of the LT_* macros and use the
PMIX_ENABLE_DLOPEN_SUPPORT variable to figure out whether
--disable-dlopen was invoked


ggouaillardet:
see open-mpi/ompi#2079 and open-mpi/ompi@041a431966a72313320d19fff7287bbcc114b75d

PR #125, opened 2016-08-11
Fix buffer unpack hang

karasevb:
In case the buffer unpack failed, we have a hang in [waiting status](https://github.com/pmix/master/blob/d8def6b020dd392ed937962dcc6f4d2eb6fdb2c1/src/client/pmix_client.c#L419).

PR #156, opened 2016-09-26
Fix buffer unpack hang

karasevb:
Сhery-picked from commit d8def6b (PR #125)


karasevb:
@artpol84 please review

artpol84:
@karasevb please update the commit comment with the as follows:
1. Name of commit should identify that you are fixing test suite, i.e. "test_suite: fir smoky test buffer unpack hang"
2. Fix the commit message describing the problem: 
* in what circumstances the problem appears
* why exactly the hang occurs.

artpol84:
Ok, I talked to @karasevb and it is not a test fix. 
Good to go.

artpol84:
@jjhursey please review and merge

PR #159, opened 2016-10-03
v2.x: sec/munge: add a missing include file

ggouaillardet:


ggouaillardet:
i will then backport into Open MPI
pmix master does things differently, so this is really a one-off commit

rhc54:
This is fine, but I don't think pmix2x is in OMPI at this time - OMPI master is at PMIx v3, and OMPI 2.x is still at PMIx v1.1.5

rhc54:
Also, FWIW: we may rebranch PMIx v2.x if a proposed change in plan for OMPI v2.1 is accepted. So I'll take this in for now, but it might be worth holding off on worrying about PMIx v2.x for a few days.

ggouaillardet:
@rhc54 in pmix/master, ```sec/munge``` has been moved into a ```mca```
but in open-mpi/master, ```sec/munge``` is still in ```src/sec``` just like in PMIx ```v2.x```
so i will backport this (at least for the next few days, until embedded pmix is refreshed)

PR #160, opened 2016-10-03
dstore: fix the extention of the initial segment

karasevb:
When added new sm segment the counter was not reset. It was
cause of error in logic operations on data elements in a segment
with incorrect counter. This fix is resets the counter of elements
for new segment.


karasevb:
@artpol84 please review.

artpol84:
:+1:

PR #165, opened 2016-10-05
dstore: fix PMIX_DSTPATH info-key type

artpol84:
So it'll be consistent with pmix_common.h


artpol84:
@karasevb please review.

I stumble upon it when I was porting dstore to v1.2.x. In the pmix_common.h we say that the type ins (char*). 

Our intentions of using "ptr" was to avoid extra copies. It then has to be stated in `pmix_common.h`. Maybe we should go that route, though I'm not sure we will gain much from this but may introduce some confusion at the user side.

karasevb:
:+1:

Issue #129, opened 2016-08-18
Reduce memory footprint


rhc54:
Fujitsu has kindly shared their memory footprint measurements at scale, and the PMIx footprint is growing at an unacceptable rate. The primary reasons are two-fold:
1. the use of fixed-length keys and namespace values is consuming a lot of empty space.
2. presence of a pmix_proc_t struct, which in turn contains a fixed-length namespace array, in the pmix_value_t data union, thereby causing that oft-used struct to be quite large

Concern was raised over the use of these in the pmix_info_t structure, but that is a red-herring. PMIx only stores pmix_kval_t objects - the pmix_info_t structure is only used for interfacing to the outside. Thus, we can resolve the problem by:
1. changing the pmix_kval_t to use flexible sized key strings. To avoid multiple tiny malloc's, we will move the key entry to the end of the struct and pre-compute size using the known string length.
2. replacing the pmix_proc_t in the pmix_value_t struct with a pointer to that struct
3. replacing fixed namespace arrays with variable-length storage in all internal objects

These steps will resolve the problem without impacting any of the RM's or applications that have implemented their integration. This is a good thing as that would have been highly disruptive.

kawashima-fj:
The presented material:
https://github.com/open-mpi/ompi/wiki/Meeting-2016-08-attachments/Fujitsu-OMPI-Dev-Meeting-2016-Aug.pdf
(also linked from [Open MPI Developer's meeting Wiki page](https://github.com/open-mpi/ompi/wiki/Meeting-2016-08))

jjhursey:
We found a memory leak in PMIx_Get in the server side code a couple weeks ago. We have a fix that we need to push upstream - @dsolt and @nysal did that work. It might not be a huge factor here (depending on the number of get calls), but would be good to push back.

rhc54:
Well, I went thru the code base and found that we already use pmix\_kval\_t everywhere internally, converting all pmix\_info\_t values to that struct. Thus, the first part of my list of fixes was already done, and nothing to be gained there.

So I did the second part in the referenced PR, squashing the pmix\_value\_t struct down to 8 bytes, and committed that to the OMPI master so folks can see how that impacts the memory footprint. I also enabled the shared memory datastore, which should also help.

@hjelmn said he could test this next week, so let's see what this now looks like.

kawashima-fj:
@rhc54 Thanks! I'll evaluate the memory footprint again using the latest master. The result will be posted here in late this week hopefully.

[...]

PR #170, opened 2016-10-06
plug misc memory leaks on finalize

ggouaillardet:


ggouaillardet:
Refs #129

ggouaillardet:
Refs open-mpi/ompi#2175

PR #172, opened 2016-10-06
V1.2 dstor pr7 (depends on #171)

artpol84:
Once #171 is merged this one has to be rebased.
Represents #141


artpol84:
bad sequence again!

PR #174, opened 2016-10-06
pmix/test: fixed tests

karasevb:
- Fixed spawn test
- Fixed fence test

### DIFF
--- a/test/pmix_test.c
+++ b/test/pmix_test.c
@@ -35,6 +35,9 @@
 
 #include "server_callbacks.h"
 #include "utils.h"
+#include "src/include/pmix_globals.h"
+
+bool spawn_wait = false;
 
 int main(int argc, char **argv)
 {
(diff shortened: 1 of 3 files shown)


test/pmix_test.c
```diff
@@ -175,6 +178,8 @@ int main(int argc, char **argv)
         test_fail = 1;
     }
 
+    PMIX_WAIT_FOR_COMPLETION(spawn_wait);
```
artpol84: What if we are running non-spawn test?

test/pmix_test.c
```diff
@@ -35,6 +35,9 @@
 
 #include "server_callbacks.h"
 #include "utils.h"
+#include "src/include/pmix_globals.h"
+
+int spawn_wait = 0;
```
artpol84: `spawn_wait` is `int` and set to 0 here

test/server_callbacks.c
```diff
@@ -263,13 +266,27 @@ typedef struct {
...
+
+static void * _release_cb(void *arg)
 {
-    release_cbdata *cb = (release_cbdata*)cbdata;
+    release_cbdata *cb = (release_cbdata*)arg;
     if (NULL != cb->cbfunc) {
         cb->cbfunc(cb->status, "foobar", cb->cbdata);
     }
     free(cb);
+    spawn_wait = false;
```
artpol84: but here `spawn_wait` is set to `false`. This will work but doesn't seem consistent.

karasevb:
@artpol84 please check

PR #181, opened 2016-10-12
configury: fix --disable-dlopen support

ggouaillardet:


ggouaillardet:

Refs open-mpi/ompi#2207

rhc54:
I _think_ that looks right - I'll add it in and update OMPI master to see if it fixes that problem. Thx!

PR #185, opened 2016-10-13
client: invoke pmix_progress_thread_init() once

ggouaillardet:
pmix_progress_thread_init() is already (indirectly) invoked
(by pmix_rte_init()) so there is no need to call it again


ggouaillardet:
currently, with Open MPI, the refcount of the PMIx thread is always greater than 0 and the thread is never free'd
with this commit, the refcount drops to 0 and the PMIx thread is stopped.

PR #191, opened 2016-10-18
usock: destroy pmix_usock_globals.posted_recvs in the progress thread

ggouaillardet:


ggouaillardet:
@rhc54 can you please have a look at this ?

this is related so MTT errors such as https://mtt.open-mpi.org/index.php?do_redir=2369

there is a race condition in ```pmix_rte_finalize```
in the main thread, ```pmix_usock_finalize``` does
```c
PMIX_LIST_DESTRUCT(&pmix_usock_globals.posted_recvs);
```
while the progress thread invokes ```pmix_usock_process_msg``` and access ```pmix_usock_globals.posted_recvs```

this PR have the ```pmix_usock_globals.posted_recvs``` list destructed in the progress thread, so the race condition is avoided.

an other (better ?) option is to invoke ```pmix_progress_thread_finalize()``` __before__ ```pmix_usock_finalize()```, but i cannot assess the impact of doing this

rhc54:
Shutting down the progress thread is the correct solution as there are other things in finalize that are similar race conditions. I've posted PR #2237 to do so.

Thanks!

rhc54:
Sorry - that was the OMPI PR number. I've committed the same change to PMIx master as well.

PR #192, opened 2016-10-18
Add in the network interface support

rhc54:
This is not used as of yet, but will be used by the upcoming cross-compatibility code

@ggouaillardet I cannot check that the Solaris components build - can you please take a look?


rhc54:
Actually, I don't have everything copied over here yet - closing this until I do.

PR #196, opened 2016-10-21
Fix misc memory leaks on finalize

ggouaillardet:


ggouaillardet:
the last commit does fixes a bug. if you run a OMPI hello world singleton under valgrind, you will see memory is used after free

src/mca/psec/base/psec_base_frame.c
```diff
@@ -43,24 +43,31 @@
...
+    PMIX_LIST_FOREACH_SAFE(active, prev, &pmix_psec_globals.actives, pmix_psec_base_active_module_t) {
+      pmix_list_remove_item(&pmix_psec_globals.actives, &active->super);
       if (NULL != active->component->finalize) {
         active->component->finalize();
       }
+      PMIX_RELEASE(active);
     }
     PMIX_DESTRUCT(&pmix_psec_globals.actives);
 
+    PMIX_LIST_FOREACH_SAFE(item, prev_item, &pmix_psec_base_framework.framework_components, pmix_list_item_t) {
```
rhc54: Are you saying that framework_close doesn't destroy this list for us? That would seem like a bug in the MCA base, yes?

ggouaillardet:
from ```pmix_mca_base_framework_close```
```c
    /* close the framework and all of its components */
    if (is_open) {
        if (NULL != framework->framework_close) {
            ret = framework->framework_close ();
        } else {
            ret = pmix_mca_base_framework_components_close (framework, NULL);
        }

        if (PMIX_SUCCESS != ret) {
            return ret;
        }
    } else {
        pmix_list_item_t *item;
        while (NULL != (item = pmix_list_remove_first (&framework->framework_components))) {
            pmix_mca_base_component_list_item_t *cli;
            cli = (pmix_mca_base_component_list_item_t*) item;
            pmix_mca_base_component_unload(cli->cli_component,
                                           framework->framework_output);
            PMIX_RELEASE(item);
        }
        ret = PMIX_SUCCESS;
    }

    framework->framework_flags &= ~(PMIX_MCA_BASE_FRAMEWORK_FLAG_REGISTERED | PMIX_MCA_BASE_FRAMEWORK_FLAG_OPEN);

    PMIX_DESTRUCT(&framework->framework_components);
```

in the case of ```psec``` ```is_open``` is ```true``` and ```framework->framework_close``` is not ```NULL```
then we end up doing

```c
PMIX_DESTRUCT(&framework->framework_components);
```

at first glance, we could
```c
PMIX_LIST_DESTRUCT(&framework->framework_components);
```
instead, so ```framework->framework_close``` does not have to manually free the components.
/* other path free the components */

fwiw, ```OPAL``` does the same thing
i do not know the history and cannot say whether this is a typo or a design choice.

@jsquyres @hjelmn @bosilca can you please comment on that ?

rhc54:
FWIW: I don't see any frameworks in OMPI that behave differently than psec does here - so are we saying that every OMPI framework is leaking this way?

ggouaillardet:
among other ```OPAL``` frameworks, ```btl``` does
```c
(void) mca_base_framework_components_close(&opal_btl_base_framework, NULL);
```

so instead of 
```c
PMIX_LIST_FOREACH_SAFE(item, prev_item, &pmix_psec_base_framework.framework_components, pmix_list_item_t)
...
```
i should likely have done
```c
(void) pmix_mca_base_framework_components_close(&pmix_psec_base_framework, NULL);
```

makes sense ?

rhc54:
Ah - yes, that should be in the close function in pmix/mca/sec/base/pmix_base_frame.c. Thanks!

rhc54:
argh - it is also missing

```
    /* Open up all available components */
    return pmix_mca_base_framework_components_open(&pmix_psec_base_framework, flags);
```

in the open function.

ggouaillardet:
i updated the PR with open and close in ```psec```

rhc54:
how odd - github won't let me commit this PR "at this time". Perhaps a github issue? I'll try again in a bit.

PR #200, opened 2016-10-24
Daves new get changes 

dsolt:
This is a re-organization and rebase of the previous pull request for get changes:

Support a client (likely a tool) to call PMIx_Get on a namespace other than
its own and in particular:

1) call PMIx_Get on a namespace that is not located locally
2) call PMIx_Get on either a specific rank or PMIX_RANK_WILDCARD
   In the later case, we expect job_info data to be collected

To fully support this, the RM code for dmodex get has to be able
to handle the case where we request a rank PMIX_RANK_WILDCARD and
has to know to figure out which host to send that request to to
get the job_info data.


jjhursey:
Should PR #139 be closed and replaced with this one?

PR #201, opened 2016-10-24
pmix/config: Enable dstore by default

jjhursey:
* This commit enables the shared memory datastore by default
   - This makes it match the `master` branch
 * No clear commit to pick from `master` so manually applied change.


jjhursey:
Merging in so we can include this as part of the RC roll.

PR #205, opened 2016-11-02
dstore: fixed pmix build fails with dstore

karasevb:
Fixed the pmix build fail with `--dstore-enable` which introduced after the commit 333efe5.

@rhc54 please review


rhc54:
Huh - that's weird. My code shows the same lines as this commit adds, yet I don't see those lines in the referenced commit. Somehow, it wasn't picked up in the commit?

Dunno. Anyway, thanks for picking it up!

jjhursey:
Looks good to me too. Note that `v1.2` does not have this issue, only `master` so no need to cherry-pick.

PR #208, opened 2016-11-04
Add the pmix performance measurement tool.

artpol84:
The following process is considered:
    - each process put's defined number (param) of keys of some size (param)
    - each process does commit and fence
    - each process fetches the keys for each other process.


artpol84:
@jjhursey this is a more convenient version of the performance evaluation tool.

rank0 will summarize the measurements from others. We can also output min/max in addition to averages if we want. But this is something to begin with.

artpol84:
@karasevb

Issue #207, opened 2016-11-03
PMIx 1.2 dstore PMIx_Get Performance issue


jjhursey:
I just got back some performance results comparing PMIx `1.1.2`, `1.1.5`, and `1.2.0rc1` (with dstore enabled). I see a performance regression when performing large numbers of `PMIx_Get` calls at high PPN on a few nodes.

For the particular test code (which I cannot share, sorry) it makes a number of `PMIx_Get` calls combined with other `PMIx_` calls (e.g., put, fence, commit -- nothing too fancy). In the output below you can see the total time spent in `PMIx_Get` then normalized per call. The diff is in comparison to `1.1.2`
```text
-------------------------------------------------------------------------------------
Nodes    NP       1.1.2  calls    per call       1.1.5  calls    per call    diff    1.2.0rc1  calls    per call    diff 
-------------------------------------------------------------------------------------
    1    80    0.158119    403    0.000392    0.156834    403    0.000389   0.99x    0.505656    403    0.001255   3.20x 
    2   160    0.399714    803    0.000498    0.404362    803    0.000504   1.01x    1.314171    803    0.001637   3.29x 
    3   240    0.656375   1203    0.000546    0.667607   1203    0.000555   1.02x    2.067067   1203    0.001718   3.15x 
    4   320    0.926069   1603    0.000578    0.952027   1603    0.000594   1.03x    2.843061   1603    0.001774   3.07x 
    5   400    1.222624   2003    0.000610    1.233513   2003    0.000616   1.01x    3.529636   2003    0.001762   2.89x 
    6   480    1.538374   2403    0.000640    1.560746   2403    0.000649   1.01x    4.208262   2403    0.001751   2.74x 
    8   640    2.095369   3203    0.000654    2.084200   3203    0.000651   0.99x    5.623388   3203    0.001756   2.68x 
   12   960    2.934969   4803    0.000611    2.995031   4803    0.000624   1.02x    8.383371   4803    0.001745   2.86x 
   16  1280    3.925197   6403    0.000613    3.992772   6403    0.000624   1.02x   11.404139   6403    0.001781   2.91x 
   18  1440    4.416280   7203    0.000613    4.494544   7203    0.000624   1.02x   12.785547   7203    0.001775   2.90x 
```

You can see an almost 3x performance lost by using dstore, which is completely unexpected. I want to do some more fine tuned experimentation early next week, but maybe others can take a look before then.

I configured PMIx with
```shell
[...]

PR #210, opened 2016-11-05
dstore/esh: fix fence

artpol84:
Ensure that the local keys are getting into the dstore during fence.

Partially address #207. Accidentally local keys wasn't submitted to the dstore at the fence stage. This was causing significant performance degradation at local data access.


artpol84:
Incorrect PR. will open the right one asap.

PR #212, opened 2016-11-05
dstore: Fix PMIx_Commit

artpol84:
Force local data to deposit directly to the dstore at PMIx_Commit.
Before this patch local keys wasn't published in the dstore causing
first access to local key to go through a long path by sending a
message to server and asking to publish the data.

This was causing performance issues.


artpol84:
bot:retest

artpol84:
bot:retest

artpol84:
bot:retest

PR #216, opened 2016-11-08
Provide support for both TCP and USOCK transports by creating a new PMIx

rhc54:
Transport Layer (PTL) framework, and implementing each transport as a
component within that framework. Support for legacy USOCK clients will
only be started if the PMIX_USOCK_ENABLE key is passed during
PMIx_server_init - otherwise, all communication will occur over the TCP
transport.

Note: Tool connections are only supported over TCP, and tools are
allowed to connect to remote servers by passing the remote server's URI
in the PMIx_tool_init call via the PMIX_TCP_URI key.

The TCP component defines its own PMIX_SERVER_URI2 envar by which the
server's connection info can be passed to clients using PMIx_Setup_fork.
The USOCK component, if enabled, will continue to pass its rendezvous
filename using the PMIX_SERVER_URI envar for backward compatibility. The
handshake protocol for USOCK remains the same as in prior PMIx versions.

The "simptest" test code has been modified to support cross-version
testing by the addition of the "-x" option. When given, this option
directs simptest to alternate TCP and USOCK transports for each client.

This has been tested running multiple simultaneous clients using a mix
of TCP and USOCK transports.


rhc54:
Note: I still need to restore the search code for discovering the server contact info when multiple PMIx servers are present. Given the MCA param for specifying the URI, this takes a little thought to figure out the easiest-to-use method.

rhc54:
@artpol84 This is ready to go. I changed the PTL selection logic so usock is always on (along with tcp). You can select either one using MCA param (PMIX\_MCA\_ptl). Priorities will auto-select tcp.

Can you take a look and let me know if it's okay to merge?

artpol84:
@rhc54 
I see that psec/none component was added. What is the purpose?

artpol84:
I guess this is for the trivial TCP case where the environment is considered protected, correct?

Important question: **Current Jenkins run demonstrates TCP or Unix socket results**? As I understand TCP is the default, so we don't test usock anymore. @rhc54 what are the parameters to activate it?

@karasevb We need to test both code paths - thus modify our Jenkins. Please add this to your TODO list.

rhc54:
@artpol84 The psec/none component was indeed added for protected environments and can only function if nothing else works or someone specifies it. Note that both sides would have to agree to use it, so a user can't just arbitrarily invoke it.

The current Jenkins will default to using TCP, not usock. To select usock, just set ```PMIX_MCA_ptl=usock``` in your environment.

artpol84:
@rhc54 I see in the commit description that you actually have a set of commits. Why not to keep them instead of merging into the giant one?

rhc54:
Need to bring this along so I can fix Intel builds and continue working on cross-version support over the holidays.

Issue #76, opened 2016-04-15
Error handler memory corruption


artpol84:
@rhc54 @annu13
Hi, I'm facing a memory corruption when stress testing PMIx through PMI2 interface (pmix plugin).
The test can be downloaded here:
https://bugs.schedmd.com/attachment.cgi?id=1470

Here is the backtrace:

```
(gdb) bt
#0  0x00007f3b19b06cc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007f3b19b0a0d8 in __GI_abort () at abort.c:89
#2  0x00007f3b19b43394 in __libc_message (do_abort=do_abort@entry=1, fmt=fmt@entry=0x7f3b19c51b28 "*** Error in `%s': %s: 0x%s ***\n")
    at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007f3b19b4e0f7 in malloc_printerr (action=<optimized out>, str=0x7f3b19c4dd1d "realloc(): invalid next size", ptr=<optimized out>) at malloc.c:4996
#4  0x00007f3b19b51937 in _int_realloc (av=<optimized out>, oldp=0x110be50, oldsize=<optimized out>, nb=<optimized out>) at malloc.c:4234
#5  0x00007f3b19b52fc9 in __GI___libc_realloc (oldmem=0x110be60, bytes=160) at malloc.c:3029
#6  0x00007f3b171d19fd in expand_array () at src/class/pmix_object.c:190
#7  0x00007f3b171d198d in save_class (cls=0x7f3b1743c940 <pmix_error_reg_info_t_class>) at src/class/pmix_object.c:177
#8  0x00007f3b171d18ca in pmix_class_initialize (cls=0x7f3b1743c940 <pmix_error_reg_info_t_class>) at src/class/pmix_object.c:145
#9  0x00007f3b171e2baf in pmix_obj_new (cls=0x7f3b1743c940 <pmix_error_reg_info_t_class>) at ./src/class/pmix_object.h:471
#10 0x00007f3b171e2a40 in pmix_obj_new_debug (type=0x7f3b1743c940 <pmix_error_reg_info_t_class>, file=0x7f3b1722bb42 "_id", line=350) at ./src/class/pmix_object.h:251
[...]

PR #219, opened 2016-11-22
Disable the errhandler support as it is broken

rhc54:
will be restored in subsequent release series


rhc54:
Fixes #76

artpol84:
@karasevb please check this branch in the environment where you were observing problems.

artpol84:
@rhc54 For me it looks good, but let's give it a try at runtime as well.

rhc54:
Just to be clear: the code referenced here is never called in Open MPI v2.0 or v2.x series branches. So this should only be an issue when testing PMIx itself.

karasevb:
I checked, smoky test is now working well :+1:

artpol84:
Thank you, @karasevb .

PR #222, opened 2016-11-24
Add network support APIs

rhc54:
Add a new network support framework so that the RM can:

* precondition an application (e.g., adding a security token to the app's environment)

* setup the local network driver to support an application (e.g., for "instant on" address resolution)

* pass directives in the environment of client procs prior to forking

* cleanup after each child terminates

* cleanup after all local children for a given application have terminated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,7 @@ examples/dmodex
 examples/dynamic
 examples/fault
 examples/pub
+examples/server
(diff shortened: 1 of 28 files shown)


rhc54:
Refs https://github.com/pmix/RFCs/pull/12

rhc54:
@afriedle-intel

PR #223, opened 2016-11-28
Perf: Added memory measurement for the perf tool

karasevb:
Added measuring of memory usage by the ["proportional set size" (PSS)](https://www.kernel.org/doc/Documentation/filesystems/proc.txt) (from `/proc/<pid>/smaps`)

```
The "proportional set size" (PSS) of a process is the count of pages it has
in memory, where each page is divided by the number of processes sharing it.
So if a process has 1000 pages all to itself, and 1000 shared with one other
process, its PSS will be 1500.
Note that even a page which is part of a MAP_SHARED mapping, but has only
a single pte mapped, i.e.  is currently used by only one process, is accounted
as private and not as shared.
```

### DIFF
--- a/contrib/perf_tools/pmi_intra_perf.c
+++ b/contrib/perf_tools/pmi_intra_perf.c
@@ -13,6 +13,8 @@
 #include <unistd.h>
 #include <time.h>
 #include <getopt.h>
+#include <limits.h>
+#include <string.h>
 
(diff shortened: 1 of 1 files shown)


karasevb:
@artpol84 Please review

artpol84:
@karasevb well done!

PR #226, opened 2016-12-01
dstor: fix PMIx_Get

karasevb:
- Fix get some data from the internal hash in dstor case.
- Silence a warning

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -560,6 +560,20 @@ static void _getnbfn(int fd, short flags, void *cbdata)
         return;
     }
 
+#if defined(PMIX_ENABLE_DSTORE) && (PMIX_ENABLE_DSTORE == 1)
+    if (PMIX_SUCCESS == (rc = pmix_hash_fetch(&nptr->internal, cb->rank, cb->key, &val))) {
+        /* found it - we are in an event, so we can
+         * just execute the callback */
+        cb->value_cbfunc(rc, val, cb->cbdata);
+        /* cleanup */
+        if (NULL != val) {
+            PMIX_VALUE_RELEASE(val);
+        }
+        PMIX_RELEASE(cb);
+        return;
+    }
+#endif
+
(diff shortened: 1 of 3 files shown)


src/client/pmix_client_get.c
```diff
@@ -564,23 +564,24 @@ static void _getnbfn(int fd, short flags, void *cbdata)
...
+        cb->value_cbfunc(rc, val, cb->cbdata);
+        /* cleanup */
+        if (NULL != val) {
+            PMIX_VALUE_RELEASE(val);
+        }
+        PMIX_RELEASE(cb);
+        return;
     }
 
     /* otherwise, the data must be something they "put" */
```
artpol84: As far as I recall the the key name that caused particular error for OMPI was "pmix.loc". But what if the key with some arbitrary name was submitted through PMIx_Store_local?

src/common/pmix_jobdata.c
```diff
@@ -136,6 +136,9 @@ static inline pmix_status_t _job_data_store(const char *nspace, void *cbdata)
     int32_t cnt;
     size_t nnodes;
     uint32_t i;
+#if !(defined(PMIX_ENABLE_DSTORE) && (PMIX_ENABLE_DSTORE == 1))
+    uint32_t j;
+#endif
```
artpol84: Why this variable was introduced but no usages are visible?

src/client/pmix_client_get.c
artpol84: I mean that we probably need to search "internal" hash in case key doesn't begin's from "pmix" too.

artpol84:
:)

artpol84:
@rhc54 - my sister sent it to me today, so I was unable to resist - it was fitting to well ;)

karasevb:
bot:retest

artpol84:
@karasevb :+1:

jladd-mlnx:
😳

Issue #204, opened 2016-10-28
Compress large PMIx values


hjelmn:
It might be beneficial to compress large PMIx strings before transferring or storing them. This could lead to a large decrease in the memory footprint.

hjelmn:
Here is a quick and dirty implementation that works:

```diff --git a/opal/mca/pmix/pmix3x/pmix/src/buffer_ops/pack.c b/opal/mca/pmix/pmix3x/pmix/src/buffer_ops/pack.c
index 2543eec..0c009e9 100644
--- a/opal/mca/pmix/pmix3x/pmix/src/buffer_ops/pack.c
+++ b/opal/mca/pmix/pmix3x/pmix/src/buffer_ops/pack.c
@@ -35,6 +35,7 @@
 #include "src/util/error.h"
 #include "src/util/output.h"
 #include "src/buffer_ops/internal.h"
+#include <zlib.h>
 
 pmix_status_t pmix_bfrop_pack(pmix_buffer_t *buffer,
                               const void *src, int32_t num_vals,
@@ -300,26 +301,53 @@ pmix_status_t pmix_bfrop_pack_string(pmix_buffer_t *buffer, const void *src,
     pmix_status_t ret = PMIX_SUCCESS;
     int32_t i, len;
     char **ssrc = (char**) src;
-
     for (i = 0; i < num_vals; ++i) {
         if (NULL == ssrc[i]) {  /* got zero-length string/NULL pointer - store NULL */
-        len = 0;
-        if (PMIX_SUCCESS != (ret = pmix_bfrop_pack_int32(buffer, &len, 1, PMIX_INT32))) {
+         len = 0;
+         if (PMIX_SUCCESS != (ret = pmix_bfrop_pack_int32(buffer, &len, 1, PMIX_INT32))) {
             return ret;
-        }
-    } else {
-        len = (int32_t)strlen(ssrc[i]) + 1;
-        if (PMIX_SUCCESS != (ret = pmix_bfrop_pack_int32(buffer, &len, 1, PMIX_INT32))) {
+         }
+       } else {
+         z_stream strm;
+         char *tmp = NULL;
+         int32_t deflated_len = 0;
+
+         len = (int32_t)strlen(ssrc[i]) + 1;
+         if (PMIX_SUCCESS != (ret = pmix_bfrop_pack_int32(buffer, &len, 1, PMIX_INT32))) {
             return ret;
-        }
-        if (PMIX_SUCCESS != (ret =
[...]

PR #232, opened 2016-12-06
dstor: reduce storage space for the key name

karasevb:
Currently in dstor the key names are stored in the arrays of 512 bytes.
This causes memory loss because real key names usually less
than 16 bytes.This commit reduces the memory footprint by using
only requested amount memory size to store the key name.

Beside memory footprint, the key access performance was also
significantly improved on PPC (Power8).
Average amount of L1-dcache-load-misses was decreased by 30%,
amount of LLC-load-misses was decreased by 59%.

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -551,26 +551,28 @@ static void _getnbfn(int fd, short flags, void *cbdata)
 
     /* the requested data could be in the job-data table, so let's
      * just check there first.  */
+    if (0 == strncmp(cb->key, "pmix", 4)) {
 #if defined(PMIX_ENABLE_DSTORE) && (PMIX_ENABLE_DSTORE == 1)
-    if (PMIX_SUCCESS == (rc = pmix_dstore_fetch(nptr->nspace, PMIX_RANK_WILDCARD, cb->key, &val))) {
+        if (PMIX_SUCCESS == (rc = pmix_dstore_fetch(nptr->nspace, PMIX_RANK_WILDCARD, cb->key, &val))) {
 #else
-    if (PMIX_SUCCESS == (rc = pmix_hash_fetch(&nptr->internal, PMIX_RANK_WILDCARD, cb->key, &val))) {
+        if (PMIX_SUCCESS == (rc = pmix_hash_fetch(&nptr->internal, PMIX_RANK_WILDCARD, cb->key, &val))) {
 #endif
-        /* found it - we are in an event, so we can
-         * just execute the callback */
-        cb->value_cbfunc(rc, val, cb->cbdata);
-        /* cleanup */
-        if (NULL != val) {
-            PMIX_VALUE_RELEASE(val);
+            /* found it - we are in an event, so we can
+             * just execute the callback */
+            cb->value_cbfunc(rc, val, cb->cbdata);
+            /* cleanup */
+            if (NULL != val) {
+                PMIX_VALUE_RELEASE(val);
+            }
+            PMIX_RELEASE(cb);
+            return;
         }
-        PMIX_RELEASE(cb);
-        return;
-    }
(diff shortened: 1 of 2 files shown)


karasevb:
Here is results of PMIx perf tool for PPC (Power8) for one node (160 process):
image

(Note:  in bottom line (`dstor + key fix`) is the results for this PR)

jjhursey:
@karasevb Does this fix also need to go into `master`?

I can test locally again, but this sounds like it might fix Issue #207 (ppc perf) and help with the overhead noted in Issue #204.

karasevb:
@jjhursey yes, sure. It should get into `master`

jjhursey:
I ran some tests against 3 ppc64le nodes today. Each node has 20 cores, 160 hwthreads.

I'm good with these improvements. I think there is more to do, but this PR is a step in the right direction.

**PMIx_Get**:
 - With dstore enabled: This PR reduces the `PMIx_Get` overhead by ~50% compared with the current `v1.2` branch
   * At 3 nodes with 160 PPN: 0.002158 (`v1.2`) vs 0.001097 (PR)
 - However, it is still 1.6x slower than if run without dstore (on this PR branch)
   * At 3 nodes with 160 PPN: 0.000669 (without) vs  0.001097 (with)
   * I think cache contention is still a problem here, but I need to dig in a bit deeper to see exactly where and what we might be able to do about it.

**Memory**:
 - dstore does improve the memory footprint on node - even without this PR.
   * At 3 nodes with 160 PPN: 2.6 GB (`v1.2` without) vs 962 MB (`v1.2` with)
 - This PR seems to save about 1-2% at 3 nodes with dstore enabled
   * At 3 nodes with 160 PPN: 962 MB (`v1.2` with) vs 949 MB (PR with)

Raw data at the gist below:
 * https://gist.github.com/jjhursey/568b3daf238b2edd9fb2c95ba9a50e4c

rhc54:
That's still a disturbingly large footprint...especially as you only have 480 procs in the entire job! We need to look into it and find a way to reduce it.

jladd-mlnx:
@karasevb Great improvement!

artpol84:
@rhc54 I think that Josh was using `pmix_perf` for the evaluation ( @jjhursey correct me if I am wrong).
And I assume ( @karasevb correct me if I am wrong) that the memory footprint is a **SUM** of all proc's shared memory consumptions: `962 MB / 480 procs = 2 MB/proc`
By default in `pmix_perf` each proc submits 10 keys of 400 bytes each, so it is `400 * 10 * 480 = 1,9 MB` total and this is only a key sizes, while you also have:
* key names
* job information
* procs are using shared memory not only for the dstore - this is OMPI!

So I guess that there is nothing bad with the numbers in case I haven't missed anything.

jjhursey:
@artpol84 Yes I was using the `pmix_perf` tool. I posted the full output from three run in the gist if you want to see the full sets of numbers.

artpol84:
@jjhursey thanks!

so yes, I am right - this was a total sum throughout all of the nodes:
```
mem:           loc 2011.00 rem 2023.67 min 2009.00 max 2075.00 total 971360.00 Kb
```

artpol84:
@karasevb we need to consider trying a multi-node setup. But anyway this is a good result. Espesially considering https://github.com/pmix/master/issues/144#issuecomment-265333180.

kawashima-fj:
I'll try evaluating the memory footprint of this PR on multi-node like #129 and #144 but it take some days because of machine resource issue.

rhc54:
@artpol84 @jjhursey Thanks for clarifying!! This sounds much more positive than I had interpreted. Let's get it into master and update OMPI so we can see the results on some large machines.

karasevb:
In part this PR corresponds to #236

jjhursey:
Discussed on the Teleconf this morning. Good to go.

PR #234, opened 2016-12-07
Update the header tests to support Open BSD

rhc54:


rhc54:
Verified by OMPI user

PR #236, opened 2016-12-12
dstor: reduce storage space for the key name

karasevb:
Currently in dstor the key names are stored in the arrays of 512 bytes.
This causes memory loss because real key names usually less
than 16 bytes.This commit reduces the memory footprint by using
only requested amount memory size to store the key name.


jjhursey:
Discussed on the Teleconf this morning. Good to go.

PR #237, opened 2016-12-14
class: correctly handle concurrent class initialization

ggouaillardet:


ggouaillardet:
@rhc54 i think this will solve issues such as https://mtt.open-mpi.org/index.php?do_redir=2381

can you please review this ?

this is basically a copy/paste from master, except i use a ```pthread_mutex_t``` so there is no need to backport all the ```opal_atomic_t``` stuff (that include a bunch of assembler)

PR #241, opened 2016-12-21
Further reduce memory footprint

### LINKED ISSUE #204: Compress large PMIx values

rhc54:
Add the ability to compress/decompress long strings per suggestion by @hjelmn. We can't do it exactly as he suggested as that approach only dealt with pack/unpack of buffers for communication. Our problem is more around the storage size, not the size of our messages. So check the length of strings that are about to be stored, and compress them if larger than a limit. Mark such strings with their decompressed size and a new PMIX_COMPRESSED_STRING data type so we know how to decompress them before delivery.

Also, remove the HWLOC support from PMIx. We really don't want to store the actual topology tree as (a) it is quite large on some architecures, and (b) the pointers in that tree require that we store a full copy of the tree in each client process. Instead, we will store the HWLOC topology xml string, properly compressed, and provide that back to the caller when requested. The caller will then have to "load" the topology tree from the xml string, free the string, and "destroy" the tree when they are done with it.

Fixes #204


rhc54:
@artpol84 @karasevb @kawashima-fj @jjhursey 

Please take a look at this change as it materially reduces the memory footprint on OMPI. I don't believe your perf tool will see any difference as I don't think it uses strings of sufficient length to trigger the compression.

I'll be posting an OMPI PR that includes this change and will provide the reference shortly. You can then test it over there to measure the impact both in memory footprint and in wireup time.

src/client/pmix_client.c
```diff
@@ -541,18 +548,104 @@ PMIX_EXPORT pmix_status_t PMIx_Abort(int flag, const char msg[],
...
+    strm.avail_out = len;
+    strm.next_out = tmp;
+
+    deflate (&strm, Z_FINISH);
+    deflateEnd (&strm);
+
+    /* allocate 4 bytes beyond the size reqd by zlib so we
+     * can pass the size of the uncompressed string to the
+     * decompress side */
+    outlen = len - strm.avail_out + 4;
```
jjhursey: Instead of using `4` here can we use `sizeof(uint32_t)`? I know it will always be the same, but that would make the code a bit more clear for future selves. (same comment for the other `4` values below)

rhc54: sure - can do

src/client/pmix_client_get.c
```diff
@@ -73,6 +77,63 @@ static void _getnb_cbfunc(struct pmix_peer_t *pr,
...
+    uint8_t *dest;
+    int32_t len2;
+    z_stream strm;
+    int rc;
+
+    /* set the default error answer */
+    *outstring = NULL;
+
+    /* the first 4 bytes contains the uncompressed size */
+    memcpy(&len2, inbytes, 4);
```
jjhursey: Same comment here. Change `4` to `sizeof(uint32_t)`

rhc54: ok, i got it in all spots, including pmix_jobdata.c

artpol84:
I want to look carefully on this an check performance penalty. Please don't quick merge.

rhc54:
It's why I asked you to do so 😄

src/client/pmix_client_get.c
jjhursey: Thanks!

src/client/pmix_client.c
```diff
@@ -541,18 +548,104 @@ PMIX_EXPORT pmix_status_t PMIx_Abort(int flag, const char msg[],
...
+    if (NULL == ptr) {
+        free(tmp);
+        return false;
+    }
+    *outbytes = ptr;
+    *nbytes = outlen;
+
+    /* fold the uncompressed length into the buffer */
+    memcpy(ptr, &inlen, sizeof(uint32_t));
+    ptr += 4;
```
jjhursey: `4` to `sizeof(unit32_t)` here too

src/common/pmix_jobdata.c
```diff
@@ -28,6 +31,73 @@
...
+    if (NULL == ptr) {
+        free(tmp);
+        return false;
+    }
+    *outbytes = ptr;
+    *nbytes = outlen;
+
+    /* fold the uncompressed length into the buffer */
+    memcpy(ptr, &inlen, sizeof(uint32_t));
+    ptr += 4;
```
jjhursey: `4` to `sizeof(unit32_t)` here too

artpol84:
@karasevb please evaluate the performance and memory footprint impact of this changes using `pmix_perf`. We will need to sync today about the details.

src/server/pmix_server.c
```diff
@@ -335,6 +336,14 @@ static void _register_nspace(int sd, short args, void *cbdata)
                             "pmix:server _register_nspace recording %s",
                             cd->info[i].key);
 
+        if (0 == strcmp(cd->info[i].key, PMIX_REGISTER_NODATA)) {
```
artpol84: @rhc54 what is the purpose of the `PMIX_REGISTER_NODATA`

artpol84:
@rhc54 please, correct me if I am wrong but in 7dcd452f1f5f85e14d411cb2983aa20c986c33a5 we are solving basically 3 (or maybe more) independent things:
1. Remove HWLOC support.
2. Introduce compression facility
3. Introduce `PMIX_REGISTER_NODATA` info key.
4. Some code cleanups.

Can we have separate commits for each of them?

rhc54:
No, I'm afraid I'm not going to break things up into non-usable chunks.

The ```PMIX_REGISTER_NODATA``` key is used to avoid having the daemon put job-level info on itself into the dstore. There is nobody on the node who can access that info, so it just eats up memory for no purpose. ORTE has been modified to use it.

src/buffer_ops/copy.c
```diff
@@ -259,6 +259,12 @@ bool pmix_value_cmp(pmix_value_t *p, pmix_value_t *p1)
         case PMIX_STRING:
             rc = strcmp(p->data.string, p1->data.string);
             break;
+        case PMIX_COMPRESSED_STRING:
+            if (p->data.bo.size != p1->data.bo.size) {
```
rhc54: it isn't - but then that function is never called, and so I didn't bother to waste the time.

rhc54:
@karasevb Please be aware that the new code will not be triggered unless a string value (not key) is > 512 bytes in length. If you are just pushing small strings, the test will not exercise any of the new code paths.

artpol84:
@rhc54, why non-usable? Those changes seem to be self-contained and independent.

artpol84:
We know about 512 threshold.

artpol84:
@rhc54,
"No, I'm afraid I'm not going to break things up into non-usable chunks": we can do this and open another PR. 
I believe this is important to have a git-way commits as they are more bisect, cherry pick and review friendly.

src/buffer_ops/copy.c
artpol84: Does it makes sense to completely remove this function then?

rhc54:
No, please do not do that. I realize you have an opinion on this matter, but so do others, as was discussed at Tues meeting. This PR is a functional unit, and shall be committed as such. Thank you.

artpol84:
HWLOC support removal seems completely orthogonal to compression introduction.
Can you please explain why this is so important to have a large commit?

src/buffer_ops/copy.c
artpol84: My concern is that we will most probably forget about this conversation when/if we will decide to use it.
Doing byte comparison may save us some debugging cycles.

artpol84: My concern is that we will most probably forget about this conversation when/if we will decide to use it.
Doing byte comparison may save us some debugging cycles.

artpol84:
@jjhursey @rhc54, by the way, I recall we discussed that on future systems hwloc info is expected to be large. Do we know how large it is expected to be?

I'm asking because hwloc is may be the right place to solve this problem. Otherwise we will still end up unpacking hwloc data on each rank and doing something with it. And if it is huge and all ranks will be doing this step simultaneously - we will still have `n x hwloc_size` amount of data for the certain period of time (during initialization).

rhc54:
Please see https://github.com/open-mpi/ompi/pull/2629 for an explanation of why this may not be a problem.

 I had a long chat with Brice about making the topology tree itself a shared memory object, but it is a very difficult problem as the tree contains pointers between the leaves that are process-specific. Making the tree a shared memory object therefore involves rewriting most of HWLOC's functional APIs so the caller can pass in required info allowing that operation to properly traverse the tree for the calling process, and requires that each process retain the required information so it can be provided when calling an HWLOC function.

Such a major code change is unlikely to happen for some time, and would almost certainly require that someone fund it. Therefore, it certainly won't happen in time to help Trinity or the CORAL machines.

karasevb:
@rhc54 pmix perf tool has some problems with `master`. We need to fix it for the evaluation of the performance and memory footprint of this PR.

artpol84:
This will close https://github.com/pmix/master/issues/224 in parallel.

rhc54:
@karasevb No worries - thanks for taking a look at it. We need to wait until the OMPI evaluations are done as well anyway, and that may not happen until after the holidays over here.

artpol84:
Thank you, I read open-mpi/ompi#2629 and would like to clarify some statements.

It is said there  
```... This PR represents a first cut at solving the memory footprint problem being seen at scale on Trinity ... ```
Does that mean that dstore itself doesn't help on Trinity? Was it evaluated? Can you share the conclusions related to this topic?

We have done some measurements related to hwloc size on POWER8 and it turns to be ~85K for the node with 20 cores and 160 hw threads.  With the dstore we store hwloc string only once per node. Given that can you comment on the following statement:
```
on complex architectures such as KNL and Power9. When multiplied by 100s 
of processes on a node, the aggregated memory footprint can reach into the gigabyte range.
```
We figured out (it is quite roughly though) that each core (**not thread**) in hwloc topology takes ~4.2-4.6K (as we scaled on the nodes that we have) . So my estimation of the memory consumption on the very large node with 1024 cores will be not gigabytes but 10s of megabytes max. This sounds fine given that such a big nodes will likely have enough memory.

As it is said in open-mpi/ompi#2629 at some stage in MPI_Init all procs will want to grab this data which will temporally increase memory consumption. But this will happen with this PR too.

Also I have concerns about 1 proc-per-hw-thread case: running cpu-intensive deflate simultaneously on all hw threads during MPI_Init may take time. But we will measure this once we will fix `pmix_perf`.

**Comment by rhc54** [comment]
```
Does that mean that dstore itself doesn't help on Trinity? Was it evaluated?
Can you share the conclusions related to this topic?
```

Dstore doesn't address the HWLOC topology tree problem as PMIx only stores the XML topology string, not the topology tree. Each process takes the XML topology string and creates its own instance of the tree. This sits in the process' memory for its entire execution, even though it is only accessed (if at all) during init. When expanded, the tree occupies over 1MByte/process, as @hjelmn has noted on several occasions.

```
As it is said in open-mpi/ompi#2629 at some stage in MPI_Init all procs will want to
grab this data which will temporally increase memory consumption. But this will
happen with this PR too.
```

I noted that in the initial comments, which is why I stated this was an initial cut at the problem. Before spending more time, I want to see if the final footprint works for Trinity. If so, then there are some remaining changes (noted in the comments above) that will eliminate the need for the application procs to ever instantiate the tree - which would be an optimal solution.

**Comment by kawashima-fj** [comment]
@rhc54 I evaluated the memory footprint on Fujitsu PRIMEHPC FX10 (16 cores). On 16 procs x 64 nodes configuration, as you expected, there is no significant difference. In job data, only two data are compression target (> 512B string) on each node. (size: 10152B -> 887B)

**Comment by rhc54** [comment]
@kawashima-fj Thank you for testing it. I suspect the primary memory savings are going to come from changes in the OMPI area, with perhaps some minor corresponding adjustments to PMIx. Still, this exercise did identify some savings worth keeping. Once we complete the OMPI work, I'll incorporate the parts that still make sense and we can look at this PR again.

**Comment by karasevb** [comment]
We tested on 2 nodes 28ppn with modified pmix_perf. The modified in follow: each key of test contains value equal pmix.ltopo. And we had found a significant memory consumption with this PR. We investigated the problem, and had found a mem leak which fixed in PR [#6](https://github.com/rhc54/pmix-master/pull/6)
Results for 10 keys (modified `pmix_perf` on 2x28ppn):

image

**Comment by hjelmn** [comment]
Seeing a SEGV when using open-mpi/ompi#2629 when running with more than one node.

Last output from the crashed orted:

```
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:1860:_get_data_region_by_offset
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2119:pmix_sm_store: for rank 1, replace flag 1 skip pmix.rank key, look for MPI_THREAD_LEVEL key
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2119:pmix_sm_store: for rank 1, replace flag 1 skip pmix.lpeers key, look for MPI_THREAD_LEVEL key
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2119:pmix_sm_store: for rank 1, replace flag 1 skip pmix.lcpus key, look for MPI_THREAD_LEVEL key
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2119:pmix_sm_store: for rank 1, replace flag 1 skip pmix.appnum key, look for MPI_THREAD_LEVEL key
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2119:pmix_sm_store: for rank 1, replace flag 1 skip pmix.aldr key, look for MPI_THREAD_LEVEL key
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2119:pmix_sm_store: for rank 1, replace flag 1 skip pmix.grank key, look for MPI_THREAD_LEVEL key
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2119:pmix_sm_store: for rank 1, replace flag 1 skip pmix.apprank key, look for MPI_THREAD_LEVEL key
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2119:pmix_sm_store: for rank 1, replace flag 1 skip pmix.lrank key, look for MPI_THREAD_LEVEL key
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2119:pmix_sm_store: for rank 1, replace flag 1 skip pmix.nrank key, look for MPI_THREAD_LEVEL key
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2119:pmix_sm_store: for rank 1, replace flag 1 skip INVALIDATED key, look for MPI_THREAD_LEVEL key
[nid00193:51239] ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2119:pmix_sm_store: for rank 1, replace flag 1 skip        key, look for MPI_THREAD_LEVEL key
[nid00193:51239] *** Process received signal ***
[nid00193:51239] Signal: Segmentation fault (11)
[nid00193:51239] Signal code:  (128)
[nid00193:51239] Failing at address: (nil)
[nid00193:51239] [ 0] /lib64/libpthread.so.0(+0xf870)[0x2aaaadc34870]
[nid00193:51239] [ 1] /lib64/libc.so.6(+0x1251ec)[0x2aaaadf671ec]
[nid00193:51239] [ 2] /lustre/ttscratch1/hjelmn/build/ompi/pmix_test/lib/libpmix.so.0(+0xb6bc3)[0x2aaaac244bc3]
[nid00193:51239] [ 3] /lustre/ttscratch1/hjelmn/build/ompi/pmix_test/lib/libpmix.so.0(+0xb76bb)[0x2aaaac2456bb]
[nid00193:51239] [ 4] /lustre/ttscratch1/hjelmn/build/ompi/pmix_test/lib/libpmix.so.0(+0xb23e0)[0x2aaaac2403e0]
[nid00193:51239] [ 5] /lustre/ttscratch1/hjelmn/build/ompi/pmix_test/lib/libpmix.so.0(pmix_dstore_store+0x45)[0x2aaaac23d649]
[nid00193:51239] [ 6] /lustre/ttscratch1/hjelmn/build/ompi/pmix_test/lib/libpmix.so.0(pmix_server_commit+0x2ed)[0x2aaaac202ad0]
[nid00193:51239] [ 7] /lustre/ttscratch1/hjelmn/build/ompi/pmix_test/lib/libpmix.so.0(+0x71edb)[0x2aaaac1ffedb]
[nid00193:51239] [ 8] /lustre/ttscratch1/hjelmn/build/ompi/pmix_test/lib/libpmix.so.0(+0x73d2a)[0x2aaaac201d2a]
[nid00193:51239] [ 9] /lustre/ttscratch1/hjelmn/build/ompi/pmix_test/lib/libpmix.so.0(pmix_ptl_base_process_msg+0x238)[0x2aaaac260848]
[nid00193:51239] [10] /lustre/ttscratch1/hjelmn/build/ompi/pmix_test/lib/libopen-pal.so.0(opal_libevent2022_event_base_loop+0xe51)[0x2aaaab1837f1]
[nid00193:51239] [11] /lustre/ttscratch1/hjelmn/build/ompi/pmix_test/lib/libpmix.so.0(+0xb9ae8)[0x2aaaac247ae8]
[nid00193:51239] [12] /lib64/libpthread.so.0(+0x80a4)[0x2aaaadc2d0a4]
[nid00193:51239] [13] /lib64/libc.so.6(clone+0x6d)[0x2aaaadf2702d]
```

Backtrace:

```
#0  0x00002aaaadf671ec in __strncmp_ssse3 () from /lib64/libc.so.6
Missing separate debuginfos, use: zypper install glibc-debuginfo-2.19-22.16.2.x86_64 libelf1-debuginfo-0.158-6.1.x86_64 libexpat1-debuginfo-2.1.0-17.1.x86_64 libglib-2_0-0-debuginfo-2.38.2-5.12.x86_64 libgthread-2_0-0-debuginfo-2.38.2-5.12.x86_64 libltdl7-debuginfo-2.4.2-14.60.x86_64 libnuma1-debuginfo-2.0.9-6.1.x86_64 libpciaccess0-debuginfo-0.13.2-3.59.x86_64 libpcre1-debuginfo-8.33-3.314.x86_64 libudev1-debuginfo-210-70.51.1.x86_64 libz1-debuginfo-1.2.8-5.1.x86_64 unixODBC-debuginfo-2.3.1-4.95.x86_64
(gdb) bt
#0  0x00002aaaadf671ec in __strncmp_ssse3 () from /lib64/libc.so.6
#1  0x00002aaaac244bc3 in pmix_sm_store (ns_info=0x7d5eb0, rank=1, kval=0x2aaab8048da0, rinfo=0x2aaaafa20530, data_exist=1)
    at ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2065
#2  0x00002aaaac2456bb in _store_data_for_rank (ns_info=0x7d5eb0, rank=1, buf=0x2aaaafa20790)
    at ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:2230
#3  0x00002aaaac2403e0 in _esh_store (nspace=0x2aaab800c408 "928710657", rank=1, kv=0x2aaab803ddf0)
    at ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_esh.c:890
#4  0x00002aaaac23d649 in pmix_dstore_store (nspace=0x2aaab800c408 "928710657", rank=1, kv=0x2aaab803ddf0)
    at ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/dstore/pmix_dstore.c:66
#5  0x00002aaaac202ad0 in pmix_server_commit (peer=0x2aaab80458c0, buf=0x2aaaafa20d50)
    at ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/server/pmix_server_ops.c:189
#6  0x00002aaaac1ffedb in server_switchyard (peer=0x2aaab80458c0, tag=3, buf=0x2aaaafa20d50)
    at ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/server/pmix_server.c:2035
#7  0x00002aaaac201d2a in server_message_handler (pr=0x2aaab80458c0, hdr=0x2aaab803e540, buf=0x2aaaafa20d50, cbdata=0x0)
    at ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/server/pmix_server.c:2192
#8  0x00002aaaac260848 in pmix_ptl_base_process_msg (fd=-1, flags=4, cbdata=0x2aaab803e460)
    at ../../../../../../../../../opal/mca/pmix/pmix2x/pmix/src/mca/ptl/base/ptl_base_sendrecv.c:593
#9  0x00002aaaab1837f1 in event_process_active_single_queue (activeq=0x684570, base=0x7d5090) at ../../../../../../opal/mca/event/libevent2022/libevent/event.c:1370
#10 event_process_active (base=<optimized out>) at ../../../../../../opal/mca/event/libevent2022/libevent/event.c:1440
#11 opal_libevent2022_event_base_loop (base=0x7d5090, flags=1) at ../../../../../../opal/mca/event/libevent2022/libevent/event.c:1644
#12 0x00002aaaac247ae8 in progress_engine (obj=0x7d4b28) at ../../../../../../../opal/mca/pmix/pmix2x/pmix/src/runtime/pmix_progress_threads.c:149
#13 0x00002aaaadc2d0a4 in start_thread () from /lib64/libpthread.so.0
#14 0x00002aaaadf2702d in clone () from /lib64/libc.so.6
```

artpol84:
@hjelmn
Does this the case for the current master? Or only with the PR.

hjelmn:
@artpol84 I was able to run ompi master with external pmix master with no issue. Not sure if that is a good comparison though since this pmix change is not compatible yet with the external pmix component.

hjelmn:
One thing I see is the first key after the INVALIDATED key is blank. Should that ever be the case?

artpol84:
No, it shouldn't can you track what keys was submitted previously?

hjelmn:
Huh, when you overwrite a key with INVALIDATED what happens if the key is shorter than the string INVALIDATED? Won't the size get messed up?

artpol84:
This may be the issue. We recently add the flexible key name size. We will check that.

artpol84:
Btw, what is your command line? I wonder why keys get overwritten. It shouldn't be the case if native PMIx is used.

hjelmn:
@artpol84 See above. Looks like pmix.hname. That is shorter than INVALIDATED.

hjelmn:
Run line ```mpirun -n 2 -N 1 ring_c```

artpol84:
So you answered my question with this output. Host name is submitted twice which is not what I would expect.
So we have 2 possible issues here:
1. Incorrect key overwrite 
2. Double hostname submission.

artpol84:
@hjelmn, yes I think you are right. We will take care of this.
The reason why we missed this is that key overwrite wasn't the case before this PR.

hjelmn:
@artpol84 Ok. Let me know when you have a fix and I will test again.

karasevb:
@hjelmn Thank you for your analysis.
We have fixed it in PR #245

karasevb:
We have evaluated  the effect of compression on highly threaded Power8 nodes. We don't see significant overhead. And with recent memory leak fix all looks good. 
One thing I'd like to suggest: lets keep compress/decompress functionality in the separate file somewhere in utils for example. This will make it easier to maintain this functionality.
:+1:

rhc54:
@karasevb thank you for checking this out. I like your suggestion and have refactored the code as suggested.

PR #243, opened 2016-12-22
contrib/perf: Fix pmix_perf for master

artpol84:


artpol84:
Address https://github.com/pmix/master/issues/224.

PR #255, opened 2017-01-10
dstore: Fixed dstor bug case of key overwrite.

karasevb:
Data corruption was observed when a key was stored second time (forcing
data overwrite). In case of overwrite, the old record is invalidated
with "INVALIDATED" string written into its key name field. The problem
was observed when space insufficient for storing "INVALIDATED" string
was reserved for the key name field at the original key submission.
The fix provides minor refactoring of the dtore record formats to
flexibly address this issue.

(cherry picked from commit f6808239e91f4308eb93d9e0562f00d234e81c12)

Conflicts:
	src/dstore/pmix_esh.c

### DIFF
--- a/src/dstore/pmix_esh.c
+++ b/src/dstore/pmix_esh.c
@@ -61,62 +61,61 @@ pmix_dstore_base_module_t pmix_dstore_esh_module = {
 #define ESH_ENV_NS_DATA_SEG_SIZE    "NS_DATA_SEG_SIZE"
 #define ESH_ENV_LINEAR              "SM_USE_LINEAR_SEARCH"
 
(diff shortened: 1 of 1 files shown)


karasevb:
This PR has been synchronized with `master` and breaks binary compatibility with v1.2. Can restore compatibility with v1.2, this issue should be discussed on PMIx telecon.

artpol84:
@karasevb need to address this as per telecon discussion.

artpol84:
:+1:

artpol84:
@jjhursey good to go.
We should merge this one and #262 and give this codebase a thorough MTT testing.

PR #261, opened 2017-01-21
dstore: Fix significant performance degradation.

artpol84:
Accidentally loop that was submitting jobinfo data was duplicated
in two nested functions. As the result computational complexity
rised from O(n) to O(n^2) which caused significant performance
hit at 8k node level (thanks to @hjelmn for noticing and doing
initial analysis!)
This bug wasn't visible at dstore data level because if for the key
being replaced the size of the data was the same - key was simply
rewritten.

### DIFF
--- a/src/common/pmix_jobdata.c
+++ b/src/common/pmix_jobdata.c
@@ -372,12 +372,9 @@ static inline pmix_status_t _job_data_store(const char *nspace, void *cbdata)
 
 #if defined(PMIX_ENABLE_DSTORE) && (PMIX_ENABLE_DSTORE == 1)
     if (NULL != cb->dstore_fn) {
-        uint32_t size = (uint32_t)pmix_value_array_get_size(cb->bufs);
-        for (i = 0; i < size; i++) {
-            if (PMIX_SUCCESS != (rc = _rank_key_dstore_store(cbdata))) {
-                PMIX_ERROR_LOG(rc);
-                goto exit;
-            }
+        if (PMIX_SUCCESS != (rc = _rank_key_dstore_store(cbdata))) {
+            PMIX_ERROR_LOG(rc);
+            goto exit;
         }
     }
 #endif


artpol84:
Intercepts #260.

artpol84:
@rhc54 @hjelmn I haven't fully tested this, but I'm quite sure that 8000^2 stores versus 8000 would look exactly like @hjelmn reported.

artpol84:
@karasevb Please test with OMPI ASAP!

hjelmn:
OUCH. Yeah. That would cause a problem. @ 4096 nodes with 68 ppn this bug takes it from 278k stores to 77b stores!

hjelmn:
4k, 1 ppm launch with this on top of the no-lock and remove extra pack/unpack patches took the launch down to 1:40 from 8+.

hjelmn:
8k launch, 1 ppn went from "i give up" to 3m. There is still some work to be done but this was a big one. Thanks!

artpol84:
@hjelmn can you test this patch alone without lock removal and buffer optimization?
I believe they are not do any harm anymore. The reason was in the fact that we were calling them too much.

artpol84:
As I recall 10k of flock/unlock cycles was taking ~ 10 ms so I am sure that locking is not a bottleneck anymore. With square complexity (which is fixed by this PR) it would grow up to 100s which is close to 60s you were observing with 8k locks/unlocks.
So I believe we don't need to remove locks anymore.
As I wrote in #260 we will switch to pthread locking later but this shouldn't be an urgent task now.

artpol84:
So for clarity.
You were observing 60s overhead of fcntl. We know that there were `~8000x8000` store operations.
Now with this PR we reduced number of operations to `8000` so I expect fcntl overhead to be
`60s / 8000 =  0.0075s = 7.5 ms` 
which should be OK.

artpol84:
The same applies to the buffer pack unpack operations. I expect this overhead to decrease down to
```
50s / 8000 = 0.00625s
```

artpol84:
We will carefully check your pack/unpack optimization to make sure that it doesn't break anything.
It seems correct to me, but since you were testing only `/bin/true` you may not seen some sort of problems that may appear when keys are actually being fetched.

PR #263, opened 2017-01-23
sm: use posix_fallocate() before mmap'ing it

ggouaillardet:
if the filesystem is full, both ftruncate() and mmap() success, but
a crash occurs when accessing the mmap'ed memory.
unlike ftruncate(), posix_fallocate() does fail when the filesystem
is full

### DIFF
--- a/src/sm/pmix_mmap.c
+++ b/src/sm/pmix_mmap.c
@@ -1,6 +1,8 @@
 /*
  * Copyright (c) 2015-2016 Mellanox Technologies, Inc.
  *                         All rights reserved.
+ * Copyright (c) 2017      Research Organization for Information Science
+ *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -16,6 +18,7 @@
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <errno.h>
 
 #include <src/include/pmix_config.h>
 #include <pmix_common.h>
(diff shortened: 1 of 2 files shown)


ggouaillardet:
@artpol84 this is related to open-mpi/ompi#2737

artpol84:
@ggouaillardet we found that lseek(SEEK_SET) + lseek(SEEK_END) works well for that purpose too.

artpol84:
And I was gonna use it instead ftruncate.
It should be less changes.

artpol84:
Any advantages of posig_fallocate ?

ggouaillardet:
@artpol84
```posix_fallocate()``` does allocate the file, and that is what we want.
```lseek()``` works too, but i cannot tell whether this is per the specs or whether this is because of the current implementation by a given linux kernel.
bottom line, ```posix_fallocate()``` looks like the right thing to do, and if it is not available, then i suggest we fallback to ```lseek()``` instead of ```ftruncate()```

**Comment by artpol84** [comment]
Ok, I agree that it fits well.
@karasevb please verify that this PR works in our environment and then it is good to be merged.

**Comment by ggouaillardet** [comment]
@artpol84 note this PR fallback to ftruncate() abd it is known not doing what we need.
I can update it tomorrow, or feel free to issue a new one

**Comment by artpol84** [comment]
@ggouaillardet I synced with @karasevb today. It seems like I misunderstood him. He came up with the same solution as in this PR and lseek cannot be used to solve this.
@karasevb will review and merge this PR later today.
@ggouaillardet thank you very much for this fix.

File: src/sm/pmix_mmap.c [hunk 1]
```diff
@@ -58,12 +60,21 @@ int _mmap_segment_create(pmix_sm_seg_t *sm_seg, const char *file_name, size_t si
         goto out;
     }
     /* size backing file - note the use of real_size here */
+#ifdef HAVE_POSIX_FALLOCATE
+    if (0 != posix_fallocate(sm_seg->seg_id, 0, size)) {
+        pmix_output_verbose(2, pmix_globals.debug_output,
+                "sys call posix_fallocate(2) fail\n");
+        rc = PMIX_ERROR;
```
karasevb: I suggest to check the `errno` value for clarity.
Calling `posix_fallocate` returns `ENOSPC` in  the case where the storage space ended. So can will check it and return the PMIx like error code: `PMIX_ERR_OUT_OF_RESOURCE`
This allows to see the root cause of failure.

It can look like:
```C
switch( errno ) {
    case ENOSPC:
        rc = PMIX_ERR_OUT_OF_RESOURCE;
        break;
    default:
        rc = PMIX_ERROR;
        break;
}
```

ggouaillardet: @karasevb sounds good to me, i will update the PR accordingly

karasevb:
:+1:

jjhursey:
@ggouaillardet Please PR to `v1.2` branch. Thanks

PR #268, opened 2017-01-25
dstore: Introduce session supports

karasevb:
Defined the key name to set session id


artpol84:
:+1:

PR #273, opened 2017-01-27
dstore: Added locking by pthread.

karasevb:
Added alternative locking by `pthread` with use shared memory region
for the lock. The code path of old locking by `flock` has been saved
for evaluate performance against locking by `pthread` and other purposes.

To change the locking type needs to enable the code paths by macro:
```C  
#define PTHREAD_LOCK_ENABLE  1
#define FCNTL_LOCK_ENABLE    1
```
In priority uses the locking by `pthread`

Thanks to @artpol84 for this proof of concept.
- The  results of locking by `pthread` here: https://github.com/pmix/master/pull/260#issuecomment-273945190
- Code base of concept: https://github.com/artpol84/poc/tree/master/benchmarks/shmem_locking

### DIFF
--- a/src/dstore/pmix_esh.h
+++ b/src/dstore/pmix_esh.h
@@ -26,21 +26,12 @@ BEGIN_C_DECLS
 
 #define PMIX_DSTORE_ESH_BASE_PATH "PMIX_DSTORE_ESH_BASE_PATH"
 
-#define PTHREAD_LOCK_ENABLE 1
-#define FCNTL_LOCK_ENABLE   0
-
-#if defined(PTHREAD_LOCK_ENABLE) || defined(FCNTL_LOCK_ENABLE)
-#if (PTHREAD_LOCK_ENABLE == 1)
-#undef FCNTL_LOCK_ENABLE
-#define PTHREAD_LOCK
+#ifdef HAVE_PTHREAD_SHARED
+#define ESH_PTHREAD_LOCK
+#elif defined HAVE_FCNTL_FLOCK
+#define ESH_FCNTL_LOCK
 #else
-#undef PTHREAD_LOCK_ENABLE
-#define FCNTL_LOCK
-#endif
-#elif !(defined(PTHREAD_LOCK_ENABLE) || defined(FCNTL_LOCK_ENABLE))
-#error not enabled any one type of locking
-#else
-#define PTHREAD_LOCK
+#error No locking mechanism was found
 #endif
 
 /* this structs are used to store information about
@@ -66,14 +57,13 @@ struct seg_desc_t {
 typedef struct ns_map_data_s ns_map_data_t;
 typedef struct session_s session_t;
 typedef struct ns_map_s ns_map_t;
-typedef struct rwlock_map_s rwlock_map_t;
 
 struct session_s {
     int in_use;
     uid_t jobuid;
     char *nspace_path;
     char *lockfile;
-#ifdef PTHREAD_LOCK
+#ifdef ESH_PTHREAD_LOCK
     pmix_sm_seg_t *rwlock_seg;
     pthread_rwlock_t *rwlock;
 #endif
--- a/src/dstore/pmix_esh.h
+++ b/src/dstore/pmix_esh.h
@@ -26,6 +26,23 @@ BEGIN_C_DECLS
 
 #define PMIX_DSTORE_ESH_BASE_PATH "PMIX_DSTORE_ESH_BASE_PATH"
 
+#define PTHREAD_LOCK_ENABLE 1
+#define FCNTL_LOCK_ENABLE   0
+
+#if defined(PTHREAD_LOCK_ENABLE) || defined(FCNTL_LOCK_ENABLE)
+#if (PTHREAD_LOCK_ENABLE == 1)
+#undef FCNTL_LOCK_ENABLE
+#define PTHREAD_LOCK
+#else
+#undef PTHREAD_LOCK_ENABLE
+#define FCNTL_LOCK
+#endif
+#elif !(defined(PTHREAD_LOCK_ENABLE) || defined(FCNTL_LOCK_ENABLE))
+#error not enabled any one type of locking
+#else
+#define PTHREAD_LOCK
+#endif
+
 /* this structs are used to store information about
  * shared segments addresses locally at each process,
  * so they are common for different types of segments
@@ -49,12 +66,17 @@ struct seg_desc_t {
 typedef struct ns_map_data_s ns_map_data_t;
 typedef struct session_s session_t;
 typedef struct ns_map_s ns_map_t;
+typedef struct rwlock_map_s rwlock_map_t;
 
 struct session_s {
     int in_use;
     uid_t jobuid;
     char *nspace_path;
     char *lockfile;
+#ifdef PTHREAD_LOCK
+    pmix_sm_seg_t *rwlock_seg;
+    pthread_rwlock_t *rwlock;
+#endif
     int lockfd;
     seg_desc_t *sm_seg_first;
     seg_desc_t *sm_seg_last;
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -35,6 +35,7 @@ EXTRA_DIST = \
 	pmix_check_package.m4 \
 	pmix_check_vendor.m4 \
 	pmix_check_visibility.m4 \
+	pmix_check_lock.m4 \
 	pmix_config_subdir.m4 \
 	pmix_ensure_contains_optflags.m4 \
 	pmix_functions.m4 \
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -655,6 +655,14 @@ AC_DEFUN([PMIX_SETUP_CORE],[
 
     PMIX_MCA
 
+    ##################################
+    # Dstore Locking
+    ##################################
+
+    pmix_show_title "Dstore Locking"
+
+    PMIX_CHECK_DSTOR_LOCK
+
     ############################################################################
     # final compiler config
     ############################################################################
@@ -867,6 +875,7 @@ AC_DEFINE_UNQUOTED([PMIX_WANT_PRETTY_PRINT_STACKTRACE],
                    [$WANT_PRETTY_PRINT_STACKTRACE],
                    [if want pretty-print stack trace feature])
 
+#
 # Do we want the shared memory datastore usage?
 #
 
@@ -886,7 +895,22 @@ AC_DEFINE_UNQUOTED([PMIX_ENABLE_DSTORE],
                  [if want shared memory dstore feature])
 
 #
+# Use pthread-based locking
+#
+DSTORE_PTHREAD_LOCK="1"
+AC_MSG_CHECKING([if want dstore pthread-based locking])
+AC_ARG_ENABLE([dstore-pthlck],
+              [AC_HELP_STRING([--disable-dstore-pthlck],
+                              [Disable pthread-based lockig in dstor (default: enabled)])])
+if test "$enable_dstore_pthlck" == "no" ; then
+    AC_MSG_RESULT([no])
+    DSTORE_PTHREAD_LOCK="0"
+else
+    AC_MSG_RESULT([yes])
+    DSTORE_PTHREAD_LOCK="1"
+fi
 
+#
 # Ident string
 #
 AC_MSG_CHECKING([if want ident string])
--- a/config/pmix_check_lock.m4
+++ b/config/pmix_check_lock.m4
@@ -1,0 +1,57 @@
+dnl -*- shell-script -*-
+dnl
+dnl Copyright (c) 2017      Mellanox Technologies, Inc.
+dnl                         All rights reserved.
+dnl $COPYRIGHT$
+dnl
+dnl Additional copyrights may follow
+dnl
+dnl $HEADER$
+dnl
+
+AC_DEFUN([PMIX_CHECK_DSTOR_LOCK],[
+    orig_libs=$LIBS
+    LIBS="-lpthread $LIBS"
+
+    _x_ac_pthread_lock_found="0"
+    _x_ac_fcntl_lock_found="0"
+
+    AC_CHECK_MEMBERS([struct flock.l_type],
+    [
+        AC_DEFINE([HAVE_FCNTL_FLOCK], [1],
+        [Define to 1 if you have the locking by fcntl.])
+        _x_ac_fcntl_lock_found="1"
+    ], [], [#include <fcntl.h>])
+
+    if test "$ESH_PTHREAD_LOCK" == "1"; then
+        AC_CHECK_FUNC([pthread_rwlockattr_setkind_np],
+            [AC_EGREP_HEADER([PTHREAD_RWLOCK_PREFER_WRITER_NONRECURSIVE_NP],
+                    [pthread.h],[
+                        AC_DEFINE([HAVE_PTHREAD_SETKIND], [1],
+                            [Define to 1 if you have the `pthread_rwlockattr_setkind_np` function.])])])
+
+        AC_CHECK_FUNC([pthread_rwlockattr_setpshared],
+            [AC_EGREP_HEADER([PTHREAD_PROCESS_SHARED],
+                    [pthread.h],[
+                        AC_DEFINE([HAVE_PTHREAD_SHARED], [1],
+                            [Define to 1 if you have the `PTHREAD_PROCESS_SHARED` definition.
+                        ])
+                        _x_ac_pthread_lock_found="1"
+            ])
+        ])
+
+        if test "$_x_ac_pthread_lock_found" == "0"; then
+            if test "$_x_ac_fcntl_lock_found" == "1"; then
+                AC_MSG_WARN([dstore: pthread-based locking not found, will use fcntl-based locking.])
+            else
+                AC_MSG_ERROR([dstore: no available locking mechanisms was found. Can not continue. Try disabling dstore])
+            fi
+        fi
+    else
+        if test "$_x_ac_fcntl_lock_found" == "0"; then
+            AC_MSG_ERROR([dstore: no available locking mechanisms was found. Can not continue. Try disabling dstore])
+        fi
+    fi
+
+    LIBS="$orig_libs"
+])
--- a/src/dstore/pmix_esh.c
+++ b/src/dstore/pmix_esh.c
@@ -15,7 +15,6 @@
 #include <stdio.h>
 #include <sys/types.h>
 #include <sys/stat.h>
-#include <fcntl.h>
 #include <sys/file.h>
 #include <dirent.h>
 #include <errno.h>
@@ -36,6 +35,14 @@
 
 #include "pmix_dstore.h"
 #include "pmix_esh.h"
+
+#ifdef FCNTL_LOCK
+#include <fcntl.h>
+#endif
+
+#ifdef PTHREAD_LOCK
+#include <pthread.h>
+#endif
 
 static int _esh_init(pmix_info_t info[], size_t ninfo);
 static int _esh_finalize(void);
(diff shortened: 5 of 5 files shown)


karasevb:
The  results of locking by `pthread` here:  https://github.com/pmix/master/pull/260#issuecomment-273945190
Code base of concept: https://github.com/artpol84/poc/tree/master/benchmarks/shmem_locking

src/dstore/pmix_esh.h
```diff
@@ -26,6 +26,23 @@ BEGIN_C_DECLS
 
 #define PMIX_DSTORE_ESH_BASE_PATH "PMIX_DSTORE_ESH_BASE_PATH"
 
+#define PTHREAD_LOCK_ENABLE 1
+#define FCNTL_LOCK_ENABLE   1
```
jjhursey: Could we make these `configure` options? Is there any reason someone would want to choose the `flock` version over the `pthread_lock`? Or should we just remove the `flock` code?

karasevb:
Results on the one node PowerPC 160 ppn (the average of 10 times):
image

artpol84:
@karasevb we need to do the analysis - init time should go down as well as fence

artpol84:
But get time looks really good.
Also need comparison with messaging

src/dstore/pmix_esh.h
artpol84: We preserve it for evaluation reasons and we want to keep it.

artpol84: Not intended to be enabled other than through the source code modification

artpol84: Not intended to be enabled other than through the source code modification

artpol84: For clarity need to change like this:

+#define PTHREAD_LOCK_ENABLE 1
+#define FCNTL_LOCK_ENABLE   0

artpol84:
Sounds like flock contention was rising a high overhead on PPC when hwthreads was used. This makes sense.

karasevb:
@artpol84 I checked a raw data of test, and one of all runs have a bad result, which decreases the average result. 

Retest:
image

artpol84:
Oh yeah!!

artpol84:
This looks really good!
@jjhursey: can you evaluate on your end?

artpol84:
I was playing with my POC code on PPC node and found potential problem. 
Short story: we need to set rwlock attribute that will give writer priority like this:
https://github.com/artpol84/poc/blob/master/benchmarks/shmem_locking/pthread.c#L20

If we won't do that - the default behavior is that reader can jump over writer sitting in the queue if read lock was already acquired. Here is a simple 3-proc example:
```
1. reader1 grabs the lock
2. writer tries to get the lock and blocks
3. reader2 grabs the lock and succeeds (as per default behavior)
4. reader1 releases the lock, but it is still held by reader2 so writer is postponed.
5. reader1 grabs the lock again and reader 2 releases it slightly later.
n. reader(i={1,2}) grabs the lock again and reader(i={2,1}) releases it slightly later.
```
in this scenario writer will never get it's lock. And this I was observing for 160 ppn runs for both flocks and pthread locks.

With the attribute mentioned above the scenario will be correct:
```
1. reader1 grabs the lock
2. writer tries to get the lock and blocks
3. reader2 grabs the lock and BLOCKS
4. reader1 releases the lock, and WRITER GETS it while reader2 will be waiting in the queue.
```

artpol84:
Update: it looks like it may not be a Linux-specific feature. I verified that it works ok on Linux but need to check portability.

artpol84:
According to a [manpage](http://man7.org/linux/man-pages/man3/pthread_rwlockattr_setkind_np.3.html) this is a non-portable feature.
So as I understand we need to test that this function is available and use it only in this case. Otherwise we should rely on the default thread/lock arbitration.

artpol84:
I'll provide some POC results shortly. 
Here are some more facts/thoughts:
* flocks have the same behavior
* we will face this problem only with direct modex
* even with direct modex readers will eventually give up the database, but there is a possibility that they will hold server from posting for a while.

artpol84:
It took me longer than I expected to collect the data.
I've modified my POC code, now it consists from 3 tests:
* Write-only (corresponds to full modex, data publish phase). Writer does 10K locks/unlocks and locking overhead is measured.
* Read-only (corresponds to full modex, data fetch phase). Multiple readers grab the lock and hold it for 10us. Locking overhead is measured.
* Read/write (corresponds to direct-modex). Writer needs to do 100 locks/unlocks sleeping `usleep(n_readers)` between them. Readers getting/releasing the locks, sleeping 10us with acquired lock. Writer lock overhead and average number of reader locks is counted.

Out of curiosity I implemented mutex-based locking which expected to be inefficient to ensure that it is the case.
From the locking mechanisms that we currently considering only rwlocks with prioritization flag was able to pass rd/wr test, for others I had to disable it. Results can be reproduced - POC github repo is up to date, all you may need is to fix MPI environment in `launch.sh`.

I short rwlocks + prioritization indeed provide the best performance among considered standard approaches. It is also most stable. Next I plan to experiment with some home-brewed schemes, but I believe that this approach is fine for `v1.2.1`.

Solution proposed by this PR + prioritization flag:
```
$ ./eval.sh core test_pthread_prio
readers wronly ovh      wronly (lck/s)  rdonly ovh      rdonly (lck/s)  rdwr (wr ovh)   rdwr (wr time)  rdwr (rlck/s)
9       0.002245        4453833.764859  0.004736        2111666.095603  0.001501        0.002419        58244.661927
19      0.002256        4433537.968561  0.151155        66157.464182    0.008125        0.009910        24780.164966
39      0.002848        3510939.989200  0.461968        21646.518492    0.023351        0.027481        8290.030086
79      0.003140        3185124.159977  1.105906        9042.360481     0.057069        0.066495        4159.607720
159     0.004928        2029051.944125  2.449973        4081.677344     0.134938        0.154491        1681.639634

$ ./eval.sh node test_pthread_prio
readers wronly ovh      wronly (lck/s)  rdonly ovh      rdonly (lck/s)  rdwr (wr ovh)   rdwr (wr time)  rdwr (rlck/s)
9       0.004837        2067495.851619  0.004941        2023910.519258  0.003398        0.005824        31688.395135
19      0.004611        2168861.355277  0.008732        1145221.727154  0.004427        0.009180        32938.292694
39      0.004619        2165104.895468  0.110454        90535.819445    0.011304        0.020955        21322.759302
79      0.004953        2018908.263870  0.660087        15149.526774    0.034060        0.047328        7272.478590
159     0.004861        2057317.787078  2.313170        4323.072422     0.150697        0.184318        1586.163716
```

Flocks (current solution):
```
$ ./eval.sh core test_flock --no-rdwr
readers wronly ovh      wronly (lck/s)  rdonly ovh      rdonly (lck/s)
9       0.008786        1138140.969968  0.014929        669844.242673
19      0.008823        1133457.583398  0.328270        30462.739260
39      0.010403        961283.523591   0.997085        10029.233507
79      0.015053        664298.492203   3.669395        2725.245059
159     0.036128        276797.423113   10.991259       909.813921

$ ./eval.sh node test_flock --no-rdwr
readers wronly ovh      wronly (lck/s)  rdonly ovh      rdonly (lck/s)
9       0.034872        286759.198522   0.023834        419576.950088
19      0.034084        293391.528265   0.042408        235805.835804
39      0.034143        292883.517742   0.422463        23670.686926
79      0.036254        275828.470993   1.717725        5821.654907
159     0.036481        274112.271459   12.046849       830.092564
```

Mutex-based locking
```
$ ./eval.sh node test_pthread_mutex --no-rdwr
readers wronly ovh      wronly (lck/s)  rdonly ovh      rdonly (lck/s)
9       0.003590        2785775.776357  0.969098        10318.877714
19      0.003492        2863386.360497  1.352956        7391.224309
39      0.003567        2803805.067281  3.230695        3095.309347
79      0.003456        2893606.822469  5.766820        1734.057902
159     0.003471        2880708.513474  4.709114        2123.541530

$ ./eval.sh core test_pthread_mutex --no-rdwr
readers wronly ovh      wronly (lck/s)  rdonly ovh      rdonly (lck/s)
9       0.001433        6980181.814400  0.436735        22897.199350
19      0.001436        6963340.120007  0.978426        10220.495777
39      0.001892        5285279.659259  2.194046        4557.790349
79      0.002260        4424978.473526  4.857809        2058.541020
159     0.003633        2752179.618542  12.238805       817.073248
```

artpol84:
After posting initial data I decided to try to see how fair rlocks acquisition is. POC was again slightly modified:
* I changed layout to be more compact (unfortunately field names are not that informative now, but they are the same)
* I'm collecting average, minimum and maximum numbers related to readers to see the fairness of locks access.

Here is what I've got (I'm providing 2 runs for each):

pthread_prio (unfortunately for read/write test fairness is bad, still it is better than others (who just hang) but again it is worth exploring self-made prioritization schemes, I'll provide some more details in my next comment):
```
$ ./eval.sh core test_pthread_prio
readers WO:ovh  WO:lk/s RO:ovh[avg/min/max]     RO:lk/s RW:wovh RW:wtm  RW:rlk/s[avg/min/max]
9       2e-03   4375398 8e-03/7e-03/1e-02       1228750 1e-03   1e-02   180483/155510/184594
19      2e-03   4402305 2e-01/2e-01/2e-01       65024   8e-03   3e-02   46267/41212/51282
39      3e-03   3732487 4e-01/4e-01/5e-01       22930   4e-02   7e-02   16485/203/19269
79      3e-03   3147193 1e+00/8e-01/1e+00       9085    7e-02   2e-01   8277/110/9777
159     5e-03   2110050 2e+00/2e+00/3e+00       4019    2e-01   4e-01   3309/397/4210

$ ./eval.sh core test_pthread_prio
readers WO:ovh  WO:lk/s RO:ovh[avg/min/max]     RO:lk/s RW:wovh RW:wtm  RW:rlk/s[avg/min/max]
9       2e-03   4372750 8e-03/8e-03/1e-02       1179569 1e-03   1e-02   181609/155269/185867
19      2e-03   4365056 2e-01/1e-01/2e-01       65178   8e-03   3e-02   47228/43121/52133
39      3e-03   3614057 4e-01/4e-01/5e-01       22633   3e-02   7e-02   18240/252/21003
79      3e-03   3170776 1e+00/9e-01/1e+00       9377    7e-02   2e-01   8401/175/10046
159     5e-03   2092159 3e+00/2e+00/3e+00       3923    2e-01   3e-01   3261/540/4445
```

Flocks:
```
$ ./eval.sh core test_flock --no-rdwr
readers WO:ovh  WO:lk/s RO:ovh[avg/min/max]     RO:lk/s
9       9e-03   1135921 2e-02/2e-02/2e-02       520479
19      9e-03   1121346 3e-01/3e-01/3e-01       30381
39      1e-02   949068  1e+00/7e-01/1e+00       10332
79      2e-02   664128  4e+00/3e+00/4e+00       2796
159     4e-02   285646  1e+01/9e+00/1e+01       849

$ ./eval.sh core test_flock --no-rdwr
readers WO:ovh  WO:lk/s RO:ovh[avg/min/max]     RO:lk/s
9       9e-03   1114055 2e-02/2e-02/3e-02       521200
19      9e-03   1131675 3e-01/3e-01/3e-01       30139
39      1e-02   954283  1e+00/7e-01/1e+00       10282
79      1e-02   666868  4e+00/3e+00/4e+00       2783
159     4e-02   282386  1e+01/1e+01/1e+01       849
```

Mutexes (you can see the because of non-FIFO nature of mutexes some procs finish in milliseconds while others take seconds to finish):
```
$ ./eval.sh core test_pthread_mutex --no-rdwr
readers WO:ovh  WO:lk/s RO:ovh[avg/min/max]     RO:lk/s
9       1e-03   6953354 4e-01/1e-02/8e-01       23953
19      1e-03   6957445 1e+00/5e-02/2e+00       10121
39      2e-03   5089733 2e+00/2e-01/4e+00       4340
79      2e-03   4367266 5e+00/1e-01/9e+00       2038
159     3e-03   2923498 1e+01/3e-02/3e+01       765

$ ./eval.sh core test_pthread_mutex --no-rdwr
readers WO:ovh  WO:lk/s RO:ovh[avg/min/max]     RO:lk/s
9       1e-03   6950475 4e-01/1e-02/8e-01       23615
19      1e-03   6945197 1e+00/2e-02/2e+00       10126
39      2e-03   5093924 2e+00/3e-02/4e+00       4376
79      2e-03   4381769 5e+00/2e-01/9e+00       1999
159     4e-03   2820461 1e+01/2e-02/3e+01       782
```

artpol84:
With slight modification of the read/write test (specifically read proc doesn't sleep while holding the lock) I can run mutex-based implementation. And here are the results:

**pthread_prio**
```
$ ./eval.sh core test_pthread_prio
readers WO:ovh  WO:lk/s RO:ovh[avg/min/max]     RO:lk/s RW:wovh RW:wtm  RW:rlk/s[avg/min/max]
9       2e-03   4400471 8e-03/7e-03/1e-02       1228542 2e-03   1e-02   231020/222912/244567
19      2e-03   4435459 2e-01/2e-01/2e-01       63152   9e-03   3e-02   95813/78172/114010
39      3e-03   3637676 5e-01/4e-01/6e-01       22195   6e-02   1e-01   29179/882/39419
79      3e-03   3167262 1e+00/9e-01/1e+00       9208    6e-02   1e-01   20522/771/26393
159     5e-03   2110582 2e+00/2e+00/3e+00       4048    2e-01   3e-01   8587/478/11994
```
**mutex**
```
$ ./eval.sh core test_pthread_mutex
readers WO:ovh  WO:lk/s RO:ovh[avg/min/max]     RO:lk/s RW:wovh RW:wtm  RW:rlk/s[avg/min/max]
9       1e-03   6967908 4e-01/1e-02/8e-01       24102   1e-03   1e-02   246918/200050/272740
19      1e-03   6941087 1e+00/6e-02/2e+00       10294   4e-03   2e-02   145818/101635/176544
39      2e-03   5244931 2e+00/2e-01/4e+00       4305    1e-02   5e-02   69396/665/89309
79      2e-03   4349556 5e+00/2e-01/9e+00       1981    3e-02   1e-01   33357/35/45475
159     3e-03   2878602 1e+01/3e-02/3e+01       785     7e-02   3e-01   11195/545/16000
```

Mutexes demonstrate better performance. Although we cannon use plain mutexes because readers **will** hold the lock for a while, this example demonstrates that if mutex is not holded during the whole access but instead used to arbitrate access to resource-busy flag  we can possibly improve the performance.

artpol84:
On portability:
I see [here](http://stackoverflow.com/questions/2782883/condition-variable-in-shared-memory-is-this-code-posix-conformant) and [here](http://alesteska.blogspot.com/2012/08/pthreadprocessshared-not-supported-on.html) that on MAC OS `PTHREAD_PROCESS_SHARED` may not be supported. At least this was the case in 2012. 
@rhc54 can you check that? I don't have MAC.
@karasevb before enabling pthread-based locking we need to check if `PTHREAD_PROCESS_SHARED` is supported. And if not - we may stick to `flock` implementation.

rhc54:
On my Mac:

```
/usr/include/pthread/pthread.h
156:#define PTHREAD_PROCESS_SHARED         1

```

artpol84:
But this is POSIX definition, in my understanding this doesn't mean it's supported.
Can you try to run my POC code there?
https://github.com/artpol84/poc/tree/master/benchmarks/shmem_locking

check [launch.sh](https://github.com/artpol84/poc/blob/master/benchmarks/shmem_locking/launch.sh) to see how to launch it. You will need to replace `module load hpcx-gcc` with env setting for your OMPI installation.

artpol84:
for `make` you also need to make sure that your `mpicc` is visible by setting `PATH` accordingly.

rhc54:
I'm afraid that code doesn't compile nor run on a Mac. I don't have time right now to correct it, but can try to do something later.

Meantime, I think the best bet is to detect availability of support for shared mutex and use it if found. Otherwise, fall back to something else. I don't understand your tables, so I can't comment on the next-best option.

bosilca:
With the following patch the code compiles (conditionally using PTHREAD_RWLOCK_PREFER_WRITER_NONRECURSIVE_NP) and mostly runs on OSX. Mostly, because the launch script enforce a binding policy, something that is not supported on OSX.

[osx.patch.txt](https://github.com/pmix/master/files/737946/osx.patch.txt)

artpol84:
Thank you @bosilca, I will incorporate your changes. This confirms that PTHREAD_RWLOCK_PREFER_WRITER_NONRECURSIVE_NP is not supported by non-linux systems as I expected.

@rhc54 if you interested in results interpretation, please read the description above. Here is explanation of the table header:
* `RO` - read-only test, only readers access segment with read locks.
* `WO` - write-only test, only writer access segment.
* `RW` - read-write test, both writer and reader access data, algorithm was provided above.
* `ovh` - locking overhead, measured time spent locking and unlocking.
* `wovh` - same as `ovh` but writer only in RW test.
* `lk/s` - locks per second according to actual number of lock/unlock cycles and the time it took to get them.
* `rlk/s` - reader locks per second.
* `wtm` - overall time that writer spent in RW test, this time includes `wovh` + processing and sleep time. I'm using this to calculate ` RW:rlk/s`
* `[avg/min/max]` means that this column represents corresponding statistics observed amongst a set of readers.

artpol84:
@bosilca, just out of curiosity:

can you check if this will help: `_XOPEN_SOURCE >= 500 || _POSIX_C_SOURCE >= 200809L`. According to [manpage](http://man7.org/linux/man-pages/man3/pthread_rwlockattr_setkind_np.3.html) those macro's should be on to enable the feature.
I doubt they are, but as I've said - out of curiosity.

bosilca:
PTHREAD_RWLOCK_PREFER_WRITER_NONRECURSIVE_NP is nowhere to be found on any on the system header files. I checked on /usr/include and /Applications/Xcode.app/Contents/Developer.

artpol84:
ok, thank you!

artpol84:
@rhc54 @jjhursey 
Configuration logic to discover available locking mechanisms under consideration was added to this PR
We still do some fine tuning like adjusting warnings etc. but in general it's ready

artpol84:
It will be ready for discussion for tomorrow's call

rhc54:
From OMPI master, with this update:

```
+-------------+-----------------+-------------+----------+------+------+----------+------+--------------------------------------------------------------------------+
| Phase       | Section         | MPI Version | Duration | Pass | Fail | Time out | Skip | Detailed report                                                          |
+-------------+-----------------+-------------+----------+------+------+----------+------+--------------------------------------------------------------------------+
| MPI Install | my installation | 3.0.0a1     | 00:01    | 1    |      |          |      | MPI_Install-my_installation-my_installation-3.0.0a1-my_installation.html |
| Test Build  | trivial         | 3.0.0a1     | 00:02    | 1    |      |          |      | Test_Build-trivial-my_installation-3.0.0a1-my_installation.html          |
| Test Build  | ibm             | 3.0.0a1     | 00:43    | 1    |      |          |      | Test_Build-ibm-my_installation-3.0.0a1-my_installation.html              |
| Test Build  | intel           | 3.0.0a1     | 01:44    | 1    |      |          |      | Test_Build-intel-my_installation-3.0.0a1-my_installation.html            |
| Test Build  | java            | 3.0.0a1     | 00:01    | 1    |      |          |      | Test_Build-java-my_installation-3.0.0a1-my_installation.html             |
| Test Build  | orte            | 3.0.0a1     | 00:02    | 1    |      |          |      | Test_Build-orte-my_installation-3.0.0a1-my_installation.html             |
| Test Run    | trivial         | 3.0.0a1     | 00:13    | 8    |      |          |      | Test_Run-trivial-my_installation-3.0.0a1-my_installation.html            |
| Test Run    | ibm             | 3.0.0a1     | 11:52    | 493  |      |          |      | Test_Run-ibm-my_installation-3.0.0a1-my_installation.html                |
| Test Run    | spawn           | 3.0.0a1     | 00:14    | 6    | 1    |          |      | Test_Run-spawn-my_installation-3.0.0a1-my_installation.html              |
| Test Run    | loopspawn       | 3.0.0a1     | 04:58    | 1    |      |          |      | Test_Run-loopspawn-my_installation-3.0.0a1-my_installation.html          |
| Test Run    | intel           | 3.0.0a1     | 19:22    | 474  |      |          | 4    | Test_Run-intel-my_installation-3.0.0a1-my_installation.html              |
| Test Run    | intel_skip      | 3.0.0a1     | 15:02    | 431  |      |          | 47   | Test_Run-intel_skip-my_installation-3.0.0a1-my_installation.html         |
| Test Run    | java            | 3.0.0a1     | 00:01    | 1    |      |          |      | Test_Run-java-my_installation-3.0.0a1-my_installation.html               |
| Test Run    | orte            | 3.0.0a1     | 00:43    | 19   |      |          |      | Test_Run-orte-my_installation-3.0.0a1-my_installation.html               |
+-------------+-----------------+-------------+----------+------+------+----------+------+--------------------------------------------------------------------------+
```

artpol84:
And how does it look like without this PR?
Does one spawn test fail as well?

rhc54:
yes - that has been around awhile. I just need to look at the details

PR #278, opened 2017-02-02
Fixed PMIx_Get

karasevb:
The `PMIx_Get` is hanging if key was stored in local hash and
key name not equal `pmix.*`. Bug reproduced with `--disable-dstore`.
This PR is fixes the get internal keys in non dstore case.

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -596,8 +596,16 @@ static void _getnbfn(int fd, short flags, void *cbdata)
         return;
     }
 
+    /* should be in the internal hash table. */
 #if defined(PMIX_ENABLE_DSTORE) && (PMIX_ENABLE_DSTORE == 1)
+    rc = pmix_hash_fetch(&nptr->internal, cb->rank, cb->key, &val);
+    if (PMIX_SUCCESS != rc) {
+        rc = pmix_dstore_fetch(cb->nspace, cb->rank, cb->key, &val);
+    }
+    if (PMIX_SUCCESS == rc) {
+#else
     if (PMIX_SUCCESS == (rc = pmix_hash_fetch(&nptr->internal, cb->rank, cb->key, &val))) {
+#endif
         /* if this is a compressed string, then uncompress it */
         if (PMIX_COMPRESSED_STRING == val->type) {
             pmix_util_uncompress_string(&tmp, (uint8_t*)val->data.bo.bytes, val->data.bo.size);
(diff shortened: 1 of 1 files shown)


rhc54:
@karasevb I'm a little puzzled as we are not seeing any such hangs in OMPI, and OMPI uses the current master. Can you please describe how you are encountering this issue?

karasevb:
@rhc54 
I made mistake in PR description. This bug reproduces with **dstore disabled** only, if key stored to local hash (using `PMIx_Store_internal`) and **key name not equal `pmix.*`**. So get the key name != `pmix.*` in internal hash it is impossible for current `PMIx_Get` implementation.

I'm not sure, that this bug can be reproduced in OMPI. This bug is observed with `./pmix_test --test-internal` in `master`.

karasevb:
I have updated the PR description.

rhc54:
Thanks! Appreciate the clarification, and the catch.

artpol84:
Bot:retest

artpol84:
Jenkins was updated with 2 new tests , one of them tests this case. Let's see ...

artpol84:
@rhc54 
I'm looking into v1.2 and master codebase and it seems that they are diverged. If the key starts from "pmix" we:
* give up do not request from server in [v1.2](https://github.com/pmix/master/blob/v1.2/src/client/pmix_client_get.c#L554:L576)
* request data from the server in [master](https://github.com/pmix/master/blob/master/src/client/pmix_client_get.c#L628:L660), this was introduced in fc65cbd148244dc754e952be787e642c29e43701.

Is this correct?

artpol84:
@rhc54 
regarding https://github.com/pmix/master/blob/master/src/client/pmix_client_get.c#L519
I searched sources briefly and I don'see any `PMIx_Get`'s with `key == NULL`. Can you remind why do we need this codepath?

artpol84:
@rhc54 
Could you also explain how `cb->checked` works in https://github.com/pmix/master/blob/master/src/client/pmix_client_get.c#L724

rhc54:
Let's see if I can answer these for you - my memory may be a little vague as some things evolved when @dsolt was working on the "unknown nspace" problem.

If a key begins with "pmix", then that is a reserved name and belongs to us. Keys that begin with that name are only installed by the RM and cannot be "put" by an application. We therefore do not go up to the server to retrieve them - if they are not found, then they have not been given.

The only exception to that rule is if the key is for a proc in a different nspace. In that case, the RM would not have provided it at startup, and we have to go up to the server and request a copy. The server will respond with the entire job-level data set for that nspace, from which we extract the piece that was requested. The rest goes into the client's storage (or is available in dstore).

As the comment states, the NULL keys come in thru the PMI-1/2 backward compatibility interfaces.

The ```cb->checked``` flag is used to prevent us from continually asking the server for data. As the comment explains, there is a race condition whereby we might ask for something just before the data arrives. The check allows us to ask twice, and then give up.

HTH

artpol84:
```
As the comment states, the NULL keys come in thru the PMI-1/2 backward compatibility interfaces.
```

But as I wrote above I don't see any calls with the zero key. In what case PMI1/PMI2 may request without specifying the key?

rhc54:
I honestly don't recall - I only know that's where it came from, and that's why the code protects it.

artpol84:
Can we find that out at some point?

rhc54:
Maybe? Someday? The protection is cheap, and it wasn't created for no reason, so we may as well leave it alone for now.

artpol84:
```
The cb->checked flag is used to prevent us from continually asking the server for data. As the comment explains, there is a race condition whereby we might ask for something just before the data arrives. The check allows us to ask twice, and then give up.
```

Still not clear - what do you mean by "continually asking"? When PMIx_Get is called multiple times? When rank runs multiple threads and all of them are asking?

Why we allow only 2 times? Why can't we proceed to request and add this to the pending list?

artpol84:
```
Maybe? Someday? The protection is cheap, and it wasn't created for no reason, so we may as well leave it alone for now.
```
This is a relatively big peace of code (compared to the rest of the function. And for me it looks confusing. I made the refactoring of the main codepath, but I have no idea how to change this portion.

rhc54:
If you don't understand that section of code, then I'd recommend that you not refactor this at all - we know it works, so let's leave it alone until we have time to fully consider the implications

artpol84:
This part is left intact in my refactoring. However I'm not comfortable having the code that no one understands.

rhc54:
Like I said - the original proposed changes looked fine. The refactoring should be deferred - things work, let's not disrupt them when we are trying to release things. There is no hurry to do this. We have many other things to worry about.

artpol84:
This is opened against master. I don't understand why are you refering release?

artpol84:
Get problem exists only in master.

rhc54:
Could you _please_ just table this for a couple of weeks? I'd like to take my vacation, and I really don't need another extended argument right now. Nor do I want to leave for a week only to see master broken by an unnecessary disruptive change. People are trying to use it!

I keep reiterating - the original fix is fine. Let's hold it there.

artpol84:
I'm not leaving and I'm capable of fixing problems - just let people know that I'm the point of contact.

artpol84:
I'm fine with deferring this as well, but on the other hand I'm ready to be responsible for the code that I've contributed, so I'm not sure what causes the problem now.

artpol84:
I'm merging this so jenkins will be happy again.

PR #285, opened 2017-02-03
Refactor PMIx_Get fixing `PMIX_store_internal` case

artpol84:

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -477,13 +477,14 @@ static void _getnbfn(int fd, short flags, void *cbdata)
     pmix_cb_t *cb = (pmix_cb_t*)cbdata;
     pmix_cb_t *cbret;
     pmix_buffer_t *msg;
-    pmix_value_t *val;
+    pmix_value_t *val = NULL;
     pmix_info_t *info, *iptr;
     pmix_pointer_array_t results;
     pmix_status_t rc;
     pmix_nspace_t *ns, *nptr;
     size_t n, nvals;
     char *tmp;
+    bool my_nspace = false, my_rank = false;
 
     pmix_output_verbose(2, pmix_globals.debug_output,
                         "pmix: getnbfn value for proc %s:%d key %s",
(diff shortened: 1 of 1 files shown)


artpol84:
@rhc54 @karasevb 
I failed to read existing `_getnbfn` and decided to refactor it to make the logic more straightforward.

artpol84:
Intended to replace https://github.com/pmix/master/pull/285

artpol84:
Agree.

artpol84:
@rhc54 however I was hoping that you can look through the refactored logic from you point of view to see if there is any unintended logic change.

artpol84:
It shouldn't be hard to do, however I feel that I may lack to understand the full picture.

rhc54:
See comments on other PR - I think the two are getting confused. What @karasevb proposed was simple and fine. It's clear the refactoring isn't well enough understood, so we probably should just defer that for now.

artpol84:
```
 It's clear the refactoring isn't well enough understood, so we probably should just defer that for now.
```
This is not correct. I understand the changed part. And the portion with key == 0 was left unchanged.

artpol84:
For the records: both PMI1 and PMI2 compat codes disallow key to be NULL:
* pmi1: https://github.com/pmix/master/blob/master/src/client/pmi1.c#L195:L197
* pmi2: https://github.com/pmix/master/blob/master/src/client/pmi2.c#L391

artpol84:
@karasevb thank you. What tests have you tried?
@rhc54 I haven't recalled when exactly you are going to vacation - I thought it's next week. If so - would be great if you can run your MTT against this branch to see if all works well.
As an alternative could you share the MTT script you are running so we can check independently?

karasevb:
@artpol84 I checked with dstor/msg, full/direct modex for ring, spawn

artpol84:
Thanks!

jjhursey:
@dsolt I know you were working on this section of code. Can you take a look at this change?

artpol84:
We ran our MTT against master and this PR and it doesn't introduce any degradation.

jjhursey:
I'm ok with the refactoring to consolidate the error path. That will help keep it from degrading.

There is a discussion about the change introduced here https://github.com/pmix/master/commit/58de093e78d9a8900650d556d760502d374606f2#diff-a644e6348df49913a3a3a5d5a2a46e25R393.
It would be good to document how this behavior is being used (`NULL` key and non-`PMIX_RANK_WILDCARD` rank). I'm not advocating to remove it, just to document it a bit more so we can make sure to support it in the future.

For now let's hold off on merging until @rhc54 is back from vacation.

rhc54:
Okay with me as well. I agree we need to further document the NULL key and non-wildcard rank code paths, but let's retain them for now and we'll work on that problem. I'm working on a branch with @hjelmn now that will likely need to use the NULL key code path, so I'll take the action item to document it.

Issue #279, opened 2017-02-02
SLURM plugin hangs


rhc54:
We have a report that the SLURM plugin is hanging job launch, even for non-MPI jobs. This is SLURM release v16.05.8, built against PMIx v1.2.0.

When launched, the slurmstepd for the job gets stuck in an infinite loop calling ```sched_yield```. This is reportedly easily reproduced with:

```shell
srun -N1 --mpi=pmix hostname
```

Can someone please take a look and see if v1.2.0 is causing the plugin to break?

artpol84:
We will check.
From my experience: I'm running the plugin with 1.2 all the time without issues.
Do you have any additional info? 
Like have they tried plugin before, was other versions working for them?
You've said it's sched yield- so they have backtrace?

rhc54:
I'm asking for more - this is from the OpenHPC guys. They are trying to update their SLURM version and add PMIx, but running into this problem. So they haven't tried anything before now.

They are building against the distro version of libevent, which is 2.0.21 and includes thread support, so that should be okay. They used strace to see the sched_yield calls - I've asked for a gdb backtrace.

rhc54:
Here you go:

```shell
(gdb) bt
#0  0x00002b8df0b2b057 in sched_yield () from /usr/lib64/libc.so.6
#1  0x00002b8df328e0b5 in pmixp_agent_start () from /usr/lib64/slurm/mpi_pmix.so
#2  0x00002b8df328d0d5 in p_mpi_hook_slurmstepd_prefork () from /usr/lib64/slurm/mpi_pmix.so
#3  0x000000000042e9ab in job_manager ()
#4  0x000000000042bc3a in main ()
(gdb) thread apply all bt full
 
Thread 6 (Thread 0x2b8df2576700 (LWP 39286)):
#0  0x00002b8df0b0d41d in nanosleep () from /usr/lib64/libc.so.6
No symbol table info available.
#1  0x00002b8df0b3e014 in usleep () from /usr/lib64/libc.so.6
No symbol table info available.
#2  0x00000000004ef1e8 in _timer_thread ()
No symbol table info available.
#3  0x00002b8df083bdc5 in start_thread () from /usr/lib64/libpthread.so.0
No symbol table info available.
#4  0x00002b8df0b461cd in clone () from /usr/lib64/libc.so.6
No symbol table info available.
 
Thread 5 (Thread 0x2b8df2677700 (LWP 39287)):
[...]

Issue #286, opened 2017-02-04
Backward compatibility with PMI1/PMI2


artpol84:
Original post by @JohnWestlund [here](https://github.com/pmix/master/issues/279#issuecomment-277184679) in issue #279.

The executable launched now -- but I'm still seeing some weird behavior (my Slurm-foo is rusty, so this could just be pilot error). I'm launching the following C code:
```
#include "mpi.h"
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <assert.h>

int main(int argc, char *argv[])
{
  int num_procs, num_local;
  char mach_name[MPI_MAX_PROCESSOR_NAME];
  int mach_len;

  MPI_Init (&argc,&argv);
  MPI_Comm_size (MPI_COMM_WORLD, &num_procs);
  MPI_Comm_rank (MPI_COMM_WORLD, &num_local);
  MPI_Get_processor_name(mach_name,&mach_len);

  /* compare avail MPI tasks to requested. If not requested via argv,
     convention is to assume 1 task */

  printf("num_procs: %d\nnum_local: %d\nargv1: %d\n\n", num_procs, num_local, atoi(argv[1]));

  if(argc > 1)
    assert(num_procs == atoi(argv[1]));
  else
    assert(num_procs == 1);

  /* Verify a quick collective */

  int local=1;
  int global;

  MPI_Allreduce(&local,&global,1,MPI_INT,MPI_SUM,MPI_COMM_WORLD);

  assert(global == num_procs);

  MPI_Finalize();
  return 0;

}
```

But `MPI_Init` keeps bailing on me (gnu or intel with impi):
```
-bash-4.2# cat /tmp/job.root.20186
#!/bin/bash
#SBATCH -J Intel_HPC_test
#SBATCH -N 1
#SBATCH -n 2
#SBATCH -t 1
#SBATCH -o job.%j.out
srun -n 2 --mpi=pmix ./C_test 2
#mpiexec.hydra -bootstrap slurm ./C_test 16
-bash-4.2# sbatch /tmp/job.root.20186
Submitted batch job 52
-bash-4.2# cat job.52.out
Fatal error in MPI_Init: Other MPI error, error stack:
MPIR_Init_thread(805).....: fail failed
MPID_Init(1783)...........: channel initialization failed
MPIDI_CH3_Init(126).......: fail failed
MPID_nem_init_ckpt(836)...: fail failed
[...]

PR #284, opened 2017-02-03
dstore: Fix extension slot allocation

artpol84:
Signed-off-by: Artem Polyakov <artpol84@gmail.com>
(cherry picked from commit d9fcd4b7ca70ce737f1579ade0cfef21c4b0716e)

Conflicts:
	src/dstore/pmix_esh.c

PR #287, opened 2017-02-05
Set of fixes targeting restoration of PMI1/PMI2 support

artpol84:
With set of changes I was able to can run [MPICH 3.2](http://www.mpich.org/static/downloads/3.2/mpich-3.2.tar.gz) with SLURM 16.05 (branch from github + the fix with "volatile" as in Issue #279 ).

### DIFF
--- a/src/client/pmi1.c
+++ b/src/client/pmi1.c
@@ -202,6 +202,30 @@ PMIX_EXPORT int PMI_KVS_Get( const char kvsname[], const char key[], char value[
     pmix_output_verbose(2, pmix_globals.debug_output,
             "PMI_KVS_Get: KVS=%s, key=%s value=%s", kvsname, key, value);
 
+    /* PMI-1 expects resource manager to set
+     * process mapping in ANL notation. */
+    if (!strcmp(key, ANL_MAPPING)) {
+        /* we are looking in the job-data. If there is nothing there
+         * we don't want to look in rank's data, thus set rank to widcard */
+        proc = myproc;
+        proc.rank = PMIX_RANK_WILDCARD;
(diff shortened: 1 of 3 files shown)


artpol84:
@jjhursey needs to be backported to master. I'll open PR shortly.

artpol84:
#286

jjhursey:
We'll wait until PR #286 is confirmed to fix the issue, and merged in. Then I'll merge in this PR.

artpol84:
@jjhursey you mean #284?

jjhursey:
Sorry. It's Issue #286 - The PR to port this back to `master` hasn't been created yet.

As soon as we get confirmation that this fixes that issue then I'm fine with merging it in. I'd also like to have the PR against `master` in place too so this doesn't get lost.

artpol84:
You are rigth, I'll create the PR asap.
Also - this PR solves the issue for me. So I guess we want it anyway. And there might be something on top for the #286.

artpol84:
To make it clear: this PR is intended to fix MPICH support in PMIx. Since IMPI is based on MPICH it is definitely needed for #286.
I agree that it would be great if @JohnWestlund or @gdeisher can double check me with MPICH to confirm that this works.
However I wasn't trying to fully solve #286 with this PR, some further effort from @JohnWestlund and @gdeisher may be needed to verify/additional fix.

artpol84:
@jjhursey I've created PR to the master. I can't check it with SLURM/pmix plugin as master is not yet supported. However the change is straightforward an I think we are fine here. Will try to check with some sample PMI tests.

jjhursey:
Per discussion on the teleconf. Merging this in.

PR #282, opened 2017-02-02
dstore: Fix extension slot allocation

artpol84:
Signed-off-by: Artem Polyakov <artpol84@gmail.com>

PR #289, opened 2017-02-07
dstore: fix errors in offset calculation

hjelmn:
This commit fixes two errors in the bounds checking in
put_data_to_the_end.

Signed-off-by: Nathan Hjelm <hjelmn@lanl.gov>

### DIFF
--- a/src/dstore/pmix_esh.c
+++ b/src/dstore/pmix_esh.c
@@ -1,9 +1,12 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2015-2016 Mellanox Technologies, Inc.
  *                         All rights reserved.
  * Copyright (c) 2016      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2016-2017 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2017      Los Alamos National Security, LLC. All rights
+ *                         reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -2107,9 +2110,8 @@ static int put_empty_ext_slot(seg_desc_t *dataseg)
 
 static size_t put_data_to_the_end(ns_track_elem_t *ns_info, seg_desc_t *dataseg, char *key, void *buffer, size_t size)
 {
-    size_t offset;
+    size_t offset, id = 0;
     seg_desc_t *tmp;
-    int id = 0;
     size_t global_offset, data_ended;
     uint8_t *addr;
 
@@ -2126,7 +2128,7 @@ static size_t put_data_to_the_end(ns_track_elem_t *ns_info, seg_desc_t *dataseg,
     offset = global_offset % _data_segment_size;
 
     /* We should provide additional space at the end of segment to place EXTENSION_SLOT to have an ability to enlarge data for this rank.*/
-    if (sizeof(size_t) + ESH_KEY_SIZE(key, size) + EXT_SLOT_SIZE(key) > _data_segment_size) {
+    if ((sizeof(size_t) + ESH_KEY_SIZE(key, size) + EXT_SLOT_SIZE(key)) > _data_segment_size) {
         /* this is an error case: segment is so small that cannot place evem a single key-value pair.
          * warn a user about it and fail. */
         offset = 0; /* offset cannot be 0 in normal case, so we use this value to indicate a problem. */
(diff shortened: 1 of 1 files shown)


hjelmn:
@artpol84 Hit the problem again @ 2k nodes. The error is pretty self-evident. Even with this commit there is still a segment overflow and I don't have time to dig into it now. Adding a small constant to the left hand side of the bounds check made it work. I am running with:

```
if ((offset + 8 + ESH_KEY_SIZE(key, size) + EXT_SLOT_SIZE()) > _data_segment_size)  {
```

So it looks like something in ESH_KEY_SIZE is wrong? I can look further tomorrow if the error hasn't been found.

src/dstore/pmix_esh.c
```diff
@@ -2126,15 +2128,15 @@ static size_t put_data_to_the_end(ns_track_elem_t *ns_info, seg_desc_t *dataseg,
...
+    if ((sizeof(size_t) + ESH_KEY_SIZE(key, size) + EXT_SLOT_SIZE(key)) > _data_segment_size) {
         /* this is an error case: segment is so small that cannot place evem a single key-value pair.
          * warn a user about it and fail. */
         offset = 0; /* offset cannot be 0 in normal case, so we use this value to indicate a problem. */
         pmix_output(0, "PLEASE set NS_DATA_SEG_SIZE to value which is larger when %lu.",
                 sizeof(size_t) + strlen(key) + 1 + sizeof(size_t) + size + EXT_SLOT_SIZE(key));
         return offset;
     }
-    if (offset + ESH_KEY_SIZE(key, size) + EXT_SLOT_SIZE(key) > _data_segment_size)  {
+    if ((offset + ESH_KEY_SIZE(key, size) + EXT_SLOT_SIZE(key)) > _data_segment_size)  {
```
artpol84: * ` EXT_SLOT_SIZE(key)` - to make sure: is this fixed in your version? should be like this:
https://github.com/pmix/master/pull/282/files#diff-ef98725cc50c8a568d8e5622d505e776R2137

src/dstore/pmix_esh.c
```diff
@@ -2107,9 +2110,8 @@ static int put_empty_ext_slot(seg_desc_t *dataseg)
 
 static size_t put_data_to_the_end(ns_track_elem_t *ns_info, seg_desc_t *dataseg, char *key, void *buffer, size_t size)
 {
-    size_t offset;
+    size_t offset, id = 0;
```
artpol84: so you say that integer was overflowed?
segment size if 4K by default. To overflow it you need to store 2G*4K which is unlikely I hope.

src/dstore/pmix_esh.c
```diff
@@ -2126,15 +2128,15 @@ static size_t put_data_to_the_end(ns_track_elem_t *ns_info, seg_desc_t *dataseg,
     offset = global_offset % _data_segment_size;
 
     /* We should provide additional space at the end of segment to place EXTENSION_SLOT to have an ability to enlarge data for this rank.*/
-    if (sizeof(size_t) + ESH_KEY_SIZE(key, size) + EXT_SLOT_SIZE(key) > _data_segment_size) {
+    if ((sizeof(size_t) + ESH_KEY_SIZE(key, size) + EXT_SLOT_SIZE(key)) > _data_segment_size) {
```
artpol84: According to this:
http://en.cppreference.com/w/c/language/operator_precedence
this line was fine.
can you explain what two errors you identified?

src/dstore/pmix_esh.c
```diff
@@ -2126,15 +2128,15 @@ static size_t put_data_to_the_end(ns_track_elem_t *ns_info, seg_desc_t *dataseg,
...
         /* this is an error case: segment is so small that cannot place evem a single key-value pair.
          * warn a user about it and fail. */
         offset = 0; /* offset cannot be 0 in normal case, so we use this value to indicate a problem. */
         pmix_output(0, "PLEASE set NS_DATA_SEG_SIZE to value which is larger when %lu.",
```
artpol84: according to the error you previously reported you should either see this message

artpol84:
@hjelmn can you also remind me what version of PMIx you are using? We need to know this because ESH code diverged in master so it's important to know.

src/dstore/pmix_esh.c
artpol84: I see that you are using updated version of `EXT_SLOT_SIZE`

src/dstore/pmix_esh.c
artpol84: What I'm trying to say is that I don't mind changing the type of  `id` but I don't believe it was the reason of the error.

artpol84:
@hjelmn And I'll duplicate my question here: what error you hit this time:
* This https://github.com/pmix/master/pull/282#issuecomment-277422536
* or this https://github.com/pmix/master/pull/282#issuecomment-277454198

jjhursey:
@hjelmn After we merge in PR #284 you will need to rebase this branch so we can pick up the `int` to `size_t` change.

jjhursey:
I just noticed this was a PR against `master` (updated the milestone). PR #282 will merge in, then you'll need to rebase. If you want this change in `v1.2` then you can cherry-pick it over in a new PR. Thanks!

PR #290, opened 2017-02-07
pmix: fix zlib protection macro usage

jsquyres:
It's possible that we can have `zlib.h` but still not have zlib support.  Use the correct macro to protect the usage of calling zlib functions.


jjhursey:
👍  This looks good to me. Thanks for catching that.

This should only impact the PMIx `master` (future `v2.0`) so it does not need to go to any release branches in OMPI.

PR #291, opened 2017-02-07
config: Fix pthread locking variable

jjhursey:
* Typo in the variable check prevents the check for the pthread locking mechanism.

### DIFF
--- a/config/pmix_check_lock.m4
+++ b/config/pmix_check_lock.m4
@@ -2,6 +2,7 @@ dnl -*- shell-script -*-
 dnl
 dnl Copyright (c) 2017      Mellanox Technologies, Inc.
 dnl                         All rights reserved.
+dnl Copyright (c) 2017      IBM Corporation.  All rights reserved.
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


artpol84:
@jjhursey does this makes a difference when pthreads are on on PPC?

jjhursey:
@artpol84 Without this fix the pthread locks were never enabled (since configure never went down this path). So I was still using `flock` in my performance tests I mentioned this morning.

With this patch the pthread locks were enabled, and performance was ***much*** better.

artpol84:
That's what I'm asking! 
Can you share the data?

jjhursey:
I probably should have included this reference to where `DSTORE_PTHREAD_LOCK` is declared.
 * https://github.com/pmix/master/blob/master/config/pmix.m4#L900-L911

Yeah, I'll share the data. I need to get a couple more data points then I'll update Issue #207 which might/should be resolved now.

artpol84:
:+1: 
Thank you!

artpol84:
Let's wait until Jenkins is done and merge.

artpol84:
Have you checked 1.2? Is it affected?

jjhursey:
@artpol84 Yeah we will need to cherry-pick over to `v1.2` as it has the same issue. Once CI is finished then I'll merge, and file the PR for `v1.2`

PR #292, opened 2017-02-07
config: Fix pthread locking variable

jjhursey:
Ref PR #291 (`master`)


artpol84:
@karasevb, please double-check.

karasevb:
@artpol84, `./configure` works right, I tried with `--disabled store-thick` and without, and verified that `config.h` generated correctly.
:+1:

PR #296, opened 2017-02-09
dstore: Fix illegal memory free

artpol84:
rinfo may point to the shared memory region or be allocated using
malloc. Release it only if it was malloc'ed.
This needs refactoring to make rinfo always point to the shared
memory ares. This will be done after v1.2.1 release.


artpol84:
bot:retest

PR #298, opened 2017-02-09
dstore: Fix access rights for pthread-based locking

karasevb:
This PR changes the access rights to locking in case when PMIx server
was started as root user and clients from normal user.


artpol84:
@karasevb
Was that tested with ORTE and our MTT?
If not - please do.

PR #309, opened 2017-02-16
make Oracle compilers happy panda again

ggouaillardet:
Thanks Siegmar Gross for the report

### DIFF
--- a/src/dstore/pmix_esh.c
+++ b/src/dstore/pmix_esh.c
@@ -135,23 +135,11 @@ __extension__ ({                                            \
 })
 
 #ifdef ESH_PTHREAD_LOCK
-#define _ESH_LOCK(rwlock, operation)                        \
+#define _ESH_LOCK(rwlock, func)                             \
 __extension__ ({                                            \
     pmix_status_t ret = PMIX_SUCCESS;                       \
     int rc;                                                 \
-    switch (operation) {                                    \
-        case F_WRLCK:                                       \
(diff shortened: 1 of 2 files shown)


ggouaillardet:
@karasevb that fixes the issue reported at https://mail-archive.com/users@lists.open-mpi.org/msg30625.html
the root cause is very likely a workaround in Oracle compilers
(a compiler should never crash !)
i revamped the macros (fortunatly, it is not un-elegant) and the compiler stopped crashing ...

src/dstore/pmix_esh.c
```diff
@@ -1092,10 +1084,11 @@ int _esh_store(const char *nspace, pmix_rank_t rank, pmix_kval_t *kv)
     }
 
     /* set exclusive lock */
-    if (PMIX_SUCCESS != (rc = _ESH_WRLOCK(_ESH_SESSION_lock(ns_map->tbl_idx)))) {
+    if (PMIX_SUCCESS != (rc = _ESH_WRLOCK(PMIX_VALUE_ARRAY_GET_BASE(_session_array, session_t)[ns_map->tbl_idx].rwlock))) {
         PMIX_ERROR_LOG(rc);
         return rc;
     }
+#if 1
```
rhc54: What does this do?

ggouaillardet: my bad, this is debug leftover ...
will fix shortly

ggouaillardet:
@rhc54 i removed the debug and an extra useless change

PR #310, opened 2017-02-16
Remove stale file reference in check_compiler_version that caused the…

rhc54:
… tests to always fail. Update the psm2 check to look for the correct renamed function in the updated libraries.


jjhursey:
The update to `config/pmix_check_compiler_version.m4` should probably be pulled into `v1.2` once this gets merged in.

PR #311, opened 2017-02-16
Ensure we retain the peer object until we are done with it, then dete…

rhc54:
…ct that the socket has closed due to a lost connection and cleanly release the message event


rhc54:
waiting for OMPI MTT results

rhc54:
OMPI MTT results look very good indeed!

PR #318, opened 2017-02-24
misc OpenBSD 6 related fixes

ggouaillardet:
Thanks Paul Hargrove for bringing this to our attention

Refs open-mpi/ompi#2106

### DIFF
--- a/config/pmix_check_package.m4
+++ b/config/pmix_check_package.m4
@@ -13,6 +13,8 @@
 # Copyright (c) 2012-2015 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2012      Oracle and/or its affiliates.  All rights reserved.
 # Copyright (c) 2014      Intel, Inc. All rights reserved.
+# Copyright (c) 2017      Research Organization for Information Science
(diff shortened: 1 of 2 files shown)


rhc54:
@ggouaillardet If I read this correctly, we are indeed restoring the original LDFLAGS when we are done with searching - correct? I think it's important that we do so.

rhc54:
I guess it also hits LIBS as well, but I gather we protect it too.

ggouaillardet:
@rhc54 @PHHargrove this is correct
```_PMIX_CHECK_PACKAGE_LIB``` is an internal subroutine only invoked by ```PMIX_CHECK_PACKAGE```.
and in ```PMIX_CHECK_PACKAGE``` we do save and restore ```CPPFLAGS```, ```LDFLAGS``` and ```LIBS```

so i think we are good

PHHargrove:
@ggouaillardet this resolves the issue that has been blocking my ability to test open-mpi/ompi#2107

rhc54:
@ggouaillardet @PHHargrove  Thanks to both of you!

PR #323, opened 2017-02-28
Fix a number of abstraction breaks relating to the assign, add, active, and del of events.

rhc54:
Thanks to Nathan Hjelmn for pointing them out


rhc54:
bot:mellanox:retest

PR #326, opened 2017-03-07
Fix some missing includes and one #ifdef that shouldn't have been included

rhc54:
Thanks to Paul Hargrove for pointing this out.


jjhursey:
Ref https://github.com/open-mpi/ompi/issues/3116

PR #327, opened 2017-03-08
Fix a problem that surfaced on NetBSD/AMD64. 

rhc54:
Just because HAVE_POSIX_FALLOCATE is defined doesn't mean the function actually works, so fall back to the ftruncate code path when it fails.

Thanks to Paul Hargrove for the patch


rhc54:
@jjhursey You will want to cherry-pick this over to the 1.2 branch

PR #330, opened 2017-03-09
pmix/test: fix mem corruption

karasevb:
Fixed the string size calculate error in arguments parsing of test


artpol84:
:+1:

PR #332, opened 2017-03-09
Update libevent configure to avoid adding flags multiple times - at s…

rhc54:
…ome point, we changed to "check_package", which automatically adds the flags to CPPFLAGS and friends so we don't need to explicitly do so. Spit out the ldflags and libs at the end of configure as well as cppflags and cflags

### DIFF
--- a/config/pmix_setup_libevent.m4
+++ b/config/pmix_setup_libevent.m4
@@ -2,7 +2,7 @@
 #
 # Copyright (c) 2009-2015 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2013      Los Alamos National Security, LLC.  All rights reserved.
-# Copyright (c) 2013-2016 Intel, Inc. All rights reserved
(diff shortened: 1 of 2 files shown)


rhc54:
@artpol84 Did something change in the test system?

```
00:00:46 checking for library containing event_config_new... -levent
00:00:46 checking for evthread_set_lock_callbacks in -levent... no
00:00:46 configure: WARNING: External libevent does not have thread support
00:00:46 configure: WARNING: PMIx requires libevent to be compiled with
00:00:46 configure: WARNING: thread support enabled
00:00:46 configure: error: Cannot continue
00:00:46 Build step 'Execute shell' marked build as failure
```

artpol84:
I will check, but it shouldn't be.

rhc54:
okay, thx - i'll recheck the changes as well

rhc54:
@artpol84 Looks like it may have been a bug - fixed now.

PR #336, opened 2017-03-23
Ensure we don't hang in PMIx_Abort if the server disappears after we …

rhc54:
…send, but before we receive the response


rhc54:
@jjhursey You may need the equivalent backported to 1.2.3.

Issue #337, opened 2017-03-23
PMIx_Abort hang possibility


jjhursey:
See PR #336 for details on the change that went to `master`. A similar fix will need to be investigated for `v1.2.3`

rhc54:
v1.2 series is closed

PR #339, opened 2017-03-25
Ensure we don't hang in PMIx_Abort if the server disappears after we …

rhc54:
…send, but before we receive the response

(cherry picked from commit d80d4f12b35879b6cd192904de765af636916b1d)


rhc54:
Fixes #337

PR #340, opened 2017-03-27
configury: allow repeated arguments with -XCClinker

ggouaillardet:


ggouaillardet:
@rhc54 when using fujitsu cross-compilers, i must
```configure LDFLAGS='=Xg -Xcclinker -Xg' ... ``` so the second ```-Xg``` must not be stripped out.

PR #341, opened 2017-03-27
configury: allow repeated arguments with -Xg

ggouaillardet:


ggouaillardet:
@rhc54 i messed up #340 
```configure LDFLAGS='-Xg -XCClinker -Xg'```
what i really wanted is to allow the argument following ```-XCClinker``` being a duplicate,
but that would require the logic being revamped
so i simply allow ```-Xg``` to be a duplicate

PR #344, opened 2017-04-02
Implement the event notification extension RFC

rhc54:
Refs https://github.com/pmix/RFCs/pull/18

### DIFF
--- a/src/event/pmix_event_notification.c
+++ b/src/event/pmix_event_notification.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2014-2016 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2014-2017 Intel, Inc.  All rights reserved.
  * Copyright (c) 2017      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * $COPYRIGHT$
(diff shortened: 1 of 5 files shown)


src/event/pmix_event_notification.c
```diff
@@ -190,124 +192,189 @@ static void progress_local_event_hdlr(pmix_status_t status,
...
+            }
+        }
     }
 
-    if (NULL != chain->def) {
-        if (pmix_list_get_end(&pmix_globals.events.default_events) != (nxt = pmix_list_get_next(&chain->def->super))) {
-            def = (pmix_default_event_t*)nxt;
-            chain->def = def;
+    /* if we registered a "last" handler, and it fits the given range
+     * and code, then invoke it now */
```
abouteiller: So, last_handler will be called after default?

src/event/pmix_event_registration.c
```diff
@@ -382,30 +489,164 @@ static void reg_event_hdlr(int sd, short args, void *cbdata)
...
+                    }
+               }
+               found = true;
+               break;
+            }
+        }
+        /* if the handler wasn't found, then it may show up later - so
+         * for now just prepend it to the list */
+        if (!found) {
+            pmix_list_prepend(cd->list, &evhdlr->super);
```
abouteiller: Maybe I missed it, but I did not see the piece of code to reorder the list when inserting the BEFORE-AFTER event

src/event/pmix_event_notification.c
rhc54: Yes - remember, there can be any number of registered "default" handlers. All "default" means is that the handler shall be invoked for any code.

src/event/pmix_event_registration.c
rhc54: You observed correctly 😄  I implemented the before-after ordering for the case where the event (B) we are ordering against is already present on the list. However, for the case where B isn't present, then we simply prepend the new event (A) to the list.

When B finally shows up, we put it wherever it needs to go. We then need to scan the list to find all events that specified ordering against B. In this case, we would find A, and reposition it to be before/after B. This would have to be done for all events ordered relative to B.

Since I expect this to be the unusual case, I left implementation to the student 😄

rhc54:
@abouteiller Please take a look to see I got the error return as we discussed

PR #345, opened 2017-04-04
Fix dmodex - eliminate race condition where data was being released while sent

rhc54:


jjhursey:
@rhc54 I see similar code in the `v1.2` series. Does it need to be fixed there as well?

rhc54:
Probably - I didn't check over there. This was seen at large scale where the race is evident.

PR #346, opened 2017-04-04
Fix dmodex - eliminate race condition where data was being released while sent

jjhursey:
Refs PR #345


rhc54:
FWIW: just heard from Nathan that the problem remains on his system at scale. So it appears I just pushed the race condition the other way for my smaller system. Sigh.

jjhursey:
Ok - well we can keep this PR open and keep it in sync with `master` until the issue is resolved.

PR #350, opened 2017-04-12
Allow a user to disable backward compatibility for PMI-1/2 as these c…

rhc54:
…an conflict with system-based equivalent libraries


jjhursey:
Question about the configure flag name `--enable-pmix-backward-compatibility` disables support for PMI1/PMI2, but not the cross-versioning support for PMIx nor backwards compatibility for PMIx versions.

Wouldn't something like `--enable-pmi-backward-compatibility` be a better name? Then we can reserve the `pmix` version for the future when we might need it.

What do you think?

rhc54:
Agreed - I changed it and re-pushed.

rhc54:
Great - feel free to take it whenever you're ready. I'll update master to match

PR #366, opened 2017-05-15
Check for the correct levels of autotools

rhc54:


rhc54:
Looks like something is going bonkers when hit with a SIGKILL during "fence", but clearly the autotools change was fine as Jenkins was able to build the software

PR #369, opened 2017-05-26
If we get an event with PMIX_RANGE_NAMESPACE, then we should set the …

dsolt:
…namespace

of the reporting proc as the target namespace for the event.

At present we ignore if the user set a custom range in the namespace, but
used PMIX_RANGE_NAMESPACE in the PMIx_Notify_event call.

### DIFF
--- a/src/event/pmix_event_notification.c
+++ b/src/event/pmix_event_notification.c
@@ -3,6 +3,8 @@
  * Copyright (c) 2014-2017 Intel, Inc.  All rights reserved.
  * Copyright (c) 2017      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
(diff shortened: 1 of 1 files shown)


rhc54:
@dsolt This makes sense - didn't you have another fix we wanted to capture? IIRC, we weren't checking ranges when pulling events out of the ring buffer?

dsolt:
Well, I think I had two ways to fix this... one was to check the range when pulling things out of the ringbuffer, the other way was to set targets when putting into the ring buffer.  So I don't think there were 2 issue, just 2 ways to fix the one issue and I chose this way.

rhc54:
Okay, that's cool - I have no issue with it. Just wanted to ensure we weren't missing something.

PR #370, opened 2017-05-26
EXPORT some buffer_op functions for use by RM's

dsolt:


rhc54:
Sounds okay to me - we'll need to redo it in the cross-version as this will create conflicts with that branch, but it's no big deal.

Only question: using this will require that the RM include a header (buffer_ops.h) that we don't normally install. Shouldn't we make them available via pmix_common.h, or perhaps pmix_server.h? Perhaps hide them behind a set of macros?

dsolt:
That would be great.  Right now we configure with devel headers and I think that makes buffer_ops.h available.

rhc54:
Yes, that would make it work. However, we probably want to avoid making people do so - if the capability is going to be generally available, then it should be installed by default.

We can commit this for now, and then create a more generalized solution in a separate PR. I'm willing to create it - I'll try to get that done for v2.0.

PR #371, opened 2017-05-26
Don't try to add data to dstore that doesn't exist to avoid warning/e…

dsolt:
…rror messages being logged.

### DIFF
--- a/src/common/pmix_jobdata.c
+++ b/src/common/pmix_jobdata.c
@@ -3,6 +3,7 @@
  * Copyright (c) 2016-2017 Mellanox Technologies, Inc.
  *                         All rights reserved.
  * Copyright (c) 2016-2017 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


rhc54:
@dsolt Just curious - why would we receive a zero-byte payload for a proc's job-level info?

dsolt:
I think it happens every time the ranks are split across more than 1 host.   Unless I'm not calling PMIx_Register_nspace correctly?   When I call PMIx_Register_nspace, I only pass down proc level details for the ranks that are local to that host, but then dstore seems to cycle through all ranks in the namespace looking for proc data to push into the dstore

rhc54:
Hmmm...well, that's the issue. We were expecting you to pass down all info for all ranks, but it is also true that we don't need all the details. ORTE currently passes down the local rank, node rank, and node ID for every rank - and that's it. We actually _do_ require that much info for every rank in order for them to auto-wireup.

Are you at least passing that much? If not, then that's a problem.

rhc54:
@dsolt I'm going to close this one as it seems that the problem here is the lack of data required by the PMIx spec. Instead, I've provided a suppressible warning of the situation - please see https://github.com/pmix/pmix/pull/376

PR #379, opened 2017-05-28
Add new PMIx_Data_xxx APIs to expose the bfrops support as potentially useful utilities for host RMs

rhc54:


rhc54:
Supports https://github.com/pmix/RFCs/pull/20

Issue #347, opened 2017-04-06
Modex occasionally misses a key, success on retry


jjhursey:
Occasionally a `PMIx_Get` call will return key not found, then if you try again immediately it will find the key. This has been seen in the following locations in Open MPI `v2.x`:
 * `OPAL_PMIX_LOCAL_RANK`
    - https://github.com/open-mpi/ompi/blob/v2.x/orte/mca/ess/pmi/ess_pmi_module.c#L140-L141
 * `OPAL_PMIX_UNIV_SIZE`
    - https://github.com/open-mpi/ompi/blob/v2.x/orte/mca/ess/pmi/ess_pmi_module.c#L158-L159
 * Something happening with locality flags (`proc_flags`) which are being set from a `OPAL_PMIX_LOCALITY` seen as a mislabeled process (labeled nonlocal - default, instead of local)
   - https://github.com/open-mpi/ompi/blob/v2.x/opal/mca/btl/sm/btl_sm.c#L479-L483
   - Set here: https://github.com/open-mpi/ompi/blob/v2.x/ompi/proc/proc.c#L148

I confirmed in the first and third case if I immediately reissue the `OPAL_MODEX_RECV_VALUE(_OPTIONAL)` then I receive the correct value. For case 2 I have not been able to trigger that failure mode locally.

I'm using PMIx `v1.2` HEAD and testing with Open MPI `v2.x`. However, there has been a report that this can happen with PMIx `master` and Open MPI `master`. This happens both with and without the dstore. I suspect that this occurs in the `v1.1.x` series as well, but have not tried to independently confirm that yet.

I've tried adding some memory barriers in places where we store data in one thread/process and read it in another. Those have not helped.

rhc54:
Ah, very good indeed! Thanks!

jjhursey:
Fixed in `master` in PR #386, will be fixed in `v1.2` in PR #387 - at which point this ticket will close.

jjhursey:
Fixed in master in PR #386 (also in v2.0), and v1.2 in PR #387. Will be in upcoming release of both branches.

PR #383, opened 2017-06-02

jjhursey:
* Fixes Issue #347
 * A PMIx_Get operation thread shifts to perform the actual lookup.
   - Thread A: PMIx_Get() -> Thread_shift
   - Thread B: Lookup value, allocate memory, copy value, release Thread A
   - Thread A: Wakeup and read the allocated value
 * The problem is that Thread B can allocate and update the pointer, but
   due to cache effects Thread A does not see that value (instead sees
   NULL).
   - The solution is to add a memory barrier in Thread B once it has
     finished updating the value.
 * This case matches the first example ("Global thread flag") at:
   - https://www.ibm.com/developerworks/systems/articles/powerpc.html
   - As such we are using the `RWB` which corresponds to `lwsync` on PPC

### DIFF
--- a/include/pmix_common.h
+++ b/include/pmix_common.h
@@ -690,6 +690,44 @@ typedef struct pmix_byte_object {
 } pmix_byte_object_t;
 
 
+/****    PMIX DATA BUFFER    ****/
+typedef struct pmix_data_buffer {
+    /** Start of my memory */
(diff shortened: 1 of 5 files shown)


jjhursey:
(Shoot I messed up by branch - I'll need to re-create this PR)

PR #384, opened 2017-06-02

jjhursey:
* Fixes Issue #347
 * A PMIx_Get operation thread shifts to perform the actual lookup.
   - Thread A: PMIx_Get() -> Thread_shift
   - Thread B: Lookup value, allocate memory, copy value, release Thread A
   - Thread A: Wakeup and read the allocated value
 * The problem is that Thread B can allocate and update the pointer, but
   due to cache effects Thread A does not see that value (instead sees
   NULL).
   - The solution is to add a memory barrier in Thread B once it has
     finished updating the value.
 * This case matches the first example ("Global thread flag") at:
   - https://www.ibm.com/developerworks/systems/articles/powerpc.html
   - As such we are using the `RWB` which corresponds to `lwsync` on PPC

### DIFF
--- a/src/atomics/sys/gcc_builtin/atomic.h
+++ b/src/atomics/sys/gcc_builtin/atomic.h
@@ -16,6 +16,7 @@
  * Copyright (c) 2016      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2017      Intel, Inc. All rights reserved.
+ * Copyright (c) 2017      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -63,6 +64,8 @@ static inline void pmix_atomic_wmb(void)
 }
 
 #define PMIXMB() pmix_atomic_mb()
+#define PMIXRMB() pmix_atomic_rmb()
+#define PMIXWMB() pmix_atomic_wmb()
 
 /**********************************************************************
  *
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -7,7 +7,7 @@
  *                         All rights reserved.
  * Copyright (c) 2016      Mellanox Technologies, Inc.
  *                         All rights reserved.
- * Copyright (c) 2016      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2016-2017 IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -190,7 +190,9 @@ static void _value_cbfunc(pmix_status_t status, pmix_value_t *kv, void *cbdata)
     if (PMIX_SUCCESS == status) {
         if (PMIX_SUCCESS != (rc = pmix_bfrop.copy((void**)&cb->value, kv, PMIX_VALUE))) {
             PMIX_ERROR_LOG(rc);
+            cb->status = rc;
         }
+        PMIXRMB();
     }
     cb->active = false;
 }


jjhursey:
I created PR #385 which PR's this change to the `v1.2` series. I decided not to wait to file the PR since I had to port the atomics/thread code over, and it would be custom. We can break it up into two PRs if needed.

rhc54:
wow - really? I believe you, but now I'm a little nervous. Not about this proposed fix as I'm sure you've beaten it to death - rather, I'm concerned that if you hit this race condition in some circumstance, then how can we be confident we won't just hit more of them in other parts of the code when on a different platform, or under different loading, or...?

Do we have any idea why this particular spot might be more sensitive than any other? Or is this just a crapshoot?

jjhursey:
@rhc54 That thought has been going through my mind the past couple days ever since I pinpointed it.

I suspect that we need this in other places in the code too, since we use this technique in a bunch of other places in the codebase. I just flagged it as "In Progress" so we can talk though it, and maybe expand the reach.

To really convince myself I created a small, simple pthread program that uses the thread shift technique. I used hwloc to pin them to the first and last cores on a heavily loaded machine. I let it run though a large number of iterations, and without the memory barrier it would get a NULL pointer in Thread A. With the memory barrier it passed clean. That was enough to convince me of the behavior.

I'll clean up that test case and attach it here in case others are interested.

rhc54:
@bosilca I was talking with @jsquyres about this, and our heads are aching from trying to figure it out. What we do is create an object and populate it in one thread and then push it into a libevent event. Another thread catches that event and attempts to access the object and its fields.

This appears to work 99.99% of the time. However, every so often it fails as per @jjhursey explanation above. We could throw RMB's around, but we fear that will negatively impact every thread on the chip. So we are now beginning to wonder if our threading model really is sound.

Do you have any thoughts on this? Is it possible that our "volatile" declarations aren't complete (I'm thinking of that "double volatile" declaration I asked you about in a recent OMPI issue). 

Moving away from the libevent-based model is a major step, so I don't want to jump down that path unless we truly convince ourselves that it is necessary. I've added this topic to the July meeting

bosilca:
volatile doesn't ensure ordering in the memory operations, it merely prevent the compiler from storing a variable in a register (so it always generates a load instruction). As @jjhursey identified, you have a pretty classical "read-after-write" concurrency issue. For correctness reasons, mainly because the hardware is allowed to reorder the memory operations, you need to have a memory flush before accessing the data from another thread. The issue arise from the fact that you have multiple memory locations to update: on one side the object that holds your data, on the other the pointer to that object you pass to another thread. If there is no memory ordering between the 2, the reader thread can notice the update on the pointer, but without the update on the content of the object. Thus, you need to signal that there is an order in the memory operations, and on the thread that writes the object, you need to add a fence between updating the object and updating the pointer. This fence has the role of indicating to the hardware that when the second update become visible (so the pointer update), all previous updates must also be visible (the update of the object itself). 

Now to make matters more complicated depending on the architecture, your need 2 barriers, a write barrier on the thread that updates the object, and a read barrier on the other.

rhc54:
Argh - I think I see, but it calls into question our threading model here and in ORTE. We had assumed (and yes, I know what they say about those who "assume") that libevent took care of such issues. So pushing an event into libevent and letting it "fire" would ensure both thread safety and ordering.

I gather that assumption isn't correct, which leads to some decisions:

* we could go thru PMIx and ORTE, adding read/write barriers everywhere we use events to switch thread contexts. Fortunately, much of that takes place inside macros, so at least the write barriers would be relatively easy to install. The reads, however, would be much more painful.

* we could attempt to modify our embedded libevent to meet our needs, but that strikes me as a bad option as it doesn't help when we are built against an external version

* we could switch our threading model to the more conventional pthread-driven one abstracted in the pmix/opal thread code. Negative here is that there is far more likelihood of hitting thread stalls, and the code will get significantly more ugly. However, those macros _do_ already have the read/write barriers in them.

* we could create our own event library that accounts for these problems. @hjelmn is working on such a thing now, and given our use of the pmix/opal event abstraction, it offers a clean "drop-in" solution.

Any thoughts/suggestions on the approach to take? I'm inclined to the last two options myself, with a slight preference for the last one only because of ease of adoption. I'm not sure of the performance tradeoff, though, but I imagine the approach taken in the event library would significantly impact it.

I'm going to have it added to the Tues telecon agenda as this is going to impact the OMPI v3.0 release schedule.

jjhursey:
To one of George's points - I'm thinking that we might need the read and the write barriers around the thread shifting.
I have been running the stress test case (it's an MPI hello world, in a loop, at large scale, heavily oversubscribed - it takes about 11-16 hours to run 2000 iterations. It'll reproduce somewhere in the middle once or twice). So far it has been working like a champ, but I did see one failure late today that I thought was a system issue unrelated, but after thinking about that comment it could be that I'm missing the other barrier on Thread A. I've restarted the test to see if it was just a systems issue - so far so good though. I'm going to spend some time with the PPC hardware docs/folks to see if it's needed on our architecture.

I think the reason I am seeing it in `PMIx_Get` and not elsewhere is because we make a ton of `PMIx_Get` calls relative to the other `PMIx` API calls. And sometimes in rapid succession.

To Ralph's other questions about the threading model - I want to think about it a bit more over the weekend. My sense it that it's not wrong to use the model we have, but we will need more of these types of synchronization mechanisms...


To put a bit of context in here for those playing at home maybe not super familiar with this section of code (or really just for Future Josh so I don't forget).

The structure we are allocating is the [`pmix_cb_t`](https://github.com/pmix/pmix/blob/v1.2/src/usock/usock.h#L86-L113)
```c
/* usock struct for tracking ops */
typedef struct {
    pmix_list_item_t super;
    pmix_event_t ev;
    volatile bool active;
    bool checked;
    int status;
    // ...
    pmix_value_t *value;
    // ...
} pmix_cb_t;
PMIX_CLASS_DECLARATION(pmix_cb_t);
```

The two parts of that structure that pertain to this discussion are:
```c
    volatile bool active;
    pmix_value_t *value;
```

Here's the timeline:
 1. Thread A : Calls `PMIx_Get` passes a [`PMIX_NEW(pmix_cb_t)`](https://github.com/pmix/pmix/blob/v1.2/src/client/pmix_client_get.c#L89) to `PMIx_Get_nb` which eventually calls [`PMIX_THREADSHIFT(cb, _getnbfn)`](https://github.com/pmix/pmix/blob/v1.2/src/client/pmix_client_get.c#L172).
The value of `cb->value` is `NULL` (set [here](https://github.com/pmix/pmix/blob/v1.2/src/usock/usock.c#L277) ).
The value of `cb->active` is `true` (set [here](https://github.com/pmix/pmix/blob/v1.2/src/include/pmix_globals.h#L293-L299) ).
 1. Thread A : Waits in [`PMIX_WAIT_FOR_COMPLETION`](https://github.com/pmix/pmix/blob/v1.2/src/client/pmix_client_get.c#L97-L99) on the `cb->active` flag.
 1. Thread B : Starts looking for the value in [`_getnbfn`](https://github.com/pmix/pmix/blob/v1.2/src/client/pmix_client_get.c#L400)
 1. Thread B : Finds the value and calls [_value_cbfunc](https://github.com/pmix/pmix/blob/v1.2/src/client/pmix_client_get.c#L177-L189)
 1. Thread B : Allocates `cb->value` and copies the value into it using the [`pmix_bfrop.copy`](https://github.com/pmix/pmix/blob/v1.2/src/client/pmix_client_get.c#L184) operation
 1. Thread B : Set `cb->active = false`
 1. Thread A : Wakes up and immediately inspects `cb`
```c
/* wait for the data to return */
    PMIX_WAIT_FOR_COMPLETION(cb->active);
    rc = cb->status;
    *val = cb->value;
    PMIX_RELEASE(cb);
```
 8. Thread A : `cb->active` will either be the valid memory value or (admittedly rarely) `NULL` - since it doesn't see the update.

I've instrumented the code such that I can see the above progression when it fails. That's what led me here.

rhc54:
Many thanks, @jjhursey for the detailed explanation and detective work!

My question from earlier: if we put a "volatile" on that cb->value pointer, would that solve the problem? Perhaps its just my poor understanding, but it appears to me that "volatile" instructs the compiler/hardware to pull the update, which is basically what the barrier is forcing. So in this case, we declared ```active``` to be volatile, and so it gets refreshed and wakes up A. However, we didn't declare ```value``` to be volatile, and so it may or may not be refreshed. As @bosilca indicated, it may well be cached in a register.

Make any sense?

jjhursey:
I tried adding the `volatile` to the pointer, and that still failed.

I've convinced myself that we need both the `RMB` on the supporting thread and a `WMB` on the main thread (at least for PPC). I'll update this PR on Monday with that change (I'll hold off doing it now in case there are other changes we need).

I've posted a test program that I've been using to experiment with memory barrier placement.
 * https://gist.github.com/jjhursey/61d502825024ab7bf9ffad83ab6c9975

This is a pthread program that depends on hwloc. It has hard coded the PPC assembly for the memory barriers, but those can be easily replaced for your architecture. It needs to be run at a large number of iterations on a machine that is heavily loaded. On a lightly loaded machine I've seen the test pass without the memory barriers. On a heavily loaded machine it fails consistently in modes `0,1,2` and passes in modes `3,4`.

rhc54:
Nuts - that would have been the simplest solution ☹️ 

What this saids to me, though, is that we _should_ revise our threading model to remove libevent from the thread shift areas (we still would use it for monitoring file descriptors, at least for now). Instead, we would use the PMIx thread acquire/release macros (borrowed from OPAL) that have the W/R MB's embedded in them:

* we replace the ```volatile bool active``` in the caddy with a mutex and condition variable
* thread A acquires the progress thread and adds its caddy to the work list. It then releases the thread and goes into a conditioned wait in the wait_for_completion macro
* thread B (the progress thread) sits on a conditioned wait so it doesn't burn cpu when idle. When woken up, it acquires the thread and cycles thru the list to execute the work items. We need the list because multiple ```thread A's``` might have acquired the thread and added work items in-between thread B getting control. As it finishes each item, thread B signals on the caddy's condition using the PMIx "signal" macro that includes a WMB
* the wait_for_completion macro releases on the condition signal and "reacquires" the caddy with an RMB before exiting the macro

This has a desirable side effect of removing those "sleep" calls in the wait_for_completion macro, which we've already been "dinged" for by a few people. It also hides the MB's so developer's don't invisibly get into this problem in the future.

Make sense? I thought I might modify your test program over the weekend to use the macros and see how that looks and works, and test it out on my Intel-based cluster

rhc54:
@jjhursey How are you loading your machine for this test? I want to ensure I replicate it accurately.

bosilca:
There are 2 possible solution that are both relatively cheap to implement.
1. add a opal_atomic_wmb before after setting the pointer but before setting the bool ```active```, and on the shifted thread an opal_atomic_rmb after noticing that the ```active``` field has changed, and before reading the pointer. I assume you can import the opal_atomic_rmb and opal_atomic_wmb from OPAL.
1. Add a mutex to the object, and use it to provide a portable read/write barrier. The POSIX mutexes have implicit memory fences inside, so one can use them as a slightly more expensive mechanism for handling RAW accesses.

That being said, you don't need to be concerned about performance here, there memory fences take tens of nanoseconds, way less than the cost of shifting to another thread.

rhc54:
Fair points - will ponder and then we can discuss. FWIW, I ran @jjhursey test on my 24-core Intel box and it passed with no memory fences:
```
$ ./threads 9000000 0
---------------------------
Iterations:    9000000
Mode R MB :   Disabled
Mode W MB :   Disabled
Mode - MB :   Disabled
Mode X MB :   Disabled
---------------------------
Main : [objs = 24] : cpuset is 0x01000001
Peer : [objs = 24] : cpuset is 0x00008000,0x00800000
  10.0 % complete : Iteration     900002 /    9000000 :   56.5 usec / iter
  20.0 % complete : Iteration    1800002 /    9000000 :   56.3 usec / iter
  30.0 % complete : Iteration    2700002 /    9000000 :   56.2 usec / iter
  40.0 % complete : Iteration    3600002 /    9000000 :   56.1 usec / iter
  50.0 % complete : Iteration    4500002 /    9000000 :   56.1 usec / iter
  60.0 % complete : Iteration    5400002 /    9000000 :   56.0 usec / iter
  70.0 % complete : Iteration    6300002 /    9000000 :   56.0 usec / iter
  80.0 % complete : Iteration    7200002 /    9000000 :   56.0 usec / iter
  90.0 % complete : Iteration    8100002 /    9000000 :   55.9 usec / iter
Success -   55.9 usec / iter

Loaded with: $ mpirun --oversubscribe --np 100 -host rhc001  ./mpi_spin

$ uptime
 11:13:21 up 9 days, 23:15,  4 users,  load average: 90.81, 38.03, 14.30
```

rhc54:
@bosilca Let me ask a probably dumb question. If I create an object that includes a pointer, set that pointer to some object, and then thread shift - can I be assured that the callback servicing the event on the other end of that shift will "see" the correct values? Or do I need to memory fence this use-case as well?

bosilca:
There is no way to ensure that the memory updates are becoming visible in the order in which they were made. In other terms, any memory updates A and B made sequentially on a core, may be visible in any order by all the other cores (the opportunity increases with the memory distance). You will need to use some constraints (acquire/release/fence) to enforce the ordering.

rhc54:
I figured you were going to say that 😄

rhc54:
Please see https://github.com/pmix/pmix/pull/386

rhc54:
This was replaced

PR #385, opened 2017-06-02

jjhursey:
* Port over the atomics and thread support from `master`
 * Bring over changes from `master` PR #384

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -56,11 +56,12 @@ include src/buffer_ops/Makefile.am
 include src/util/Makefile.am
 include src/usock/Makefile.am
 include src/client/Makefile.am
-include src/atomics/Makefile.am
-include src/threads/Makefile.am
 include src/server/Makefile.am
 include src/sec/Makefile.am
 include src/common/Makefile.am
+
+include src/atomics/asm/Makefile.am
+include src/atomics/sys/Makefile.include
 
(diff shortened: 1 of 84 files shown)


rhc54:
Ugh - you really want to bring that much over? I'll support you if you think it's really necessary. Just a lot of code  for what ultimately is a tiny change, and I can't help but wonder if there might be some smaller (maybe more "hacky") solution for the older release series (e.g., just spin for a little bit?).

jjhursey:
Yeah I'll think about that over the weekend. Maybe we can just grab the atomics portion, and not the threaded portion. In my development build while debugging I just hard coded the PPC instructions, but for portability of pmix in general I'd want to pull over the whole atomics discovery mechanism.

I need to port this back to the `v1.2` series. Actually I also need to port it back to the `v1.1` series. Since `v1.1` is closed I'll just keep that patch internal - but since the two branches are similar the patch will likely look the same.

rhc54:
I have no issue with pulling it into v1.1 if needed - I don't think it's being used outside of you folks, and so I doubt there is an issue. If you need the PPC instructions, then pulling the whole thing over may be the best option. See my comments on #384 - we may have to pull it over just to fix the threading model anyway.

jjhursey:
PR #387 Replaces this PR - streamlining a bit.

PR #386, opened 2017-06-04
Update all cases where we threadshift to use memory fences to ensure data sync.

rhc54:
Per discussion on https://github.com/pmix/pmix/pull/384, update all cases where we threadshift to use memory fences to ensure data sync. Define new PMIX_POST_OBJECT and PMIX_ACQUIRE_OBJECT macros that (for now) contain only the atomic memory fence commands so we can later, if we choose, use them to change the threading model.

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -53,6 +53,7 @@
 
 #include "src/class/pmix_list.h"
 #include "src/buffer_ops/buffer_ops.h"
+#include "src/threads/threads.h"
 #include "src/util/argv.h"
 #include "src/util/compress.h"
 #include "src/util/error.h"
@@ -186,12 +187,14 @@ static void _value_cbfunc(pmix_status_t status, pmix_value_t *kv, void *cbdata)
     pmix_cb_t *cb = (pmix_cb_t*)cbdata;
     pmix_status_t rc;
 
+    PMIX_ACQUIRE_OBJECT(cb);
     cb->status = status;
     if (PMIX_SUCCESS == status) {
         if (PMIX_SUCCESS != (rc = pmix_bfrop.copy((void**)&cb->value, kv, PMIX_VALUE))) {
             PMIX_ERROR_LOG(rc);
         }
     }
+    PMIX_POST_OBJECT(cb);
     cb->active = false;
 }
 
@@ -238,12 +241,12 @@ static pmix_buffer_t* _pack_get(char *nspace, pmix_rank_t rank,
     return msg;
 }
 
-/* this callback is coming from the usock recv, and thus
+/* this callback is coming from the ptl recv, and thus
  * is occurring inside of our progress thread - hence, no
  * need to thread shift */
 static void _getnb_cbfunc(struct pmix_peer_t *pr,
                           pmix_ptl_hdr_t *hdr,
-                         pmix_buffer_t *buf, void *cbdata)
+                          pmix_buffer_t *buf, void *cbdata)
 {
     pmix_cb_t *cb = (pmix_cb_t*)cbdata;
     pmix_cb_t *cb2;
@@ -485,6 +488,9 @@ static void _getnbfn(int fd, short flags, void *cbdata)
     size_t n, nvals;
     char *tmp;
     bool my_nspace = false, my_rank = false;
+
+    /* cb was passed to us from another thread - acquire it */
+    PMIX_ACQUIRE_OBJECT(cb);
 
     pmix_output_verbose(2, pmix_globals.debug_output,
                         "pmix: getnbfn value for proc %s:%d key %s",
@@ -739,11 +745,12 @@ request:
         rc = PMIX_ERROR;
         goto respond;
     }
-
+    /* we made a lot of changes to cb, so ensure they get
+     * written out before we return */
+    PMIX_POST_OBJECT(cb);
     return;
 
-respond:
-
+  respond:
     /* if a callback was provided, execute it */
     if (NULL != cb->value_cbfunc) {
         if (NULL != val)  {
--- a/src/include/pmix_globals.h
+++ b/src/include/pmix_globals.h
@@ -38,7 +38,7 @@
 #include "src/class/pmix_list.h"
 #include "src/class/pmix_ring_buffer.h"
 #include "src/event/pmix_event.h"
-
+#include "src/threads/threads.h"
 #include "src/mca/psec/psec.h"
 #include "src/mca/ptl/ptl.h"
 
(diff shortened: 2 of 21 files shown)


rhc54:
As I said in #384, I think we need to apply this approach globally to ensure we don't hit this problem under different circumstances else we will wind up playing "whack-a-mole". I've taken a stab at it here - @jjhursey can you please try this on your machine and see if it fixes the problem?

This would replace #384.

jjhursey:
Yeah I think this is the right thing to do (for now at least). I'll setup a build this morning and let you know if it addresses the issue on our system. Thanks!

src/threads/threads.h
```diff
@@ -116,6 +116,19 @@ PMIX_EXPORT PMIX_CLASS_DECLARATION(pmix_thread_t);
     } while(0);
 
 
+/* provide a macro for forward-proofing the shifting
+ * of objects between threads - at some point, we
+ * may revamp our threading model */
+
+/* post an object to another thread - for now, we
+ * only have a memory barrier */
+#define PMIX_POST_OBJECT(o)     pmix_atomic_wmb()
```
jjhursey: Shouldn't we use the `PMIXRMB` / `PMIXWMB` macros here?

rhc54: I wasn't sure - don't they just define to their atomic versions?

rhc54: (laugh) it all works out the same:
```
static inline void pmix_atomic_rmb (void)
{
    PMIXRMB();
}
```

jjhursey:
I am manually porting this PR to the `v1.2` series since that is where I can most easily reproduce. In doing so I am performing a close examination of this PR's changes.

For PPC we need both the `WMB` and `RMB` in place. Two threads Main Thread and Support Thread.

For the Support Thread we need a `RMB` (`lwsync` on PPC) just before we set the `->active` flag to `false`:
```c
// Allocate memory and update structure
// Memory Barrier - Prevent hoisting of code above this instruction
RMB(); // lwsync on PPC
// Currently calls PMIX_POST_OBJECT() - which is a WMB not a RMB
cb->active = false; // Release Main Thread
```

For the Main Thread we need a `WMB` (`eieio` on PPC) immediately after it sees the `cb->active == false` from `PMIX_WAIT_FOR_COMPLETION`
```c
// Memory Barrier - Prevent sinking of code below this instruction
#define PMIX_WAIT_FOR_COMPLETION(a)             \
    do {                                        \
        while ((a)) {                           \
            usleep(10);                         \
        }                                       \
        WMB();                              \
// Currently calls PMIX_ACQUIRE_OBJECT() - which is a RMB not a WMB
    } while (0)
```

In the current version of the PR these are swapped:
```c
#define PMIX_POST_OBJECT(o)     pmix_atomic_wmb()
#define PMIX_ACQUIRE_OBJECT(o)  pmix_atomic_rmb()
```
```c
// Instead they should be defined as
#define PMIX_POST_OBJECT(o)     pmix_atomic_rmb()
#define PMIX_ACQUIRE_OBJECT(o)  pmix_atomic_wmb()
```

The Support Thread calls `PMIX_POST_OBJECT(cb);` before setting `cb->active` to `false` which should be defined as a `RMB` is defined as a `WMB` instead of a `RMB`.

The Main Thread calls `PMIX_ACQUIRE_OBJECT(cb);` before accessing the data which is defined as `RMB` instead of a `WMB`.

The [gcc builtin](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html) has been helpful for me to keep the barrier straight. I have to admit that it is easy to get them turned around.

It looks like there are a few places where we need better placement of the `PMIX_POST_OBJECT()` call in this PR. The one I noticed in the review was:
 * In `pmix_client_get.c` we need a `PMIX_POST_OBJECT(cb)` in `_value_cbfunc` just before the `cb->active = false;`

I'll finish my port to `v1.2` and, while that is running, I'll take a look at `master` and setup a reproducing environment there. I'll also work up a patch for this PR as I do so.

rhc54:
Maybe I'm confused, but "post" was intended to mean "ensure that this info is pushed out" and "acquire" was intended to mean "ensure that this info has been refreshed". I would therefore have thought that we wanted a "write" barrier for post (to ensure the write has completed) and a "read" barrier for acquire (to ensure the read completes). Do we have our mb's defined backwards, perhaps?

jjhursey:
Yeah that's what has been making me go a bit confused as well. I know the PPC instruction placement I described is correct, and the GCC builtin version are also correct (from my reading of their docs). Our macro descriptions/definitions are getting turned around in my mind. Maybe we can discuss on the OMPI call tomorrow - since they use the same macros.

rhc54:
sounds like a plan!

jjhursey:
After talking to @hjelmn and @rhc54 about this on the phone here are our notes:
 * The `WMB` vs `RMB` placement in this PR is correct.
    - We need a `WMB` in the Support Thread to ensure all writes complete before any reads.
    - We might, but probably not 100% of the time, need a `RMB` in the Main Thread once it becomes active to prevent any reads from crossing the barrier.
 * The PPC atomics in [PMIx](https://github.com/pmix/pmix/blob/master/src/atomics/sys/powerpc/atomic.h#L31-L33) and [Open MPI](https://github.com/open-mpi/ompi/blob/master/opal/include/opal/sys/powerpc/atomic.h#L30-L32) are incorrect which lead to confusion.
    - A `lwsync` is a **write memory barrier**
    - A `eieio` is a read memory barrier (we think - need to verify)
    - PRs to fix the PPC atomics are in development.

rhc54:
That would be much appreciated, if you have time

src/client/pmix_client_get.c
```diff
@@ -186,12 +187,14 @@ static void _value_cbfunc(pmix_status_t status, pmix_value_t *kv, void *cbdata)
     pmix_cb_t *cb = (pmix_cb_t*)cbdata;
     pmix_status_t rc;
 
+    PMIX_ACQUIRE_OBJECT(cb);
```
jjhursey: I don't think this Acquire is required since it's called from `_getnbfn`, but it's not hurting anything and I suppose it protects us in the future if we call this from another location.

src/include/pmix_globals.h
```diff
@@ -351,6 +352,7 @@ PMIX_CLASS_DECLARATION(pmix_info_caddy_t);
     do {                                        \
         while ((a)) {                           \
             usleep(10);                         \
+            PMIX_ACQUIRE_OBJECT((a));           \
```
jjhursey: The `PMIX_ACQUIRE_OBJECT` should be just outside of the loop. We don't need a memory barrier on `(a)` since it's volatile (and if it takes a while for the cache to flush it's not a big slowdown). Before we leave the macro though we need to do the `PMIX_ACQUIRE_OBJECT` once to force the barrier.

rhc54: okay, will change

PR #388, opened 2017-06-05
client: Missing unlock in PMIx_Finalize path

jjhursey:


jjhursey:
That's fine. I was just working though that code and noticed the missing unlock.

PR #390, opened 2017-06-06
Minor cleanups

rhc54:


rhc54:
Mellanox is hung, spinning forever while it tries to report

PR #387, opened 2017-06-05
Memory barrier fixes for thread shifting

jjhursey:
* Replaces PR #385
 * Port over the atomics code
 * Add memory barriers from PR #386 (which is based upon discussion in PR #384
 * Fixes Issue #347

PR #391, opened 2017-06-06
atomics/powerpc: Fix WMB instruction

jjhursey:
* `lwsync` is a write memory barrier.
    - `eieio` is really not meant for this type of operation.
 * `lwsync` can also be used for the read memory barrier according to
   my reading of the of the Power 8 ISA docs (v2.07)
    - https://www-01.ibm.com/marketing/iwm/iwm/web/reg/download.do?source=swg-opower&S_PKG=dl&lang=en_US&cp=UTF-8


jjhursey:
References PR #386, PR #387, Issue #347

Note that this change will also be pushed to Open MPI for their atomics.

PR #395, opened 2017-06-09
Remove all the sleep-based thread syncs

rhc54:
Replace them with the standard conditioned-wait/signal methods


rhc54:
Sigh - once again, I am at a loss to understand this test result. The log shows nothing but "OK" on the tests. Yet at the very end, it saids "exit 2" and marks it as failed.

Can someone tell me what is supposedly wrong?

PR #407, opened 2017-06-23
Protect the PMIX_INFO_FREE macro from NULL data arrays. 

rhc54:
Add missing support for pmix_info_t objects in pmix_data_array_t as these are typically returned from calls to PMIx_Query

This needs to go over to the 2.x branch


rhc54:
I guess this Jenkins system has not yet been fixed

PR #417, opened 2017-07-05
Add regex framework so we can handle multiple different regular expression formats

rhc54:
Move all the PMIx internal regex support to the "native" component, and make it always be the highest priority as it is the default. When parsing a regex, each module will check for the starting string to indicate the regex generator that was used. "pmix" will indicate native.

We probably won't be generating other formats, so allow any given API to be NULL.


rhc54:
Next steps will add a slurm component, and (optionally) remove automatic expansion of the regex into local storage. We'll provide an option to replace the latter with calls to extract the info from the regex so we can reduce the footprint at scale.

PR #418, opened 2017-07-07
configury: use default libevent headers unless specified otherwise

ggouaillardet:

### DIFF
--- a/config/pmix_setup_libevent.m4
+++ b/config/pmix_setup_libevent.m4
@@ -3,6 +3,8 @@
 # Copyright (c) 2009-2015 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2013      Los Alamos National Security, LLC.  All rights reserved.
 # Copyright (c) 2013-2017 Intel, Inc. All rights reserved.
+# Copyright (c) 2017      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
@@ -39,8 +41,11 @@ AC_DEFUN([_PMIX_LIBEVENT_EMBEDDED_MODE],[
     AC_MSG_CHECKING([for libevent])
     AC_MSG_RESULT([assumed available (embedded mode)])
 
-    PMIX_EVENT_HEADER="$with_libevent_header"
-    PMIX_EVENT2_THREAD_HEADER="$with_libevent_header"
+    AS_IF([test -z "$with_libevent_header" || test "$with_libevent_header" = "yes"],
+          [PMIX_EVENT_HEADER="<event.h>"
+           PMIX_EVENT2_THREAD_HEADER="<event2/thread.h>"],
+          [PMIX_EVENT_HEADER="$with_libevent_header"
+           PMIX_EVENT2_THREAD_HEADER="$with_libevent_header"])
 
  ])
 


ggouaillardet:
initially, i tried to fix some ompi warnings (```MB``` is defined twice in both ompi and pmix atomics)
then i found that pmix can directly include ```libevent``` headers instead of the ompi libevent glue.

once merged and it has landed in ompi master, the following trivial patch will be needed
```patch
diff --git a/opal/mca/pmix/pmix2x/configure.m4 b/opal/mca/pmix/pmix2x/configure.m4
index 109491d..ec25383 100644
--- a/opal/mca/pmix/pmix2x/configure.m4
+++ b/opal/mca/pmix/pmix2x/configure.m4
@@ -14,7 +14,7 @@
 #                         All rights reserved.
 # Copyright (c) 2010-2017 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2013-2017 Intel, Inc.  All rights reserved.
-# Copyright (c) 2015-2016 Research Organization for Information Science
+# Copyright (c) 2015-2017 Research Organization for Information Science
 #                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
 #
@@ -61,7 +61,7 @@ AC_DEFUN([MCA_opal_pmix_pmix2x_CONFIG],[
         opal_pmix_pmix2x_timing_flag=--disable-pmix-timing
     fi
 
-    opal_pmix_pmix2x_args="--with-pmix-symbol-rename=OPAL_MCA_PMIX2X_ $opal_pmix_pmix2x_sm_flag $opal_pmix_pmix2x_timing_flag --without-tests-examples --disable-pmix-backward-compatibility --disable-visibility --enable-embedded-libevent --with-libevent-header=\\\"opal/mca/event/$opal_event_base_include\\\" --enable-embedded-mode"
+    opal_pmix_pmix2x_args="--with-pmix-symbol-rename=OPAL_MCA_PMIX2X_ $opal_pmix_pmix2x_sm_flag $opal_pmix_pmix2x_timing_flag --without-tests-examples --disable-pmix-backward-compatibility --disable-visibility --enable-embedded-libevent --enable-embedded-mode"
     AS_IF([test "$enable_debug" = "yes"],
           [opal_pmix_pmix2x_args="--enable-debug $opal_pmix_pmix2x_args"
            CFLAGS="$OPAL_CFLAGS_BEFORE_PICKY $OPAL_VISIBILITY_CFLAGS -g"],
```

rhc54:
Actually, I believe the current configure (which, just to be clear for any reader, is in OMPI, not here) is correct. We don't want to use an OPAL abstraction. We want to go directly to the libevent headers as that is what we build against.

Is there some problem this is trying to address?

ggouaillardet:
well, right now, we are using the OPAL abstraction layer.
from ompi, pmix configury is invoked with
```--enable-embedded-libevent --with-libevent-header=\\\"opal/mca/event/$opal_event_base_include\\\"```
and ```$opal_event_base_include``` is either ```libevent2022/libevent.h``` or ```external/external.h```, which are both OPAL abstractions.
when building ompi, you can double check the ```PMIX_EVENT_HEADER``` and ```PMIX_EVENT2_THREAD_HEADER``` macros in the generated ```pmix_config.h```.

so as far as i am concerned, the OPAL abstraction layer is currently used.
this PR plus the inline patch for ompi do what you want.

(and based on your comment, you might even want to get rid of the ```--with-libevent-header``` option)

makes sense ?

**Comment by rhc54** [comment]
IIRC, the reason we did this was because we were accidentally picking up headers from a system-installed version of libevent even though we linked against the internal version. I'm a tad nervous about changing it as it seems to be working correctly - is there problem you are trying to address?

**Comment by ggouaillardet** [comment]
from my point of view, it is not working correctly.
if it were, the ```MB``` macro would not be redefined when building the embedded PMIX since the OPAL abstraction would have not been pulled in the first place.

an other option is not to change pmix (embedded pmix nor its glue) but instead split ```libevent2022/libevent.h``` and ```external/external.h``` into a very minimal (e.g. no reference to OPAL) and the full header so ```$opal_event_base_include``` can point to the minimal header instead
in case of ```libevent2022/libevent.h``` this minimal header file would only contain
```c
#include <libevent/event.h>
#include <libevent/include/event2/thread.h>
```

rhc54:
Fair point - let me check first to ensure that we don't recreate the original problem as it was a real pain to chase down.

rhc54:
I'm not seeing any issues, so I think this is fine - not sure if enough other things have changed that caused it. Regardless, I prefer this solution to what we currently have as it seems cleaner to me.

I also agree about just removing that ```--with-libevent-header``` configure option and will look into it.

PR #419, opened 2017-07-07
GDS Fixes

karasevb:
This PR fixes a modex problems, which appears for more than two nodes.
Refs [#3787](https://github.com/open-mpi/ompi/pull/3787)
There are our MTT results with `dstore` used: 
```
   Phase    Section           MPI Version   Duration Pass Fail Time  Skip
                                                               Out
   Test Run atest             topic/pmix210 22:21    41        1
   Test Run ctxalloc          topic/pmix210 00:03    2
   Test Run trivial           topic/pmix210 20:16    11        1
   Test Run mpi-test-suite    topic/pmix210 00:13    2
   Test Run fp-sum-test       topic/pmix210 02:10    2
   Test Run mpi-small-tests   topic/pmix210 43:25    50

Totals

    Tests            Failed           Passed           Duration
    116              2                114              01:51:42
```
_Note:_ tests uses both modex methods (full/direct)

The results without `dstore` are passed all of tests. I will investigate a `dstore` hangs.

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -270,11 +270,8 @@ static void _getnb_cbfunc(struct pmix_peer_t *pr,
     pmix_status_t rc, ret;
     pmix_value_t *val = NULL;
     int32_t cnt;
-    pmix_proc_t proc, proct;
-    pmix_byte_object_t bo;
-    pmix_buffer_t pbkt;
+    pmix_proc_t proc;
     pmix_kval_t *kv;
-    pmix_peer_t *peer;
 
     pmix_output_verbose(2, pmix_globals.debug_output,
                         "pmix: get_nb callback recvd");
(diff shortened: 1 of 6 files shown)


src/server/pmix_server_get.c
```diff
@@ -631,7 +638,7 @@ static pmix_status_t _satisfy_request(pmix_nspace_t *nptr, pmix_rank_t rank,
         cb.proc = &proc;
         cb.scope = scope;
         cb.copy = false;
-        PMIX_GDS_FETCH_KV(rc, peer, &cb);
+        PMIX_GDS_FETCH_KV(rc, pmix_globals.mypeer, &cb);
```
rhc54: Something about this doesn't feel quite right. In the above code, we store the incoming data in the GDS module for a peer in this nspace if the scope was LOCAL, and in ```pmix_globals.mypeer``` otherwise - we point the ```peer``` object at the GDS module where we store the data. So why then are we looking for the data strictly in ```pmix_globals.mypeer```? Won't this only be correct for scope != LOCAL?

src/server/pmix_server_get.c
```diff
@@ -531,7 +531,9 @@ static pmix_status_t _satisfy_request(pmix_nspace_t *nptr, pmix_rank_t rank,
      * the corresponding GDS to retrieve the data. Otherwise,
      * the data will have been stored under our GDS */
     if (0 < nptr->nlocalprocs) {
-        *local = true;
+        if (local) {
```
karasevb: There is here: [pmix_server_get.c:271](https://github.com/pmix/pmix/blob/ef707ab63511bbe38bdb684d09e8cbb5ff63cf17/src/server/pmix_server_get.c#L721)

```C
pmix_status_t rc;
    rc = _satisfy_request(nptr, rank, scd, req->cbfunc, req->cbdata, NULL);
    if( PMIX_SUCCESS != rc ){
```

rhc54: Ah, okay - thanks! I missed that line.

karasevb:
bot:retest

src/server/pmix_server_get.c
karasevb: Right, I agree. This change has been canceled.

src/client/pmix_client_get.c
```diff
@@ -523,16 +524,22 @@ static void _getnbfn(int fd, short flags, void *cbdata)
         }
     }
 
+    my_nspace = (0 == strncmp(cb->pname.nspace, pmix_globals.myid.nspace, PMIX_MAX_NSLEN));
+    my_rank = (pmix_globals.myid.rank == cb->pname.rank);
```
karasevb: Fixed

src/client/pmix_client_get.c
```diff
@@ -394,7 +394,7 @@ static void _getnb_cbfunc(struct pmix_peer_t *pr,
             /* fetch the data from my peer module - since we are passing
              * it back to the user, we need a copy of it */
             cb->copy = true;
-            PMIX_GDS_FETCH_KV(rc, pmix_globals.mypeer, cb);
+            PMIX_GDS_FETCH_KV(rc, pmix_client_globals.myserver, cb);
```
karasevb: The comment was changed. Here the server's GDS is used, because the server puts all the requested keys into `dstore` and responds to the client with an empty message (only status, without keys), then the client finds the requested keys in `dstore`. Otherwise, without `dstore`, the server's GDS will be the same as the client.

karasevb:
bot:retest

src/client/pmix_client_get.c
rhc54: I see - makes sense. Thanks!

PR #428, opened 2017-07-14
Update VERSION to 2.1 as there have been no API changes thus far. Fix visibility issues

rhc54:

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -13,8 +13,8 @@
 # major, minor, and release are generally combined in the form
 # <major>.<minor>.<release>.
 
-major=3
-minor=0
+major=2
+minor=1
 release=0
 
 # greek is used for alpha or beta release tags.  If it is non-empty,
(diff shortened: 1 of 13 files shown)


rhc54:
@artpol84 We see this periodically from Jenkins:
```
10:10:25 libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, './config'.
10:10:25 libtoolize:   error: copying '/hpc/local/share/libtool/build-aux/ltmain.sh' to './config/' failed
10:10:25 libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, './config'.
```
Any idea what might be causing it? Re-running the tests seems to always resolve the problem, so it looks like some kind of race condition maybe?

rhc54:
bot:retest

artpol84:
@rhc54 sorry for the late response.
I will try to investigate.
@Di0gen, please have a look

PR #429, opened 2017-07-20
Fix GDS

rhc54:
This combines patches from @karasevb #419, rebased to master and with conflicts resolved. I have also temporarily disabled the ds12/GDS component so we can commit this downstream to OMPI for testing. Will post MTT results shortly.


karasevb:
@rhc54 Let's close this PR. I have got final fixes which will commit today in my original PR. Then PR #419 will be ready to merge.

rhc54:
Excellent! Thanks - that was going to be my proposal. If necessary, I can disable ds12 in the OMPI tree.

PR #435, opened 2017-07-21
Silence error log of "success"

rhc54:

### DIFF
--- a/src/mca/gds/ds12/gds_dstore.c
+++ b/src/mca/gds/ds12/gds_dstore.c
@@ -2886,15 +2886,19 @@ static pmix_status_t dstore_store_modex(struct pmix_nspace_t *nspace,
         PMIX_DESTRUCT(&pbkt);
         return rc;
     }
+    /* don't store blobs to the sm dstore from local clients */
+    if (_my_client(proc.nspace, proc.rank)) {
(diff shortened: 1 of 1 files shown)


rhc54:
@karasevb @artpol84 Just an FYI - upstreaming from OMPI

src/mca/gds/ds12/gds_dstore.c
```diff
@@ -2886,15 +2886,19 @@ static pmix_status_t dstore_store_modex(struct pmix_nspace_t *nspace,
         PMIX_DESTRUCT(&pbkt);
         return rc;
     }
+    /* don't store blobs to the sm dstore from local clients */
+    if (_my_client(proc.nspace, proc.rank)) {
+        bo->bytes = pbkt.base_ptr;
```
artpol84: @rhc54 thank you.
@karasevb 
* I'd rather do a "goto" to the end of the function (https://github.com/pmix/pmix/pull/435/files#diff-d609ed5f77a8e88bbd6f266f4776fd25L2923) here to avoid code duplication.
* I'm concerned about comment "don't store blobs to the sm dstore from local clients", what blobs are mentioned here? I think that PMIX_LOCAL scope should go to SM, while REMOTE indeed should not.
*

PR #439, opened 2017-07-26
Silence coverity

karasevb:
Address:
CID 1375987: Time of check time of use (TOCTOU)
CID 1415599: Resource leaks  (RESOURCE_LEAK)
CID 1415602: Null pointer dereferences  (FORWARD_NULL)
CID 1415601: Resource leaks  (RESOURCE_LEAK)    
CID 1415600: Uninitialized variables  (UNINIT)

Was marked as ignored:
CID 1375986: Security best practices violations (TOCTOU)

### DIFF
--- a/src/mca/gds/ds12/gds_dstore.c
+++ b/src/mca/gds/ds12/gds_dstore.c
@@ -2396,7 +2396,7 @@ inline pmix_status_t _dstore_fetch(const char *nspace, pmix_rank_t rank, const c
         kval_cnt = rinfo->count;
 
         /*  Initialize array for all keys of rank */
-        if ((NULL == key) || (kval_cnt > 0)) {
+        if ((NULL == key) && (kval_cnt > 0)) {
             kval = (pmix_value_t*)malloc(sizeof(pmix_value_t));
             if (NULL == kval) {
                 return PMIX_ERR_NOMEM;
@@ -2543,7 +2543,7 @@ done:
     }
 
     if( rc != PMIX_SUCCESS ){
-        if( NULL == key ) {
+        if ((NULL == key) && (kval_cnt > 0)) {
             if( NULL != info ) {
                 PMIX_INFO_FREE(info, ninfo);
             }
--- a/src/mca/gds/ds12/gds_dstore.c
+++ b/src/mca/gds/ds12/gds_dstore.c
@@ -2304,8 +2304,10 @@ inline pmix_status_t _dstore_fetch(const char *nspace, pmix_rank_t rank, const c
         return rc;
     }
 
-    if (kvs) {
-        *kvs = NULL;
+    if (NULL == kvs) {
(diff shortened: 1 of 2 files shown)


src/mca/gds/ds12/gds_dstore.c
```diff
@@ -2304,10 +2304,6 @@ inline pmix_status_t _dstore_fetch(const char *nspace, pmix_rank_t rank, const c
         return rc;
     }
 
-    if (kvs) {
-        *kvs = NULL;
```
artpol84: This doesn't sound right to me. it doesn't address NULL-pointer dereference error.
The problem is in the case where `kvs == NULL` (which is assumed by this if) and not with `*kvs == NULL`.
Instead of removing this you need to require `kvs != NULL` and error exit if this is not true. 
It doesn't make sense to call this function if you don't want to receive the answer.

artpol84: And this has to be done at the very begining of the function.

src/mca/gds/ds12/gds_dstore.c
```diff
@@ -2396,23 +2392,20 @@ inline pmix_status_t _dstore_fetch(const char *nspace, pmix_rank_t rank, const c
...
-            if (NULL == info) {
-                rc = PMIX_ERR_NOMEM;
-                goto done;
-            }
-
-            PMIX_VALUE_CONSTRUCT(kval);
             kval->type = PMIX_DATA_ARRAY;
             kval->data.darray = (pmix_data_array_t*)malloc(sizeof(pmix_data_array_t));
             if (NULL == kval->data.darray) {
+                PMIX_VALUE_RELEASE(kval);
```
artpol84: I don't think you can `PMIX_VALUE_RELEASE(kval)` if you didn't `PMIX_VALUE_CONSTRUCT(kval)`

src/mca/gds/ds12/gds_dstore.c
```diff
@@ -2396,23 +2392,20 @@ inline pmix_status_t _dstore_fetch(const char *nspace, pmix_rank_t rank, const c
...
             }
-
-            ninfo = kval_cnt;
-            PMIX_INFO_CREATE(info, ninfo);
-            if (NULL == info) {
-                rc = PMIX_ERR_NOMEM;
-                goto done;
-            }
-
-            PMIX_VALUE_CONSTRUCT(kval);
```
artpol84: I think we still need it

src/mca/gds/ds12/gds_dstore.c
artpol84: For the reference: `	** CID 1415602:  Null pointer dereferences  (FORWARD_NULL)`

artpol84:
@karasevb, please see if we need to port this to the release branches.

PR #442, opened 2017-08-04
Fix incorrect syntax in several places where "test" is used in .m4 files - equality is "=", not "=="

rhc54:
(cherry picked from commit aa15a87855209961cc6926f8699a7d8750eda29d)


config/pmix.m4
```diff
@@ -945,14 +944,15 @@ AC_DEFINE_UNQUOTED([PMIX_ENABLE_DSTORE],
                  [if want shared memory dstore feature])
 
 #
+#
```
jjhursey: There is a `if test "$enable_dstore" == "no" ; then` just above this line that was missed.

rhc54:
Got it - thanks!

PR #446, opened 2017-08-04
Check for incorrect install locations and no usable components

rhc54:
Check for failure to find any usable components in frameworks that require at least one, print an informative error message, and return an error.

If the user provides us with a PMIX_INSTALL_PREFIX that points to an unusable location (either nonexistent or not available to us), or none of the default search locations exist (because the installation was moved), then provide a nice message so they know what is wrong and return an error.

### DIFF
--- a/src/mca/pdl/pdlopen/pdl_pdlopen_module.c
+++ b/src/mca/pdl/pdlopen/pdl_pdlopen_module.c
@@ -4,7 +4,7 @@
  * Copyright (c) 2015      Los Alamos National Security, LLC. All rights
  *                         reserved.
  * Copyright (c) 2016      IBM Corporation.  All rights reserved.
- * Copyright (c) 2016      Intel, Inc. All rights reserved
+ * Copyright (c) 2016-2017 Intel, Inc. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
--- a/src/runtime/help-pmix-runtime.txt
+++ b/src/runtime/help-pmix-runtime.txt
@@ -40,3 +40,20 @@ a fatal situation, but may negatively impact your launch performance.
 If you feel you have received this warning in error, or wish to ignore
 it in the future, you can disable it by setting the PMIx MCA parameter
 "pmix_suppress_missing_data_warning=1"
+#
+[no-plugins]
+We were unable to find any usable plugins for the %s framework. This PMIx
+framework requires at least one plugin in order to operate. This can be caused
+by any of the following:
+
+* we were unable to build any of the plugins due to some combination
+  of configure directives and available system support
+
+* no plugin was selected due to some combination of MCA parameter
+  directives versus built plugins (i.e., you excluded all the plugins
+  that were built and/or could execute)
+
+* the PMIX_INSTALL_PREFIX is pointing to a location that doesn't include
+  any plugins for this framework.
+
+Please check your installation and environment.
--- a/src/runtime/help-pmix-runtime.txt
+++ b/src/runtime/help-pmix-runtime.txt
@@ -53,7 +53,8 @@ by any of the following:
   directives versus built plugins (i.e., you excluded all the plugins
   that were built and/or could execute)
 
-* the PMIX_INSTALL_PREFIX is pointing to a location that doesn't include
-  any plugins for this framework.
+* the PMIX_INSTALL_PREFIX environment variable, or the MCA parameter
+  "mca_base_component_path", is set and doesn't point to any location
+  that includes at least one usable plugin for this framework.
 
 Please check your installation and environment.
--- a/src/mca/base/pmix_mca_base_component_repository.c
+++ b/src/mca/base/pmix_mca_base_component_repository.c
@@ -15,7 +15,7 @@
  *                         reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
- * Copyright (c) 2016      Intel, Inc. All rights reserved.
+ * Copyright (c) 2016-2017 Intel, Inc. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -179,9 +179,12 @@ static int file_exists(const char *filename, const char *ext)
 
 int pmix_mca_base_component_repository_add (const char *path)
 {
+    int rc=PMIX_SUCCESS;
+
 #if PMIX_HAVE_PDL_SUPPORT
     char *path_to_use = NULL, *dir, *ctx;
     const char sep[] = {PMIX_ENV_SEP, '\0'};
+    bool found_one = false;
 
     if (NULL == path) {
         /* nothing to do */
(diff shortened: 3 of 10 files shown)


rhc54:
Refs https://github.com/open-mpi/ompi/issues/3980

src/mca/pdl/pdlopen/pdl_pdlopen_module.c
```diff
@@ -175,7 +175,26 @@ static int pdlopen_foreachfile(const char *search_path,
 
         dp = opendir(dirs[i]);
         if (NULL == dp) {
-            ret = PMIX_ERR_IN_ERRNO;
+            if (0 == i) {
```
artpol84: If my understanding is correct we are ok if some of the directories are missing:
```shell
PMIXP_INSTALL_PREFIX="x:y:z"
```
As it is implemented now - you are failing if `x` can't be opened. But what if `x` is missing, but `y` is what we need?

src/runtime/help-pmix-runtime.txt
```diff
@@ -40,3 +40,20 @@ a fatal situation, but may negatively impact your launch performance.
...
+by any of the following:
+
+* we were unable to build any of the plugins due to some combination
+  of configure directives and available system support
+
+* no plugin was selected due to some combination of MCA parameter
+  directives versus built plugins (i.e., you excluded all the plugins
+  that were built and/or could execute)
+
+* the PMIX_INSTALL_PREFIX is pointing to a location that doesn't include
```
artpol84: I mean that user could be surprised if you tell him `the PMIX_INSTALL_PREFIX is pointing to a location...` while he has nothing to do with this var :)

src/mca/pdl/pdlopen/pdl_pdlopen_module.c
rhc54: This is in keeping with what is in OPAL. If x can't be opened, we immediately return and never look beyond that point. All I'm doing is emitting the warning if that happens with the very first entry as this would correspond with the typical use case of PMIX_INSTALL_PREFIX being incorrectly set. Rather than doing something different here, I'd suggest taking that question upstream to the OPAL community - there was a reason behind the original logic.

src/runtime/help-pmix-runtime.txt
rhc54: I'm not sure I comprehend the concern. We have no way of determining who set the envar, or why they set it - all we can say is that the envar is set and pointing to an incorrect location. What more would you have us say?

src/mca/pdl/pdlopen/pdl_pdlopen_module.c
artpol84: Sound good.

jjhursey: Yeah this was what I wanted to test today. In Open MPI `mca_base_component_path` provides a `:` separated list of paths. By default it contains two - one to the prefix install dir and the other to the user's home directory

> MCA mca base: parameter "mca_base_component_path" (current value: "/u/jjhursey/local/ompi/current-dev/lib/openmpi:/u/jjhursey/.openmpi/components", data source: default, level: 9 dev/all, type: string, synonyms: mca_component_path)

I just want to make sure we preserve the functionality that the `${HOME}/.openmpi` version is allowed to not exist, but at least one of the values must exist. I need to setup a build to confirm that this was preserved (I believe it was). I agree with Ralph that if we want something better we should talk to the OMPI community. But the OMPI version of this commit has not been merged yet, so we might be able to tackle it all at once.

rhc54: I chatted with Nathan and he sees no issue with continuing the search beyond the first failure, so I've added that here and fixed the bug being seen in the OMPI version

jjhursey: So I did some testing. I think this PR (and it's upstream version) are doing the right thing now.

Assume I have three different paths:
 * **A** `/u/jjhursey/local/current-ompi-dev/lib/openmpi` : Directory exists, and contains components
 * **B** `/u/jjhursey/.openmpi/components` : Directory does not exist
 * **C** `/u/jjhursey/` : Directory exists, but contains no components

Here are the behaviors that I observed:

### With this branch (before Ralph's most recent change):

 * `A:B` (Open MPI default). Everything works fine.
 * `B:A` Error message displayed by this PR
```
-------------------------------------------------------
No usable directories were found in the provided path
when searching for available plugins. This usually indicates
either that the OMPI installation was moved, or that the
OPAL_PREFIX environmental variable is set and pointing
to an incorrect or non-existent location.

Please correct the situation and try again.
-------------------------------------------------------
```
 * `C:A:B` `mpirun` does not work correctly
```
shell$ mpirun -np 4 -mca mca_base_component_path C:A:B hello_c
mpirun: Error: unknown option "-np"
```

### Comparing that with OMPI `master`:
 * `A:B` (Open MPI default). Everything works fine.
 * `B:A` `mpirun` does not work correctly
```
shell$ mpirun -np 4 -mca mca_base_component_path C:A:B hello_c
mpirun: Error: unknown option "-np"
```
 * `C:A:B` `mpirun` does not work correctly
```
shell$ mpirun -np 4 -mca mca_base_component_path C:A:B hello_c
mpirun: Error: unknown option "-np"
```

### After Ralph's most recent change to the upstream PR

 * `A:B` (Open MPI default). Everything works fine.
 * `B:A` Everything works fine.
 * `B` Error message displayed by this PR
```
-------------------------------------------------------
No usable directories were found in the provided path
when searching for available plugins. This usually indicates
either that the OMPI installation was moved, or that the
OPAL_PREFIX environmental variable is set and pointing
to an incorrect or non-existent location.

Please correct the situation and try again.
-------------------------------------------------------
```
 * `C:A:B` Everything works fine.
 * `C:B` `mpirun` does not work correctly
```
shell$ mpirun -np 4 -mca mca_base_component_path C:A:B hello_c
mpirun: Error: unknown option "-np"
```

rhc54: Hmmm...only thing I don't like is that C:B didn't give you a reasonable error message. I would never go from an unknown np option to guessing that my component path was wrong. I guess the reason my error message didn't get triggered is that at least one directory existed - but no components could be found.

Problem is that some frameworks are okay without components, so I don't know how we can generalize an error message. I'm guessing mpirun came up with that error because the ORTE schizo framework found no usable components, and thus the cmd line was never defined. I'll see if I can at least deal with that one.

rhc54: I updated that scenario in OMPI so you now get a better error message:
```
$ mpirun -np 2 -mca mca_base_component_path /home/common hostname
--------------------------------------------------------------------------
We were unable to find any usable plugins for the SCHIZO framework. This
framework requires at least one plugin in order to operate. This can be caused
by any of the following:

* we were unable to build any of the plugins due to some combination
  of configure directives and available system support

* no plugin was selected due to some combination of MCA parameter
  directives versus built plugins (i.e., you excluded all the plugins
  that were built and/or could execute)

* the OPAL_PREFIX environment variable is doesn't point to any location that
  includes at least one usable plugin for this framework.

Please check your installation and environment.
--------------------------------------------------------------------------
```

rhc54:
I don't think the upstream OPAL problem translates to us here as we never do a partial init of the opal-transferred infrastructure. Once we see the problem up there, we can see if something needs to be done here and transfer it across.

src/mca/pdl/pdlopen/pdl_pdlopen_module.c
jjhursey: Awesome 👍  Thanks!

PR #449, opened 2017-08-09
Change the error to a warning if --enable-builtin-atomics is given and builtin atomics aren't available, and default to the inline atomics

rhc54:
Per recommendation from upstream

### DIFF
--- a/config/pmix_config_asm.m4
+++ b/config/pmix_config_asm.m4
@@ -923,7 +923,7 @@ AC_DEFUN([PMIX_CONFIG_ASM],[
     AS_IF([test "$pmix_cv_asm_builtin" = "BUILTIN_NO" && test "$enable_builtin_atomics" != "no"],
           [PMIX_CHECK_SYNC_BUILTINS([pmix_cv_asm_builtin="BUILTIN_SYNC"], [])])
     AS_IF([test "$pmix_cv_asm_builtin" = "BUILTIN_NO" && test "$enable_builtin_atomics" = "yes"],
-          [AC_MSG_ERROR([__sync builtin atomics requested but not found.])])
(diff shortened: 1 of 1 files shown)


ibm-ompi:
The IBM CI (PGI Compiler) build failed! Please review the log, linked below.

 Gist: https://gist.github.com/c5a0a6308c6827c40508cc23f6136caf

rhc54:
@jjhursey do you know where this -pthread is coming from, and how we get rid of it? I can't find any differences in the config_threads and config_pthreads.m4 files from what is in opal.

rhc54:
@ggouaillardet Perhaps you remember how we resolved this -pthread issue?

ibm-ompi:
The IBM CI (PGI Compiler) build failed! Please review the log, linked below.

 Gist: https://gist.github.com/0a080abd257807691025c136aa509368

jjhursey:
I remember having to fix this recently and it was relatively easy... I'll look today

jjhursey:
bot:ibm:pgi:retest

jjhursey:
You were correct that the PMIx side is doing the right thing. So it wasn't the easy fix I was thinking about, it was a different fix.

The problem was when PMIx invoked `libtool` and added the `libevent` dependency which was compiled with gcc. It's profile hard coded the `-pthread`. So in the CI setup I build a separate `libevent` with the PGI compilers and that made everything happy. I'm not exactly sure how difficult it would be to add a configure check to the PMIx side of things, but we can leave that for future work.

rhc54:
Maybe we just need to look at OMPI's external libevent integration? It must deal with it, yes?

jjhursey:
I'm not sure. I've not tried to use PGI to build Open MPI with an external component. It might just error like this. I agree that's a good place to look. I'll create a ticket to figure that out since it's a PGI issue and that's something we can help sort through.

PR #454, opened 2017-08-10
Per 08/10/2017 telecon: enable test suite build when visibility is ON.

artpol84:
Now pmix utility functions like containers (lists) and opal-like OS functions
are exported to allow plugins to work. This means that tests are ok with
enabled visibility.


artpol84:
Let's wait for the CI and merge then.

PR #459, opened 2017-08-11
Fixed provide the unpacked status of fence

karasevb:
The error status is not detected for clients when server
responded with error message for Fence request.


artpol84:
@karasevb please check to what release branches we need to port this.

PR #460, opened 2017-08-12
Export the class instances

rhc54:


rhc54:
I think so, yes - let's see

PR #471, opened 2017-08-15
ptl/tcp: do not use an already free'd pointer

ggouaillardet:
gds is a pointer within msg, so msg can only be free'd after gds is not more used


ggouaillardet:
fwiw, this issue can easily be evidenced with ompi master and the following command
```
valgrind mpirun -np 1 ./hello_c
```

PR #477, opened 2017-08-19
Backport conversion of PMIX_CMD to PMIX_COMMAND wherever it was used …

rhc54:
…to support backward compatibility


rhc54:
This breaks compatibility with v2.0.0 because it changes the data type of pmix_cmd_t from int to uint8.

PR #490, opened 2017-08-30
Provide a mechanism by which we can check a peer's version - we'll ne…

rhc54:
…ed it for places where versions differ in behavior

### DIFF
--- a/src/server/pmix_server.c
+++ b/src/server/pmix_server.c
@@ -2103,7 +2103,8 @@ static pmix_status_t server_switchyard(pmix_peer_t *peer, uint32_t tag,
         PMIX_ERROR_LOG(rc);
         return rc;
     }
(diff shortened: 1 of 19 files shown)


src/server/pmix_server.c
```diff
@@ -2103,7 +2103,8 @@ static pmix_status_t server_switchyard(pmix_peer_t *peer, uint32_t tag,
         PMIX_ERROR_LOG(rc);
         return rc;
     }
-    pmix_output_verbose(2, pmix_globals.debug_output,
+   // pmix_output_verbose(2, pmix_globals.debug_output,
+                        pmix_output(0,
```
artpol84: @rhc54 Is this intentional or for debug purposes?

rhc54:
believe this has since been cleaned up, but thanks for noticing!

PR #492, opened 2017-08-30
Complete the (basic) support for v1 client working with a v2 server.

rhc54:
The simpclient test now passes

### DIFF
--- a/src/mca/bfrops/pmix1/bfrop_pmix1.c
+++ b/src/mca/bfrops/pmix1/bfrop_pmix1.c
@@ -417,6 +417,11 @@ int pmix1_v2_to_v1_datatype(pmix_data_type_t v2type)
                 v1type = 22;
                 break;
 
+            case 40:
+                /* proc rank is just an int in v1 */
+                v1type = 6;
+                break;
+
             case 22:
             case 23:
             case 24:
--- a/src/mca/bfrops/pmix1/pack.c
+++ b/src/mca/bfrops/pmix1/pack.c
@@ -641,20 +641,20 @@ pmix_status_t pmix1_bfrop_pack_pdata(pmix_buffer_t *buffer, const void *src,
 pmix_status_t pmix1_bfrop_pack_buf(pmix_buffer_t *buffer, const void *src,
                                    int32_t num_vals, pmix_data_type_t type)
 {
-    pmix_buffer_t **ptr;
+    pmix_buffer_t *ptr;
     int32_t i;
     pmix_status_t ret;
 
-    ptr = (pmix_buffer_t **) src;
+    ptr = (pmix_buffer_t *) src;
 
     for (i = 0; i < num_vals; ++i) {
         /* pack the number of bytes */
-        if (PMIX_SUCCESS != (ret = pmix1_bfrop_pack_sizet(buffer, &ptr[i]->bytes_used, 1, PMIX_SIZE))) {
+        if (PMIX_SUCCESS != (ret = pmix1_bfrop_pack_sizet(buffer, &ptr[i].bytes_used, 1, PMIX_SIZE))) {
             return ret;
         }
         /* pack the bytes */
-        if (0 < ptr[i]->bytes_used) {
-            if (PMIX_SUCCESS != (ret = pmix1_bfrop_pack_byte(buffer, ptr[i]->base_ptr, ptr[i]->bytes_used, PMIX_BYTE))) {
+        if (0 < ptr[i].bytes_used) {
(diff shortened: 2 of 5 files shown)


rhc54:
Should also note that this is true _only_ with dstore disabled in v1 and PMIX_MCA_gds=hash on the v2 server.

@karasevb @artpol84 The dstore support doesn't appear to work across v1/v2 at this time. I get errors if I don't disable it on both sides. I suspect we'll need a GDS component to specifically support v1 dstore, and we need to figure out how to at least detect the corner cases (hopefully deal with them).

An example error: if ds12 component is active, but v1 was configured with dstore disabled, then
```
$ ./simptest -u -n 2 -e /home/common/pmix/v1.2/test/simple/.libs/simpclient
Testing version 2.1.0
[rhc001:242370] SERVER: ERRHANDLER REGISTRATION CALLBACK CALLED WITH STATUS 0, ref=0
[rhc001:242386] Security mode none is not available
[rhc001:242387] Security mode none is not available
[rhc001:242386] Client ns foobar rank 0: Running
[rhc001:242387] Client ns foobar rank 1: Running
[rhc001:242386] Client ns foobar rank 0: PMIx_Get universe size failed: -46
[rhc001:242386] Client ns foobar rank 0: Finalizing
[rhc001:242387] Client ns foobar rank 1: PMIx_Get universe size failed: -46
[rhc001:242387] Client ns foobar rank 1: Finalizing
[rhc001:242370] SERVER: FINALIZED foobar:0 WAKEUP 2
[rhc001:242370] SERVER: FINALIZED foobar:1 WAKEUP 2
Client ns foobar rank 1:PMIx_Finalize successfully completed
	pmix.pgm.model:	PMIX
	pmix.mdl.name:	test
	pmix.evname:	SIMPTEST-MODEL
Test finished OK!
$
```
So the client's get thru PMIx_Init and are identified as being v1, but we don't know that dstore was disabled and so we still try to use it (which is why the client fails to get the data). Not sure how we can detect/support that situation.

artpol84:
@rhc54 Let me check if I understood correctly: the problem is that dstor is on on the server but off on the client. Is that correct?

rhc54:
For that specific example, yes - dstore is available on the server, but disabled on the client. Note that the v1 connection handshake doesn't pass any info about dstore being enabled/disabled.

artpol84:
And the problem is that v1 client initiation message does not contain info about dstore support so server has no way to figure out if the remote side is looking into the dstore or not.

I guess what we can do - is to guess it based on client behavior: if v1 client supports dstore it will not request job-level info keys from the server. If it does - it most probably don't have gds support. In this case we need to switch back to messaging. It will require ability to do this switch after the original selection which might require some additional work.

rhc54:
Is that correct? I believe v1 client does actually request the job-level info - at least, that's what I'm seeing happen. Looking at PMIx_Init in v1.2, I see the following code starting at line 231:
```c
    /* send a request for our job info - we do this as a non-blocking
     * transaction because some systems cannot handle very large
     * blocking operations and error out if we try them. */
    req = PMIX_NEW(pmix_buffer_t);
    if (PMIX_SUCCESS != (ret = pmix_bfrop.pack(req, &cmd, 1, PMIX_CMD))) {
        PMIX_ERROR_LOG(ret);
        PMIX_RELEASE(req);
        return ret;
    }
    PMIX_ACTIVATE_SEND_RECV(&pmix_client_globals.myserver, req, job_data, cbdata);
```
If we are using dstore, we just get a marker message sent back - if not, then we get the full data load returned. It assumes that the server and client were both built the same way and so there is no ambiguity.

artpol84:
You may be right. My assumption was that when v1 cli won't find a specific key locally it will try to request it from the server before failing. As in your example it tries to get universe size. And if server sees this request after it published data in the dstore - this is a sign that dstore is not used by the client.
This has two problems:
a) I'm not sure if cli will request job-level info from the server if it can't find it locally.
b) I'm not sure that v1 clients are requesting any of job-level info in PMIx_Init. If they don't - relying on the app to do this is too fragile.

rhc54:
I checked the v1.2 code, and I'm afraid it won't work (though it was a good idea). If the key starts with "pmix" and the rank is wildcard (i.e., we are looking for a job-level piece of data), then we simply fail if it isn't locally found in the client. The assumption was that the job-level info was always provided during PMIx_Init and therefore no reason existed to request it if not found.

I relaxed that in v2.1 as there are times when the RM might not provide optional info at the beginning of time (e.g., to save footprint). However, we are stuck for now with whatever v1.2 and v2.0 will let us do. I'm beginning to think we'll just have to document these corner-cases and ensure there is some kind of error message emitted.

PR #509, opened 2017-09-13
jobinfo: avoid duplication of job information

karasevb:
This fix avoids double unpacking of the jobinfo array, which occurs in nspace register, and then when clients are connected. To provide job data to clients the local cache is used instead of unpacking data from jobinfo array. Since the jobinfo array is not needed now so it has been removed.


rhc54:
I believe this is okay - will need to check the "no-data" directive to ensure that is observed as it is hard to decipher from the diff.

Thanks!

PR #536, opened 2017-10-04
Support singleton operations under PMI-1 and PMI-2

rhc54:


rhc54:
@koomie - see if this solves the problem

PR #562, opened 2017-10-20
Roll to v2.0.3

rhc54:


rhc54:
You make a good point - I had forgotten the a1. Updated now.

PR #564, opened 2017-10-22
Fix an error uncovered by cross-version tests

rhc54:
Fix an error uncovered when trying to execute OMPI v2.x apps (using PMIx v1.2) using a server from OMPI master (using PMIx v2.1). Note that the updated OMPI passes pmix_info_array_t values and has an additional security component not present in PMIx v1.2. Prior versions of OMPI didn't pass pmix_info_array_t values, and hence the bug in the unpacking routine wasn't exercised. Likewise, the presence of an unrecognized security component shouldn't cause the app to exit, so long as at least one common component is found


rhc54:
Let's chat about it. My understanding is that OMPI's promise is only to go backward to OMPI v3.0, which is based on PMIx v2.0, so there may not be a pressing need for a v1.2.5 release.

PR #565, opened 2017-10-22
Fix errors uncovered by cross-version testing

rhc54:
Fix errors uncovered when trying to execute OMPI v2.x apps (using PMIx v1.2) using a server from OMPI master (using PMIx v2.1). Get the packing/unpacking of pmix_info_array_t's and pmix_proc correct. Modex blobs are always provided based on the local peer since they were "fetched" locally. Avoid a segfault when a remote daemon fails to start its procs, thereby causing a collective to execute locally without any local callbacks.


ibm-ompi:

 Gist: https://gist.github.com/035b422d149ce2e46e57256035c3a39d

PR #582, opened 2017-10-31
Make pmix_common.h be a generated header file so we actually take vis…

rhc54:
…ibility into account. As things were, we ignored that directive because PMIX_HAVE_VISIBILITY wasn't defined in pmix_common.h and we don't want to include pmix_config.h in a user-facing file.


rhc54:
Just noting the exception: SRPM build is failing due to an inability to write a temporary header file

PR #583, opened 2017-10-31
Make pmix_common.h be a generated header file

rhc54:
Make pmix_common.h be a generated header file so we actually take visibility into account. As things were, we ignored that directive because PMIX_HAVE_VISIBILITY wasn't defined in pmix_common.h and we don't want to include pmix_config.h in a user-facing file.

(cherry picked from commit 531a9f6ebfa62fb29fa2bd0532c66b20c4716ec5)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,7 @@ examples/tool
 
 include/pmix_version.h
 include/pmix_rename.h
+include/pmix_common.h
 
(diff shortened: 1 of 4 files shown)


jjhursey:
Mellanox failure is unrelated - SPRM creation issue fails due to what looks like disk space issue.

PR #593, opened 2017-11-12
Update the usock component to support cross-version. 

rhc54:
There is still something wrong with usock support, but it is internal to v2.0 and not associated with cross-version ops.

Note that cross-version works fine with the default tcp transport.


rhc54:
@jjhursey I guess we'll need to decide if we care about usock support in the 2.0 series - not sure of the answer. If we do, then we'll need to figure out what is wrong (there is a segfault on usock messages that looks like the ```peer->nptr->info``` object is losing ref count and being released).

Issue #434, opened 2017-07-21
Performance degradation 2.1 against 2.0


karasevb:
There are results of PMIx perf tools, measured for versions 2.1 (master) and 2.0 for messages and dstore. The results show an increase memory consumption and an increase timings for PMIx v2.1 as for `message` and for `dstore`.  All tests were got on 4x16 ppn.

![image](https://user-images.githubusercontent.com/17932065/28464540-eb073f0e-6e4f-11e7-8de2-0cd7e82e80e2.png)

rhc54:
Interesting. The memory footprint is, of course, the bigger concern because of the enormous jump. There is some overhead of using components instead of direct code, but nowhere near enough to account for that difference.

Any thoughts?

karasevb:
The `dstore` looks like it does not work correctly, because memory has grown strongly. I have no supposition now, this requires investigation.

artpol84:
Our plan was:
1. Make it work
2. Do the perf optimization.
We shifted to the phase 2 now.

karasevb:
There are updated results:
![image](https://user-images.githubusercontent.com/17932065/29266021-5be6bbb2-810d-11e7-9fcc-6899f7b97660.png)
As seen, the footprint problem of v2.1 was addressed. Oldest measure of `375498.75` Kb was reduced to `44682.85` Kb, that looks like as for v2.0.
The latest measurements was did with included of PRs: #439, #461

karasevb:
@artpol84 @rhc54 
Can we close an issue?

artpol84:
Thanks @karasevb .
However all performance characteristics are worse compared to v2.0. We need to keep this issue until we resolve all of them.

artpol84:
Please, provide comparison to v1.2.1 as well.

karasevb:
There is results for v1.2:
![image](https://user-images.githubusercontent.com/17932065/29305527-5da4e152-81c3-11e7-908a-0a3aca401622.png)

rhc54:
I found the source of at least some of this performance issue. In v2.1, we cannot cache the job-level info in a packed buffer because we do not know in advance the GDS component that the client will be using. So we use a GDS function to cache the job-level info in the server's hash component.

[...]

PR #609, opened 2017-12-14
dstor: avoid the use of macro function

karasevb:

### DIFF
--- a/src/mca/gds/ds12/gds_dstore.c
+++ b/src/mca/gds/ds12/gds_dstore.c
@@ -52,75 +52,76 @@
 #include "src/mca/pshmem/base/base.h"
 
 #ifdef ESH_PTHREAD_LOCK
-#define _ESH_LOCK(rwlock, func)                             \
-__extension__ ({                                            \
-    pmix_status_t ret = PMIX_SUCCESS;                       \
-    int rc;                                                 \
-    rc = pthread_rwlock_##func(rwlock);                     \
-    if (0 != rc) {                                          \
-        switch (errno) {                                    \
-            case EINVAL:                                    \
-                ret = PMIX_ERR_INIT;                        \
-                break;                                      \
-            case EPERM:                                     \
-                ret = PMIX_ERR_NO_PERMISSIONS;              \
-                break;                                      \
-        }                                                   \
-    }                                                       \
-    if (ret) {                                              \
-        pmix_output(0, "%s %d:%s lock failed: %s",          \
-            __FILE__, __LINE__, __func__, strerror(errno)); \
-    }                                                       \
-    ret;                                                    \
-})
+static int inline _esh_pthread_lock(pthread_rwlock_t *rwlock, int rwlock_func(pthread_rwlock_t *rwlock));
 
-#define _ESH_WRLOCK(rwlock) _ESH_LOCK(rwlock, wrlock)
-#define _ESH_RDLOCK(rwlock) _ESH_LOCK(rwlock, rdlock)
-#define _ESH_UNLOCK(rwlock) _ESH_LOCK(rwlock, unlock)
+static int inline _esh_pthread_lock(pthread_rwlock_t *rwlock, int rwlock_func(pthread_rwlock_t *rwlock)) {
+    pmix_status_t ret = PMIX_SUCCESS;
+    int rc;
+    rc = rwlock_func(rwlock);
+    if (0 != rc) {
+        switch (errno) {
+            case EINVAL:
+                ret = PMIX_ERR_INIT;
+                break;
+            case EPERM:
+                ret = PMIX_ERR_NO_PERMISSIONS;
+                break;
+        }
+    }
+    if (ret) {
+        pmix_output(0, "%s %d:%s lock failed: %s",
+            __FILE__, __LINE__, __func__, strerror(errno));
+    }
+    return ret;
+}
+
+#define _ESH_WRLOCK(rwlock) _esh_pthread_lock(rwlock, pthread_rwlock_wrlock)
+#define _ESH_RDLOCK(rwlock) _esh_pthread_lock(rwlock, pthread_rwlock_rdlock)
+#define _ESH_UNLOCK(rwlock) _esh_pthread_lock(rwlock, pthread_rwlock_unlock)
 #endif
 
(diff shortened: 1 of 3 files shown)


karasevb:
refs https://github.com/open-mpi/ompi/pull/4606

rhc54:
Is there some reason as to why all this code is declared inline? I'm asking because it tends to cause optimization issues, especially when the inlined sections are long and/or involve conditional jump (e.g., "if" and "switch") statements.

karasevb:
I use inline, because those functions are called frequently, and some of them has inbuilt in other functions calls which in its turn can be invoked separately as well.
For example here: https://github.com/pmix/pmix/pull/609/files#diff-2dfe3f5a8c9f7e18a7a720a85ef91714R69
I guess it shouldn't be lead to degradation.

rhc54:
It may actually slow things down as it can interfere with the optimizer. What we have observed is that lots of inline code can cause sudden changes in performance when even tiny changes are made in the code as it trips the optimizer into a cache miss. Drives people nuts trying to track it down as even a debug statement can cause the miss.

So we tend to stay away from inline except for very small, linear code snippets. I'd suggest doing so here as well unless you have something really short and without any conditionals in it.

rhc54:
Testing these changes on https://github.com/open-mpi/ompi/pull/4623

karasevb:
@rhc54 I will return to #434 later, and recheck the `inline` influence

artpol84:
@rhc54 do you remember what compiler[s] had this issue?
My naive understanding is that in principle inline shouldn't bring any issues unless compiler has problems.

rhc54:
Afraid it failed again. There is just something in the way the code is written that is causing it to fail when embedded in a deeper tree. I can take a crack at it tomorrow as well.

rhc54:
@hjelmn @bwbarrett  can probably answer your question better than I as they were more involved in tracking down the wild performance swings. I don't think it is a compiler "bug" so much as an issue that inline code blocks optimization paths, thus causing cache lines to become brittle.

hjelmn:
Not a compiler bug. Excessive inlining of common code paths can cause the instruction cache to thrash. I noticed this while optimizing vader. Deleting a couple instructions from the code path (should always speed things up, right?) caused the latency to spike. PAPI showed that the average icache miss went from 0.4 per MPI_Send to 10+. It is completely non-intuitive but unless you can show a very real improvement it is best to avoid excessive inlining.

hjelmn:
FWIW this applies to macros as well.

artpol84:
hmmm, 
* were you able to pinpoint a problem in the assembly code? 
* Does that mean that if you will copy-paste the content of the inline/macro into the place where they are invoked you will get the same issue? Otherwise it doesn't make sense to me.

hjelmn:
@artpol84 Saw nothing wrong with the assembly. I had simply caused the common paths through ob1 and vader to cause the icache to thrash. I ended up removing a lot of inlining in vader to fix the issue. Yes, there are more instructions now but with less duplicated code the cache can stay hot.

hjelmn:
@artpol84 And, yes. C&P of the code causes the same issue. I also bloats the binary code.

artpol84:
So you mean that the following case was an issue:

```C
...
if( condition ){
    func();
}
<code below>
```
if `func()` is long and compiler assumes `taken` by default you will get only a function call in the iCache and the rest will be `<code below>`. But if you will copy/paste you'll get icache filled with those instructions.

It sound very case-specific. I mean if your fastpath is usually `taken` then I still see the benefit of C&P the code there. If it's 50/50 - then yes it will be a problem.

Am I getting it right?

artpol84:
@hjelmn I read carefully what you wrote and I think that my previous question perhaps extra. Thanks.

hjelmn:
@artpol84 Ok. Hope it makes sense. It really comes down to needing a high-level understanding of flow through your code. No compiler can do this unfortunately and the code flow can be complicated to work out so it is best to be cautious when inlining.

rhc54:
Looks like the root cause of the issue here is that there is something not quite correct in the way the dstore_mod_t's are being handled. Simply removing one of them solves the problem.

rhc54:
@karasevb @artpol84 After playing with this on the OMPI repo and talking about this situation on the telecon today, I would like to ask that we revert c4957fa in place of this PR. It appears that the move to function pointers has introduced a recursion loop that is breaking at least the optimizer in the PGI compiler, and may impact others (since we don't test every version of every compiler out there). I know there is an aesthetics preference in retaining that modification, but it isn't something that impacts actual code operation.

I think it would be fine if you chose to continue working with it on a branch, and I think we all are open to someday returning to it. However, it is holding up the release process on both PMIx and OMPI.

Would you please push a revert of that commit to the master branch? We can then go ahead with a PR of the remaining code to the v2.1 branch.

artpol84:
@rhc54  I am not sure what do you mean "that we revert c4957fa in place of this PR".  Can you rephrase it and explain in more details what are you proposing?

We don't want to hold OMPI release, however we invested noticeable amount of time into this change. Before reverting a good piece of work I would like to see more detailed explanation of what we observed. My impression from https://github.com/open-mpi/ompi/pull/4606 was the you have more guess than a real detailed understanding.
Also I honestly can't figure out what has been changed in https://github.com/open-mpi/ompi/pull/4606 to make it pass the CI. PR is huge and I don't see what exactly helped.

Unfortunately I was unable to join the call today, however I'm open to have a separate call to discuss this particular issue.

artpol84:
In other words: are we talking about cosmetic change to make compiler happy or you want to revert the whole work and release with broken dstore?

artpol84:
Checking open-mpi/ompi#4606 I see that function pointers are there but CI is happy:
* https://github.com/open-mpi/ompi/pull/4606/commits/5d4c5621a6b4bc72047a6f67bf069452f8015085#diff-7b422726a1247e9038a793edc7fc8efcR92
* https://github.com/open-mpi/ompi/pull/4606/commits/5d4c5621a6b4bc72047a6f67bf069452f8015085#diff-887eefeed6fcbb126e3b59c91c01f616R135

What has been done to achieve this?

rhc54:
I don't know what you are talking about - I reverted the function pointer code completely out of that PR. It is the PMIx master minus the commit we discussed elsewhere. This is why it now builds.

PR #613, opened 2017-12-20
 Fixed server-side get logic when there is no key in GDS

karasevb:
- added the flag indicating the rank has invoked `commit`
- the request is moved to pending till the requested rank does not invoke `commit`


karasevb:
@artpol84 please review

PR #614, opened 2017-12-21
Fixed server-side get logic when there is no key in GDS

artpol84:
- added the flag indicating the rank has invoked `commit`
- the request is moved to pending till the requested rank does not invoke `commit`

(cherry picked from commit c63b1528e9550c0c581f363c104db92eae2a4f94)


artpol84:
@karasevb : FYI

rhc54:
I believe we also need dab23fe4a4002f48485f26cc5682a40d0051f3c7?

artpol84:
Done: #616

PR #617, opened 2017-12-25
Ensure pmi/pmi2 compat libs have all symbols

pkovacs:
This patch allows you to `dlopen` the libpmi/libpmi2 compat libraries using `RTLD_NOW` which is desirable for hardened environments.   There were a plethora of undefined `pmix_` symbols in those libraries which I've eliminated by building the same code for all 3 libraries.   There are no doubt more surgically precise ways to ensure libpmi and libpmi2 contain no extraneous code, but this method is simple and seems like a good first step.  Test program below.

```
#include <stdio.h>
#include <stdlib.h>
#include <dlfcn.h>

#define PMI "/usr/lib64/libpmi.so"
#define PMI2 "/usr/lib64/libpmi2.so"

int main()
{
        void *handle = dlopen(PMI, RTLD_NOW);
        if (!handle) {
                fprintf(stderr, "%s\n", dlerror());
                exit(EXIT_FAILURE);
        }
        dlerror();
        dlclose(handle);
        handle = dlopen(PMI2, RTLD_NOW);
        if (!handle) {
                fprintf(stderr, "%s\n", dlerror());
                exit(EXIT_FAILURE);
        }
        dlerror();
        dlclose(handle);
        return 0;
}

```

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -74,8 +74,12 @@ if WANT_PMI_BACKWARD
 lib_LTLIBRARIES += libpmi.la libpmi2.la
 libpmi_la_SOURCES = $(headers) $(sources)
 libpmi_la_LDFLAGS = -version-info $(libpmi_so_version)
+libpmi_la_LIBADD = $(libpmix_la_LIBADD)
+libpmi_la_DEPENDENCIES = $(libpmi_la_LIBADD)
 libpmi2_la_SOURCES = $(headers) $(sources)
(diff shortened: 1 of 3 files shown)


rhc54:
This isn't quite complete, but a much-appreciated first step. We always commit changes to master and then cherry-pick them to the release branches, so I have ported this to master and added the missing cleanup.

Thanks for the assist! Please take a look at what is in #618 and suggest changes

PR #618, opened 2017-12-27
Cleanup separation of PMI-1/2 and PMIx symbols

rhc54:
Port the proposed patch in #617 from @pkovacs to master. It is not quite complete, however, as we shouldn't have PMI-1 or PMI-2 symbols in libpmix, nor should we have PMI-1 symbols in libpmi2 and vice versa. So fix all that and adjust the test linkages to match

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -43,6 +43,10 @@ nodist_headers =
 EXTRA_DIST =
 dist_pmixdata_DATA =
 
+# place to capture sources for backward compatibility libs
+pmi1_sources =
+pmi2_sources =
+
 libpmix_la_LIBADD = \
 	mca/base/libpmix_mca_base.la \
 	$(MCA_pmix_FRAMEWORK_LIBS) \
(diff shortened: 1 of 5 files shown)


pkovacs:
@rhc54 Ralph, I've built and tested your revision to my changes and they work fine.  I am able to dlopen all 3 libraries with RTLD_NOW, so the symbols are in order.  I had noticed on the v2.0 branch a related problem that appears to be gone in this branch.   Let me summarize.

PMIx itself uses dlopen to open its own internal modules.   In certain build scenarios, any attempt to use RTLD_LAZY (which pmix does use) will be ignored in favor of "now" loading.  For example, if we wish to harden the libraries, we would force the global offset table to read-only and require "now" loading:

`./configure CFLAGS="-Wl,-z,relro,-z,now"  ...`

Such a build does seem to work on this new branch here, but it does not work (`pmix_test` fails) back on v2.0 because some symbols are undefined in the modules.   For example:

In `src/mca/base/pmix_mca_base_var.h`:

`v2.0: int pmix_mca_base_var_env_name`
`master: PMIX_EXPORT int pmix_mca_base_var_env_name`

I think some of those PMIX_EXPORT's might need to be cherry picked back to v2.x.  I haven't checked everywhere. 

Anyway, this one here looks good to me.
Phil

PR #619, opened 2017-12-29
psec/munge: use munge configure.m4 vars; use C99 designated initializers

pkovacs:

- Link psec/munge to munge library for `munge_decode` and `munge_encode`.
- Use C99 designated initializers for `pmix_psec_module_t`, just like the other modules.

### DIFF
--- a/src/mca/psec/munge/Makefile.am
+++ b/src/mca/psec/munge/Makefile.am
@@ -19,6 +19,8 @@
 # $HEADER$
 #
 
+AM_CPPFLAGS = $(psec_munge_CPPFLAGS)
+
 headers = psec_munge.h
 sources = \
         psec_munge_component.c \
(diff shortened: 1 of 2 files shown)


ggouaillardet:
Though this is not related to this PR, it seems `munge` is searched in the default path unless instructed otherwise. `libmunge` is `LGPL` so i guess that should be fine with respect to `PMIx` license, , but IANAL ...

rhc54:
Yes, it is just fine for it to be found by default so far as licensing is concerned. However, the community specifically asked that we _not_ do so because munge is frequently included in distributions and people didn't want it "silently" used. So we should only be building that component if someone specifically asks us to do so.

rhc54:
I have confirmed that munge does not build by default, so that at least is fine.

pkovacs:
My motivation for this PR I have mentioned elsewhere: if you build pmix with hardened linker settings, (specifically -z now) modules won't load with unresolved symbols. 
```
./configure CFLAGS="-Wl,-z,relro,-z,now" --with-munge ...
```
Such a build requires that psec/munge be linked to munge or it won't load.

rhc54:
No argument - I was just responding to @ggouaillardet to indicate that we don't build it by default, and it is okay to search the default path if someone _does_ ask us to build it.

rhc54:
Thanks for testing these options! They aren't used very often, and so tend to lie untested.

Issue #628, opened 2018-01-04
Components not found when "standard" install location used.


amckinstry:
pmix fails to search for and find the MCA components when standard Debian paths are used.
i.e. for the configure line:
./configure --build=x86_64-linux-gnu --prefix=/usr --includedir=\${prefix}/include --mandir=\${prefix}/share/man --infod\
ir=\${prefix}/share/info --sysconfdir=/etc --localstatedir=/var --disable-silent-rules --libdir=\${prefix}/lib/x86_64-linux-gnu --libexecdir=\${prefix}/lib/x8\
6_64-linux-gnu 

with components in /usr/lib/$(build)/pmix/lib/pmix/ , this directory is never searched.
If the components are in $HOME/.pmix/components then pmix tests and openmpi work fine.
Also, the same code works fine with --prefix=$(HOME)/pmix.

This is for V2.1.0rc1 and V2.1 head-of-tree.

amckinstry:
Ok, the problem appears to be that component_path in /etc/pmi-mca-params.conf is ignored if $(LIBDIR)/pmix exists.

The layout I had in Debian is/was: prefix=/usr/lib/x86_64-linux-gnu
  $prefix/pmix
  $prefix/pmix/lib
  $prefix/pmix/include
with libpmix.so , etc in $prefix/pmix/lib and components in
 $prefix/pmix/lib/pmix

This was so that --with-pmix=$(prefix)/pmix could be used in OpenMPI  (it needs to be a directory containing the include and lib dirs), and component_path in /etc points to the components.

rhc54:
Just to clarify: are you saying you have a param set in the pmix-mca-params.conf file, and we are ignoring it under the given condition? Can you post your .conf file?

pkovacs:
I see the problem.   The parameter `mca_base_component_path` is initialized in `pmix_mca_base_open.c`,  but its value is not updated when you use the short name `component_path` in one of the `mca-params.conf` files.   The api `save_value()` in `pmix_mca_base_parse_paramfile.c` is not matching them.   Maybe there's a synonym problem.  

Using the longer parameter name works, so that's a workaround for now.

```
# This doesn't work
component_path = /path/to/mca/components

# This works
mca_base_component_path = /path/to/mca/components
```

amckinstry:
I confirm this workaround works for me

rhc54:
[...]

PR #629, opened 2018-01-05
Correct the comment in the default MCA param template

rhc54:
Correct the comment in the default MCA param template - we do not support a param called "component_path". The correct syntax is "mca_base_component_path"

[skip ci]
bot:notest


rhc54:
Fixes #628

PR #634, opened 2018-01-17
Ensure the buffer type gets set

rhc54:
Ensure we set the buffer type when embedding user-level buffers into
PMIx internal buffers.

(cherry picked from commit 89408baf927cf8bcdfed2eee18962e55cb96b303)


rhc54:
@jjhursey This is now ready to go

PR #646, opened 2018-01-23
General cleanup of tool and failure code paths

rhc54:
Silence trivial warning

Protect receive callbacks from NULL and/or empty buffers as this
can occur when the peer on a connection disappears.

Fix tool connection search so it properly descends into the directory
tree while searching for the server's contact file.

Fix store_local so it doesn't reject a new nspace as that can happen
when working with tools

Ensure we _always_ complete PMIx_Finalize - don't return if something
goes wrong in the middle of the procedure


rhc54:
@jjhursey I have added quite a bit of cleanup here, but have hit a brick wall when trying to run a v2.0 server with a v1.2.5 client. I just cannot get the return communication from server to client to wind up with the correct message size - I just get garbage on the v1.2.5 side, even though everything on the server side is correct.

Testing this is pretty easy:

* build both versions --enable-static --disable-shared --enable-debug

* go into the test/simple directory of each version and do "make"

* from the v2.0 test/simple directory, execute "./simptest -u -e path-to-v1.2-test/simple/simpclient"

If you set PMIX_DEBUG=5 in the environment, you'll get debug output showing the problem. The protection for v1.2.5 (PR #647) will prevent it from segfaulting.

I'll take a break for a bit, and then retry. My head hurts.

@ggouaillardet Don't know if you are interested, but you're pretty good at debugging this kind of thing - feel free to take a fresh look if you like!

ggouaillardet:
i tried with the latest `v1.2` and `v2.0` branches.

in the `v2.0` branch, in `src/mca/ptl/base/ptl_base_sendrecv.c` at line 434
```c
            peer->recv_msg->hdr.pindex = ntohl(hdr.pindex);
            peer->recv_msg->hdr.tag = ntohl(hdr.tag);
            peer->recv_msg->hdr.nbytes = ntohl(hdr.nbytes);
```

the issue is there is no `htonl()` equivalent in the `v1.2` branch

i let you decide how to fix that, since the fix might alter the compatibility matrix ...

rhc54:
Yeah, I already changed that in this PR (we don't go thru the ptl/base functions for usock), but it didn't help. I think I need to just look at the byte streams to see what is going on.

Thx for looking!

ggouaillardet:
oops, i should have tested this PR instead of the stock `v2.x` branch

in the `v1.2` branch

`pmix_bfrop_pack()` calls `pmix_bfrop_store_data_type(buffer, PMIX_INT32)`

but in your `v2.0` branch

`pmix_bfrop_unpack()` calls `pmix_bfrop_get_data_type()` that calls `pmix_bfrop_unpack_datatype(buffer, type, &cnt, PMIX_DATA_TYPE)` that ends up calling `pmix_bfrop_unpack_int16(buffer, dest, num_vals, type)`

bottom line, `v1.2` packs 32 bits, but `v2.0` unpacks only 16 bits, and things go wrong after that.

rhc54:
Aha!!! Thanks!

ggouaillardet:
my previous reply was misleading, please read

in the `v1.2` branch

`pmix_bfrop_pack()` calls `pmix_bfrop_store_data_type(buffer, PMIX_INT32)` that calls `pmix_bfrop_pack_datatype(buffer, &type, 1, PMIX_INT)` that ends up calling `pmix_bfrop_pack_int32(buffer, src, num_vals, type)`

conclusion is the same, 32 bits are packed in `v1.2` when only 16 bits are unpacked in `v2.0`

rhc54:
Okay, I chased all this down, and the bottom line is that v2.0 cannot support v1.2 clients. There are just too many data type changes to let it work without backporting the entire v2.1 bfrops framework, and that is probably too much change.

@jjhursey I think this is as ready as it can be.

jjhursey:
(There is a merge conflict on this PR now)
It's too bad about v2.0 servers not being able to support v1.x clients, but it's good that v2.1 will have that support. It's the v2.1 series that we really have focused on cross version compatibility. So I think that is fine - adding the note to the `NEWS` is good.

src/mca/ptl/base/ptl_base_frame.c
```diff
@@ -67,8 +69,18 @@ pmix_ptl_API_t pmix_ptl = {
     .stop_listening = pmix_ptl_base_stop_listening
 };
 
+static int max_msg_size = PMIX_MAX_MSG_SIZE;
```
jjhursey: We should probably change this from an `int` to a `size_t` since that is what we compare against later.
I don't expect us to ever have a user ask for a `max_msg_size` of greater than 2GB.

rhc54:
I love those copyright merge conflicts - can you approve again? Modifying it auto-rejects the prior review.

src/mca/ptl/base/ptl_base_frame.c
rhc54: oh - sure! will do right now

rhc54: it's going to be the same in the v2.1 branch, so i'll push a change there too

rhc54:
I'm going to take this one due to the time differences, and the fact that the final change made was trivial.

rhc54:
Ah, you are still around! Thanks!

PR #647, opened 2018-01-26
Protect against overly-large messages

rhc54:


rhc54:
just pulled it out of the air 😆  Honestly, happy to accept any finite number. Security folks just don't like an unconstrained malloc in a system-level server.

jjhursey:
I think 16 MB is ok - and it matches the other release branches. I like that in the v2.1 series we can adjust the limit via MCA variable.

PR #649, opened 2018-01-27
Protect against overly-large messages

rhc54:


rhc54:
I'm going to take this one due to the time differences, and because the final change made was trivial

PR #654, opened 2018-01-31
This causes assert failures for me in debug if I do not comment it out.

dsolt:
I'm not 100% sure if it is because the kv has been added to a list above
or if there is a better fix.


rhc54:
This will cause a memory leak - I'll look into fixing the underlying problem.

PR #655, opened 2018-01-31
Remove the dstore path from when job data is retrieved from another n…

dsolt:
…amespace

using a modex call.   In this case the server is not using dstore, so one
fix is to update the server code to use dstore.  The other fix which I have
taken here (at least temporarily) is to never use dstore for this.  Once the
data is pulled down from the server, it is placed in the dstore for other
local ranks, but it could be improved by having the server place it directly
in the dstore.

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -7,7 +7,7 @@
  *                         All rights reserved.
  * Copyright (c) 2016      Mellanox Technologies, Inc.
  *                         All rights reserved.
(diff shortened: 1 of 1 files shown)


rhc54:
@dsolt @jjhursey Looks like this breaks the fence operation when data is collected.

@karasevb @artpol84 Can you perhaps address the underlying problem? This looks more like a workaround.

artpol84:
We are waiting for the reproducer now and will investigate.

PR #657, opened 2018-02-01
preg: fix and enhanced test

ggouaillardet:


jjhursey:
@ggouaillardet Thanks for porting that over!
@rhc54 Do you see any issue in getting this bug fix into the v2.1 and v2.0 series?

rhc54:
Should be okay - doesn't break compatibility and is small.

jjhursey:
Ok. I'll try to get that done today so we can include it in the next rc's

jjhursey:
I filed PR #658 (v2.1) and #659 (v2.0) for this PR.

PR #662, opened 2018-02-03
Fix dmodex when receiving new nspace

rhc54:


artpol84:
@rhc54 do you think that @dsolt was facing this issue?

rhc54:
Not sure - need to check to see if this is present in the v2.x series

rhc54:
definitely in v2.1 - not in v2.0

PR #664, opened 2018-02-04
Resolve dmodex of job-level info

rhc54:
When a process requests job-level info for a namespace unknown to the PMIx server, the server generates a dmodex request to its local host. The request specifies the target nspace, but with a WILDCARD rank since the data being requested is from the job-level info. At least one way the host can respond is to register the nspace using the PMIx server's register nspace API. However, this results in the data being "cached" in the server's GDS module, not in that of the requesting client.

I took a shot at transferring the data in pmix_server_get.c, but the buffer format created by PMIX_GDS_REGISTER_JOB_INFO doesn't match that returned in a regular dmodex request, and so this fails. Alternatively, we could have the host create a job-level dmodex response buffer and give it to us - I chose not to go that way because it would require the host to basically duplicate the code already used to feed register_nspace.

Perhaps someone could take a look at resolving the problem?

Also included a few minor cleanups that cropped up when compiling and trying to commit (e.g., whitespace)

### DIFF
--- a/contrib/pmix_jenkins.sh
+++ b/contrib/pmix_jenkins.sh
@@ -395,4 +395,3 @@ if [ -n "$JENKINS_RUN_TESTS" -a "$JENKINS_RUN_TESTS" -ne "0" ]; then
         set -e
     fi
 fi
-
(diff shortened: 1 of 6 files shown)


rhc54:
@artpol84 @karasevb @jjhursey @dsolt 

Consider this a "checkpoint" on the dmodex repair. The problem was partly in pmix, and partly in ORTE. I will bring the ORTE fixes into OMPI master.

rhc54:
Updated with solution that supports both the register_nspace and return of job-info via the dmodex callback function

rhc54:
Closes #655

PR #670, opened 2018-02-16
preg: fixed object destructor: free(): invalid pointer

karasevb:
The object destructor tries to free the `nspace` name pointer when it not NULL:
https://github.com/pmix/pmix/blob/master/src/include/pmix_globals.c#L286-L287
But this pointer was not allocated, and instead it points to another data that
shouldn't free here.

Signed-off-by: Boris Karasev <karasev.b@gmail.com>

PR #671, opened 2018-02-16
preg: fixed object destructor: free(): invalid pointer

artpol84:
The object destructor tries to free the `nspace` name pointer when it not NULL:
https://github.com/pmix/pmix/blob/master/src/include/pmix_globals.c#L286-L287
But this pointer was not allocated, and instead it points to another data that
shouldn't free here.

(cherry picked from commit ee313a337eebb58347f93d39d5f8ac1e70ab3e03)


artpol84:
See #670 for some details

PR #673, opened 2018-02-18
ptl/base: Fix preliminary peer->info release

artpol84:
It is needed for PMIX_REPORT_EVENT.


src/mca/ptl/base/ptl_base_sendrecv.c
```diff
@@ -155,6 +154,10 @@ void pmix_ptl_base_lost_connection(pmix_peer_t *peer, pmix_status_t err)
              * an event. If not, then we do */
             PMIX_REPORT_EVENT(err, peer, PMIX_RANGE_NAMESPACE, _notify_complete);
```
artpol84: info is used implicitly here.

artpol84:
This is a blocker for OMPI v3.1.
We are getting segfaults in our MTT.

PR #691, opened 2018-03-11
Provide spawned keys to resulting procs

rhc54:
For consistency, always provide the PMIX_SPAWNED and PMIX_PARENT_ID keys
in the job-level info of procs started by PMIx_Spawn from a client. Do
not do so for procs started by a tool as the tool isn't part of the
resulting nspace.

(cherry picked from commit c0e2ec0f854ac50312b9f592d0fcfedd10c53ce7)


rhc54:
[pmix-2.1.2a1.tar.gz](https://github.com/pmix/pmix/files/1802227/pmix-2.1.2a1.tar.gz)

Providing the tarball so the reporting user can verify it.

rhc54:
Not really - just posted the tarball so they wouldn't have to wait for an official release as that might take awhile.

Issue #586, opened 2017-11-06
Direct modex broken when dstore GDS used


rhc54:
Please see https://github.com/pmix/pmix/pull/585 for an explanation of the problem.

rhc54:
Transferring this description here so it doesn't get lost:

When requesting modex data without it having already been collected, the client issues a request to the server for data "put" by the given rank. Since the server always returns all data for that proc, we don't pass the key being requested. Thus, the server operates with a NULL key for the request.

When a proc "commits" its modex data, the server stores it in different locations based on its scope. Local scope data goes into the GDS component for that peer, but remote scope data is stored under the server's GDS component. This is done to allow proper response to remote vs local procs - i.e., local procs don't see remote scope data, and vice versa.

The GDS/hash component is working correctly. When a request for data arrives, we look in the appropriate place based on whether the requested proc is local or remote. If we find it, we return that data. If we don't find it, then we create a tracker and wait for it to appear.

The GDS/ds12 component, however, is NOT working correctly. Because a NULL key is provided, it only looks to see if the job-level data is available, and then immediately returns with SUCCESS and the shared memory location. This is okay IF the requested proc has already "committed" its data, but causes the client proc to generate an error if the requested proc has NOT provided its data yet.

This is seen when running the following test in the test/simple directory:
```
$ export PMIX_SIMPDMODEX_ASYNC=1
$ ./simptest -n 2 -e ./simpdmodex
```
@artpol84 @karasevb I think this is probably a blocker for v2.1 - can you please verify?

artpol84:
Indeed it looks like a blocker. We will take a look next week.

rhc54:
It looks like this is going to hold up OMPI v3.1 release (as it would break your latest SLURM plugin enhancement), so please keep us updated

artpol84:
Yes, we will.

@karasevb this is a high prio!

artpol84:
@rhc54 Can you provide more details on the connection with Slurm plugin?

artpol84:
Also I do you know if this is something that was a problem from the day 1st or this is related to the recent cross-version implementation?

rhc54:
I think this may have been around for awhile - at least, I observed it prior to the recent commit. It impacts your direct modex support in the SLURM plugin because of the race condition - if someone asks for modex info prior to the peer having committed it, then the request will return an error.

artpol84:
[...]

PR #697, opened 2018-03-14
Fixed the direct modex request

karasevb:
Fixed checking the commit counter which should be not checked for
server, otherwise clients have get error when direct modex.

### DIFF
--- a/src/server/pmix_server_get.c
+++ b/src/server/pmix_server_get.c
@@ -689,7 +689,14 @@ static pmix_status_t _satisfy_request(pmix_nspace_t *nptr, pmix_rank_t rank,
 
     /* retrieve the data for the specific rank they are asking about */
     if (PMIX_RANK_WILDCARD != rank) {
-        if (!peer->commit_cnt) {
+        if (!PMIX_PROC_IS_SERVER(peer) && !peer->commit_cnt) {
+            /* this condition works only for local requests, server does 
+             * count commits for local ranks, and check this count when 
+             * local request.
+             * if that request performs for remote rank on the remote 
+             * node (by direct modex) so `peer->commit_cnt` should be ignored,
+             * it is can not be counted for the remote side and this condition 
+             * does not matter for remote case */
             return PMIX_ERR_NOT_FOUND;
         }
         proc.rank = rank;


karasevb:
Waiting MTT results

src/server/pmix_server_get.c
```diff
@@ -689,7 +689,7 @@ static pmix_status_t _satisfy_request(pmix_nspace_t *nptr, pmix_rank_t rank,
 
     /* retrieve the data for the specific rank they are asking about */
     if (PMIX_RANK_WILDCARD != rank) {
-        if (!peer->commit_cnt) {
+        if (!PMIX_PROC_IS_SERVER(peer) && !peer->commit_cnt) {
```
rhc54: I know you guys added this commit_cnt for a reason, but I honestly don't remember it any more. Can you add a brief comment explaining why it is needed? Or is that comment somewhere else in the code?

karasevb: @rhc54 `commit_cnt` was introduced in PR #613 that fixes issue #586. 
This works for local requests where `peer` is corresponds to client, 
but for remote requests `peer` corresponds to server peer: https://github.com/pmix/pmix/blob/master/src/server/pmix_server_get.c#L805-L810:
```
        scd = PMIX_NEW(pmix_server_caddy_t);
        PMIX_RETAIN(pmix_globals.mypeer);
        scd->peer = pmix_globals.mypeer;
        PMIX_LIST_FOREACH(req, &ptr->loc_reqs, pmix_dmdx_request_t) {
            pmix_status_t rc;
            rc = _satisfy_request(nptr, rank, scd, req->cbfunc, req->cbdata, NULL);
```
For server `commit_cnt` should not be taken into account, this always will be return [PMIX_ERR_NOT_FOUND](https://github.com/pmix/pmix/blob/1207023c664ffe5b6c3cd101fceef425944979b8/src/server/pmix_server_get.c#L693).

karasevb:
Here is MTT results for variants of tests for direct and full modex. Without this PR half of results are failed for variants with direct modex. This PR fixes this problem.
```
Summary

   Phase   Section                        MPI     Duration Pass Fail Time Skip
                                          Version                    Out

   Test    gtest                          master  01:27:53      1
   Run
   Test    ibm-tests                      master  06:34:10 1126 1    1
   Run
   Test    osu_micro_benchmark_collective master  01:28:30 551  1
   Run
   Test    graph500                       master  28:29    16
   Run
   Test    osu_micro_benchmark_one_sided  master  06:38    144
   Run
   Test    ext-bugs-regres                master  00:30    24
   Run
   Test    mpich_tests_mpi_coll           master  06:21:12 628       1
   Run

Totals

    Tests            Failed           Passed           Duration
    2582             5                2577             16:47:45
```

src/server/pmix_server_get.c
rhc54: Okay, let me see if I understand. The server is the one tracking this commit_cnt as it is the one who updates it. It updates that count whenever the client "commits" its data to the server. However, the server itself doesn't do the commit, and so the server's "peer" object will never have a non-zero commit_cnt.

The server stores committed data from its clients in its local storage. When someone requests that data, the server goes through the pmix_server_get code using its own peer object since that is where the data is stored. The current check prevented us from going thru the code path because it looked at the server's peer object, whose commit_cnt is always zero.

The correct check would have been to look at the commit_cnt of the proc whose info is being requested - yes? Wouldn't that be the correct fix here, instead of just ignoring the flag?

artpol84: @rhc54 Indeed this looks like a better solution.
@karasevb lets sync on this today.

artpol84: @rhc54 
I just talked to @karasevb and we think that this change is correct. Here is the reason.
* This portion originally worked ok if the key was requested from another local process.
* The issue was when we were applying this check to the case when the key was requested from the process on the remote node. In this case we simply don't have corresponding commit count as it is maintained by another PMIx server.

rhc54: Ah - thank you! Could you capture that explanation in a comment in an appropriate place in the code? I'm sure future developers working thru the code won't have that context.

karasevb: I'll add it to the commit message and in the code comment.

karasevb: @rhc54 , @artpol84
Code comment and commit message have been updated, PR ready to go.

PR #698, opened 2018-03-15
Fixed the direct modex request

karasevb:
This commit fixes the case when local client asks for the key from the
process on the remote node. The local server don't have commit count for
remote ranks, it is maintained by another PMIx server, so commit count
should be ignored for remote requests.

(cherry picked from commit 0a988c1dfa7c469f8a7ab46bbe0f3509b06b964e)


jladd-mlnx:
@karasevb would this bug impact the v1.2.x branch?

karasevb:
@jladd-mlnx this bug does not impact v1.2. It was introduced since v2.1 version.

karasevb:
@rhc54 @jjhursey good to go

PR #700, opened 2018-03-29
Fabric pre-configuration support

rhc54:
Implement a pnet/tcp component and demonstrate how the existing PMIx_server_setup_application and PMIx_server_setup_local_support APIs can be used to pre-configure the network - in this example, by assigning static ports to the application's processes. Add a new "simple" test (both server and client) to illustrate the process.

Per the developer's conference, allow the caller to define the key by which the assigned network resources can be accessed. Note that some providers will define the network library themselves and won't need to necessarily expose the resource assignments.

Allow multiple network resource allocations per application to support the case where multiple libraries operate in parallel, each seeking its own network resources.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,6 +114,8 @@ test/simple/test_pmix
 test/simple/simptool
 test/simple/simpdie
 test/simple/simptimeout
+test/simple/gwclient
+test/simple/gwtest
 
 # coverity
 cov-int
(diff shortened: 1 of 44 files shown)


rhc54:
@jsquyres Looks like the checkers are still jammed

rhc54:
This includes extension of the pnet framework to add a "collect_inventory" API in accordance with https://github.com/pmix/RFCs/pull/30, but does not (yet) implement it.

jsquyres:
@rhc54 Sometimes I hate PHP.  😦  This PR is ok now.

rhc54:
only sometimes?? you are a glutton for punishment 😄

jjhursey:
bot:ibm:retest

Issue #704, opened 2018-04-18
PMIX_HOSTNAME missing from job data in PMIx 2.1 series


jjhursey:
I'm running the simple PMIx example in the Gist below:
 * https://gist.github.com/jjhursey/b93a3c1e7d1ffe9827442d5aa114cc8e

If I run with PMIx 2.0.3 then it runs fine, but if I run with PMIx 2.1.1 (head of branch) then it fails on the `PMIx_Get(&proc, PMIX_HOSTNAME, NULL, 0, &val);` with a `-17` (`PMIX_ERR_PROC_ENTRY_NOT_FOUND`).

It seems that the job data is missing the per-process hostname in the v2.1 series. Is that expected?

jjhursey:
Note: I thought it might be something with the `gds` component. I tried just `hash` and it gave the same behavior. If I tried just `ds12` then it failed on the 'job size' before it gets to the hostname.

rhc54:
Where are you trying it? PMIx itself doesn't supply the value, so all this means is that whomever is acting as the server didn't set it.

jjhursey:
Sorry I forgot that detail. I was using Open MPI v3.0.x branch (HEAD) with mpirun launch across at least 2 nodes (if I only ran one node, then it was fine).

Initially, I was trying to swap out the pmix install from pmix 2.0.3 to pmix 2.1.x (HEAD) by swapping the pmix install that LD_LIBRARY_PATH was pointing to. That highlighted the problem. I'm now trying to rebuild some fresh builds to see if that LD_LIBRARY_PATH trick was contributing to the problem.

Maybe this is an Open MPI issue. I posted here since, in my original experiment, I kept the Open MPI part constant and only changed the pmix library between the runs.

jjhursey:
Fresh build completed.
 * Open MPI v3.0.x (HEAD) built with external PMIx 2.0 (HEAD) - works
 * Open MPI v3.0.x (HEAD) built with external PMIx 2.1 (HEAD) - fails

Nothing really special about the Open MPI build:
`./configure --prefix=/share/install/ompi-v3.0.x-dbg-v2.1 --with-libevent=/share/libevent --with-hwloc=/share/hwloc --with-pmix=/share/install/pmix-v2.1 --enable-debug`
or the PMIx build:
`./configure --prefix=/share/install/pmix-v2.1 --with-libevent=/share/libevent --disable-debug`

[...]

PR #707, opened 2018-04-18
Transfer all cached data to client gds

rhc54:
Upon first connection, the server transfers its cached nspace registration data to the gds being used by the client's nspace. Any data registered for a specific proc must go as well, even if the proc is remote. Otherwise, the client library won't find it, and won't ask the server for it because the key starts with "pmix", and must therefore have been provided at startup.


rhc54:
Refs #704

PR #723, opened 2018-05-19
Fix deadlock in PMIx_Server_init in the error case

artpol84:


artpol84:
@karasevb please cherry-pick to the target branches.

karasevb:
@artpol84 required only for v2.1, it already has it.

artpol84:
:+1:

PR #724, opened 2018-05-19
v2.1: Fix deadlock in PMIx_Server_init in the error case

artpol84:


artpol84:
#723 is in the master

PR #748, opened 2018-06-07
Plug misc memory leaks

ggouaillardet:


ggouaillardet:
Once this PR is applied, the remaining leaks is some shared memory segments are not unmapped.
Other than that, there is a very small memory leaks that is independent of the number of iterations (which is good enough or even pretty good)

PR #749, opened 2018-06-07
Revert connect/disconnect to v2 signatures

rhc54:
Although we are unaware of anyone who has implemented the server-side backend for PMIx_Connect/Disconnect, we really need to begin enforcing the backward compatibility guarantee. We had changed these signatures to accommodate the work being done in the MPI sessions working group, burt we will instead meet that request by adding a set of new APIs (PMIx_Group_*) in the future v4 of the standard.

Note: this effectively "rejects" RFC https://github.com/pmix/RFCs/pull/3 - it will be so marked and replaced by a new RFC for PMIx_Group


jjhursey:
This needs to be PR'ed to the `v3.0` branch right?

rhc54:
yes - will do so shortly, along with a couple of other things

PR #766, opened 2018-06-19
Add the stability test for ease of use

rhc54:


rhc54:
Merging so we move things along

PR #773, opened 2018-06-20
Simplify cython detection

rhc54:
Thanks to @ggouaillardet for the suggestion!


rhc54:
Ah, okay - I never use VPATH builds so I don't see it. IIRC, the simple solution is to replace top_srcdir with top_builddir

ggouaillardet:
it seems `top_srcdir` is only to be used in `Makefile.am` by default, but not in `m4` files.
so I think the right fix is to use `$srcdir` instead.

PR #774, opened 2018-06-21
Use srcdir instead of top_srcdir

rhc54:
Fix VPATH builds when looking for Cython


ggouaillardet:
thanks !

PR #775, opened 2018-06-21
bindings/python: fix make install when VPATH is in use

ggouaillardet:
look for setup.py in the source directory instead of
the current directory


ggouaillardet:
This is `Makefile.am` and **not** a `m4` file, so `$(top_srcdir)` is fine here (and I tested it)

PR #777, opened 2018-06-25
dstore: fixed the logic of unmaping dstore segments

karasevb:

### DIFF
--- a/src/mca/gds/ds12/gds_dstore.c
+++ b/src/mca/gds/ds12/gds_dstore.c
@@ -2885,6 +2885,8 @@ static pmix_status_t dstore_del_nspace(const char* nspace)
     ns_map_t *ns_map;
     session_t *session_tbl = NULL;
     ns_track_elem_t *trk = NULL;
+    int dstor_track_idx;
+    size_t session_tbl_idx;
 
     PMIX_OUTPUT_VERBOSE((10, pmix_gds_base_framework.framework_output,
         "%s:%d:%s delete nspace `%s`", __FILE__, __LINE__, __func__, nspace));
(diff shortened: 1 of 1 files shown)


artpol84:
@karasevb please, PR to the relevant release branches

ggouaillardet:
fwiw, this fixes the PMIx test program

I did a quick check with PMIx `master` as an external PMIx for Open MPI, and a crash occur with a simple test.

At this stage, I cannot tell whether the root cause is PMIx or Open MPI

```
$ mpirun -np 1 hostname
c7
mpirun: ../../../../../../src/pmix/src/mca/gds/ds12/gds_dstore.c:2928: dstore_del_nspace: Assertion `((0xdeafbeedULL << 32) + 0xdeafbeedULL) == ((pmix_object_t *) (trk))->obj_magic_id' failed.
Aborted (core dumped)
```

karasevb:
Will check

ggouaillardet:
a similar/related (?) issue can be reproduced without Open MPI

```
./stability -n 1 -reps 1 -e /bin/true
```

that looks like a bozo case which is not correctly handled.

karasevb:
@ggouaillardet Thanks!
The problem is clear, there is a dstor issue. I'm going to fix.

Could you clarify, `./stability -n 1 -reps 1 -e /bin/true` is reproduces the crash? I able to catch it only with `mpirun -np 1 hostname`.

ggouaillardet:
yes, I confirm I do not need Open MPI to reproduce the issue

```
$ ./stability -n 1 -reps 1 -e /bin/true
Testing version 4.0.0
Running cycle 0
lt-stability: ../../../../../../src/pmix/src/mca/gds/ds12/gds_dstore.c:2928: dstore_del_nspace: Assertion `((0xdeafbeedULL << 32) + 0xdeafbeedULL) == ((pmix_object_t *) (trk))->obj_magic_id' failed.
Aborted (core dumped)
```

`dstore_del_nspace` is invoked by a PMIx thread, so there could be a race condition involved with `-reps 1`.

PR #780, opened 2018-06-26
v2.0 dstore: fixed the logic of unmaping dstore segments

karasevb:
Corresponds to #777, #781


karasevb:
The unmapping of dstore segments was verified at the runtime.

PR #792, opened 2018-07-19
Provide a numeric version definition

rhc54:
Enable easier check of full version number. The numeric fields are:

bytes 0-3: major version number (hex)

bytes 4-5: minor version number (hex)

bytes 6-7: release number (hex)


rhc54:
Moving ahead

PR #795, opened 2018-07-20
Complete initial pass thru prototype Py bindings

rhc54:
Require Python 3 for bindings

Output configuration summary report. Show Python and Cython version info

Create a python script to automatically generate Cython input files from
PMIx headers. This gets executed during "make install" as part of
generating the bindings. The conversion routines to shift information
between C and Python need to be manually written for now. Fortunately,
these don't change as we move across PMIx versions and so they should
only be modified when something gets added, which is rare.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,8 @@ config/py-compile
 bindings/python/pmix.c
 bindings/python/*.so
 bindings/python/build/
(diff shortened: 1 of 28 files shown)


ibm-ompi:

 Gist: https://gist.github.com/8c9b2555c03c459b73af86e0c7dbb313

PR #797, opened 2018-07-24
Transfer all cached data to client dstore gds

### LINKED ISSUE #704: PMIX_HOSTNAME missing from job data in PMIx 2.1 series

jjhursey:
* Fixes #704 for dstore

### DIFF
--- a/src/mca/common/dstore/dstore_base.c
+++ b/src/mca/common/dstore/dstore_base.c
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2015-2018 Intel, Inc. All rights reserved.
- * Copyright (c) 2016      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2016-2018 IBM Corporation.  All rights reserved.
  * Copyright (c) 2016-2017 Mellanox Technologies, Inc.
  *                         All rights reserved.
  * Copyright (c) 2018      Research Organization for Information Science
@@ -2804,7 +2804,7 @@ PMIX_EXPORT pmix_status_t pmix_common_dstor_register_job_info(pmix_common_dstore
     char *msg;
     pmix_status_t rc;
     pmix_proc_t proc;
-    pmix_rank_info_t *rinfo;
+    pmix_rank_t rank;
 
     pmix_output_verbose(2, pmix_gds_base_framework.framework_output,
                         "[%s:%d] gds:dstore:register_job_info for peer [%s:%d]",
@@ -2836,8 +2836,8 @@ PMIX_EXPORT pmix_status_t pmix_common_dstor_register_job_info(pmix_common_dstore
             return rc;
         }
 
-        PMIX_LIST_FOREACH(rinfo, &ns->ranks, pmix_rank_info_t) {
-            proc.rank = rinfo->pname.rank;
+        for (rank=0; rank < ns->nprocs; rank++) {
+            proc.rank = rank;
             rc = _store_job_info(ds_ctx, ns_map, &proc);
             if (PMIX_SUCCESS != rc) {
                 PMIX_ERROR_LOG(rc);


rhc54:
Let me check my understanding: the hash function now returns all job info from the cache and passes it to the dstore, but the dstore currently only stores the data for local procs from that cache - correct?

jjhursey:
From my tracing the `hash` wasn't being used at all. When `PMIX_GDS_REGISTER_JOB_INFO` is called it goes directly to `dstore_register_job_info` instead of `register_info` in `hash`.

From what I remember, when I tested last time it went to `hash` first. But I'm not seeing  `register_info` in `hash` being called at all. I thought we needed it store in the `hash` for cross version support?

In essence, yeah that's what happening. dstore is only storing the local rank's information instead of all rank's information when it's `dstore_register_job_info` was being called. This PR forces it to store the information for all of the ranks.

rhc54:
Something got trashed in the logic then. The data must go into the hash cache first and held there until the first local process connects so we can determine the gds component to use for that job.

Looking at the code, when the host calls register_nspace we call ```gds_add_nspace``` to ensure there is an nspace record in each active component, and then we call ```gds_cache_job_info``` to cache the data until first connect. When a proc connects, we call ```register_job_info``` which should retrieve the cached info from the hash and move it to the dstore.

Unless someone broke those macros, I don't see how the data could go directly to dstore.

jjhursey:
I did a fresh build of Open MPI `master` (with a tweak to get the `ext3x` component to work with PMIx `master` versioned `4.0.0`) and PMIx `master`. The `hash` (priority 10) and `ds12` (priority 20) components are loaded. I added a print statement to the top of both `dstore_register_job_info` and `hash_register_job_info`. Only `dstore_register_job_info` was called by the ORTE daemons.

For one of the daemon's I see this interaction:
```
shell$ mpirun --np 2 --host node01:1,node02:1 -x LD_LIBRARY_PATH ./pmix_init_finalize
159149) ====== ds12 ========== dstore_add_nspace
159149) ====== HASH ========== nspace_add
159149) ====== HASH ========== hash_cache_job_info
159149) ====== HASH ========== hash_cache_job_info
159149) ====== ds12 ========== dstore_register_job_info
```

So `hash_register_job_info` is not called. Instead we call directly into `dstore_register_job_info` (from the `PMIX_GDS_REGISTER_JOB_INFO` macro). I can only think that the `ptr->compat.gds` was directed at the `ds12`.

I'm not familiar with the logic of the `hash` to `dstore` handoff to dive into what might have gone wrong :/

I *think* this PR is solid though, aside from the `hash` to `dstore` handoff. Right?

rhc54:
I'd have to look into the dstore code - it may very well do a gds "fetch" of all info instead of calling "register_job_info". I would assume this PR is correct and if the data shows up, then the dstore must be doing the right thing. @karasevb will have to confirm that portion as I'm not familiar with the dstore logic.

karasevb:
@jjhursey `hash_register_job_info` should not be called for dstore case, and it is true that `PMIX_GDS_REGISTER_JOB_INFO` calls directly `dstore_register_job_info`.

src/mca/gds/ds12/gds_dstore.c
```diff
@@ -3187,8 +3187,8 @@ static pmix_status_t dstore_register_job_info(struct pmix_peer_t *pr,
             return rc;
         }
 
-        PMIX_LIST_FOREACH(rinfo, &ns->ranks, pmix_rank_info_t) {
-            proc.rank = rinfo->pname.rank;
+        for (rank=0; rank < ns->nprocs; rank++) {
+            proc.rank = rank;
```
karasevb: For each nspaces `_store_job_info` will invoke only for ranks started from 0, it looks not correct. Some nspaces do not have rank 0.

rhc54: I can't think of how an nspace would start from a rank other than 0 - can you provide some insight into that scenario?

jjhursey: Yeah. I'm not sure I understand that scenario either. If that was the case then the `hash` version should fail as well.

karasevb: @rhc54 when the number of nspaces more than one, each nspace will start rank counting from 0.

I have added printouts of ranks with nspaces:
```patch
diff --git a/src/mca/gds/ds12/gds_dstore.c b/src/mca/gds/ds12/gds_dstore.c
index 580f018..83e9af0 100644
--- a/src/mca/gds/ds12/gds_dstore.c
+++ b/src/mca/gds/ds12/gds_dstore.c
@@ -3187,8 +3187,10 @@ static pmix_status_t dstore_register_job_info(struct pmix_peer_t *pr,
             return rc;
         }
 
+        pmix_output(0, "dstore_register_job_info: nspace=%s, nprocs=%d", ns->nspace, ns->nprocs);
         for (rank=0; rank < ns->nprocs; rank++) {
             proc.rank = rank;
+            pmix_output(0, "_store_job_info for nspace=%s, rank=%d", ns->nspace, proc.rank);
             rc = _store_job_info(&proc);
             if (PMIX_SUCCESS != rc) {
                 PMIX_ERROR_LOG(rc);
```

Here is run `./pmix_test` for two ranks, each rank in different nspace (`smoky_nspace-0` with rank 0, and `smoky_nspace-1` with rank 1):
```
user@fe ~/pmix/master/test $ ./pmix_test -n 2 --ns-dist 1:1 -e ./pmix_init_finalize
[fe:04085] dstore_register_job_info: nspace=smoky_nspace-0, nprocs=1
[fe:04085] _store_job_info for nspace=smoky_nspace-0, rank=0                   <---- rank 0
                                       ^^^^^^^^^^^^^^^^^^^^^
[fe:04085] dstore_register_job_info: nspace=smoky_nspace-1, nprocs=1
[fe:04085] _store_job_info for nspace=smoky_nspace-1, rank=0                   <---- rank 0
                                       ^^^^^^^^^^^^^^^^^^^^^
Client smoky_nspace-0:0 Running...
Client smoky_nspace-1:1 Running...
Client smoky_nspace-1:1 num procs 1
Client smoky_nspace-1:1 PMIx_Get hostname of 0 failed: -17
Client smoky_nspace-0:0 num procs 1
Client smoky_nspace-1:1 Finalizing...
Client smoky_nspace-0:0 hostname of rank 0 is node1
Client smoky_nspace-0:0 Finalizing...
Client smoky_nspace-1:1 PMIx_Finalize successfully completed
Client smoky_nspace-0:0 PMIx_Finalize successfully completed
pmix_test.c:main: Test finished OK!
```
For this case `_store_job_info` was not invoked for rank 1, it stored jdata for ns `smoky_nspace-1` with rank 0, but rank 0 is not assigned for `smoky_nspace-1`.

rhc54: It looks to me like your test is incorrect. Every nspace must start with rank=0 within that nspace. It makes no sense to have an nspace that doesn't have a zero rank.

artpol84: * I agree with @rhc54 that namespace should have a sequential rank numbering starting from 0.
* It looks like PMIx interface currently allows to register clients differently (as this "incorrect" test is working).

@karasevb even given that our implementation is more permissive then it is required I still don't understand why `PMIX_LIST_FOREACH(rinfo, &ns->ranks, pmix_rank_info_t)` doesn't cover all of the ranks. Can you please check why not all of the ranks are in this list.
I'd like to have a root cause.

rhc54: The list only includes the ranks that are _local_ to that PMIx server. We need to have the data from _all_ ranks of the nspace so that things like hostname can be resolved.

artpol84:
I see, then indeed it is a problem and the fix looks right to me.

artpol84:
Thanks for tracking this, Josh!

jjhursey:
Per discussion - I think we have verbal approval of this PR.

src/mca/gds/ds12/gds_dstore.c
karasevb: I take my statements back, I mistakenly believed that we can have multiple namespaces with different ranks. Thanks for the clarification.

PR #804, opened 2018-08-15
Dstore optimizations

karasevb:
1. Added support for multiple dstore format module.
2. Optimize key search:
2.1. Use flags to identify Extension and Invalid 
2.2. Use hash function as a faspath for non-matching strings strings
3. Remove extra extention slots. 
4. Remove extra mallocs.

### DIFF
--- a/src/mca/common/dstore/dstore_file.h
+++ b/src/mca/common/dstore/dstore_file.h
@@ -1,0 +1,108 @@
+#ifndef DSTORE_FORMAT_H
+#define DSTORE_FORMAT_H
+
+typedef size_t (*pmix_common_dstore_kv_size_fn)(uint8_t *addr);
+typedef char* (*pmix_common_dstore_key_name_ptr_fn)(uint8_t *addr);
+typedef size_t (*pmix_common_dstore_key_name_len_fn)(char *key);
+typedef uint8_t* (*pmix_common_dstore_data_ptr_fn)(uint8_t *addr);
+typedef size_t (*pmix_common_dstore_data_size_fn)(uint8_t *addr, uint8_t* data_ptr);
+typedef size_t (*pmix_common_dstore_key_size_fn)(char *key, size_t data_size);
+typedef size_t (*pmix_common_dstore_slot_size_fn)(void);
+typedef void (*pmix_common_dstore_put_key_fn)(uint8_t *addr, char *key, void *buf,
+                                              size_t size);
+
+typedef struct {
+    const char *name;
+    pmix_common_dstore_kv_size_fn kval_size;
+    pmix_common_dstore_key_name_ptr_fn kname_ptr;
+    pmix_common_dstore_key_name_len_fn kname_len;
+    pmix_common_dstore_data_ptr_fn data_ptr;
+    pmix_common_dstore_data_size_fn data_size;
+    pmix_common_dstore_key_size_fn key_size;
+    pmix_common_dstore_slot_size_fn slot_size;
+    pmix_common_dstore_put_key_fn put_key;
+} pmix_common_dstore_file_cbs_t;
+
+#define ESH_REGION_EXTENSION        "EXTENSION_SLOT"
+#define ESH_REGION_INVALIDATED      "INVALIDATED"
+#define ESH_ENV_INITIAL_SEG_SIZE    "INITIAL_SEG_SIZE"
+#define ESH_ENV_NS_META_SEG_SIZE    "NS_META_SEG_SIZE"
+#define ESH_ENV_NS_DATA_SEG_SIZE    "NS_DATA_SEG_SIZE"
+#define ESH_ENV_LINEAR              "SM_USE_LINEAR_SEARCH"
+
+#define ESH_MIN_KEY_LEN             (sizeof(ESH_REGION_INVALIDATED))
+
+#define PMIX_DS_PUT_KEY(rc, ctx, addr, key, buf, size)              \
+    do {                                                            \
+        rc = PMIX_ERROR;                                            \
+        if ((ctx)->file_cbs && (ctx)->file_cbs->put_key) {          \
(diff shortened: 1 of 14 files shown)


karasevb:
image

karasevb:
@artpol84 please review

src/client/pmix_client_get.c
```diff
@@ -83,14 +83,6 @@ PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc, const char key[],
...
     pmix_status_t rc;
 
-    PMIX_ACQUIRE_THREAD(&pmix_global_lock);
-
-    if (pmix_globals.init_cntr <= 0) {
-        PMIX_RELEASE_THREAD(&pmix_global_lock);
-        return PMIX_ERR_INIT;
-    }
-    PMIX_RELEASE_THREAD(&pmix_global_lock);
-
```
rhc54: I take it you are then relying on checking for initialization in the non-blocking call? If so, we probably should review the code that executes prior to calling the nb function to ensure that nothing there relies on PMIx being initialized

rhc54: Actually, you don't have to look far - line 95 (in your changed version) calls PMIX_NEW which requires that the object system be initialized. I'm afraid the lock has to remain.

karasevb: This change has been rejected.

rhc54:
@karasevb It occurs to me that one could replace that PMIX_ACQUIRE_THREAD code with an atomic lock, which might (should?) be faster. We would have to do it everywhere we check for initialization, but I _believe_ we don't have to worry about thread stall here as we aren't engaging in any operations (just checking the init flag).

src/mca/gds/ds12/gds_ds12_base.c
```diff
@@ -23,15 +23,22 @@
...
 static pmix_common_dstore_ctx_t *ds12_ctx;
 
 static pmix_status_t ds12_init(pmix_info_t info[], size_t ninfo)
 {
     pmix_status_t rc = PMIX_SUCCESS;
+    pmix_common_dstore_file_cbs_t *dstore_file_cbs = NULL;
 
-    ds12_ctx = pmix_common_dstor_init(&pmix_ds12_lock_module, "ds12", info, ninfo);
+    if (!PMIX_PROC_IS_SERVER(pmix_globals.mypeer)) {
+        dstore_file_cbs = &pmix_ds20_file_module;
```
artpol84: Why no ds12 here?

src/mca/common/dstore/dstore_file.h
```diff
@@ -0,0 +1,161 @@
+#ifndef DSTORE_FORMAT_H
+#define DSTORE_FORMAT_H
+
+typedef size_t (*pmix_common_dstore_kv_size_fn)(uint8_t *addr);
+typedef char* (*pmix_common_dstore_key_name_ptr_fn)(uint8_t *addr);
+typedef size_t (*pmix_common_dstore_key_name_len_fn)(char *key);
+typedef uint8_t* (*pmix_common_dstore_data_ptr_fn)(uint8_t *addr);
+typedef size_t (*pmix_common_dstore_data_size_fn)(uint8_t *addr, uint8_t* data_ptr);
+typedef size_t (*pmix_common_dstore_key_size_fn)(char *key, size_t data_size);
+typedef size_t (*pmix_common_dstore_slot_size_fn)(void);
```
artpol84: Is this for the extension slot? If so this should be reflected in the name of a function

src/mca/gds/ds12/gds_ds12_base.c
karasevb: The ds12 module has two dstore formats for 12 and 20. By default, it uses v20 format for dstor, but if a client will v12 then it is changed to v12.

src/mca/common/dstore/dstore_file.h
karasevb: Updated.

artpol84:
@karasevb what is the status of this?

rhc54:
Is this PR no longer required?

karasevb:
We have discussed with @artpol, this PR should go. I will resolve merge conflicts.

artpol84:
@rhc54 we will need this to be released along with the new dstor component as these changes are changing the format of the dstor.
Has "ds21" component been already released as part of 2.1 or 3.0?

artpol84:
I see that 2.2 and 3.0 are @ the RC stages.
Would want to go add it there.

karasevb:
Please do not merge until I complete the double check with MTT.

karasevb:
Ready to go.

jjhursey:
@karasevb Can you add a high level description of the optimization contained in this PR? Can you add it to the PR description so it makes it into the git logs?

The performance graph shows improvement, but I don't know what the design change was.

karasevb:
@jjhursey Done. The description of PR has been updated.

jjhursey:
Have you all checked the cross-version compatibility for this change?

It's fine to merge into master (I'll let @artpol84 review/merge), but before we merge to the release branches we need an assessment of cross-version compatibility.

artpol84:
@jjhursey thanks for not merging it.
I wanted to check additionally commit that avoid extra extension slots and I think though it is correct it is more a hack to unroll undesired behaviour rather than a root-cause fix.
I'm going to prepare a fix now and will open a new PR with the set of commits from this PR + a proper extension slot fix.

rhc54:
> @jjhursey do we have an automatic way of checking cross-version compatibility?
> As we only change the format of the ds21 we are not expecting to break it, but it would be good to double-check this.

We have the following script: https://github.com/pmix/pmix-tests/blob/master/crossversion/xversion.py. Would love to see it in MTT tests, though we might want to update it to work with posted tarballs or some other means of detecting that a branch changed 😄

karasevb:
Closed due #908

PR #810, opened 2018-08-17
Missed one spot where we need to avoid the cbfunc

rhc54:


rhc54:
Fixes https://github.com/pmix/prrte/issues/94

Issue #817, opened 2018-08-21
inline is not correctly defined in v2.1.3


ggouaillardet:
In `v2.1.3`, there are tons of warnings when building, for example

```
In file included from /home/gilles/src/pmix-v2.1/src/mca/base/pmix_mca_base_var.h:66:0,
                 from /home/gilles/src/pmix-v2.1/src/mca/gds/gds.h:21,
                 from /home/gilles/src/pmix-v2.1/src/include/pmix_globals.h:43,
                 from ../../../src/pmix-v2.1/examples/server.c:31:
/home/gilles/src/pmix-v2.1/src/class/pmix_value_array.h:67:19: warning: ‘pmix_value_array_init’ defined but not used [-Wunused-function]
 static inline int pmix_value_array_init(pmix_value_array_t *array, size_t item_sizeof)
                   ^
/home/gilles/src/pmix-v2.1/src/class/pmix_value_array.h:85:19: warning: ‘pmix_value_array_reserve’ defined but not used [-Wunused-function]
 static inline int pmix_value_array_reserve(pmix_value_array_t* array, size_t size)
                   ^
/home/gilles/src/pmix-v2.1/src/class/pmix_value_array.h:108:22: warning: ‘pmix_value_array_get_size’ defined but not used [-Wunused-function]
 static inline size_t pmix_value_array_get_size(pmix_value_array_t* array)
                      ^
/home/gilles/src/pmix-v2.1/src/class/pmix_value_array.h:161:21: warning: ‘pmix_value_array_get_item’ defined but not used [-Wunused-function]
 static inline void* pmix_value_array_get_item(pmix_value_array_t *array, size_t item_index)
                     ^
/home/gilles/src/pmix-v2.1/src/class/pmix_value_array.h:226:19: warning: ‘pmix_value_array_append_item’ defined but not used [-Wunused-function]
 static inline int pmix_value_array_append_item(pmix_value_array_t *array, const void *item)
                   ^
/home/gilles/src/pmix-v2.1/src/class/pmix_value_array.h:244:19: warning: ‘pmix_value_array_remove_item’ defined but not used [-Wunused-function]
 static inline int pmix_value_array_remove_item(pmix_value_array_t *array, size_t item_index)
```

I think the root cause is `inline` is not correctly macro-defined
From `config.status`
```
D["inline"]=" "
```

A possible fix is to backport pmix/pmix@ca510b3bf15975b848ee4b65c0565a89f085dabc

In `config.status` we end up we
```
[...]

PR #818, opened 2018-08-21
configury: define C11 macros once

rhc54:
Revamp PMIX_PROG_CC_C11 macro in order to define macros only once.
Otherwise, macros get redefined during the configure process and
issue a bunch of warning in config.log. That would also cause
PMIx fail to build if compiled with "-Werror"

(back-ported from upstream commit open-mpi/ompi@c9d3c81cbf97df97ebe0de8e4761e8c7a9278f48)

Signed-off-by: Gilles Gouaillardet <gilles@rist.or.jp>`
(cherry picked from commit ca510b3bf15975b848ee4b65c0565a89f085dabc)

### DIFF
--- a/config/pmix_setup_cc.m4
+++ b/config/pmix_setup_cc.m4
@@ -25,52 +25,44 @@ dnl $HEADER$
 dnl
 
 AC_DEFUN([PMIX_CC_HELPER],[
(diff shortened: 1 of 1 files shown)


rhc54:
Fixes #817

PR #827, opened 2018-09-11
common/pmix_strings: silence warnings

ggouaillardet:
Since the answer string is large enough, there is no need to
initialize it nor use strncpy().
This also silences warnings emitted by -Wformat-truncation

### DIFF
--- a/src/common/pmix_strings.c
+++ b/src/common/pmix_strings.c
@@ -12,6 +12,8 @@
  * Copyright (c) 2007-2012 Los Alamos National Security, LLC.
  *                         All rights reserved.
  * Copyright (c) 2014-2018 Intel, Inc. All rights reserved.
+ * Copyright (c) 2018      Research Organization for Information Science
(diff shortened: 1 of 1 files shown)


rhc54:
I don't disagree with the changes, but I'm unable to understand how you can get a format truncation warning out of this code - which may be why I'm not seeing any warnings. Can you explain?

ggouaillardet:
I observed the warning with gcc 8.2.0

Previously, we `strncpy(..., "FOO", strlen("FOO"))`, so my guess is the compiler complains we do not write the NULL terminating character. We could `strncpy(..., "FOO", strlen("FOO)+1)` but this is imho equivalent to the shorter `strcpy(..., "FOO")` form.
To be complete, that did not matter because we previously `memset(..., 0)` the result string, so the result string was always NULL terminated.

rhc54:
Wow, that seems like an absurdly strict compiler! Think I'll avoid going up there for awhile. Thx for the explanation.

PR #843, opened 2018-09-27
Convert the code base to use pmix_strncpy

rhc54:
We have been seeing lots and lots of warnings appearing on certain
combinations of compiler and environment due to unused return value from
strncpy and the possibility that the resulting string could fill the
provided array without leaving room for a NULL terminator. In addition,
the security risk from non-NULL-terminated strings being input by the
user has been flagged.

Therefore, transition all uses of "strncpy" to a new "pmix_strncpy"
function that resolves the warnings and closes the security gap.

### DIFF
--- a/bindings/python/construct.py
+++ b/bindings/python/construct.py
@@ -13,6 +13,7 @@ def signal_handler(signal, frame):
 
 def harvest_constants(options, src, constants, definitions):
     global takeconst, takeapis, takedtypes
+
     path = os.path.join(options.src, src)
     # open the file
     try:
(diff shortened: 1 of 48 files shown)


rhc54:
@karasevb Could you please review the dstore code areas? Note that ```pmix_strncpy``` requires a length parameter that is one less than the size of the input destination array so it has room for the NULL terminator.

rhc54:
@ggouaillardet Would you mind applying your eyes to this for sanity check?

ggouaillardet:
My first impression is that `strncpy()` does not do what most of us expect.

From the man page

```
Copies the first num characters of source to destination. If the end of the source C string (which is signaled by a null-character) is found before num characters have been copied, destination is padded with zeros until a total of num characters have been written to it.

No null-character is implicitly appended at the end of destination if source is longer than num. Thus, in this case, destination shall not be considered a null terminated C string (reading it as such would overflow).
```

so `strncpy()` does **not** guarantee the string will be null terminated, but generally pads with more zero than necessary.

so we generally want to use `strlcpy()`. Unfortunately, this has never been standardized and is available on (some) BSD distros.

Back to this PR, I'd rather rename the subroutine `pmix_strlcpy()` and it could be #define'd to `strlcpy()` when available.

We can also take this opportunity to implement `pmix_strlcat()` and check if it should be used.

Issue #824, opened 2018-08-23
PMIx dstore (ds21) runtime issues


jjhursey:
Creating a ticket to follow the discussion from this thread to make it easier to find:
 * https://github.com/pmix/pmix/pull/786#issuecomment-413939275

We are trying to use the `ds21` (new dstore) component with the PMIx 2.x branch, and hitting some crashes inside the server side, like the following:
```
#8  0x00002000004d6a3c in __strlen_power8 () from /lib64/libc.so.6
#9  0x00002000024137f4 in connection_handler () from /gpfs/alpinetds/ven201/scratch/jjhursey/local/smpi/current/lib/pmix/mca_ptl_usock.so
#10 0x00002000002c2b10 in event_process_active_single_queue (activeq=0x212e9380, base=0x2137f8f0) at event.c:1368
#11 event_process_active (base=<optimized out>) at event.c:1438
#12 event_base_loop (base=0x2137f8f0, flags=<optimized out>) at event.c:1642
#13 0x00002000001d6bac in progress_engine () from /gpfs/alpinetds/ven201/scratch/jjhursey/local/smpi/current/lib/libpmix.so.2
#14 0x0000200000388af4 in start_thread () from /lib64/libpthread.so.0
#15 0x0000200000548814 in clone () from /lib64/libc.so.6
```

This was testing an older port of the `ds21` to the PMIx 2.x branch. Since that data point new PRs have been opened:
 * v3.x branch: https://github.com/pmix/pmix/pull/814
 * v2.x branch: https://github.com/pmix/pmix/pull/815

We need to retry with the updated v2.x branch to see if the problem is resolved.

jjhursey:
@karasevb @artpol84 I'm hoping to be able to retry in the next few days. I'll update this ticket with what I find. If I can still reproduce with the branch, hopefully I'll come back with some more details to help us dig in.

jjhursey:
I'm working on a build today. Hopefully be able to test later today. Sorry for the delay.

jjhursey:
I hit this same error testing https://github.com/pmix/pmix/pull/815 on the second run against the same PMIx server. Well just a touch different:
```
#7  <signal handler called>
#8  0x00002000004cb1e8 in __memcpy_power7 () from /lib64/libc.so.6
#9  0x0000200002435a40 in connection_handler () from /gpfs/alpinetds/ven201/scratch/jjhursey/local/smpi/current/lib/pmix/mca_ptl_tcp.so
[...]

PR #861, opened 2018-10-05
ptl/tcp: Make try connect retry/timout MCA adjustable

jjhursey:
* Instead of retrying 2 times and waiting for 2 seconds, allow the user
   to adjust these values via new MCA parameters
   - `ptl_tcp_handshake_wait_time`
   - `ptl_tcp_handshake_max_retries`

### DIFF
--- a/src/mca/ptl/tcp/ptl_tcp.c
+++ b/src/mca/ptl/tcp/ptl_tcp.c
@@ -14,6 +14,7 @@
  * Copyright (c) 2011-2013 Los Alamos National Security, LLC.  All rights
  *                         reserved.
  * Copyright (c) 2013-2018 Intel, Inc. All rights reserved.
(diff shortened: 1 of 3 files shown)


jjhursey:
Ref Issue #824 

I'll update the ticket with details, but the issue we were seeing was related to these values not being set high enough for that highly oversubscribed scenario. We are still trying to see why the error path lead to corruption, but this change allows us to dynamically adjust values for the clients when we know that we are in that scenario.

rhc54:
Hmmm...I have no problem with doing this, and it appears those values aren't actually being used. However, their description says that they relate to a "connection file" - a little different from how you are using them. Would it make sense to update the descriptions so the help message is more accurate?

jjhursey:
@rhc54 I added 2 new MCA's for this PR instead of using the old values. Let me know what you think.

PR #863, opened 2018-10-06
ptl/tcp: Make try connect retry/timout MCA adjustable

jjhursey:
Ref Issue #824


rhc54:
@jjhursey Just an FYI: the commit I added fixed a memory corruption problem. Might be related to what you were seeing earlier.

PR #864, opened 2018-10-08
Allow for multiple children in the waitpid loop

jjhursey:
- If we get 1 SIGCHLD for multiple children exiting at the same time,
   then the `simtest` will hang because it called `return` instead of
   `break`.

Signed-off-by: Joshua Hursey <jhursey@us.ibm.com>


jjhursey:
@nysal Found this while working on the cross-version testing.

PR #869, opened 2018-10-10
Pull the gds locking up to server level to improve performance

jjhursey:
* Move the GDS (in particular the dstore) locks up to the `_mdxcbfunc`
   level to avoid the over head of locking and unlocking for each key
   set in the modex data.

### DIFF
--- a/src/mca/common/dstore/dstore_base.c
+++ b/src/mca/common/dstore/dstore_base.c
@@ -325,6 +325,12 @@ static inline pmix_peer_t * _client_peer(pmix_common_dstore_ctx_t *ds_ctx);
 
 static inline int _my_client(const char *nspace, pmix_rank_t rank);
 
+static pmix_status_t _acquire_ns_lock_by_name(pmix_common_dstore_ctx_t *ds_ctx,
+                                              const char *nspace);
+
+static pmix_status_t _release_ns_lock_by_name(pmix_common_dstore_ctx_t *ds_ctx,
+                                              const char *nspace);
+
 static pmix_status_t _dstore_store_nolock(pmix_common_dstore_ctx_t *ds_ctx,
                                    ns_map_data_t *ns_map,
                                    pmix_rank_t rank,
@@ -662,6 +668,7 @@ static inline int _esh_session_tbl_add(pmix_common_dstore_ctx_t *ds_ctx, size_t 
 
     for(idx = 0; idx < size; idx ++) {
         if (0 == s_tbl[idx].in_use) {
+            s_tbl[idx].active_locks = 0;
             goto done;
         }
     }
@@ -2031,7 +2038,7 @@ PMIX_EXPORT pmix_status_t pmix_common_dstor_store(pmix_common_dstore_ctx_t *ds_c
     }
 
     /* set exclusive lock */
-    rc = _ESH_LOCK(ds_ctx, ns_map->tbl_idx, wr_lock);
+    rc = _acquire_ns_lock_by_name(ds_ctx, proc->nspace);
     if (PMIX_SUCCESS != rc) {
         PMIX_ERROR_LOG(rc);
         goto exit;
@@ -2044,7 +2051,7 @@ PMIX_EXPORT pmix_status_t pmix_common_dstor_store(pmix_common_dstore_ctx_t *ds_c
     }
 
     /* unset lock */
-    rc = _ESH_LOCK(ds_ctx, ns_map->tbl_idx, wr_unlock);
+    rc = _release_ns_lock_by_name(ds_ctx, proc->nspace);
     if (PMIX_SUCCESS != rc) {
         PMIX_ERROR_LOG(rc);
         goto exit;
@@ -2625,6 +2632,89 @@ static inline int _my_client(const char *nspace, pmix_rank_t rank)
     return local;
 }
 
+static pmix_status_t _acquire_ns_lock_by_name(pmix_common_dstore_ctx_t *ds_ctx,
+                                              const char *nspace)
+{
+    pmix_status_t rc = PMIX_SUCCESS;
+    ns_map_data_t *ns_map;
+    session_t *s_tbl;
+
+    if (NULL == (ns_map = ds_ctx->session_map_search(ds_ctx, nspace))) {
+        rc = PMIX_ERROR;
+        PMIX_ERROR_LOG(rc);
+        return rc;
+    }
+
+    s_tbl = PMIX_VALUE_ARRAY_GET_BASE(ds_ctx->session_array, session_t);
+
+    if (0 == s_tbl[ns_map->tbl_idx].active_locks) {
+        rc = _ESH_LOCK(ds_ctx, ns_map->tbl_idx, wr_lock);
+        if (PMIX_SUCCESS != rc) {
+            PMIX_ERROR_LOG(rc);
+            return rc;
+        }
+    }
+
+    /*
+     * Maintain a reference count instead of a boolean 
+     * so we hang if someone misses an unlock.
+     */
+    s_tbl[ns_map->tbl_idx].active_locks++;
+
+    return PMIX_SUCCESS;
+}
+                                            
+
+PMIX_EXPORT pmix_status_t pmix_common_dstor_acquire_ns_lock(pmix_common_dstore_ctx_t *ds_ctx,
+                                                            struct pmix_namespace_t *nspace)
+{
+    pmix_namespace_t *ns = (pmix_namespace_t*)nspace;
+
+    return _acquire_ns_lock_by_name(ds_ctx, ns->nspace);
+}
+
+static pmix_status_t _release_ns_lock_by_name(pmix_common_dstore_ctx_t *ds_ctx,
+                                              const char *nspace)
+{
+    pmix_status_t rc = PMIX_SUCCESS;
+    ns_map_data_t *ns_map;
+    session_t *s_tbl;
+
+    if (NULL == (ns_map = ds_ctx->session_map_search(ds_ctx, nspace))) {
+        rc = PMIX_ERROR;
+        PMIX_ERROR_LOG(rc);
+        return rc;
+    }
+
+    s_tbl = PMIX_VALUE_ARRAY_GET_BASE(ds_ctx->session_array, session_t);
+    s_tbl[ns_map->tbl_idx].active_locks--;
+
+    if (0 == s_tbl[ns_map->tbl_idx].active_locks) {
+        rc = _ESH_LOCK(ds_ctx, ns_map->tbl_idx, wr_unlock);
+        if (PMIX_SUCCESS != rc) {
+            s_tbl[ns_map->tbl_idx].active_locks++;
+            PMIX_ERROR_LOG(rc);
+            return rc;
+        }
+    }
+    else if (0 > s_tbl[ns_map->tbl_idx].active_locks) {
+        s_tbl[ns_map->tbl_idx].active_locks++;
+        rc = PMIX_ERROR;
+        PMIX_ERROR_LOG(rc);
+        return rc;
+    }
+
+    return PMIX_SUCCESS;
+}
+
+PMIX_EXPORT pmix_status_t pmix_common_dstor_release_ns_lock(pmix_common_dstore_ctx_t *ds_ctx,
+                                                            struct pmix_namespace_t *nspace)
+{
+    pmix_namespace_t *ns = (pmix_namespace_t*)nspace;
+
+    return _release_ns_lock_by_name(ds_ctx, ns->nspace);
+}
+
 /* this function is only called by the PMIx server when its
  * host has received data from some other peer. It therefore
  * always contains data solely from remote procs, and we
@@ -2680,6 +2770,17 @@ PMIX_EXPORT pmix_status_t pmix_common_dstor_store_modex(pmix_common_dstore_ctx_t
         PMIX_DESTRUCT(&pbkt);
         return PMIX_SUCCESS;
     }
+
+    rc = _acquire_ns_lock_by_name(ds_ctx, proc.nspace);
+    if (PMIX_SUCCESS != rc) {
+        PMIX_ERROR_LOG(rc);
+        bo->bytes = pbkt.base_ptr;
+        bo->size = pbkt.bytes_used; // restore the incoming data
+        pbkt.base_ptr = NULL;
+        PMIX_DESTRUCT(&pbkt);
+        return rc;
+    }
+
     /* unpack the remaining values until we hit the end of the buffer */
     cnt = 1;
     kv = PMIX_NEW(pmix_kval_t);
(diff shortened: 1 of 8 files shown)


ibm-ompi:

 Gist: https://gist.github.com/0413a092764148806bcd8338cf78eac4

ibm-ompi:

 Gist: https://gist.github.com/eff2b126c21440eb61b7cca4168633c8

ibm-ompi:

 Gist: https://gist.github.com/7730a12150106828b55e7c06dd819b8f

jjhursey:
We are performance tuning at full scale with high PPN, and I noticed that the accumulation of the locking overhead inside the new dstore on the server side was considerable when processing a full modex. At smaller scales, it's hardly noticeable (10's to 100's of usec of accumulation), but at large scale it was 10's of seconds.

In our scenario we are doing a full modex and each process is contributing 1-2 key/value pairs. Each node has 168 PPN. The numbers below are using `hash,ds21`.
The main loop in `pmix_server.c:_mdxcbfunc` will have an entry for each host. The time quoted below is the time to process 1 of those per-host entries.
```
Without this PR:
0.012326 usec x nodes
  -  6.310 sec @  512
  - 56.798 sec @ 4608

With this PR:
0.004025 usec x nodes
  -  2.06 sec @  512
  - 18.547 sec @ 4608
```

The numbers above were projected from small scale measurement to the node counts noted. The performance benefit at 512 nodes was later confirmed on a scale machine.

src/mca/common/dstore/dstore_base.c
```diff
@@ -2695,10 +2761,12 @@ PMIX_EXPORT pmix_status_t pmix_common_dstor_store_modex(pmix_common_dstore_ctx_t
             PMIX_DESTRUCT(&pbkt);
             return rc;
         }
-        if (PMIX_SUCCESS != (rc = pmix_common_dstor_store(ds_ctx, &proc, PMIX_REMOTE, kv))) {
+
+        if (PMIX_SUCCESS != (rc = pmix_common_dstor_store(ds_ctx, &proc, PMIX_REMOTE, kv, true))) {
```
karasevb: Since the `_mdxcbfunc` always sets `holding_ns_locks = true`: why we cannot introduce `pmix_common_dstor_store_nolock` here,  it allows to avoid add the new argument for `pmix_common_dstor_store` and `pmix_common_dstor_store_modex`? Also in this way, `pmix_common_dstor_store_modex` may not use locks.
I think it reduces the commit changes.

karasevb:
Our Jenkins has issues (`not ok`):
```
18:07:33 Checking with ds21,hash:
18:07:33 1..14
18:07:33 not ok 1 blocking fence w/ data all (./pmix_test -n 4 --ns-dist 3:1 --fence "[db | 0:0-2;1:3]" -o $OUTDIR/out)
18:07:33 not ok 2 blocking fence w/ data all (./pmix_test -n 4 --ns-dist 3:1 --fence "[db | 0:;1:3]" -o $OUTDIR/out)
18:07:33 not ok 3 blocking fence w/ data all (./pmix_test -n 4 --ns-dist 3:1 --fence "[db | 0:;1:]" -o $OUTDIR/out)
18:07:33 ok 4 non-blocking fence w/o data (./pmix_test -n 4 --ns-dist 3:1 --fence "[0:]" -o $OUTDIR/out)
18:07:33 ok 5 blocking fence w/ data (./pmix_test -n 4 --ns-dist 3:1 --fence "[b | 0:]" -o $OUTDIR/out)
18:07:33 ok 6 non-blocking fence w/ data (./pmix_test -n 4 --ns-dist 3:1 --fence "[d | 0:]" --noise "[0:0,1]" -o $OUTDIR/out)
18:07:33 ok 7 blocking fence w/ data on the same nspace (./pmix_test -n 4 --job-fence -c -o $OUTDIR/out)
18:07:33 ok 8 blocking fence w/o data on the same nspace (./pmix_test -n 4 --job-fence -o $OUTDIR/out)
18:07:33 ok 9 publish (./pmix_test -n 2 --test-publish -o $OUTDIR/out)
18:07:33 ok 10 spawn (./pmix_test -n 2 --test-spawn -o $OUTDIR/out)
18:07:33 ok 11 connect (./pmix_test -n 2 --test-connect -o $OUTDIR/out)
18:07:33 ok 12 resolve peers (./pmix_test -n 5 --test-resolve-peers --ns-dist "1:2:2" -o $OUTDIR/out)
18:07:33 ok 13 key replacement (./pmix_test -n 5 --test-replace 100:0,1,10,50,99 -o $OUTDIR/out)
18:07:33 ok 14 local store (./pmix_test -n 5 --test-internal 10 -o $OUTDIR/out)
```
Let me check what is wrong.

karasevb:
bot:mellanox:retest

src/server/pmix_server.c
```diff
@@ -2243,6 +2244,12 @@ static void _mdxcbfunc(int sd, short argc, void *cbdata)
         }
     }
 
+    // Lock down all of the namespaces
+    PMIX_LIST_FOREACH(nptr, &nslist, pmix_nspace_caddy_t) {
+        PMIX_GDS_ACQUIRE_NS_LOCK(rc, nptr->ns);
```
sam6258: We aren't checking this return code anywhere

sam6258:
@karasevb Is the ds21 component designed to be re-entrant?

It looks like that is the key difference with this PR's CI failures with ds21.

With ds12 we see the server perform `pmix_ds12_lock_wr_get` without a release of the lock:

`[c685f10n17:130869] Locking namespace 0x200004000a90, id is smoky_nspace-0
[c685f10n17:130869] ns_map->tbl_idx = 0
[c685f10n17:130869] obtain lock in pmix_common_dstor_acquire_ns_lock
[c685f10n17:130869] pmix_ds12_lock_wr_get
[c685f10n17:130869] PMIX_GDS_ACQUIRE_NS_LOCK rc is 0
[c685f10n17:130869] Locking namespace 0x200004010b40, id is smoky_nspace-1
[c685f10n17:130869] ns_map->tbl_idx = 0
[c685f10n17:130869] obtain lock in pmix_common_dstor_acquire_ns_lock
[c685f10n17:130869] pmix_ds12_lock_wr_get
[c685f10n17:130869] PMIX_GDS_ACQUIRE_NS_LOCK rc is 0`

With ds21, we hang on first re-entry:

`[c685f10n17:131854] Locking namespace 0x200004000a90, id is smoky_nspace-0
[c685f10n17:131854] ns_map->tbl_idx = 0
[c685f10n17:131854] obtain lock in pmix_common_dstor_acquire_ns_lock
[c685f10n17:131854] pmix_ds21_lock_wr_get
[c685f10n17:131854] PMIX_GDS_ACQUIRE_NS_LOCK rc is 0
[c685f10n17:131854] Locking namespace 0x200004010c50, id is smoky_nspace-1
[c685f10n17:131854] ns_map->tbl_idx = 0
[c685f10n17:131854] obtain lock in pmix_common_dstor_acquire_ns_lock`

rhc54:
IIRC, none of the GDS components are intended or designed for re-entrant use. We utilize the event library to serialize data access. The locks here are solely intended to coordinate between the server and its clients, not between threads in either entity.

jjhursey:
Let's refactor the 'acquire/release_ns' locks to take a list of namespaces. Move the loop from pmix_server.c to inside the dstore components. Then they are resolve the subset of `ns_map`s that need locking to prevent it from trying to grab the same lock twice.

I think that would give us the control we need at that layer, right?

sam6258:
There is one problem with that solution. Once you enter the `PMIX_GDS_ACQUIRE_NS_LOCK` macro, you need to look at each namespace's personality (compat) to get the ds component that it is using. You would need to aggregate multiple lists (one list for each component), and then call corresponding `acquire_ns_lock` function. After all that is said and done, that seems like it would be pretty hacky.

Another solution could be to keep the current structure of function calls, but in the dstore base, add logic that tracks outstanding locks, so we don't re-enter a lock. Perhaps an `active_lock` value could be added to `ns_map_data_t`

karasevb:
@sam6258  keeping the `active_lock` in `ns_map_data_`t is not a good idea. Since `ns_map_data_t` is an entry for tracking of the nspace session data, if someone sets the `active_lock` it will not propagate to others nspace `ns_map_data_t` entries with the same session. There will be the same deadlock.
The first way would be better.

sam6258:
bot:mellanox:retest

jjhursey:
@karasevb Scott refined the locking in the second commit on this PR which seems to address your concern (and cleanup the 'holding_lock' field to the gds modex operation).

I think this is good to go for re-review.

src/mca/common/dstore/dstore_base.c
```diff
@@ -2680,6 +2770,13 @@ PMIX_EXPORT pmix_status_t pmix_common_dstor_store_modex(pmix_common_dstore_ctx_t
         PMIX_DESTRUCT(&pbkt);
         return PMIX_SUCCESS;
     }
+
+    rc = _acquire_ns_lock_by_name(ds_ctx, proc.nspace);
+    if (PMIX_SUCCESS != rc) {
```
sam6258: Looks like we might need this code for the error case:

`        bo->bytes = pbkt.base_ptr;
        bo->size = pbkt.bytes_used; // restore the incoming data
        pbkt.base_ptr = NULL;
        PMIX_DESTRUCT(&pbkt);
return PMIX_SUCCESS;`

karasevb: :+1:

jjhursey:
bot:ibm:retest

artpol84:
@jjhursey this looks good.
As the new writer lock is heavy we need to reduce the amount of locks. I was thinking about this optimizatio, but forgot to schedule its implementation. Thanks for capturing this.

src/server/pmix_server.c
```diff
@@ -2200,6 +2200,7 @@ static void _mdxcbfunc(int sd, short argc, void *cbdata)
     char byte;
     bool found;
     pmix_collect_t ctype;
+    bool holding_ns_locks = false;
```
karasevb: Let's remove `holding_ns_locks`

src/server/pmix_server.c
karasevb: Please add this check

src/server/pmix_server.c
```diff
@@ -2309,6 +2316,14 @@ static void _mdxcbfunc(int sd, short argc, void *cbdata)
     }
 
   finish_collective:
+
+    // Unlock all of the namespaces
+    if( holding_ns_locks ) {
+        PMIX_LIST_FOREACH(nptr, &nslist, pmix_nspace_caddy_t) {
+            PMIX_GDS_RELEASE_NS_LOCK(rc, nptr->ns);
```
karasevb: Please add the check

src/server/pmix_server.c
sam6258: The reason I didnt remove `holding_ns_locks` here is for the error/early exit case before we obtain the lock. If some code path goes to `finish_collective` before acquiring the lock, then we will be unlocking something that wasn't already locked, which will in itself trigger an error.

src/server/pmix_server.c
sam6258: Check added

src/server/pmix_server.c
sam6258: Added

src/mca/common/dstore/dstore_base.c
sam6258: Added

artpol84:
This was sounding very familiar and I finally realized why.
We moved the lock up for the job-level info store: a38d149311e9d17a068f46215d2849ab08debbbe. But we missed this modex store.
@jjhursey how do those changes go together? Are they independent or you are doing this for the job-level as well?

artpol84:
So looking on the way we deal with job-level info I was thinking that we can aproach this problem from the other angle. Exactly the same way we do with job-level info: just pass the whole thing to the GDB and let it store the full set of keys allowing coarse-grained locking as opposed to storing each key individually.
This could be lighter solution requiring one additional callback to GDS and will allow to avoid exposure of locking semantic to the upper layers.

jjhursey:
I just pushed the squash, and a documentation update to `gds.h` instead of renaming the operation. The "lock" already expresses exclusivity, so just make it clear in the header that it's an exclusive lock on the segment(s) associated with that namespace.

So I think this is good to go once CI is done.

artpol84:
> We did this just for the modex. It looks like the job level data was doing the right thing. It's passed the 
> whole 'reply' buffer and dstore is locking only once around it's storage.
@jjhursey exactly. Because I figured out we can optimize this and did that, but forgot to optimize the same thing for modex codepath.

What I am proposing is to do the same for modex as I did for the job info. So ...
Instead of exposing locking to the upper layers I suggest to introduce gds.store_modex primitive that will store the data for the set of blobs belonging to ranks as opposed to storing a single key.

It seems like a clearer solution to me.

Let's go over it on the Telecon tomorrow, or I am open to discussing by the phone if you'd like. It seems like you want this feature ASAP.

jjhursey:
@artpol84 My versioning concern is the driver. We want this change in the v2.2 series. If we modify existing `gds.h` functions then we can't backport it. If we add a new function we can.

artpol84:
**Update**: We chat with @jjhursey and I think we have some progress. I think adding a new function approach is doable. 
Will discuss this tomorrow to see who is going to implement this.

The idea. For master:
* Extract and move [this code](https://github.com/pmix/pmix/blob/master/src/server/pmix_server.c#L2244:L2322) to the `pmix_gds_batch_store` or smth similar.
* Introduce a new `gds.batch_store` routine that will subsequently use the `pmix_gds_batch_store`  function with or without locks (depending on particular GDS).
* call the GDS batch store primitive to store the whole modex allowing GDS to optimize locking if needed

For existing release branches.

* Extract and move [this code](https://github.com/pmix/pmix/blob/master/src/server/pmix_server.c#L2244:L2322) to the `pmix_gds_batch_store` or smth similar. **NOTE**: assume an additional parameter for GDS store primitive. 
* Introduce a new `gds.batch_store` routine that will subsequently use the `pmix_gds_batch_store`  function passing a **lockless** version of the GDS single-key store function. 
* in the `_mdxcbfunc` detect if `store_batch` is supported. IF IT IS - call it with the modex data, OTHERWISE - call a  `pmix_gds_batch_store` directly, but passing a lock-enabled version of the single-key GDS store.

@jjhursey please confirm that we are on the same page.

rhc54:
> The proposal is to create a new GDS function to replace `pmix_gds_base_module_store_modex_fn_t` called `pmix_gds_base_module_batch_store_modex_fn_t` that, similar to `pmix_gds_base_module_register_job_info_fn_t`, will pass the whole data packet into this new GDS function for processing. The GDS component can then lock or not as it sees fit.
> 
> The performance implications are the same. This is just a bit of code movement for a better design.
> 
> For `master`:
> 
>     * The `pmix_gds_base_module_store_modex_fn_t` will be replaced by `pmix_gds_base_module_batch_store_modex_fn_t`.
> 
>     * This will change the API, so the major version will need to change and developers need to be informed.

I don't believe this statement is true - this is not an external API and therefore has no impact on versioning.

jjhursey:
Per the teleconf, @rhc54 had a good idea. Let's just reuse the existing interface:
```c
typedef pmix_status_t (*pmix_gds_base_module_store_modex_fn_t)(struct pmix_namespace_t *ns,
                                                               pmix_list_t *cbs,
                                                               pmix_byte_object_t *bo);
```

But instead of feeding the buffer in one part at a time, send it the entire `bo` object from here:
https://github.com/pmix/pmix/blob/master/src/server/pmix_server.c#L2243

And move [this loop](https://github.com/pmix/pmix/blob/master/src/server/pmix_server.c#L2244-L2299) into the GDS component's logic. Since the function signature is the same, we are just changing the semantics of it.

Since the unpacking of this blob requires a fix ordering, it would be nice to not have to duplicate the unpacking loop portion of this code in every GDS component. An effort will be needed to see how we might be able to do that with maybe a base function with a callback for the inner loop.

We will hold this PR open until the alternative is ready.

rhc54:
Closed by #897

PR #875, opened 2018-10-10
ptl/usock: Guard against assumption that zero-length credential is NULL

sam6258:
This guards against corruption seen in https://github.com/pmix/pmix/issues/824.

This will need some more discussion as to whether we should bail out of the server here or not.

### DIFF
--- a/src/mca/ptl/usock/ptl_usock_component.c
+++ b/src/mca/ptl/usock/ptl_usock_component.c
@@ -15,6 +15,7 @@
  * Copyright (c) 2016-2018 Intel, Inc.  All rights reserved.
  * Copyright (c) 2017      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2018      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -481,6 +482,10 @@ static void connection_handler(int sd, short args, void *cbdata)
             cred.bytes = ptr;
             ptr += cred.size;
             len -= cred.size;
+        } else {
+            /* set cred pointer to NULL to guard against validation
+             * methods that assume a zero length credential is NULL */
+            cred.bytes = NULL;
         }
     }
 


sam6258:
@rhc54 When you say 'as you fail the larger "if" statement', what line of code are you referring to? The way I'm seeing the code, I think a 2.1 client could enter a path that validates security, and some of those validations (mostly in older versions of pmix like v2.1) assume that if the credential is non-NULL then security is being used. In later versions (master), this scheme is replaced with a security component, which may/may not have this issue. Wouldn't returning a zero-length payload for a credential be a default type of behavior for a client acting under sec null? 

The security validation paths I see are:

master: https://github.com/pmix/pmix/blob/80d1b221fe4b9748acf82762fdd61a5a3882bff9/src/mca/ptl/usock/ptl_usock_component.c#L686 (so maybe we don't need this fix for master, as the spot we originally segv at in v2.1 is factored out)
v2.1: https://github.com/pmix/pmix/blob/a9045a7829edaae3a988bbcdb8b164b661203443/src/mca/ptl/usock/ptl_usock_component.c#L673-L681 (I believe we segv at line 677 here)

Anyways, I'm open to discussion on this. I'm not sure the best way to fix this issue. This may need to be a two part fix. One that fixes the corruption such as this, and another that determines if a client isn't sending a security credential when it should be. I'm not very familiar with the PMIx versioning -- are any versions of PMIx required to use security by default?

rhc54:
I think you may be misunderstanding me. I'm trying to understand the code path you implied with the problem description detailed in #824. My questions were motivated by my inability to replicate the problem here based on those descriptions.

It sounds like you may not be describing a problem you encountered, but rather laying out "what if" code paths. This is fine and I'm willing to help by going thru the various scenarios instead of trying to guess on best solution. I can take some time tomorrow to do so.

sam6258:
Ah I see what you mean. So, let me give a little more background, so we both understand properly. 

We originally encountered the problem at large scale with high PPN when testing the new ds21 component. We saw that tcp was failing/timing out in some instances and backing up to usock on the client. The request coming to the server was a usock request, so I duplicated this behavior by forcing the client to use usock, and skip over the tcp component as if it had failed (I did this by modifying code path for the client's rte_init). This reproduced the same stack we were able to hit with the testing of high PPN.

Here is a link to the debug branch I used to hit this issue (. I don't think I'm doing anything too malicious in the code that would make forcing this error invalid. Let me know your thoughts :) 

https://github.com/sam6258/pmix/tree/usock_repro

I changed `pmix_ptl_base_connect_to_peer` to `pmix_ptl_base_connect_to_peer_scott` and `pmix_rte_init` to `pmix_rte_init_server.` `pmix_rte_init` will call into `pmix_ptl_base_connect_to_peer_scott.` `pmix_rte_init` will be called on client init, and `pmix_rte_init_server` will be called on server and tool inits.

sam6258:
I guess the question I have yet to answer is "How is the client getting into the usock component?". And I think we need to answer that before merging this.

sam6258:
It looks like we could enter usock if a function like `connect_to_peer` failed inside of mca/ptl/base/ptl_base_stubs.c

`    PMIX_LIST_FOREACH(active, &pmix_ptl_globals.actives, pmix_ptl_base_active_t) {
        if (NULL != active->module->connect_to_peer) {
            if (PMIX_SUCCESS == active->module->connect_to_peer(peer, info, ninfo)) {
                pr->nptr->compat.ptl = active->module;
                return PMIX_SUCCESS;
            }
        }
    }`

PR #882, opened 2018-10-16
Do not error out if topo provided...

rhc54:
...and hwloc support was not built into PMIx


rhc54:
@hjelmn

PR #895, opened 2018-10-23
Enable PMIx_Group_construct to return a context ID

rhc54:
Update the PMIx_Group_construct[_nb] APIs to allow return of a requested
context ID for the new group.


rhc54:
@hjelmn

PR #897, opened 2018-10-25
Pull gds locking of modex up to improve performance.

sam6258:
Move the GDS (in particular the dstore) locks up to avoid
the over head of locking and unlocking for each key
set in the modex data.

Signed-off-by: Scott Miller <scott.miller1@ibm.com>

### DIFF
--- a/src/mca/common/dstore/dstore_base.c
+++ b/src/mca/common/dstore/dstore_base.c
@@ -325,6 +325,11 @@ static inline pmix_peer_t * _client_peer(pmix_common_dstore_ctx_t *ds_ctx);
 
 static inline int _my_client(const char *nspace, pmix_rank_t rank);
 
+static pmix_status_t _dstor_store_modex_cb(pmix_common_dstore_ctx_t *ds_ctx,
+                                                struct pmix_namespace_t *nspace,
+                                                pmix_list_t *cbs,
+                                                pmix_byte_object_t *bo);
+
 static pmix_status_t _dstore_store_nolock(pmix_common_dstore_ctx_t *ds_ctx,
                                    ns_map_data_t *ns_map,
                                    pmix_rank_t rank,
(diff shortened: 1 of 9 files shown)


sam6258:
This PR is the redesign requested in https://github.com/pmix/pmix/pull/869. With this PR, we pull the locking up, but keep locking hidden from the server.

src/mca/common/dstore/dstore_base.c
```diff
@@ -2630,6 +2635,46 @@ static inline int _my_client(const char *nspace, pmix_rank_t rank)
...
+    }
+
+    /* set exclusive lock */
+    rc = _ESH_LOCK(ds_ctx, ns_map->tbl_idx, wr_lock);
+    if (PMIX_SUCCESS != rc) {
+        goto exit;
+    }
+
+    rc = pmix_gds_base_store_modex(nspace, cbs, buf, (pmix_gds_base_store_modex_cb_fn_t)_dstor_store_modex_cb, ds_ctx);
+    if (PMIX_SUCCESS != rc) {
```
artpol84: Minor comment. I think that the way below.

In this way we don't duplicate unlocks but also preserve ulock error logging even in the case where store modex failed as well
```suggestion
rc = pmix_gds_base_store_modex(nspace, cbs, buf, (pmix_gds_base_store_modex_cb_fn_t)_dstor_store_modex_cb, ds_ctx);
if (PMIX_SUCCESS != rc) {
    PMIX_ERROR_LOG(rc);
}
/* unset lock */
rc1 = _ESH_LOCK(ds_ctx, ns_map->tbl_idx, wr_unlock)
if( PMIX_SUCCESS !=  rc1 )
    PMIX_ERROR_LOG(rc1);
    if( PMIX_SUCCESS == rc ){
        rc = rc1;
     }
}
return rc;
```

src/mca/gds/base/base.h
```diff
@@ -98,6 +105,12 @@ PMIX_EXPORT pmix_gds_base_module_t* pmix_gds_base_assign_module(pmix_info_t *inf
 PMIX_EXPORT pmix_status_t pmix_gds_base_setup_fork(const pmix_proc_t *proc,
                                                    char ***env);
 
+PMIX_EXPORT pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
+                                                           pmix_list_t *cbs,
+                                                           pmix_buffer_t *xfer,
+                                                           pmix_gds_base_store_modex_cb_fn_t cb_fn,
```
artpol84: I think it lacks indentation here.

artpol84: Also extra empty lines

src/mca/common/dstore/dstore_base.c
sam6258: I modified the code to do something similar to what you suggested

src/mca/gds/base/base.h
sam6258: Fixed indentation here and on a couple other functions.

artpol84:
Good to go

PR #900, opened 2018-10-25
Pull gds locking of modex up to improve performance.

sam6258:
Move the GDS (in particular the dstore) locks up to avoid
the over head of locking and unlocking for each key
set in the modex data.

Signed-off-by: Scott Miller <scott.miller1@ibm.com>
(cherry picked from commit 4323836d2858146252e666584912baf0a7678c52)

Conflicts:
	src/mca/common/dstore/dstore_common.h
	src/server/pmix_server.c


sam6258:
The conflict in the cherry pick were only copyright conflicts.

PR #902, opened 2018-10-27
Wrap the inner dstore kv in an object.

jjhursey:
* Dstore needs an outer wrapper for the kv pair.
   - Without this we corrupt memory on the modex operation.
 * Bug introduced in PR #897
   - The call was switched from `pmix_common_dstor_store`, which adds the
     extra wrapping (seen in this PR), to `_dstore_store_nolock` which
     expects the `kv` to already be wrapped.

### DIFF
--- a/src/mca/common/dstore/dstore_base.c
+++ b/src/mca/common/dstore/dstore_base.c
@@ -2686,6 +2686,8 @@ static pmix_status_t _dstor_store_modex_cb(pmix_common_dstore_ctx_t *ds_ctx,
     pmix_proc_t proc;
     pmix_kval_t *kv;
     ns_map_data_t *ns_map;
+    pmix_kval_t *kv2;
+    pmix_buffer_t tmp;
 
     pmix_output_verbose(2, pmix_gds_base_framework.framework_output,
                         "[%s:%d] gds:dstore:store_modex for nspace %s",
@@ -2742,6 +2744,15 @@ static pmix_status_t _dstor_store_modex_cb(pmix_common_dstore_ctx_t *ds_ctx,
             return rc;
         }
 
+        kv2 = PMIX_NEW(pmix_kval_t);
+        PMIX_VALUE_CREATE(kv2->value, 1);
+        kv2->value->type = PMIX_BYTE_OBJECT;
+
+        PMIX_CONSTRUCT(&tmp, pmix_buffer_t);
+
+        PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, &tmp, kv, 1, PMIX_KVAL);
+        PMIX_UNLOAD_BUFFER(&tmp, kv2->value->data.bo.bytes, kv2->value->data.bo.size);
+
         if (NULL == (ns_map = ds_ctx->session_map_search(ds_ctx, proc.nspace))) {
             rc = PMIX_ERROR;
             PMIX_ERROR_LOG(rc);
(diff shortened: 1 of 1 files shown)


artpol84:
I have a couple of questions here 
* Was this bug introduced by your PR's making locks more coarse grained?
* Can you explain why we had CI happy with those PRs?
* Can dstore_nolock be changed instead?

jjhursey:
This bug was caused by PR #897 (on this line)[https://github.com/pmix/pmix/pull/897/files#diff-78b2ff28c149129cd480d5032189c6f8L2698].

The old way was to call `pmix_common_dstor_store` which adds this kv wrapper (on this line)[https://github.com/sam6258/pmix/blob/4323836d2858146252e666584912baf0a7678c52/src/mca/common/dstore/dstore_base.c#L2023-L2030].

The change was to replace the `pmix_common_dstor_store` call with a direct call to `_dstore_store_nolock` however the PR forgot to include that kv wrapper. This lead to data corruption in the full modex.

The reason Scott switched to use the `_dstore_store_nolock` was to avoid a double locking situation that would be caused by calling `pmix_common_dstor_store` instead. 

I am also surprised that CI didn't catch it. I found it when I tried to pick up the `v2.2` branch and it was broken. Making this change made it pass.

jjhursey:
@artpol84 I assume your primary concern is not the correctness of this PR, but why CI missed it. Correct?

It exhibited as a memory corruption causing the application/client process to release early from the full modex fence while the server was processing the keys. I was able to reproduce on 3-4 PPC nodes.

jjhursey:
@artpol84 I'd like to get this fix into the master and the release branches so they are not broken when checked out. Any objection to merging the PR and then investigating CI as a separate activity?

artpol84:
@jjhursey I'm taking it now. My concern is that we introducing unneeded packing/unpacking cycle instead of providing a right primitive for that.

Also, I didn't quite understand the phenomena of CI not seeing this issue. Can you provide more details?

jjhursey:
If there is a way to avoid that packing/unpacking that would be great. But that started to get into core details of dstore that I didn't want to touch. If you have an alternative approach that avoids it, that would be nice.

As I mentioned in [this comment](https://github.com/pmix/pmix/pull/902#issuecomment-433937090)
> It exhibited as a memory corruption causing the application/client process to release early from the full modex fence while the server was processing the keys. I was able to reproduce on 3-4 PPC nodes.

I don't have much more than that. Sorry. It was 100% reproducible on the v2.2 branch. I'm not setup to try other branches easily at the moment.

We need to get a fix in the release branches this week, even if it's not optimal. So if the alternative fix is going to take a while then we should consider doing this as a 2 step fix. (1) this PR, (2) better version.

artpol84:
@jjhursey Ok, now I see what has happened.

We had (and still have) this packing [here](https://github.com/pmix/pmix/blob/master/src/mca/common/dstore/dstore_base.c#L2029:L2030) in `pmix_common_dstor_store` that is being called when direct modex reply arrives.

But this was missed in `pmix_common_dstor_store_modex` -> `_dstor_store_modex_cb` sequence.

I would prefer to move 
```C
 PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, &tmp, kv, 1, PMIX_KVAL);
 PMIX_UNLOAD_BUFFER(&tmp, kv2->value->data.bo.bytes, kv2->value->data.bo.size);
```
into `_dstore_store_nolock()` and change `pmix_common_dstor_store` accordingly. So we won't have 2 places where we do the same.

The downside of bringing the packing inside the lock in dmdx response is small as  it will be a minor overhead compared to an overal overhead of direct modex request.
Also as @rhc54 mentioned we are looking forward to avoid this packing in the future.

artpol84:
The CI coverage is still to be reconsidered as this was a major bug that should have been caught.

@karasevb please run our Jenkins manually to see why our tests haven't failed.

jjhursey:
@artpol84 If we move the duplicated:
```
 PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, &tmp, kv, 1, PMIX_KVAL);
 PMIX_UNLOAD_BUFFER(&tmp, kv2->value->data.bo.bytes, kv2->value->data.bo.size);
```
code into `_dstore_store_nolock` then that would work for:
 * `pmix_common_dstor_store` [here](https://github.com/pmix/pmix/blob/master/src/mca/common/dstore/dstore_base.c#L2023-L2045)
 * `_dstor_store_modex_cb ` [here](https://github.com/pmix/pmix/blob/master/src/mca/common/dstore/dstore_base.c#L2023-L2045)

But the similar mechanism for the job data doesn't match (missing the `PMIX_BFROPS_PACK` call):
 * https://github.com/pmix/pmix/blob/master/src/mca/common/dstore/dstore_base.c#L2843-L2844

Is that going to cause a problem?

PR #905, opened 2018-11-01
v3.1 Dstore optimizations

karasevb:
Corresponds to #804


artpol84:
Will need slightly different one.

PR #908, opened 2018-11-02
Dstor/fmt optimizations

artpol84:


artpol84:
Same as #804 but without extension slot optimization which is incorrect there.
Soon will provide an appropriate fix in a separate PR

artpol84:
@jjhursey I believe this changeset is well tested and can be merged.
The commit that was removed c1d1399c6996cd899ec82ccb9caf1764c34074f1 is self-contained and provides an optimization and not a bugfix.
So I believe that this changeset is good to go.
I will work on the proper replacement for c1d1399c6996cd899ec82ccb9caf1764c34074f1 today and will PR shortly.

karasevb:
Compatibility of `ds21` is broken before this PR #909, it will be fixed in another PR.

artpol84:
Partially fixes https://github.com/pmix/pmix/issues/665

Issue #665, opened 2018-02-04
PMIx/dstore performance issues


artpol84:
I'm creating this issue as the reference for the dstore locking performance issue.
POC for evaluation: https://github.com/artpol84/poc/tree/master/benchmarks/shmem_locking

Currently dstore is using global lock to arbitrate between PMIx server writing and PMIx clients reading at the same time (this is mainly for direct modex mode).

* Originally dstore was using flock but it was found to perform very poor on high core count systems like PPC and KNL: https://github.com/pmix/pmix/pull/273#issuecomment-275944934
* To partially address that issue pthread-based locking was introduced
* However detailed analysis showed that performance of atomic operations is not scalable:
![image](https://user-images.githubusercontent.com/7534910/35774780-a23d2f70-092e-11e8-9345-c52f2ef256bc.png)
![image](https://user-images.githubusercontent.com/7534910/35774788-0d701c76-092f-11e8-9399-3e235b20c392.png)
Those results was obtained with the following POC benchmarks:
  * PPC, POWER8: https://github.com/artpol84/poc/tree/master/atomics/power8
  * For KNL corresponding asm primitives derived from OPAL was used,
* On the Telecon (Jan, 22 2018) we agreed to try the locking scheme where server will maintain independent locks for each client. I've updated my POC code to support independent locking scheme (https://github.com/artpol84/poc/commit/8396509d4cb4f810b40ab8c6e41b2b1a72d9c146, https://github.com/artpol84/poc/commit/fe31bbd3b94e7ab6cb28c06fee985620b2f81da9).
Here are the first results on Power:

`pthread_prio` test corresponds to the solution currently used in PMIx:
```
$ mpirun -np 160 -mca coll ^hcoll -mca pml ob1 -mca btl self,vader --use-hwthread-cpus \
        -map-by hwthread --bind-to hwthread ./test_pthread_prio
readers WO:ovh  WO:lk/s RO:ovh[avg/min/max]     RO:lk/s RW:wovh RW:wtm  RW:rlk/s[avg/min/max]
159     5e-03   2120645 4e+00/3e+00/4e+00       2611    1e-01   1e-01   1517/950/3028
```
`pthread_nlocks` is using 2x159 independent pthread rwlock's (2 per each client) to ensure correctness
```
$ mpirun -np 160 -mca coll ^hcoll -mca pml ob1 -mca btl self,vader --use-hwthread-cpus \
        -map-by hwthread --bind-to hwthread ./test_pthread_nlocks
[...]

PR #910, opened 2018-11-02
Dstore: fix key publishing logic

artpol84:
1. Make dstor modex to be more efficient.
    Dstor _dstore_store_nolock function is able to receive multiple keys on input (packed in a buffer).
    _dstor_store_modex_cb originally was feeding it with one key-value at a time, but when possible it is preferred to pass dstor with the whole set of keys (for performance reasons).

1. Avoid extra extension slots
    Extra extension slots were placed when the key-values were passed individually. Examples are:
    - direct modex 
    - local keys at the commit time.
    
    We need to place extension slots in order to make sure that
    they are separating keys from different ranks (and thus can
    provide the "next" link if new keys are to be appended).
    
    This solution detects unneeded extension slot when keys from the same rank are being placed. These unneeded extension slots are removed prior to placing a new key.


artpol84:
This PR depends on #908

artpol84:
@jjhursey this PR affects/improves what we did in https://github.com/pmix/pmix/pull/902.

artpol84:
Partially fixes #665

PR #931, opened 2018-11-07
v2.1/psensor/file: add missing PMIX_EXPORT on mca_psensor_file_component

benmenadue:
Add missing `PMIX_EXPORT` decoration on `mca_psensor_file_component` so that it can be resolved at runtime by `dlsym`. It was already added to master by af64c7d.

See open-mpi/ompi#6056.


rhc54:
Thanks!

PR #938, opened 2018-11-10
Default the v2.0 GDS to "ds12,hash"

rhc54:


rhc54:
@karasevb This fixes the v2.0 compatibility issue - needs to be ported to v2.2 and v3.1 branches

PR #948, opened 2018-11-21
Avoid sending messages to procs that have finalized

rhc54:
Ensure we mark that a proc has finalized. Check the finalized flag prior
to each message sent from server to client


rhc54:
Fixes https://github.com/pmix/prrte/issues/109

PR #949, opened 2018-11-21
Report events only to members of the failed procs nspace

rhc54:
Flag the event for non-default event handlers. Include the affected proc's ID.

### DIFF
--- a/src/event/pmix_event.h
+++ b/src/event/pmix_event.h
@@ -192,15 +192,22 @@ void pmix_event_timeout_cb(int fd, short flags, void *arg);
             ch = PMIX_NEW(pmix_event_chain_t);                                      \
             ch->status = (e);                                                       \
             ch->range = (r);                                                        \
-            pmix_strncpy(ch->source.nspace,                                        \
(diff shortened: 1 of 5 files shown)


rhc54:
@jjhursey I believe this will fix your problem

jjhursey:
Thanks @rhc54 We will give it a try.

jjhursey:
I backported this to the 2.2 branch and tested the original reproducer there, and this PR fixed the issue. Thanks!

The test case was:
 1. Launch the RM
 2. Launch a job that calls PMIx_Abort (or terminates abnormally) in one of the ranks (call this Namespace A)
 3. Once Namespace A is complete, launch another job that would complete successfully (call this Namespace B)

The problem was that Namespace B would abort because it received the event from Namespace A, which was cached. With this PR Namespace B (and all following) worked correctly.

rhc54:
@jjhursey I'll leave it to you to backport this PR, then. Thx for confirmation!

jjhursey:
Yep. I'll work on that on Monday when I'm back in the office.

PR #953, opened 2018-11-27
Add process set name key

rhc54:


rhc54:
Looks like the Mellanox jenkins tests are out-of-sync with the code base.

Issue #929, opened 2018-11-07
PMIx server crash with undefined event callbacks


karasevb:
PMIx server got a crash for the case when it does not support register/deregister event callbacks.
There is patch for reproduce:
```patch
diff --git a/test/simple/simptest.c b/test/simple/simptest.c
index 2fc243a..99b1497 100644
--- a/test/simple/simptest.c
+++ b/test/simple/simptest.c
@@ -132,9 +132,9 @@ static pmix_server_module_t mymodule = {
     .spawn = spawn_fn,
     .connect = connect_fn,
     .disconnect = disconnect_fn,
-    .register_events = register_event_fn,
-    .deregister_events = deregister_events,
-    .notify_event = notify_event,
+    .register_events = NULL, //register_event_fn,
+    .deregister_events = NULL,//deregister_events,
+    .notify_event = NULL,//notify_event,
     .query = query_fn,
     .tool_connected = tool_connect_fn,
     .log = log_fn,
```

Run details:
```
$./simptest -n 1 -e simpclient
Testing version 4.0.0
[node-01:04814] SERVER: ERRHANDLER REGISTRATION CALLBACK CALLED WITH STATUS 0, ref=0
Collecting inventory
[node-01:04814] INVENTORY RECEIVED
[node-01:04814] INVENTORY READY FOR DELIVERY
Inventory collected: 0
[node-01:04831] Client ns foobar rank 0: Running
lt-simptest: server/pmix_server.c:3093: server_switchyard: Assertion `((0xdeafbeedULL << 32) + 0xdeafbeedULL) == ((pmix_object_t *) (cd))->obj_magic_id' failed.
[node-01:04831] PMIX ERROR: UNPACK-PAST-END in file event/pmix_event_registration.c at line 98
Aborted (core dumped)
simpclient EVENT HANDLER REGISTRATION RETURN STATUS -10, ref=4294967295
[node-01:04831] Client: ERRHANDLER REGISTRATION CALLBACK CALLED WITH STATUS 0, ref=1
[node-01:04831] Client foobar:0 universe size 1
boriska@node-01 ~/pmix_test/pmix/test/simple
[node-01:04831] Client ns foobar rank 0 cnt 0: PMIx_Commit failed: UNREACHABLE
[node-01:04831] Client ns foobar rank 0: Finalizing
Client ns foobar rank 0:PMIx_Finalize successfully completed
```

GDB details:
```
(gdb) bt
[...]

PR #954, opened 2018-11-27
Update v3.1 with all relevant commits from master

rhc54:


rhc54:
Only failures are in the Mellanox tests that are known to be broken at this time.

artpol84:
@rhc54 the way I understood @jjhursey Mellanox jenkins wasn't broken but the repo was changed in some way that broke it. I think this is an important event that needs investigation rather than ignorance.
Also, it wasn't "known" to be broken (at least we didn't know that). I think it would be better in future if you can let us know before declaring it broken and ignoring.

artpol84:
Ref. #929

rhc54:
I think you misunderstood @jjhursey, or perhaps he was also misunderstanding things as he didn't have any code contributions. The fact is that Mellanox Jenkins was running, but the tests it was executing were "out-of-sync" with the repository tests. This was due to @karasevb and yourself working on the new multi-node feature.

I assume that you are monitoring the repo and paying attention when you look at your own PR to see that all PRs are failing, including your own. I wasn't particularly concerned as I would look into the Mellanox output, see the tests that were failing, and then verified that the cmd line options couldn't possibly work with the tests as-is in the repo. Since you guys were busy working on those in a separate PR, it was easy enough to just ensure the rest of the tests were passing.

PR #959, opened 2018-11-30
Ensure the tracker is marked for lost connection

rhc54:


src/mca/ptl/base/ptl_base_sendrecv.c
```diff
@@ -126,7 +127,6 @@ void pmix_ptl_base_lost_connection(pmix_peer_t *peer, pmix_status_t err)
                 /* if there are other participants waiting for a response,
                  * we need to let them know that this proc has disappeared
                  * as otherwise the collective will never complete */
-                trk->lost_connection = true;  // mark that a peer's connection was lost
```
ggouaillardet: should you move the comments right above altogether with the single line of code ?

rhc54: I don't think so - the following check includes a comment about nobody waiting, and the check after that is to take care of the case where other participants are waiting. So I think the comments are okay where they are.

PR #963, opened 2018-12-04
Use fastpath to get keys without thread shifting, if supported by GDS

karasevb:

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -105,10 +105,8 @@ PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc,
                         (NULL == key) ? "NULL" : key);
 
     /* try to get data directly, without threadshift */
-    if (PMIX_LIKELY(true == pmix_client_globals.get_fastpath)) {
-        if (PMIX_SUCCESS == (rc = _getfn_fastpath(proc, key, info, ninfo, val))) {
-            goto done;
-        }
+    if (PMIX_SUCCESS == (rc = _getfn_fastpath(proc, key, info, ninfo, val))) {
+        goto done;
     }
 
     /* create a callback object as we need to pass it to the
@@ -500,6 +498,7 @@ static pmix_status_t _getfn_fastpath(const pmix_proc_t *proc, const pmix_key_t k
         for (n=0; n < ninfo; n++) {
             if (0 == strncmp(info[n].key, PMIX_DATA_SCOPE, PMIX_MAX_KEYLEN)) {
                 cb->scope = info[n].value.data.scope;
+                break;
             }
         }
     }
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -5,7 +5,7 @@
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2014      Artem Y. Polyakov <artpol84@gmail.com>.
  *                         All rights reserved.
- * Copyright (c) 2016      Mellanox Technologies, Inc.
+ * Copyright (c) 2016-2018 Mellanox Technologies, Inc.
  *                         All rights reserved.
  * Copyright (c) 2016      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
@@ -75,6 +75,12 @@ static void _getnb_cbfunc(struct pmix_peer_t *pr,
 
 static void _value_cbfunc(pmix_status_t status, pmix_value_t *kv, void *cbdata);
 
+static pmix_status_t _getfn_fastpath(const pmix_proc_t *proc, const pmix_key_t key,
+                                     const pmix_info_t info[], size_t ninfo,
+                                     pmix_value_t **val);
+
+static pmix_status_t process_values(pmix_value_t **v, pmix_cb_t *cb);
+
 
 PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc,
                                    const pmix_key_t key,
@@ -98,6 +104,13 @@ PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc,
                         (NULL == proc) ? PMIX_RANK_UNDEF : proc->rank,
                         (NULL == key) ? "NULL" : key);
 
+    /* try to get data directly, without threadshift */
+    if (PMIX_LIKELY(true == pmix_client_globals.get_fastpath)) {
+        if (PMIX_SUCCESS == (rc = _getfn_fastpath(proc, key, info, ninfo, val))) {
+            goto done;
+        }
+    }
+
     /* create a callback object as we need to pass it to the
      * recv routine so we know which callback to use when
      * the return message is recvd */
@@ -116,6 +129,7 @@ PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc,
     }
     PMIX_RELEASE(cb);
 
+  done:
     pmix_output_verbose(2, pmix_client_globals.get_output,
                         "pmix:client get completed");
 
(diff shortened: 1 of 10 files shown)


karasevb:
Refs #665

karasevb:
Performance measurements:
- image
- image

src/runtime/pmix_params.c
```diff
@@ -157,6 +157,13 @@ pmix_status_t pmix_register_params(void)
                                        PMIX_INFO_LVL_1, PMIX_MCA_BASE_VAR_SCOPE_ALL,
                                        &pmix_client_globals.base_verbose);
 
+    pmix_client_globals.get_fastpath = true;
+    (void) pmix_mca_base_var_register ("pmix", "pmix", "client", "get_fastpath",
+                                       "Enable fastpath for client get operations",
+                                       PMIX_MCA_BASE_VAR_TYPE_BOOL, NULL, 0, 0,
+                                       PMIX_INFO_LVL_1, PMIX_MCA_BASE_VAR_SCOPE_ALL,
+                                       &pmix_client_globals.get_fastpath);
```
rhc54: I assume this means that the user is declaring their app to be single-threaded? We looked at that case with an eye to removing threadshifting on inbound API calls, but ran into a problem as it still meant that the user's thread might be simultaneously accessing an object inside PMIx while the PMIx progress thread was also accessing it. See #951, #956, and #959 for one thread on the issue.

Seems to me you have some similar issues here. Even if you proposed code path stays inside the user's thread, you still have the problem of the PMIx progress thread touching overlapping code paths (e.g., in the dstore) stemming from requests for GDS data from other parts of the of the user's code. For example, a call to ```PMIx_Query_info_nb```.

I'm honestly not sure I see how we could make this code path work without turning "off" the PMIx internal progress thread - which would almost certainly break quite a few things.

artpol84: No, users are not required to be single-threaded. Dstore protects it's state.

artpol84: PMIx progress thread will also be protected with locks.

artpol84:
@rhc54 
> I believe this violates the Standard -you cannot execute the callback function prior to returning from the API. 
Can you provide the motivation for this restriction?

artpol84:
Here is what I found in the standard:

>\item \refconst{PMIX_SUCCESS} indicating that the request has been accepted for processing and the provided callback function will be executed upon completion of the operation
\item a non-zero \ac{PMIx} error constant indicating a reason for the request to have been rejected. In this case, the provided callback function will \emph{not} be executed
\end{itemize}
> If executed, the status returned in the provided callback function will be one of the following constants:

The only issue I see here is the word "will"

>The callback function will be executed once the specified data becomes available from the identified process and retrieved by the local server.
The \argref{info} array is used as described by the \refapi{PMIx_Get} routine.

No conflict here.

jjhursey:
From teleconf:
 - The problem is when calling the callback from inside the function - if the callback makes a call into the PMIx library (e.g., another `PMIx_Get`) then you need to be careful about exceeding the stack limit.
 - Suggestion: Move the fast path option to the blocking `PMIx_Get` to avoid the callback issue. Leave the fastpath for nonblocking `PMIx_Get_nb` for future work. We might need to change some semantics to make the nonblocking work.

karasevb:
bot:mellanox:retest

artpol84:
bot:mellanox:retest

karasevb:
Following the suggestion above:
the get fastpath was removed from `PMIx_Get_nb` and added to `PMIx_Get`.

karasevb:
@artpol84 please review the updated version

artpol84:
@rhc54 we changed this PR as we agreed on the call - only blocking PMIx_Get operation can avoid thread shifting for now. Could you take a look once again?

As the future steps to enable non-blocking version to work. In Slurm/pmix plugin I had a similar issue. I solved it by providing an additional flag to the callback that was signaling whether or not the callback is was called immediately.
Here is the implementation for the reference:
https://github.com/SchedMD/slurm/blob/c0fe0b23c0267ed7b4c4a0b32f174ee62f055f77/src/plugins/mpi/pmix/pmixp_coll_tree.c#L321
```C
static void _ufwd_sent_cb(int rc, pmixp_p2p_ctx_t ctx, void *_vcbdata)
```
Here ctx defines the context in which the callback is being called. Can be `REGULAR` of `INLINE`. In this particular callback we decide whether or not to take the lock in this case (though can be solved with recursive locks as well).

rhc54:
As for trying to extend "fastpath" to the non-blocking get operation, I'm not sure I see a mechanism for doing so as we cannot alter the callback function signature. At this point, I'm inclined to declare this as "good enough" until someone comes forward with a compelling use-case for faster get_nb operations. I'm unaware of anyone expressing that need as Spectrum uses blocking "get" calls and is (so far as I know) the only user with a performance issue.

artpol84:
> I'm not sure I understand the purpose of the MCA parameter. If you are declaring components to be thread-safe, then the param seems somewhat unnecessary. Is this provided just as a "safe opt-out" in case this approach winds up causing problems?

Correct

> it looks like there is a new lock being introduced, yes? If so, won't that be incompatible with prior releases of ds12 that don't know anything about that lock? Would it be better to exclude ds12 from the new fastpath - if nothing else, to preserve a "known good" component?

This is an internal lock that is not visible outside of the process (not in the shared memory). So we are changing internal implementation without changing the protocol.

rhc54:
> > it looks like there is a new lock being introduced, yes? If so, won't that be incompatible with prior releases of ds12 that don't know anything about that lock? Would it be better to exclude ds12 from the new fastpath - if nothing else, to preserve a "known good" component?
> 
> This is an internal lock that is not visible outside of the process (not in the shared memory). So we are changing internal implementation without changing the protocol.

I guess I'm confused, then - what is it locking? Since this is a read-only memory in the client, there is no need to lock the dstore from multiple accesses on that end - you only need to lock it if the server is doing an update, yes? In which case, isn't it something the server and client have to share?

artpol84:
A client might need to map new segments into its virtual memory. This has to be an atomic operation with respect to other local threads doing PMIx_Get.

artpol84:
in short: the lock protects client's internal bookkeeping data structures in case of multi-threaded access.

rhc54:
I think I understand. Would it make sense to:

* remove the new mca param and just default to this being used, and

* exclude ds12 from the fastpath?

I suggest this for simplicity for users. We know that ds12 is fine with the thread-shift method, so we always have a fallback position that will work. We offer the faster path for the new component where it might potentially run into problems (though we don't expect it to). If someone encounters problems, then they can always exclude the new component with the existing mca param ```gds=^ds21```

artpol84:
This sounds good to me. We will update and ping you once again.

src/client/pmix_client_get.c
```diff
@@ -473,6 +485,54 @@ static void infocb(pmix_status_t status,
...
+    pmix_cb_t *cb = PMIX_NEW(pmix_cb_t);
+    pmix_status_t rc = PMIX_SUCCESS;
+    size_t n;
+
+    /* scan the incoming directives */
+    if (NULL != info) {
+        for (n=0; n < ninfo; n++) {
+            if (0 == strncmp(info[n].key, PMIX_DATA_SCOPE, PMIX_MAX_KEYLEN)) {
+                cb->scope = info[n].value.data.scope;
+            }
```
rhc54: Could you please put a "break" statement here? There is no point in continuing to cycle thru the directives once you have found the scope.

artpol84:
@rhc54 @jjhursey would it be possible to cherry-pick this to v3.x/v2.x at some point?

rhc54:
Yes - I'll pull it into v3.1 in the morning. I just started working on v2.2 this afternoon, and I'll be sure to pull it over there as well (probably Thurs or Fri)

artpol84:
@rhc54, can you reference corresponding commits here? (For 2.2 & 3.1)

rhc54:
For 3.1, it went in here: https://github.com/pmix/pmix/pull/1001

We haven't done 2.2 yet - it won't be a cherry-pick as things drifted too much. Still working on a path forward there.

PR #966, opened 2018-12-07
 dstor: fix detection of invalidated key

karasevb:
- fix detection of invalidated key
- moved the return code of compare and check flag operations to boolean type


artpol84:
This bug was uncovered in by PR #963.
We updated jenkins and CI should bark now.
Will need this in v3.1 and v2.2

PR #989, opened 2018-12-18
cbdate of first request will be released before all requests are satisfied

dong0321:
retain cbdata for first request, otherwise this memory will be reallocate to other process the memory will rewrite

### DIFF
--- a/src/server/pmix_server_get.c
+++ b/src/server/pmix_server_get.c
@@ -490,6 +490,7 @@ static pmix_status_t create_local_tracker(char nspace[], pmix_rank_t rank,
     if (NULL == lcd){
         return PMIX_ERR_NOMEM;
     }
+    PMIX_RETAIN(cbdata);
     pmix_strncpy(lcd->proc.nspace, nspace, PMIX_MAX_NSLEN);
     lcd->proc.rank = rank;
     lcd->info = info;


src/server/pmix_server_get.c
```diff
@@ -490,6 +490,7 @@ static pmix_status_t create_local_tracker(char nspace[], pmix_rank_t rank,
     if (NULL == lcd){
         return PMIX_ERR_NOMEM;
     }
+    PMIX_RETAIN(cbdata);
```
rhc54: Hold on a second - you cannot do a retain on cbdata as it is a void*, which means you cannot know for certain that it is a PMIX_OBJECT

rhc54: I'm not seeing the code path that is causing a problem. Each peer that requests some info has its own object added to the tracker, and that object has a pointer to the cbdata provided at the time of request. There is nobody releasing it unless the request fails to be accepted.

Can you explain what problem you are actually seeing?

dong0321: The problem is after the first request is added to loc_reqs,  for the following requests' cbdata (cd at pmix_server.c:3071) has the same memory address as the first request's cd , then the loc_reqs context will be changed.  

I will dig more to see why this happens.

rhc54: Yeah, that doesn't sound right at all. We do a PMIX_NEW each time, so hard to see how it comes back with the same address.

dong0321: I think the memory is released somehow and then get reused, i am debugging to see when and where it is released.

rhc54: it will be released when the callback function gets executed and we release the overall object

dong0321: I think it is still in local_reqs list when it is released.

rhc54: ?? Then it would assert, yes? Are you sure you are at the top of master?

dong0321: For the first request if it is not satisfied and then create a local_tracker, this request is added in loc_reqs, if the return value of pmix_server_get (pmix_server.c:3072 ) is not PMIX_SUCCESS, can we do a PMIX_RELEASE(cd) below?

rhc54: Why don't you bring this to the meeting on Thurs so we can discuss it? I'm having trouble following the logic of the problem - I still don't understand how a memory address is being free'd and then reused.

dong0321: Sure

rhc54:
Resolved in a different PR

PR #990, opened 2018-12-19
Enhance group construct for local ops

rhc54:
Detect that a group construct operation involves only local procs and
optimize the operation - override it if a context ID must be assigned.
Allow user to force local-only operation.

Ensure that modex information is returned for non-local operations so
that group members can immediately communicate.


rhc54:
@hjelmn

PR #991, opened 2018-12-19
Attempt to address the case where data isn't found

rhc54:
If the server is asked for a piece of information that isn't immediately
available, we create a request tracker and add the requesting peer to
it. The precise return status, however, must be carefully controlled to
ensure that the peer object gets released when it should, and doesn't
get release twice.

### DIFF
--- a/src/server/pmix_server_get.c
+++ b/src/server/pmix_server_get.c
@@ -245,14 +245,22 @@ pmix_status_t pmix_server_get(pmix_buffer_t *buf,
         if (PMIX_ERR_NOMEM == rc) {
             PMIX_INFO_FREE(info, ninfo);
             return rc;
-        } else if (PMIX_ERR_NOT_FOUND != rc) {
-            return rc;
         }
-
-        /* do NOT create the nspace tracker here so any request
-         * by another local client that hits before the RM responds
-         * to our request will get added to the local tracker so
-         * they receive their data upon completion */
+        if (PMIX_SUCCESS == rc) {
+            /* if they specified a timeout for this specific
+             * request, set it up now */
+            if (0 < tv.tv_sec) {
+                pmix_event_evtimer_set(pmix_globals.evbase, &req->ev,
+                                       get_timeout, req);
+                pmix_event_evtimer_add(&req->ev, &tv);
+                req->event_active = true;
+            }
+            /* we already asked for this info - no need to
+             * do it again */
+            return PMIX_SUCCESS;
+        }
+        /* only other return code is NOT_FOUND, indicating that
+         * we created a new tracker */
 
         /* Its possible there will be no local processes on this
          * host, so lets ask for this explicitly.  There can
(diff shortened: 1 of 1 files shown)


rhc54:
@ggouaillardet If you get a chance, could you please look this over? I'd like to get it into v3.1.0 though I won't hold for it as it should only happen if data isn't available, which is a fairly rare corner-case.

ggouaillardet:
@rhc54 I will do that. is there an already available test case for this ?
if not, how shall I test it ?
 - proc 0 `PMIx_Get()` and then proc 1 `PMIx_Put()` later ?
- proc 0 `PMIx_Get()` and then proc 1 never `PMIx_Put()` (and proc 0 gets a reply after a timeout) ?

rhc54:
I'm afraid I don't have one - @dong0321 ?

rhc54:
Verified ok with MTT. Based on the telecon, it appears the primary driver was an incorrect return status from PRRTE - it was returning IN_PROGRESS instead of SUCCESS. We'll need to see what happens in real-world dmodex scenarios.

src/server/pmix_server_get.c
```diff
@@ -429,20 +459,18 @@ pmix_status_t pmix_server_get(pmix_buffer_t *buf,
...
-        /* if they specified a timeout, set it up now */
-        if (0 < tv.tv_sec) {
-            pmix_event_evtimer_set(pmix_globals.evbase, &req->ev,
-                                   get_timeout, req);
-            pmix_event_evtimer_add(&req->ev, &tv);
-            req->event_active = true;
+        if (PMIX_SUCCESS != rc) {
+            /* may have a function entry but not support the request */
+            PMIX_INFO_FREE(info, ninfo);
+            pmix_list_remove_item(&pmix_server_globals.local_reqs, &lcd->super);
```
dong0321: I am not sure this can be released here(at line 835 will trigger warning this is not in the list),  line 836 will cause segment fault

dong0321: Also, from what i tested,  the return value rc at line:461 is IN_PROGRESS, then this returns to pmix_server.c at line:3332 and cd will be released. Since the request is not satisfied, can we release cd ?

rhc54: I don't understand what you are describing - PRRTE master clearly returns PMIX_SUCCESS, not IN_PROGRESS, as that is what the standard requires. What are you looking at?

rhc54: > I am not sure this can be released here(at line 835 will trigger warning this is not in the list), line 836 will cause segment fault

Again, I don't understand what you are describing. We remove it from the list and release it, then return. Line 835 gets the number of elements on the list, and line 836 goes to the "cleanup" label - so how can that cause a segfault??

dong0321: Sorry i was in review mode, the line number is not correct,  when i did run the code,  at line 866 867 will cause segfault.

dong0321: I did rebased my code on master, but i realized i should do that on V3.0, I will test after i fix my repo.  Sorry about my mistake.

rhc54: No problem - let me know if you still see a problem once you get sync'd.

dong0321:
I tested with prrte V3.0 and pmix master, for V3.0 host still return PMIX_OPERATION_IN_PROGRESS.

I cherrypicked the fix commit from master to V3.0, tested with 2 nodes, 4 nodes and all works well.

rhc54:
Very good - thx! We are going to "re-branch" v3.0, so I see no purpose in updating the current branch. Thanks for your patience in tracking it down.

PR #992, opened 2018-12-19
Plug misc memory leaks

ggouaillardet:


rhc54:
bot:mellanox:retest

src/server/pmix_server_ops.c
```diff
@@ -3955,6 +3960,13 @@ static void scaddes(pmix_setup_caddy_t *p)
     if (NULL != p->peer) {
         PMIX_RELEASE(p->peer);
     }
+    PMIX_PROC_FREE(p->procs, p->nprocs);
+    if (NULL != p->apps) {
```
rhc54: I just checked - every one of these "free" macros protects against a NULL pointer input. So all of these "if (NULL !=" clauses are unnecessary. Not saying we need to resolve it in this PR, but something we probably should clean up over time.

PR #996, opened 2018-12-20
gds/ds12: plug a memory leak in pmix_ds12_lock_finalize()

ggouaillardet:
Do release the lockfile.

### DIFF
--- a/src/mca/gds/ds12/gds_ds12_lock_pthread.c
+++ b/src/mca/gds/ds12/gds_ds12_lock_pthread.c
@@ -208,6 +208,10 @@ void pmix_ds12_lock_finalize(pmix_common_dstor_lock_ctx_t *lock_ctx)
         PMIX_ERROR_LOG(PMIX_ERROR);
         return;
     }
+    if (NULL == pthread_lock->lockfile) {
+        PMIX_ERROR_LOG(PMIX_ERROR);
+        return;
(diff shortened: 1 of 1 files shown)


ggouaillardet:
Refs. pmix/prrte#156

src/mca/gds/ds12/gds_ds12_lock_pthread.c
```diff
@@ -216,6 +216,7 @@ void pmix_ds12_lock_finalize(pmix_common_dstor_lock_ctx_t *lock_ctx)
     pmix_pshmem.segment_detach(pthread_lock->segment);
 
     free(pthread_lock->segment);
+    free(pthread_lock->lockfile);
```
rhc54: Probably should set this to NULL after free?

karasevb:
But I would suggest checking them separately:
```C
    if (NULL == pthread_lock->lockfile) {
        PMIX_ERROR_LOG(PMIX_ERROR);
        return;
    }
    if (NULL == pthread_lock->segment) {
        PMIX_ERROR_LOG(PMIX_ERROR);
        return;
    }
```

ggouaillardet:
@rhc54 @karasevb I made the suggested changes.

karasevb:
@ggouaillardet Thanks!

PR #997, opened 2018-12-20
Expand the local procs/peers wildcards for group construct

rhc54:


rhc54:
@hjelmn Forgot to expand those new wildcards - will commit once it passes CI

PR #998, opened 2018-12-20
dstore/base: plug a memory leak in pmix_common_dstor_del_nspace()

ggouaillardet:
remove the tracker from the tracking array after being destructed.


ggouaillardet:
Refs. pmix/prrte#156

without this PR, the tracking array grows forever on `prte` (one new element is allocated for each new PMIx job)

PR #1011, opened 2018-12-25
Coverity fixes for test suite

artpol84:


artpol84:
@rhc54 
I think I've addressed all Coverity issues related to the test suite.
I've updated Coverity defects accordingly:
* Performance-related bugs I've sat to be "Ignored" as they are not critical for the test suite.
* Insecure data handling (related to environment variables) issues were set to "Ignore" as well.
All other issues were addressed in this PR.

I'm going to sync with @karasevb tomorrow and if all is good will merge this PR after review.

Issue #1025, opened 2019-01-06
Adding CI with Sanitizers


ax3l:
Hi,

are you planning to add CI to the PMIx repo?

One build could include a build and run with `-fsanitize=address` in order to avoid introducing memory leaks and violations in updates.

jjhursey:
PR #1265 is re-enabling the `-fsanitize=address` version in the Travis config:
 * https://github.com/openpmix/openpmix/blob/master/.travis.yml#L32-L36

We need community help on plugging the rest of the leaks to get it to pass.

ax3l:
Little note: my PR is not re-enabling them, but enabling them for the first time ;) I just added them as comments when adding CI in the first place.

rhc54:
@ax3l I finally got around to doing this in the code base, enabling the "picky" compiler options and adding the `sanitize=address` option. Even got it all to build and work on my Mac.

However, when trying to get it thru CI, I ran into problems with compilers that support that sanitize option, but appear to require the addition of another library (believe it might be -lasan?). I'm having trouble finding the correct autoconf code to locate that library.

Any suggestions?

PR #1027, opened 2019-01-06
Travis CI Script

ax3l:
Draft for adding travis-ci for memory checks: #1025 https://github.com/open-mpi/ompi/issues/6242

## Open Questions

- [x] are there automated unit tests? `make check` or `cd test && make all`? They seem to need an installed `help-pmix-runtime.txt` but can one test those without the need to install to a system path first (aka: avoid `sudo`)?

## To Do

- [x] add badge
- [x] add libevent
- [ ] general flags: `-Werror -Wall -Wextra` and maybe `-pedantic -Wshadow`
- [ ] add `-fsanitize=address` for tests in an entry

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+sudo: false
+dist: xenial
+
+cache:
+  apt: true
+
+env:
+  global:
+    # TODO: -Werror
+    # TODOL -fsanitize=address -fsanitize=undefined
+    - CFLAGS: "-Wall -Wextra"
+    - CXXFLAGS: "-Wall -Wextra"
+
+addons:
+  apt:
+    packages:
+      - gfortran
+      - perl
+      - libevent-dev
+
+jobs:
+  fast_finish: true
+  include:
+    - script:
+        - ./autogen.pl
+        - ./configure
+        - make all
+        - make check
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![Build Status](https://travis-ci.org/pmix/pmix.svg?branch=master)](https://travis-ci.org/pmix/pmix)
+
 The Process Management Interface (PMI) has been used for quite some time as a means of exchanging wireup information needed for interprocess communication. Two versions (PMI-1 and PMI-2) have been released as part of the MPICH effort. While PMI-2 demonstrates better scaling properties than its PMI-1 predecessor, attaining rapid launch and wireup of the roughly 1M processes executing across 100k nodes expected for exascale operations remains challenging.
 
 PMI Exascale (PMIx) represents an attempt to resolve these questions by providing an extended version of the PMI standard specifically designed to support clusters up to and including exascale sizes. The overall objective of the project is not to branch the existing pseudo-standard definitions - in fact, PMIx fully supports both of the existing PMI-1 and PMI-2 APIs - but rather to (a) augment and extend those APIs to eliminate some current restrictions that impact scalability, and (b) provide a reference implementation of the PMI-server that demonstrates the desired level of scalability.


ax3l:
@rhc54 can you recommend any executables/tests that I can use to check for memleaks?

rhc54:
@ggouaillardet Can you please take a look?

artpol84:
Hello,
We are using unit tests located in tests directory.
Our Jenkins scripts can be used as the reference.

ax3l:
Hi @artpol84,

is there a common target to run all unit tests? Somehow it does not look self-explanatory to me. Where are the jenkins scripts located?

artpol84:
This is the script we are using.
https://github.com/mellanox-hpc/jenkins_scripts/tree/master/jenkins/pmix
I don’t think we have Make target for it, but we can create it if needed.

ax3l:
@artpol84 thank you. yes, can you please add a simple `make check` or `make test` target for it? That would be great, also for local tests! :)

artpol84:
@karasevb, please add to your TODO list

ax3l:
Just out of curiosity, did you find time to add the `test` targets?

artpol84:
Not yet unfortunately. @karasevb is occupied with more urgent thing.

ax3l:
Ok, no problem. Just ping we when it's ready :)

rhc54:
@ax3l Sorry for the delay - I have added "make check" support in the "tests" subdirectory with PR #1197 

HTH

ax3l:
@rhc54 can you activate Travis-CI.org for this repo under https://travis-ci.org/account/repositories ?

rhc54:
I _think_ I did it - let me know

rhc54:
fwiw: i was able to manually trigger a Travis test on master, which failed of course as the travis.yml file isn't in it yet

rhc54:
@ax3l Do you want me to commit this script so we can see what Travis does with it?

ax3l:
We can try it here so far: https://travis-ci.org/ax3l/pmix
Seams there is one test failing or at least some return code is non-zero. Does the output therein make sense to you?

rhc54:
Yes, that is the correct output. I'm not sure why, but I saw 1-2 tests failing for some reason. One failed every time, but the other looked like a race condition as it only failed occasionally. Our CI supposedly is also running those tests and yet passes, so I don't know if the tests are supposed to fail (which my script doesn't recognize yet), or if the CI isn't running those tests (and so we don't realize we have a problem), or...? We are going to discuss it on today's devel call.

For now, though, I'd say your yml looks good!

rhc54:
@ax3l We good to go? I'm not worried about the test failing at the moment as we'll get that fixed.

jjhursey:
Per teleconf:
 * @artpol84 is [this version](https://github.com/pmix/pmix/blob/master/contrib/pmix_jenkins.sh) of the jenkins script up to date with what Mellanox is running?
 * There were two tests failing in `make check` that the Jenkins script was not flagging in CI. Someone should take a look. Can Mellanox confirm their CI setup and make sure this isn't silently slipping through? Thanks!

rhc54:
I have reviewed the output from a Mellanox Jenkins run and corrected the list of tests - there were indeed differences. The test script now passes. The remaining issue I'm working on is resolving the shared library problem - the tests require linking against the PMIx plugins (which are shared libraries), and so those libraries must be "found" when the tests execute.

For now, if you configure with ```--disable-dlopen``` everything will work fine. I have a PR (#1202) to link everything up correctly, but ```make tarball``` is having issues with it that still need to be worked out. Hope to have that resolved soon. Meantime, it works fine for simply building and running ```make check```, so I am going to leave ```make tarball``` hardcoded to disable dlopen and bring the rest of the changes into the repo after a little more testing (hopefully tomorrow).

ax3l:
@rhc54 thank you for the updates, the script is passing now.
Somehow it looks like only one test is run, is that correct?

rhc54:
No, there are actually closer to 10 tests being run by the test script. However, since it is a script running it, `make check` only sees it as a single reporting executable. I'll follow up by having the script only run one executable and then have `make check` invoke it multiple times so they report separately.

rhc54:
@ax3l I have updated the tests so that they report independently - there are 16 of them. I note that a couple of your check boxes were still open. Please feel free to re-open this or create a new PR to complete what you wanted to do.

PR #1029, opened 2019-01-07
epilog: fix a typo in pmix_execute_epilog()

ggouaillardet:

### DIFF
--- a/src/include/pmix_globals.c
+++ b/src/include/pmix_globals.c
@@ -2,8 +2,8 @@
 /*
  * Copyright (c) 2014-2019 Intel, Inc.  All rights reserved.
  * Copyright (c) 2014-2017 Research Organization for Information Science
- * Copyright (c) 2014-2018 Intel, Inc.  All rights reserved.
- *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2014-2019 Intel, Inc.  All rights reserved.
+ *                         and Technology (RIST).  All rights reserved.
  * Copyright (c) 2014-2015 Artem Y. Polyakov <artpol84@gmail.com>.
  *                         All rights reserved.
  * Copyright (c) 2016      IBM Corporation.  All rights reserved.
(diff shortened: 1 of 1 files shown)


src/include/pmix_globals.c
```diff
@@ -402,7 +402,7 @@ void pmix_execute_epilog(pmix_epilog_t *epi)
          * matches that of the peer - we do this to provide at least
          * some minimum level of protection */
         tmp = pmix_argv_split(cf->path, ',');
-        for (n=0; NULL != tmp[n]; n++) {
+        for (n=0; NULL != tmp && NULL != tmp[n]; n++) {
```
rhc54: It isn't possible for tmp to be NULL - if there were no epilog, then there wouldn't be an entry on the list. What am I missing?

ggouaillardet: I used the latest `master` branches from open-mpi and pmix, use pmix as an external component (both configure'd with `--disable-dlopen`, not sure this is relevant though) and `mpirun` crashed.
This patch fixed the issue and I did not dig on how we ended up here.

rhc54: We should find the root cause instead, then, as this should not be possible. It sounds like either someone is registering a NULL epilog path (which is an error), or something is getting destructed too soon.

ggouaillardet: could there be a typo at line 404 and we should `pmix_argv_split(cd->path, ...) instead (e.g. `cd` instead of `cf` which is a loop invariant) ?

rhc54: Perhaps we should add a check in the job control code for registering epilogs - if someone gives us a NULL path, then we should return an error. Might help us spot the source.

rhc54: Ah - indeed!! A copy/paste error! Good eyes - that should definitely be the fix

ggouaillardet: I will try this and update the PR

ibm-ompi:

ibm-ompi:
The IBM CI (PGI Compiler) build failed! Please review the log, linked below.

PR #1031, opened 2019-01-07
common/dstore: plug a memory leak in pmix_common_dstor_finalize()

ggouaillardet:
free the dstore context at the end


ggouaillardet:
@karasevb could you please review this ?
I am not sure whether the `dstore` context should be free'd here or in the caller.

PR #1034, opened 2019-01-08
Fix the greek version (make it 'a1')

ggouaillardet:


ggouaillardet:

I noted that when refreshing the embedded PMIx of the Open MPI `master`. the embedded version is `4.0.0` instead of `4.0.0a1` which is misleading.

PR #1044, opened 2019-01-10
event: plug a memory leak in _notify_client_event()

ggouaillardet:
do release the event evicted from the cache.

### DIFF
--- a/src/event/pmix_event_notification.c
+++ b/src/event/pmix_event_notification.c
@@ -1,7 +1,7 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2014-2019 Intel, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@rhc54 this is the leak we discussed in PMIx v3.1.

I am not sure I am being extra careful or not checking for `-1 != room` (e.g. the event was not initially hold).
If I am correct, then we might have some misc leaks in the error paths.

rhc54:
You are correct - pmix_hotel_checkout is not protected against a negative room number. Instead of a leak, I fear we would likely crash if it is called with a -1 index.

rhc54:
Alternatively, we could add the protection just to be more robust - just ignore if the index is negative. Won't hurt anything as we don't return the object anyway.

rhc54:
I'll take this as-is and then add the protection. Thx!

PR #1050, opened 2019-01-11
Fixes PMIx_generate_regex ordering

### LINKED ISSUE #1049: PMIx_generate_regex may not preserve ordering

`v2.2rc2` (it doesn't look like this changed much on `master`) tarball

-----------------------------

## Details of the problem

A PMIx server must use `PMIx_generate_regex` to generate the string value to pass for `PMIX_NODE_MAP`, and pair it with `PMIx_generate_ppn` to generate the string for `PMIX_PROC_MAP`. The ordering of these two maps provide PMIx with the node-to-rank mapping of the job.

On the client side these two keys are the backing for `PMIx_Resolve_nodes` and `PMIx_Resolve_peers` which provide access to this mapping.

If the hostnames contain more than one block of digits (e.g., `a28n03`) then the ordering is not preserved. This leads to an incorrect node-to-rank mapping being exposed to the client. If there is only one block of numbers then it seems to work ok.

#### Test program
```c
#define _GNU_SOURCE
#include <stdbool.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <time.h>

#include <pmix.h>
#include <pmix_server.h>

static pmix_proc_t myproc;

{
    int rc;
    char *regex = NULL, *input;

    if (PMIX_SUCCESS != (rc = PMIx_Init(&myproc, NULL, 0))) {
        fprintf(stderr, "Client %s:%d PMIx_Init failed: %d\n", myproc.nspace, myproc.rank, rc);
        exit(0);
    }

    input = strdup("a28n01,a28n02,a28n03,a28n04,a29n01,a29n02,a29n03");
    PMIx_generate_regex(input, &regex);

    printf("INPUT: %s\n", input);
    printf("REGEX: %s\n\n", regex);
    free(regex); free(input);

    input = strdup("a28n01,a29n02,a28n03,a28n04,a29n01,a28n02,a29n03");
    //                     ------                      ------
    PMIx_generate_regex(input, &regex);

    printf("INPUT: %s\n", input);
    printf("REGEX: %s\n\n", regex);
    free(regex); free(input);

    input = strdup("a2801,a2901,a2802,a2902,a2804,a2903");
    PMIx_generate_regex(input, &regex);

    printf("INPUT: %s\n", input);
    printf("REGEX: %s\n", regex);
    free(regex); free(input);


    if (PMIX_SUCCESS != (rc = PMIx_Finalize(NULL, 0))) {
        fprintf(stderr, "Client %s:%d PMIx_Finalize failed: %d\n", myproc.nspace, myproc.rank, rc);
    }

    fflush(stderr);
    return(0);
}
```

#### Example output

```shell
shell$ prun -np 1 ./pmix_node_regex
INPUT: a28n01,a28n02,a28n03,a28n04,a29n01,a29n02,a29n03
REGEX: pmix[a[2:28-29]n01,a[2:28-29]n02,a[2:28-29]n03,a[2:28]n04]

INPUT: a28n01,a29n02,a28n03,a28n04,a29n01,a28n02,a29n03
REGEX: pmix[a[2:28-29]n01,a[2:29,28]n02,a[2:28-29]n03,a[2:28]n04]

INPUT: a2801,a2901,a2802,a2902,a2804,a2903
REGEX: pmix[a[4:2801,2901,2802,2902,2804,2903]]
```

jjhursey:
* The problem is that `PMIx_generate_regex` must preserve ordering of the
   list passed into it. This allows for a proper mapping of nodes-to-ranks
   to occur inside of PMIx (which is then exposed to the application).
   - If the hostnames contain multiple sets of digits then the list
     can be reordered causing the mapping to be incorrect.
 * This patch prevents the regex from 'matching' an earlier prefix/suffix
   if that would mean skipping over another match just before it.
 * Fixes #1049

### DIFF
--- a/src/mca/preg/base/preg_base_frame.c
+++ b/src/mca/preg/base/preg_base_frame.c
@@ -14,6 +14,7 @@
  * Copyright (c) 2014-2017 Intel, Inc. All rights reserved.
  * Copyright (c) 2015-2016 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2019      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -98,6 +99,7 @@ static void rvcon(pmix_regex_value_t *p)
     p->prefix = NULL;
     p->suffix = NULL;
     p->num_digits = 0;
+    p->skip = false;
     PMIX_CONSTRUCT(&p->ranges, pmix_list_t);
 }
 static void rvdes(pmix_regex_value_t *p)
--- a/src/mca/preg/native/preg_native.c
+++ b/src/mca/preg/native/preg_native.c
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2015-2018 Intel, Inc.  All rights reserved.
- * Copyright (c) 2016      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2016-2019 IBM Corporation.  All rights reserved.
  * Copyright (c) 2018      Research Organization for Information Science
  *                         and Technology (RIST).  All rights reserved.
  *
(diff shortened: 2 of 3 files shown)


jjhursey:
Note that for hostnames with multiple sets of digits (e.g., `a21n05`) the regex is not optimal, and is often longer than the comma-separated list. However, with this PR the ordering is preserved which is critical to the node-to-rank mapping inside of PMIx. Optimizing the compression of the list is being left as a secondary activity.

gpaulsen:
Thanks @jjhursey and @rhc54 v3.1 and v2.2 is great!

PR #1058, opened 2019-01-13
Update v2.2 branch to meet the v2.2 Standard

rhc54:
This was too complex to cherry-pick individual commits. Instead, it was
accomplished by copying the v3.1 branch onto the v2.2 branch and then
editing out the v3-specific parts.

Signed-off-by: Ralph Castain <rhc@pmix.org>

PR #1060, opened 2019-01-13
test/simple/simptest: do not PMIX_THREADSHIFT() an object allocated o…

ggouaillardet:
…n the stack

But dynamically allocate it instead


ggouaillardet:
Refs. #1058 

@rhc54 the bug was in `simplest` indeed.
once merged, you can cherry-pick this fix into all the branches, and it will fix this issue.

this is really a memory issue that only popped up on mac with `ptl_base_verbose` set. CentOS looked fine though `valgrind` reported this (and this is how I found the root cause).

PR #1083, opened 2019-01-22
Shift the non-standard macros to pmix_extend.h

### LINKED ISSUE #1082: Slurm integration broken with PMIx v3.1
According to our MTT results PMIx v3.1 was broken for Slurm integration started from Dec 17:
https://mtt.open-mpi.org/index.php?do_redir=3159

This started from PR https://github.com/pmix/pmix/pull/984, where `PMIX_VAL_FIELD_x` macroses were moved  from `include/pmix_common.h.in` to `test/test_common.h`. These macros are used for the Slurm PMIx plugin that currently lost an integration with PMIx v3.1

rhc54:
We moved the non-standard macros out of pmix_common.h so that the
public-facing header would fully comply with the standard. However, this
meant that the Slurm plugin no longer compiled as they use those macros
there. Shifting them back to pmix_extend.h preserves the "standard only"
nature of pmix_common.h, but still exposes the non-standard macros for
now. We will remove them in later major releases after the Slurm plugin
has been updated.

Fixes #1082 

### DIFF
--- a/include/pmix_extend.h
+++ b/include/pmix_extend.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2018 Intel, Inc. All rights reserved.
+ * Copyright (c) 2013-2019 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Artem Y. Polyakov <artpol84@gmail.com>.
  *                         All rights reserved.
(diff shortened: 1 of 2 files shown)


karasevb:
@rhc54 Thanks you!

This PR restores integration with Slurm.

rhc54:
no problem!

PR #1093, opened 2019-02-04
Update the job_ctrl return code

ggouaillardet:
Return OPERATION_SUCCEEDED for cleanup registrations

(cherry picked from commit pmix/pmix@516fe69ef2ac99df032fde1c02a572524dec56ee)


ggouaillardet:
@rhc54 this fixes the hang I observes in ORTE-less Open MPI with PMIx v3.0
Refs. open-mpi/ompi#6317

rhc54:
Hmmm...well, we can certainly take it. However, I don't know when it will be included in a release. I'll look at what else (if anything) is pending.

PR #1094, opened 2019-02-04
Update configure logic for clock_gettime

rhc54:
Port some configure changes from upstream. Protect the one place where
we call clock_gettime from systems where it isn't available.


rhc54:
Refs https://github.com/open-mpi/ompi/pull/6355

Issue #1100, opened 2019-02-06
unable to build with --enable-embedded-libevent


raffenet:
### What version of the PMIx Reference Library are you using? (e.g., v1.0, v2.1, git master @ hash, etc.)

git master @ f752b0f62cba1135bf55336c2d53305523fa5dea

### Describe how PMIx was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

git clone

### Please describe the system on which you are running

* Operating system/version: Ubuntu 16.04.5 LTS
* Computer hardware: Intel Xeon 6130
* Network type: Ethernet

-----------------------------

## Details of the problem

My system does not have a libevent development package installed. I tried adding `--enable-embedded-libevent` to my configure (config.log attached), but met the following error at make time.

```
Making all in mca/bfrops
make[2]: Entering directory '/sandbox/raffenet/pmix/src/mca/bfrops'
  CC       base/bfrop_base_frame.lo
In file included from /sandbox/raffenet/pmix/src/include/pmix_globals.h:26:0,
                 from /sandbox/raffenet/pmix/src/mca/bfrops/base/base.h:40,
                 from base/bfrop_base_frame.c:39:
/sandbox/raffenet/pmix/src/include/types.h:45:27: fatal error: event.h: No such file or directory
compilation terminated.
Makefile:728: recipe for target 'base/bfrop_base_frame.lo' failed
make[2]: *** [base/bfrop_base_frame.lo] Error 1
make[2]: Leaving directory '/sandbox/raffenet/pmix/src/mca/bfrops'
Makefile:1864: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/sandbox/raffenet/pmix/src'
Makefile:748: recipe for target 'all-recursive' failed
make: *** [all-recursive] Error 1

```
[config.log](https://github.com/pmix/pmix/files/2838101/config.log)

rhc54:
There is no embedded libevent in the PMIx library - it requires that there be an external one installed on the system. Ditto for hwloc, for that matter.

raffenet:
OK. It's easy enough to build libevent on my own and point PMIx at it. However, my first instinct to look in `./configure --help` led me to believe there was an embedded one. https://github.com/pmix/pmix/blob/master/config/pmix_setup_libevent.m4#L22

rhc54:
[...]

PR #1102, opened 2019-02-07
Update configure logic

rhc54:
Remove the --enable-embedded-libevent and --enable-embedded-hwloc flags as they are confusing users. Instead, we will use the --enable-embedded-mode to handle both of these options.


rhc54:
Fixes #1100

PR #1104, opened 2019-02-07
Update configure logic

rhc54:
Remove the --enable-embedded-libevent and --enable-embedded-hwloc flags as they are confusing users. Instead, we will use the --enable-embedded-mode to handle both of these options.

(cherry picked from commit 43f06e0afc93fea4895759074c9d22448a1b5dff)


rhc54:
@ggouaillardet We will need to adjust the OMPI ext2x configure logic to properly handle the revised flags - probably need your automated version generator in here first to make it easier to ensure correct version info.

PR #1117, opened 2019-02-14
Remove the "restrict" keyword to avoid requiring C99 for apps

rhc54:

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -2710,8 +2710,8 @@ static inline void pmix_darray_destruct(pmix_data_array_t *m)
  * @param len Size of the dest array - 1
  *
  */
-static inline void pmix_strncpy(char *restrict dest,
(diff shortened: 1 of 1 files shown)


jdinan:
You can also do something like:
```
#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
#define PMIX_RESTRICT restrict
#else
#define PMIX_RESTRICT
#endif

static inline void pmix_strncpy(char * PMIX_RESTRICT dest,
                                const char * PMIX_RESTRICT src,
                                size_t len)
```

jdinan:
Was too hasty with copy/paste.  The C99 check is:
```
#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
```
`restrict` is apparently not part of C++, so removing it or gating on C99 support is probably a good idea.

PR #1120, opened 2019-02-18
Fix dstore of unknown namespace.

dsolt:
If a namespace is not registered on every node in an allocation, then
when a node requests dmodex from a remote server and the reply comes back
it will fail when it can't find a place to store it in the dstore
instead of creating a place to store it.

### DIFF
--- a/src/mca/common/dstore/dstore_base.c
+++ b/src/mca/common/dstore/dstore_base.c
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2015-2018 Intel, Inc. All rights reserved.
- * Copyright (c) 2016-2018 IBM Corporation.  All rights reserved.
+ * Copyright (c) 2016-2019 IBM Corporation.  All rights reserved.
  * Copyright (c) 2016-2018 Mellanox Technologies, Inc.
  *                         All rights reserved.
  * Copyright (c) 2018-2019 Research Organization for Information Science
@@ -481,7 +481,7 @@ static inline ns_map_data_t * _esh_session_map_search_server(pmix_common_dstore_
                 return &ns_map[idx].data;
         }
     }
-    return NULL;
+    return _esh_session_map(ds_ctx, nspace, 0, 0);
 }
 
 static inline ns_map_data_t * _esh_session_map_search_client(pmix_common_dstore_ctx_t *ds_ctx,


rhc54:
@dsolt Could you please add a sign-off line to the commit message?

karasevb:
@dsolt could you provide a simple reproducer of the problem?

dsolt:
I cannot give a reproducer with PRRTE  because PRRTE will register the namespace on all nodes, even nodes it is not launching processes on.   Not sure if there is a simple way to turn that off for testing/experimental purposes?

rhc54:
Not immediately, no - probably could add the capability when someone has time.

artpol84:
Thanks, @dsolt 
I will need to go over it with @karasevb to make sure that all is correct.
From reading your description I have a question:
> If a namespace is not registered on every node in an allocation, then when a node requests dmodex from a remote server and the reply comes back it will fail when it can't find a place to store it in the dstore instead of creating a place to store it.

 @rhc54 my understanding is that you cannot request data from the unknown namespace. If you want to work to another namespace you have to connect to it which will result in jobinfo exchange and thus namespace will become known.

rhc54:
@artpol84 Hmmm...it appears it is not quite that simple. The Standard requires that the PMIx_Connect operation result in all involved procs having access to the job-level info from all involved nspaces. However, it doesn't say that this is the _only_ way that job-level info for another nspace can be obtained.

@dsolt : how is your proc getting the name of the other nspace so it can ask for job-level info from it? Are you doing it thru some publish/lookup mechanism? I gather you are trying to communicate with a proc to which you are not "connected"? Just wondering how that can work via MPI as that would seem to violate the MPI standard.

rhc54:
I'd like us to revisit this PR as I think it actually should be committed. There are a couple of factors to consider here. First, the notion of a group "connect" operation prior to allowing communication is decidedly an MPI concept. There are plenty of non-MPI programming models that allow arbitrary messaging between procs. We should therefore be able to support somebody obtaining - thru whatever mechanism - contact information for another process and storing it.

Note that the gds/hash component already supports this behavior as it was required for use by tools, which don't have access to the dstore components since they aren't started by the host RM. As a side effect, that component now also supports non-MPI models such as Spark.

The other question here is: should PMIx_Connect absolutely _require_ that job-level information be exchanged during the operation itself for all participating namespaces? The standard states the following:
```
The PMIx definition of _connected_ solely implies the following:

- job-level information for each namespace involved in the operation
is to be made available to all processes in the connected assemblage

- any data posted by a process in the connected assemblage (via calls
to PMIx_Put committed via PMIx_Commit) prior to execution of the
PMIx_Connect operation is to be made accessible to all processes in
the assemblage - any data posted after execution of the connect
operation must be exchanged via a separate PMIx_Fence operation
spanning the connected processes

- the host environment should treat the failure of any process in the
assemblage as a reportable event, taking action on the assemblage
as if it were a single application. For example, if the environment
defaults (in the absence of any application directives) to terminating
an application upon failure of any process in that application, then the
environment should terminate all processes in the connected assemblage
upon failure of any member.
```
Thus, the standard only states that data be made _available_ to the participants - it doesn't state that the data must be exchanged during the PMIx_Connect operation so that it resides on the local host of each participant. I take this to mean that using a "direct modex" operation to fetch the data from a remote location upon first request is in fact compliant with the standard as written.

I suspect that isn't what was in our minds when we wrote this section, but in retrospect I cannot see any problem with this interpretation. It doesn't break anything and extends some flexibility in terms of implementation and programming model support. We can list the idea of performing the exchange during PMIx_Connect as a potential optimization, but not required.

Any thoughts on the matter? As I said, I believe (based on the above) that this should be committed so that the dstore and hash components behave the same way.

artpol84:
@rhc54 can you provide a usecase that is typical for Spark?

artpol84:
>Any thoughts on the matter? As I said, I believe (based on the above) that this should be committed so that the dstore and hash components behave the same way.

So the hash component creates a new namespace whenever an unknown one it is being requested?

Couple more thoughts on this:
1. As I understand, a client is free to call PMIx_Get with a bogus namespace and receive negative response. If we will create a namespace in this case this will create a bunch of useless entries increasing resource consumption.
1. For namespace registered with MPIx_Server_ function there will be a corresponding cleanup counterpart. When will we cleanup namespaces created sporadically? May cause resource leak.
1. In order to perform direct-modex request for the process from the unknown namespace, RM has to have an information about this namespace. 
The way I was interpreting the standard this knowledge is explicitly obtained throu a Connect callback.
So the RM-portion of the standard has to make it clean that direct-modex request for unconnected namespace is legal.

dsolt:
* As I understand, a client is free to call PMIx_Get with a bogus namespace and receive negative response. If we will create a namespace in this case this will create a bunch of useless entries increasing resource consumption.

* For namespace registered with MPIx_Server_ function there will be a corresponding cleanup counterpart. When will we cleanup namespaces created sporadically? May cause resource leak.

* In order to perform direct-modex request for the process from the unknown namespace, RM has to have an information about this namespace.
  The way I was interpreting the standard this knowledge is explicitly obtained throu a Connect callback.
  So the RM-portion of the standard has to make it clean that direct-modex request for unconnected namespace is legal.

> > Any thoughts on the matter? As I said, I believe (based on the above) that this should be committed so that the dstore and hash components behave the same way.
> 
> So the hash component creates a new namespace whenever an unknown one it is being requested?
> 
> Couple more thoughts on this:
> 
>     1. As I understand, a client is free to call PMIx_Get with a bogus namespace and receive negative response. If we will create a namespace in this case this will create a bunch of useless entries increasing resource consumption.

I understand this concern.  In this case, the convenience library asked the RM for modex information for a namespace and then we provide the response and the pmix CL (dstore) returns an error because they never  heard of the namespace.   This seems kind of silly from a RM perspective since the CL is the one who asked us for the namespace modex.

>     2. For namespace registered with MPIx_Server_ function there will be a corresponding cleanup counterpart. When will we cleanup namespaces created sporadically? May cause resource leak.

This seems like a valid concern.    However, I'm not sure how this is different from a tool trying to query information about a namespace.   Does a tool have to connect to the namespace it is trying to debug/etc?   I don't think that is possible since the app doesn't know to connect with the tool.  It seems to me that namespace data has to be reference counted inside the CL.  When a namespace asked about data from another non-local namespace, it adds a reference count to that block of namespace data.  When the process goes away (or is no longer a pmix client), then the reference count is decremented?

>     3. In order to perform direct-modex request for the process from the unknown namespace, RM has to have an information about this namespace.
>        The way I was interpreting the standard this knowledge is explicitly obtained throu a Connect callback.
>        So the RM-portion of the standard has to make it clean that direct-modex request for unconnected namespace is legal.

I'm not sure what you mean by this.   If the RM doesn't know about a namespace, or doesn't know how to find out about a namespace, then how could it implement connect in the first place?    What is the difference between server_connect + server_dmodex requests coming up to the RM and just server_dmodex coming up to the RM?   The server_connect request doesn't provide the RM with any information that helps it in any way to locate information about the namespace being connected to, so how does it make the job any easier for the RM?

rhc54:
> > ```
> 1. As I understand, a client is free to call PMIx_Get with a bogus namespace and receive negative response. If we will create a namespace in this case this will create a bunch of useless entries increasing resource consumption.
> ```


> I understand this concern.  In this case, the convenience library asked the RM for modex information for a namespace and then we provide the response and the pmix CL (dstore) returns an error because they never  heard of the namespace.   This seems kind of silly from a RM perspective since the CL is the one who asked us for the namespace modex.

I'm not sure this is a valid concern. We wouldn't create the namespace storage until we received the response from the RM anyway. If the RM indicates "not found", then nothing gets created.

> ```
> 2. For namespace registered with MPIx_Server_ function there will be a corresponding cleanup counterpart. When will we cleanup namespaces created sporadically? May cause resource leak.
> ```


> This seems like a valid concern.    However, I'm not sure how this is different from a tool trying to query information about a namespace.   Does a tool have to connect to the namespace it is trying to debug/etc?   I don't think that is possible since the app doesn't know to connect with the tool.  It seems to me that namespace data has to be reference counted inside the CL.  When a namespace asked about data from another non-local namespace, it adds a reference count to that block of namespace data.  When the asking namespace goes away (no more procs from that namespace are still clients on the node), then the reference count is decremented?

Tools don't "connect" to applications even when debugging. PMIx_Connect is a collaborative operation - i.e., all participants must call it. There is no way the application can know a tool might do so.

Tools are only allowed to connect to a server, just like a client proc does. If a tool asks for an unknown namespace, the expectation is that the RM will go fetch the data and return it. This is why the "hash" component supports that operation (since the tool cannot use dstore as it wasn't started by the RM).

It's true that the library doesn't today know when to remove that information. Probably the thing to do is link the info to the requestor, and then delete the info when all requestors of it have departed. Shouldn't be too hard to do.

>     3\. In order to perform direct-modex request for the process from the unknown namespace, RM has to have an information about this namespace.
        The way I was interpreting the standard this knowledge is explicitly obtained throu a Connect callback.
        So the RM-portion of the standard has to make it clean that direct-modex request for unconnected namespace is legal.

I don't think this is quite accurate. True, the RM must have some way of obtaining info about the namespace, but the local daemon doesn't necessarily have to always have it on-hand - it just needs to have some way to get it. As I noted above, the standard only requires that info be "made available", which implies that there must be some way of accessing it. The standard doesn't state that the info must be known to the host RM daemon prior to first request.

I do agree that the standard needs some clarification around these points. I'll address that and link it here.

dsolt:
Just a clarification.. my point on tools was only that tools do not have to connect to ask for information about a namespace, so what benefit is there to requiring connect to do extra steps that have to be handled anyway in the case of tools.   To put it another way, the standard attempts to layout a well defined definition of what "connected means", but in some ways that definition is useless because tools do not use the connected concept and yet they may also do many of the things that we expect "connected" namespaces to do.   Ignoring this PR for a moment, I wonder if the standard needs to either 1) relax the notion of "connected" to just be related to event notification or 2) somehow incorporate the concept of "connected-ness" to tools (which will be difficult since currently connected-ness is a collective operation.   Just something to think about.

rhc54:
@dsolt Good points - I think your first option is the best. I really don't want "connectedness" to extend to tools as I think that opens a real can of worms. Maybe if we just make it clear that "connected" means coverage for event notification, and that data on the procs that are "connected" must be obtainable (but not necessarily exchanged in the connect operation) - would that suffice to clear this up?

I really do think that we need to support the ability to "get" information on a process that may not be connected to us. Tools certainly require it - no reason why an application process should be any different.

rhc54:
Let me correct myself. The point here is that "connected" procs implies event notification coverage, and that data can be obtained on _any_ proc (connected or not) if you know the namespace and rank of that proc. The latter clarification belongs in the PMIx_Get description.

Better?

artpol84:
That sounds ok. @rhc54 can you point me to the clarifications once they are ready?

>I'm not sure this is a valid concern. We wouldn't create the namespace storage until we received the response from the RM anyway. If the RM indicates "not found", then nothing gets created.

This is certainly a valid concern, but might be that it is handled in a "lazy" way as you described and then it is not a problem. I'll discuss this with @karasevb to double check.

With respect to the implementation - the only concern I still have here is resource cleanup. It seems to be somethig on the higher level compared to dstor.

dsolt:
I'm not sure if PMIx_Get needs to be clarified.   I think user's assume something is valid as long as they are not explicitly told otherwise.   The confusion for me comes down to these two lines in the standard (under 6.3):

* job-level information for each namespace involved in the operation is to be made available to all processes in the connected assemblage 

* any data posted by a process in the connected assemblage (via calls to PMIx_Put committed via PMIx_Commit) prior to execution of the PMIx_Connect operation is to be made accessible to all processes in the assemblage - any data posted after execution of the connect operation must be exchanged via a separate PMIx_Fence operation spanning the connected processes

The first one we already discussed.  Because it doesn't say how that should happen, it really just means, "The RM should have the capability to provide job level information to namespaces that are able to connect".   The only real implication of this that I can think of is that a RM should not allow namespaces to connect to one another if there is some reason why the RM could not also satisfy modex requests between those namespaces.    I can't think of a real situation where calling this out is really beneficial for an implementer.  Maybe two processes in different high-level allocations are trying to connect and the RM isn't able to make job-level data available across different allocations, though it is capable of delivering events across different allocation, then the connect should be denied?    

The second bullet seems actually have some weight to it however.   Doesn't the second bullet mean that a connect() call between namespaces has the affect of doing a PMIx_Fence() across those namespaces?   If that is what that is saying, then what does that mean for the RM?   The RM can't magically turn a pmix_server_connect_fn_t call into a pmix_server_fencenb_fn_t call... If calling PMIx_Connect has the effect of calling PMIx_Fence, then the CL would have to enforce that, right?  Currently if clients call PMIx_Connect, it does not result in a pmix_server_fencenb_fn_t call coming up to the RM, so I'm not sure what this second bullet is saying.   Thoughts?

rhc54:
> I'm not sure if PMIx_Get needs to be clarified. I think user's assume something is valid as long as they are not explicitly told otherwise.

True - let me reread it and see if there might be any confusion over cross-namespace access.

> The first one we already discussed. Because it doesn't say how that should happen, it really just means, "The RM should have the capability to provide job level information to namespaces that are able to connect". The only real implication of this that I can think of is that a RM should not allow namespaces to connect to one another if there is some reason why the RM could not also satisfy modex requests between those namespaces. I can't think of a real situation where calling this out is really beneficial for an implementer. Maybe two processes in different high-level allocations are trying to connect and the RM isn't able to make job-level data available across different allocations, though it is capable of delivering events across different allocation, then the connect should be denied?

Fair enough - I believe the reason we originally added that language was due to us having "instant on" in our minds. The thinking was to provide an "instant on" like behavior for processes that do a "connect" across namespaces. Thus, the thought was to have the RM provide all the data that would be available to you when your job started, and to do so for all the procs to which you connected.

Ditto for the second bullet - again, the thought behind it was "instant on". Perhaps the best solution would be to move those bullets to an "advice" section, making clear that these are things one could do to provide an "instant on"-like behavior at the completion of a "connect" operation?

> Currently if clients call PMIx_Connect, it does not result in a pmix_server_fencenb_fn_t call coming up to the RM, so I'm not sure what this second bullet is saying.

Effectively, it was saying that since the connect operation requires a collective across all participants, then you might as well include any currently-available modex data in that collective. This is why ORTE calls "register_nspace" in its implementation of "connect". However, we don't need to require that in the standard - I think if we do as I suggest above and move the bullets into an "advice" as a potential optimization, then we can include the use of either an embedded fence or collecting the data and downloading it via "register_nspace".

rhc54:
@artpol84 I agree about cleanup. Let me poke at the library a little and see how we might do it - as I said, this actually impacts tools as well. Definitely the layer above dstore. We'll need to cleanup all the tracking for that nspace.

dsolt:
I see..  Yes, I agree.  Now that you describe more about what is behind these bullets, I agree they are implementation details that would be nice for performance reasons, but not functional issues.  Moving these to advice to implementers (with maybe some clarification) would make the *requirements* of the standard clearer while still providing the idea that for performance reasons, connect-time is an idea time for the RM to move job-level data around since it is presumably about to do some communication to implement the connect call anyway.   

One thing though... the RM could choose to call PMIx_server_register_nspace to make job-level data available locally (and avoid a dmodex) for performance reasons, but that would not help with data that was put by PMIx_Put.  That data has be requested by CL since RM doesn't know how to access that data.  (I think?)  Any optimization around PMIx_Put data being exchanged at connect-time has to be done in the CL (and isn't being by the CL)?

rhc54:
Sorry for delay - been digging out from blizzard all day. Yes, the RM actually does know how to access the modex data by calling PMIx_server_dmodex_request. This returns all the data provided by that process via PMIx_Put. You can then include that in the "connect" exchange - but we do lack a way of delivering that to the PMIx server. Right now, the only mechanism would be a callback function provided with a dmodex_request, and the CL hasn't requested the data yet.

So if we truly did want to enable that optimization, we'd have to think of a way to deliver that data without waiting for a request. Perhaps an attribute so it can be included in the registration data, maybe for register_client? Dunno - have to ponder that step.

FWIW: turns out this is why MPI_Connect/accept between different mpirun's isn't working 😄  Neither side knows how to make a dmodex request to the other, so I guess I'll need to solve this one way or the other.

dsolt:
Yes, of course.  I forgot we can get dmodex data from CL.  It would just be the sending back when CL is not expecting it that can't be done with current API's.

rhc54:
This was discussed on the telecom. @artpol84 and @karasevb will look to see if this change is in the right place - we would like the dstore to add the new nspace storage when the host returns dmodex info to us for a namespace we don't yet know about. Thus, a call to "store this for me" would check to see if the nspace storage already exists, and if not, then create it. We'll hold this PR until they get back to us on it.

artpol84:
@karasevb, as @rhc54 mentioned in the prior comment, on the call we agreed that this PR needs to be double-checked to see if it has no negative impact.

Need to analyze if there is a better way of doing that. The target is to see if the namespace will only be added at a the direct modex response (not request). We don't want to create a namespace in the dstore until we are 100% sure that this namespace actually exists.
So we need to analyze all the places that call `_esh_session_map_search_server`. And if we do that at modex request we need to introduce a flag to the function telling whether we only want to check the presense of the namespace vs when we want to actually create it.

jjhursey:
bot:ibm:xver:retest

artpol84:
Per telecon (Mar 21)
We agreed with @dsolt that he will investigate the question of a cleanup of a newly created namespace for his use case and will advise additionally.
This is my main concern regarding this PR.

rhc54:
@dsolt Do you require any assistance here? I'm prepping a 3.1.3 release and would welcome this fix there - happy to help complete it if you'd like

dsolt:
Well, @artpol84 made a good point last Thursday that we have nothing to trigger this PMIx_Get data from ever being flushed from the dstore.   I think we really have to answer the question of whether a process should be allowed to PMIx_Get from an arbitrary namespace.  The other option is that the two processes have to be connected before you can ask to PMIx_Get from the other.  

If we decide to require processes to connect to one another before doing PMIx_Get then 
1) the standard really needs to be updated to make this clear for clients
2) The standard needs to be updated to specifically say that PMIx_server_register_nspace has to be called on the node if it hasn't been already for server implementers AND that PMIx_server_deregister_nspace has to be called when the last disconnect happens to that namespace (also needs to account for not de-registering a namespace if there are still processes of that namespace running on the node)
3) The CL needs to be updated to make sure that it denies PMIx_Get calls to unconnected processes (Currently it will work just fine as long as both namespaces are registered on the node and if the target namespace is not registered it sends the request up the the RM which is at the least inefficient)
4) Tools have big problems.  Tools would have no way to probe another namespace to debug it or whatever since the application would have not be able/aware of making the PMIx_Connect call with the tool.  

If we decide that you can call PMIx_Fence/PMIx_Get on an arbitrary namespace, then

1) We need a way to clear namespace/modex data from a server that does not have processes from that namespace running on the local node.   I suppose the CL could record reference counts to modex data.   The first time a processes of a namespace on a node calls PMIx_Get() data from remote namespace, they increment the reference count to that data.  When the requesting namespace is deregistered, it decrements the reference count to all namespace data that it has called PMIx_Get on.   The problem with this approach is that it allows a process in namespace A to successfully PMIx_Get() data from an existing namespace B, but then when namespace B exits, process A can continue to PMIx_Get() data from namespace B using the cached values that have not been purged yet because process A is holding a reference to them.   

OR 

1) When you call PMIx_Get on a namespace that has no processes running on the local host and has not been connected to, it results in a server_dmodex_req call up to the RM, but when that data comes back is not stored locally, it is only used to satisfy the outstanding PMIx_Get call.   This means that if a process in namespace A calls PMIx_Get() 100x on a process running remotely in namespace B, it will result in the data being fetched remotely each time.  This is going to be really rare though and there is an easy performance workaround.... PMIx_Connect() the two processes and now the data can be stored locally and removed when PMIx_Disconnect() is called.   Even in this case, I think there is more work to be done in the CL and the PMIx_Standard to clarify how all of this should work.   I don't think it is just an implementation detail if we go this way.



I'm honestly not sure what is best and at this point I don't care what we do, but I think we need something that makes sense and neither of these options seem great to me.

rhc54:
Hmmm...well, I can offer what I am doing for tools as that is pretty similar to this use-case. If data is obtained for a namespace that has no local procs, then I track who asked for that data and remove it once that proc (or procs) go away. It leaves a little bit of a race condition as the original requestor might terminate just before another one asked for the same info, but that seemed an acceptable penalty as that would likely be an unusual case.

If that sounds reasonable, it would be relatively easy to extend to the non-tool case. I am rather sympathetic to your initial concern that imposing a "connection" requirement may prove overly burdensome to some host environments.

dsolt:
So for your tool case, your approach means you can sometimes PMIx_Get data successfully from a namespace that has actually exited?

dsolt:
BTW, I'm not opposed to having a connection requirement, but I just don't know what it would mean for tools.  I don't think it is a big deal for a host environment performance to require namespaces to be registered at connect time since it is a relatively rare situation.  (I do think the standard needs to be way clearer though about the responsibilities associated with connect/disconnect though if we want to got the route of requiring processes connect/accept).   I'm more concerned about it immediately breaking tools.

rhc54:
> So for your tool case, your approach means you can sometimes PMIx_Get data successfully from a namespace that has actually exited?

I believe this is a natural race condition in the system, yes? This is true for all namespaces, not just those unknown to the local daemon, because the "deregister_nspace" call has to come from the host RM, thus leaving a time lag between when the job in the nspace terminated and the server is told to remove the nspace data. Therefore, just because you successfully retrieved info about a namespace - or a process, for that matter - does not mean it is still alive.


> BTW, I'm not opposed to having a connection requirement, but I just don't know what it would mean for tools. I don't think it is a big deal for a host environment performance to require namespaces to be registered at connect time since it is a relatively rare situation. (I do think the standard needs to be way clearer though about the responsibilities associated with connect/disconnect though if we want to got the route of requiring processes connect/accept). I'm more concerned about it immediately breaking tools.

I agree about tools as that creates some significant issues. We would probably have to come up with some other way of defining "connect" - as things stand, it imposes behaviors on event notification, termination, etc. In MPI terms, you might effectively be tying the tool into the application's communicator (at least, so far as the host environment is concerned) - which is not what we would want to do. For example, if the tool unexpectedly "died" this approach would automatically generate a "peer died" event to the application's participants.

Only way around that (I think) would be to come up with some other server-to-host interface (or define an appropriate attribute) to make clear that these participants are not to be considered as a collection by the host. This imposes more logic on the host - I suspect it will be easier if we can keep it internal to the library.

I think the "delete if no requestor remains" should provide a clean solution if that is acceptable to folks. Might be worth prototyping to see what you think?

dsolt:
I agree... there is a always a window where you can call PMIx_Get() successfully even though the remote process has gone away, but it is a limited window.   With the approach of "delete when no requestor remains", the window is indefinite.... the namespace might have exited 2 days ago and you can still PMIx_Get() on it because your process did a PMIx_Get() on it 3 days ago.   I'm not saying that is unacceptable,  I just want people to be aware of it and also think about if/how to document this.  I don't think the standard really addresses what happens right now when you call PMIx_Get on a proc that has exited or on a proc that has exited, but other procs in the namespace still exist, or a on a proc who's entire namespace is gone.   Maybe we don't want to specify?    In general, I think I'm in favor of either "delete if no requestor remains" or "don't cache if we don't know the namespace, just fulfill outstanding PMIx_Get calls and then discard the data".   I'm not sure which approach I prefer, but I'm ok with "delete if no requestor remains" if you have already been doing that in some contexts.

rhc54:
Good point. Based on your comments, I'd vote for "don't cache" as the simplest solution. You want me to prototype it?

dsolt:
Well, how long would it take you?  Its another one of those cases where I would love to dig in and try to give it a shot, but I wouldn't have time to really try it out for at least 2 weeks.   There is no rush on my end for this change, as we can use the approach in this PR as a temporary fix (leaking some memory if an app does this).

rhc54:
Feel free - I have no rush and am happy to let you give it a shot 😄

dsolt:
Ok, Let me play with it.  I think that approach is better than this approach as @artpol84 points out.

dsolt:
I finally got back into digging into this this week.  I tried my "second" approach I gave above:  

> When you call PMIx_Get on a namespace that has no processes running on the local host and has not been connected to, it results in a server_dmodex_req call up to the RM, but when that data comes back is not stored locally, it is only used to satisfy the outstanding PMIx_Get call.

I used PMIX_GDS_ADD_NSPACE and PMIX_GDS_DEL_NSPACE to temporarily add the namespace if it did not exist and then delete it if it was just added for this temporary use.  

This doesn't work because for the ds data store modules, the data is not sent back to the requester, a reference to the data is sent back to the requester and by the time it gets there, the data has already been deleted from the GDS.   I'm not sure when I can safely delete the namespace from the GDS.  Now there is an flag to pmix_common_dstor_fetch to indicate that it should return a copy of the data, but it doesn't do anything, so if I pass in copy=true at the high level call it doesn't actually return a copy.  

I was able to force the server to store this special namespace data in the hash DS instead of the ds21 data store, but then when the client is given the data, it still tries to look it up in ds21.


 I can try to use the first method I proposed:

> I suppose the CL could record reference counts to modex data. The first time a processes of a namespace on a node calls PMIx_Get() data from remote namespace, they increment the reference count to that data. When the requesting namespace is deregistered, it decrements the reference count to all namespace data that it has called PMIx_Get on. 

but this seems quite involved and has the side effects I mentioned above.  Any thoughts?

rhc54:
> I was able to force the server to store this special namespace data in the hash DS instead of the ds21 data store, but then when the client is given the data, it still tries to look it up in ds21.

I encountered the same problem in #1193 and fixed it over there. Let me try to complete that PR and then let's see if we can utilize your proposed approach.

> there is an flag to pmix_common_dstor_fetch to indicate that it should return a copy of the data, but it doesn't do anything, so if I pass in copy=true at the high level call it doesn't actually return a copy.

This is a bug - @karasevb can you please address it?

ibm-ompi:

 Gist: https://gist.github.com/4b91f68ca20c0bb9890153bd148f6adb

jjhursey:
Maybe related #1193

artpol84:
> This is a bug - @karasevb can you please address it?

I’ve added it to the @karasevb queue so it will be resolved

dsolt:
I did some testing on this and I am still hitting issues, though they seem to occur with my JSM PMIx-based launcher using pmix_master and with Open MPI mpirun/orte launcher using PMIx 3.1.3.  I have attached a test case to reproduce an issue we saw with our PAMI client.  At one time, I believe this code worked when parent and children processes were located on the same node, but now it seems to fail in every layout I tried.  Is my test case valid? 

[pmix_hostname_get.txt](https://github.com/pmix/pmix/files/3503240/pmix_hostname_get.txt)

dsolt:
It does look like this code works for me when my parent and child are on the same node as it did in the past, but on separate nodes, it still hangs in the PMIx_Get call for the PMIX_HOSTNAME of a process on another namespace.

rhc54:
FWIW: I'm able to replicate this on PRRTE due to a race condition (I think). I can trace that the dmodex call for the remote hostname is being passed up to the PRRTE level by the PMIx server library, so that part of the code path seems correct. However, it appears that PRRTE isn't handling it correctly. I'll try to track that down and see what happens on the remote end.

dsolt:
another data point... if I cherry-pick this PR to master, my JSM still does not pass the test I posted when parent and child are launched to different nodes.  So I'm not sure the code change with this PR is useful.

rhc54:
Yeah, we should move this discussion to an "issue" until we identify a potential fix.

PR #1126, opened 2019-02-21
bfrops: redefinition regtypes for internal usage

### LINKED ISSUE #1099: Generalize `bfrops` components
According to Issue #681, I'm working on the implementation of new flexible packing with a new component `bfrops`.
My question is to avoid duplication of code and simplify future bfrops multi-component support.

**Details:**
I forked bfrops v3 to v4 and modified it for use new flexible packing by changing pack/unpack registration functions. We want to use new pack/unpack functions for all integer types. For integer-based PMIx types (like `PMIX_INT16`, `PMIX_INT32`, `PMIX_SIZE` etc) is not a problem to replace registration functions to new:
```C
    PMIX_REGISTER_TYPE("PMIX_SIZE", PMIX_SIZE,
                       pmix4_bfrops_base_pack_int_flex, <- replaced
                       pmix4_bfrops_base_unpack_int_flex, <-replaced
                       ...
                       &mca_bfrops_v4_component.types);
```
The complex PMIx types, such as `PMIX_DATA_ARRAY`, `PMIX_KVAL`, etc., directly call into its implementation a simple-integer packing functions. 
Let's consider a packing for `PMIX_DATA_ARRAY`:
```C
    PMIX_REGISTER_TYPE("PMIX_DATA_ARRAY", PMIX_DATA_ARRAY,
                       pmix_bfrops_base_pack_darray,
                       pmix_bfrops_base_unpack_darray,
                       ...
```
The function `pmix_bfrops_base_pack_darray` uses `pmix_bfrops_base_pack_int64` direclty:
https://github.com/pmix/pmix/blob/0f7075f7d204b3cf46ef2aa43120632696cd858a/src/mca/bfrops/base/bfrop_base_pack.c#L883

**The issue is in following:** to implement the new bfrops component, that has foreign simple-packing integer methods it requires a full duplicate of implementation of complex-packing PMIx types. For example: for `pmix_bfrops_base_pack_darray` will require to call new `pmix_bfrops_base_pack_int64`, that require a duplicate of `pmix_bfrops_base_pack_darray` with new int64 packing. 

The multiple duplications is not a right way. So I would like to propose another way, that requires to modify a prototype of pack/unpack methods by adding `pmix_pointer_array_t *regtypes` to each one.
Example of `pmix_bfrops_base_pack_darray` modification:
```patch
 pmix_status_t pmix_bfrops_base_pack_darray(pmix_buffer_t *buffer, const void *src,
-                                           int32_t num_vals, pmix_data_type_t type)
+                                           int32_t num_vals, pmix_data_type_t type,
+                                           pmix_pointer_array_t *regtypes)
 {
     pmix_data_array_t *p = (pmix_data_array_t*)src;
     pmix_status_t ret;
     int32_t i;
+    pmix_bfrop_type_info_t *info;
 
     for (i=0; i < num_vals; i++) {
+        /* Lookup the pack function for this type and call it */
+        if (NULL == (info = (pmix_bfrop_type_info_t*)pmix_pointer_array_get_item(regtypes, type))) {
+            PMIX_ERROR_LOG(PMIX_ERR_UNKNOWN_DATA_TYPE);
+            return PMIX_ERR_UNKNOWN_DATA_TYPE;
+        }
         /* pack the actual type in the array */
         if (PMIX_SUCCESS != (ret = pmix_bfrop_store_data_type(buffer, p[i].type))) {
             return ret;
@@ -805,190 +812,10 @@ pmix_status_t pmix_bfrops_base_pack_darray(pmix_buffer_t *buffer, const void *sr
             /* nothing left to do */
             continue;
         }
-        /* pack the actual elements - have to do this the hard way */
-        switch(p[i].type) {
-            case PMIX_UNDEF:
-                break;
-            case PMIX_BOOL:
-                if (PMIX_SUCCESS != (ret = pmix_bfrops_base_pack_bool(buffer, p[i].array, p[i].size, PMIX_BOOL))) {
-                    return ret;
-                }
-                break;
-            case PMIX_BYTE:
-            case PMIX_STRING:
-            case PMIX_STRING:
-                if (PMIX_SUCCESS != (ret = pmix_bfrops_base_pack_string(buffer, p[i].array, p[i].size, PMIX_STRING))) {
-                    return ret;
-                }
-                break;
-            case PMIX_SIZE:
-            case PMIX_PID:
-            case PMIX_INT:
...
-            case PMIX_ENVAR:
-                if (PMIX_SUCCESS != (ret = pmix_bfrops_base_pack_envar(buffer, p[i].array, p[i].size, PMIX_ENVAR))) {
-                    return ret;
-                }
-                break;
-
-            default:
-                pmix_output(0, "PACK-PMIX-VALUE[%s:%d]: UNSUPPORTED TYPE %d",
-                            __FILE__, __LINE__, (int)p[i].type);
-            return PMIX_ERROR;
+        ret = info->odti_pack_fn(buffer, p[i].array, p[i].size, p[i].type);
+        if (PMIX_SUCCESS != ret) {
+            PMIX_ERROR_LOG(ret);
+            return ret;
         }
     }
     return PMIX_SUCCESS;

```
` *regtypes` must be passed from the upper abstract level of pack/unpack functions, also should be updated pack/unpack methods for all existed bfrops. 

This way allows keeping a common `pmix_bfrops_base_pack_darray` in base-code, and `odti_pack_fn` will call a specific registered packing method, that can be different between the version of bfrops. This allows reducing of code duplication, and simplify the support. The likewise modification is required other comlex-packing types.

karasevb:
This commit changes the structure of bfrops component by adding an ability to use registered pack/unpack callbacks for internal purposes.

Closes #1099

### DIFF
--- a/src/mca/bfrops/base/base.h
+++ b/src/mca/bfrops/base/base.h
@@ -226,7 +226,7 @@ PMIX_EXPORT extern pmix_bfrops_globals_t pmix_bfrops_globals;
     do {                                                                    \
         int32_t i;                                                          \
         tmptype *tmpbuf = (tmptype*)malloc(sizeof(tmptype) * (*num_vals));  \
-        ret = unpack_gentype(buffer, tmpbuf, num_vals, tmpbfroptype);       \
+        PMIX_BFROPS_UNPACK_TYPE(ret, buffer, tmpbuf, num_vals, tmpbfroptype, reg_types);        \
         for (i = 0 ; i < *num_vals ; ++i) {                                 \
             ((unpack_type*) dest)[i] = (unpack_type)(tmpbuf[i]);            \
         }                                                                   \
--- a/src/mca/bfrops/base/base.h
+++ b/src/mca/bfrops/base/base.h
@@ -14,6 +14,8 @@
  * Copyright (c) 2014-2019 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2019      Mellanox Technologies, Inc.
+ *                         All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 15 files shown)


karasevb:
PR is on testing.

src/mca/bfrops/base/base.h
```diff
@@ -222,9 +260,9 @@ typedef struct pmix_info_array {
     /** Debugging string name */
     char *odti_name;
     /** Pack function */
-    pmix_bfrop_pack_fn_t odti_pack_fn;
+    pmix_bfrop_odti_pack_fn_t odti_pack_fn;
     /** Unpack function */
-    pmix_bfrop_unpack_fn_t odti_unpack_fn;
+    pmix_bfrop_odti_unpack_fn_t odti_unpack_fn;
```
karasevb: Check usage the PMIX_BFROPS_REGISTER macro

src/mca/bfrops/base/bfrop_base_pack.c
```diff
@@ -68,51 +71,39 @@ pmix_status_t pmix_bfrops_base_pack_buffer(pmix_pointer_array_t *regtypes,
...
-        PMIX_ERROR_LOG(PMIX_ERR_UNKNOWN_DATA_TYPE);
-        return PMIX_ERR_UNKNOWN_DATA_TYPE;
-    }
-
-    return info->odti_pack_fn(buffer, src, num_vals, type);
+    return rc;
 }
 
-static pmix_status_t pack_gentype(pmix_buffer_t *buffer, const void *src,
+static pmix_status_t pack_gentype(pmix_pointer_array_t *regtypes,
```
karasevb: Remove this function, use the packing macro instead

karasevb:
bot:retest

karasevb:
This PR is a part of #1130, so no make sense to keep it separatly.

PR #1125, opened 2019-02-21
modex: optimization of rank identification size reduction

karasevb:
Packing the nspace string with each proc is redundant and not need for unpacker side. This commit avoids packing `nspace:rank` tuples into the modex blob, instead of that all ranks are renumbered throughtout procs involved in Modex and packed as integers.

Example procs paking:
```
{ "nspace1", 0} : 0
{ "nspace1", 1} : 1 => procs = { {0}, {1}, {2} }
{ "nspace2", 0} : 2
```

Refs https://github.com/pmix/pmix/issues/681

PR #1130, opened 2019-02-25
bfrops: added flexible-size integer packing to bfrops v4 

### LINKED ISSUE #681: Modex size analysis
## Short description
This Issue is to track the analysis of the modex size and searching for the ways to optimize it.
For the reference: https://github.com/open-mpi/ompi/pull/4826

## First measurements
The first observation is to clarify on @rhc54 [comment](https://github.com/open-mpi/ompi/pull/4826#issuecomment-366464181) about the effect of extra information that is being packed when PMIx is configured with `--enable-debug`.

I went back and double-checked our build and indeed `--enable-debug` was there. So here are some details. Details on how to reproduce below.

`--disable-debug`:
```
[jazz06:34916] recvd FENCE
[jazz06:34916] recvd fence from 1110704129:1 with 1 procs
[jazz06:34916] fence complete
[jazz06:34916] fence - assembling data: 5
[jazz06:34916] rank_info: rank=1, size = 23
[jazz06:34916] rank_data: rank=1, size = 143
[jazz06:34916] commit_message: rank=1 added, size = 162 # (162 - 143 - 5 = 14)
[jazz06:34916] commit_message: final size = 176          # (176 - 62 = 14)
```
`--enable-debug`
```
[jazz09:167487] recvd FENCE
[jazz09:167487] recvd fence from 1440743425:0 with 1 procs
[jazz09:167487] fence complete
[jazz09:167487] fence - assembling data: 9                 # (9 - 5 = 4)
[jazz09:167487] rank_info: rank=0, size = 27               # (27 - 23 = 4)
[jazz09:167487] rank_data: rank=0, size = 156              # (156 - 143 = 13)
[jazz09:167487] commit_message: rank=0 added, size = 183   # (183 - 9 - 156 = 18; 18 - 14 = 4)
[jazz09:167487] commit_message: final size = 201           # (201 - 176 = 25, 25 - 4 - 4 - 13 - 4 = 0 )
```
## Analysis
* Indeed debugging flag adds ~25B on top of the modex size (12%)
* Release version size is 176B which is still high
* In the rank contribution (release) 23B are taken by the rank identification (`rank_info`: 23B/176B = 13%). And this will be the case for each additional rank. Reasonable optimization here would be (as @rhc54 [suggested](https://github.com/open-mpi/ompi/pull/4826#issuecomment-366466299)) is to "cite" namespace only once and then only provide rank number. In case of multiple namespace we will need to group ranks by namespace.
The issue here is that once this is at the RM side there is no way to reduce it further (as namespaces will be duplicated in messages from different nodes. On the other hand this will only be an issue for the tree-based modex implementations that are not very efficient for the modex exchange.
* Knowing that UCX endpoint size was ~50B in this particular run we observe that `rank-data` field is also too large and equal to 120B (143B - 23B; (120B-50B) / 120B = 58%). I'm going to investigate that further.
* Finally some sort of "footer" is being applied to the buffer twice:
  * once in `commit_message: rank=1 added`
  * once when packing the final buffer: `commit_message: final size`
As the first "footer" is per rank it contributes 14/176 = 8% to the modex size.

## Reproduction 
The following patch was applied to the built-in PMIx in OMPI https://github.com/open-mpi/ompi/commit/b6e825f83ebf34003dda922cd3f9237d10571312:
```patch
diff --git a/opal/mca/pmix/pmix2x/pmix/src/server/pmix_server_ops.c b/opal/mca/pmix/pmix2x/pmix/src/server/pmix_server_ops.c                                                                                                                  
index 29fb193..c773db4 100644                                                                                                                                                                                                                 
--- a/opal/mca/pmix/pmix2x/pmix/src/server/pmix_server_ops.c                                                                                                                                                                                  
+++ b/opal/mca/pmix/pmix2x/pmix/src/server/pmix_server_ops.c                                                                                                                                                                                  
@@ -614,7 +614,7 @@ pmix_status_t pmix_server_fence(pmix_server_caddy_t *cd,                                                                                                                                                                  
.                                                                                                                                                                                                                                             
         if (PMIX_COLLECT_YES == trk->collect_type) {                                                                                                                                                                                         
             pmix_output_verbose(2, pmix_globals.debug_output,                                                                                                                                                                                
-                                "fence - assembling data");                                                                                                                                                                                  
+                                "fence - assembling data: %u", (unsigned int) bucket.bytes_used);                                                                                                                                            
                 /* get any remote contribution - note that there                                                                                                                                                                             
                  * may not be a contribution */                                                                                                                                                                                              
@@ -635,6 +635,9 @@ pmix_status_t pmix_server_fence(pmix_server_caddy_t *cd,                                                                                                                                                                  
                         PMIX_DESTRUCT(&cb);                                                                                                                                                                                                  
                         goto cleanup;                                                                                                                                                                                                        
                     }                                                                                                                                                                                                                        
+                    pmix_output_verbose(2, pmix_globals.debug_output,                                                                                                                                                                        
+                                "rank_info: rank=%d, size = %u", pcs.rank , (unsigned int) pbkt.bytes_used);                                                                                                                                 
+                                                                                                                                                                                                                                             
                     /* pack the returned kval's */                                                                                                                                                                                           
                     PMIX_LIST_FOREACH(kv, &cb.kvs, pmix_kval_t) {                                                                                                                                                                            
                         PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, &pbkt, kv, 1, PMIX_KVAL);                                                                                                                                                  
@@ -644,12 +647,19 @@ pmix_status_t pmix_server_fence(pmix_server_caddy_t *cd,                                                                                                                                                                
                             goto cleanup;                                                                                                                                                                                                    
                         }                                                                                                                                                                                                                    
                     }                                                                                                                                                                                                                        
+                                                                                                                                                                                                                                             
+                    pmix_output_verbose(2, pmix_globals.debug_output,                                                                                                                                                                        
+                                "rank_data: rank=%d, size = %u", pcs.rank , (unsigned int) pbkt.bytes_used);                                                                                                                                 
+                                                                                                                                                                                                                                             
                     /* extract the blob */                                                                                                                                                                                                   
                     PMIX_UNLOAD_BUFFER(&pbkt, bo.bytes, bo.size);                                                                                                                                                                            
                     PMIX_DESTRUCT(&pbkt);                                                                                                                                                                                                    
                     /* pack the returned blob */                                                                                                                                                                                             
                                      &bo, 1, PMIX_BYTE_OBJECT);                                                                                                                                                                              
+                    pmix_output_verbose(2, pmix_globals.debug_output,                                                                                                                                                                        
+                                "commit_message: rank=%d added, size = %u", pcs.rank , (unsigned int) bucket.bytes_used);                                                                                                                    
+                                                                                                                                                                                                                                             
                     PMIX_BYTE_OBJECT_DESTRUCT(&bo);                                                                                                                                                                                          
                     if (PMIX_SUCCESS != rc) {                                                                                                                                                                                                
                         PMIX_ERROR_LOG(rc);                                                                                                                                                                                                  
@@ -677,6 +687,10 @@ pmix_status_t pmix_server_fence(pmix_server_caddy_t *cd,                                                                                                                                                                 
         /* now unload the blob and pass it upstairs */                                                                                                                                                                                       
         PMIX_UNLOAD_BUFFER(&bucket, data, sz);                                                                                                                                                                                               
         PMIX_DESTRUCT(&bucket);                                                                                                                                                                                                              
+                                                                                                                                                                                                                                             
+                                "commit_message: final size = %u", sz);                                                                                                                                                                      
+                                                                                                                                                                                                                                             
         pmix_host_server.fence_nb(trk->pcs, trk->npcs,                                                                                                                                                                                       
                                   trk->info, trk->ninfo,                                                                                                                                                                                     
```

Command line (2-node allocation):
```shell
$ mpirun -n 2 --debug-daemons --map-by node -mca btl self -mca pmix_base_verbose 2 -mca coll ^hcoll \
    -mca pml ucx --report-bindings -x UCX_TLS=dc -x UCX_NET_DEVICES=mlx5_0:1 \
    ./ring_c 2>&1 | tee -a tmp.data
```

karasevb:
Closes #681

### DIFF
--- a/src/mca/bfrops/base/base.h
+++ b/src/mca/bfrops/base/base.h
@@ -226,7 +226,7 @@ PMIX_EXPORT extern pmix_bfrops_globals_t pmix_bfrops_globals;
     do {                                                                    \
         int32_t i;                                                          \
         tmptype *tmpbuf = (tmptype*)malloc(sizeof(tmptype) * (*num_vals));  \
-        ret = unpack_gentype(reg_types, buffer, tmpbuf, num_vals, tmpbfroptype);       \
+        PMIX_BFROPS_UNPACK_TYPE(ret, buffer, tmpbuf, num_vals, tmpbfroptype, reg_types);        \
         for (i = 0 ; i < *num_vals ; ++i) {                                 \
             ((unpack_type*) dest)[i] = (unpack_type)(tmpbuf[i]);            \
         }                                                                   \
--- a/src/mca/bfrops/base/base.h
+++ b/src/mca/bfrops/base/base.h
@@ -199,6 +199,31 @@ do {                                                        \
 } while (0)
 
 /**
+ * Unpacking conversion from a flexible representation to a
+ * signed integer value.
+ *
+ * type - C-type of a signed integer value
+ *
+ * val - flexible representation (uint64_t)
+ *
+ * ptr - pointer to a 64-bit output buffer for the upacked value
+ *
+ * (see a comment to `pmix_bfrops_pack_flex` for additional details)
+ */
+#define PMIX_BFROPS_UNPACK_FLEX_CONVERT_SIGNED(type, val, ptr)  \
+do {                                                            \
+    type __tbuf = 0;                                            \
+    uint64_t __tmp = val;                                       \
+    int sign = (__tmp) & 1;                                     \
+    __tmp >>= 1;                                                \
+    if (sign) {                                                 \
+        __tmp = ~__tmp;                                         \
+    }                                                           \
+    __tbuf = (type)__tmp;                                       \
+    memcpy(ptr, &__tbuf, sizeof(type));                         \
+} while (0)
+
+/**
  * Packing conversion of a signed integer value to a flexible representation.
  * For unsigned types it is reduced to a memcopy.
  *
@@ -216,6 +241,25 @@ do {                                                        \
 do {                                                            \
     out = 0;                                                    \
     memcpy(&out, (ptr), sizeof(type));                          \
+} while (0)
+
+/**
+ * Unpacking conversion of a flexible representation value
+ * to an unsigned integer.
+ *
+ * type - C-type of unsigned integer value
+ *
+ * val - flexible representation value (uint64_t)
+ *
+ * ptr - pointer to a 64-bit output buffer for the upacked value
+ *
+ * (see a comment to `pmix_bfrops_pack_flex` for additional details)
+ */
+#define PMIX_BFROPS_UNPACK_FLEX_CONVERT_UNSIGNED(type, val, ptr)\
+do {                                                            \
+    type __tbuf = 0;                                            \
+    __tbuf = (type)val;                                         \
+    memcpy(ptr, &__tbuf, sizeof(type));                         \
 } while (0)
 
 /**
@@ -300,51 +344,6 @@ do {                                                                \
             (r) = PMIX_ERR_BAD_PARAM;                               \
     }                                                               \
 } while(0)
-
-/**
- * Unpacking conversion from a flexible representation to a
- * signed integer value.
- *
- * type - C-type of a signed integer value
- *
- * val - flexible representation (uint64_t)
- *
- * ptr - pointer to a 64-bit output buffer for the upacked value
- *
- * (see a comment to `pmix_bfrops_pack_flex` for additional details)
- */
-#define PMIX_BFROPS_UNPACK_FLEX_CONVERT_SIGNED(type, val, ptr)  \
-do {                                                            \
-    type __tbuf = 0;                                            \
-    uint64_t __tmp = val;                                       \
-    int sign = (__tmp) & 1;                                     \
-    __tmp >>= 1;                                                \
-    if (sign) {                                                 \
-        __tmp = ~__tmp;                                         \
-    }                                                           \
-    __tbuf = (type)__tmp;                                       \
-    memcpy(ptr, &__tbuf, sizeof(type));                         \
-} while (0)
-
-/**
- * Unpacking conversion of a flexible representation value
- * to an unsigned integer.
- *
- * type - C-type of unsigned integer value
- *
- * val - flexible representation value (uint64_t)
- *
- * ptr - pointer to a 64-bit output buffer for the upacked value
- *
- * (see a comment to `pmix_bfrops_pack_flex` for additional details)
- */
-#define PMIX_BFROPS_UNPACK_FLEX_CONVERT_UNSIGNED(type, val, ptr)\
-do {                                                            \
-    type __tbuf = 0;                                            \
-    __tbuf = (type)val;                                         \
-    memcpy(ptr, &__tbuf, sizeof(type));                         \
-} while (0)
-
 
 /**
  * Unpacking conversion of a flexible representation value
--- a/src/mca/bfrops/base/base.h
+++ b/src/mca/bfrops/base/base.h
@@ -160,6 +160,11 @@ PMIX_EXPORT extern pmix_bfrops_globals_t pmix_bfrops_globals;
 #error Unsupported pid_t size!
 #endif
 
+#define PMIX_BFROPS_FLEX_BASE7_MAX_BUF_SIZE (SIZEOF_SIZE_T+1)
+#define PMIX_BFROPS_FLEX_BASE7_MASK ((1<<7) - 1)
+#define PMIX_BFROPS_FLEX_BASE7_SHIFT 7
+#define PMIX_BFROPS_FLEX_BASE7_CONT_FLAG (1<<7)
+
 /* Unpack generic size macros */
 #define PMIX_BFROP_UNPACK_SIZE_MISMATCH(reg_types, unpack_type, remote_type, ret)                     \
     do {                                                                        \
--- a/src/mca/bfrops/base/base.h
+++ b/src/mca/bfrops/base/base.h
@@ -169,7 +169,7 @@ PMIX_EXPORT extern pmix_bfrops_globals_t pmix_bfrops_globals;
 /**
  * Packing conversion of a signed integer value to a flexible representation.
  * The main idea is to split a signed negative value onto an absolute value
- * and a sing bit stored in the special location.
+ * and a sign bit stored in the special location.
  * This allows efficient representetion of negative values in the
  * flexible form.
  *
--- a/src/mca/bfrops/base/base.h
+++ b/src/mca/bfrops/base/base.h
@@ -160,10 +160,238 @@ PMIX_EXPORT extern pmix_bfrops_globals_t pmix_bfrops_globals;
 #error Unsupported pid_t size!
 #endif
 
+/* Flexible packing */
 #define PMIX_BFROPS_FLEX_BASE7_MAX_BUF_SIZE (SIZEOF_SIZE_T+1)
 #define PMIX_BFROPS_FLEX_BASE7_MASK ((1<<7) - 1)
 #define PMIX_BFROPS_FLEX_BASE7_SHIFT 7
 #define PMIX_BFROPS_FLEX_BASE7_CONT_FLAG (1<<7)
+
+/**
+ * Packing conversion of a signed integer value to a flexible representation.
+ * The main idea is to split a signed negative value onto an absolute value
+ * and a sing bit stored in the special location.
+ * This allows efficient representetion of negative values in the
+ * flexible form.
+ *
+ * type - type (pmix_data_type_t) of integer value
+ *
+ * ptr - pointer to the signed integer value
+ *       with the type defined as (type)
+ *
+ * out - flexible representation of *ptr,
+ *       extended to uint64_t if needed
+ *
+ * (see a comment to `pmix_bfrops_pack_flex` for additional details)
+ */
+#define PMIX_BFROPS_PACK_FLEX_CONVERT_SIGNED(type, ptr, out)\
+do {                                                        \
+    type __tbuf = 0;                                        \
+    uint64_t __tmp;                                         \
+    int __sign = 0;                                         \
+    memcpy(&__tbuf, (ptr), sizeof(type));                   \
+    __tmp = __tbuf;                                         \
+    (out) = (uint64_t)__tmp;                                \
+    if (__tmp & (1UL << (sizeof(__tmp)*8 - 1))) {           \
+        __sign = 1;                                         \
+        out = ~(out);                                       \
+    }                                                       \
+    (out) = ((out) << 1) + __sign;                          \
+} while (0)
+
+/**
+ * Packing conversion of a signed integer value to a flexible representation.
+ * For unsigned types it is reduced to a memcopy.
+ *
+ * type - usual integer C-type of integer value
+ *
+ * ptr - pointer to the signed integer value
+ *       with the type defined as (type)
+ *
+ * out - flexible representation of *ptr,
+ *       extended to uint64_t if needed
+ *
+ * (see a comment to `pmix_bfrops_pack_flex` for additional details)
+ */
+#define PMIX_BFROPS_PACK_FLEX_CONVERT_UNSIGNED(type, ptr, out)  \
+do {                                                            \
+    out = 0;                                                    \
+    memcpy(&out, (ptr), sizeof(type));                          \
+} while (0)
+
+/**
+ * Sizeof by PMIx type integer values.
+ *
+ * r - return status code
+ *
+ * t - type (pmix_data_type_t) of integer value
+ *
+ * s - size of type in bytes
+ *
+ * (see a comment to `pmix_bfrops_pack_flex` for additional details)
+ */
+#define PMIX_BFROPS_TYPE_SIZEOF(r, t, s)                    \
+do {                                                        \
+    (r) = PMIX_SUCCESS;                                     \
+    switch (t) {                                            \
+        case PMIX_INT16:                                    \
+        case PMIX_UINT16:                                   \
+            (s) = SIZEOF_SHORT;                             \
+            break;                                          \
+        case PMIX_INT:                                      \
+        case PMIX_INT32:                                    \
+        case PMIX_UINT:                                     \
+        case PMIX_UINT32:                                   \
+            (s) = SIZEOF_INT;                               \
+            break;                                          \
+        case PMIX_INT64:                                    \
+        case PMIX_UINT64:                                   \
+            (s) = SIZEOF_LONG;                              \
+            break;                                          \
+        case PMIX_SIZE:                                     \
+            (s) = SIZEOF_SIZE_T;                            \
+            break;                                          \
+        default:                                            \
+            (r) = PMIX_ERR_BAD_PARAM;                       \
+    }                                                       \
+} while (0)
+
+/**
+ * Packing conversion from integer value to a flexible representation.
+ *
+ * r - return status code
+ *
+ * t - type (pmix_data_type_t) of integer value, it is determines
+ *     which type of integer is converted
+ *
+ * s - pointer to the integer value with the type defined as (t)
+ *
+ * d - flexible representation output value (uin64_t)
+ *
+ * (see a comment to `pmix_bfrops_pack_flex` for additional details)
+ */
+#define PMIX_BFROPS_PACK_FLEX_CONVERT(r, t, s, d)                   \
+do {                                                                \
+    (r) = PMIX_SUCCESS;                                             \
+    switch (t) {                                                    \
+        case PMIX_INT16:                                            \
+            PMIX_BFROPS_PACK_FLEX_CONVERT_SIGNED(int16_t, s, d);    \
+            break;                                                  \
+        case PMIX_UINT16:                                           \
+            PMIX_BFROPS_PACK_FLEX_CONVERT_UNSIGNED(uint16_t, s, d); \
+            break;                                                  \
+        case PMIX_INT:                                              \
+        case PMIX_INT32:                                            \
+            PMIX_BFROPS_PACK_FLEX_CONVERT_SIGNED(int32_t, s, d);    \
+            break;                                                  \
+        case PMIX_UINT:                                             \
+        case PMIX_UINT32:                                           \
+            PMIX_BFROPS_PACK_FLEX_CONVERT_UNSIGNED(uint32_t, s, d); \
+            break;                                                  \
+        case PMIX_INT64:                                            \
+            PMIX_BFROPS_PACK_FLEX_CONVERT_SIGNED(int64_t, s, d);    \
+            break;                                                  \
+        case PMIX_SIZE:                                             \
+            PMIX_BFROPS_PACK_FLEX_CONVERT_UNSIGNED(size_t, s, d);   \
+            break;                                                  \
+        case PMIX_UINT64:                                           \
+            PMIX_BFROPS_PACK_FLEX_CONVERT_UNSIGNED(uint64_t, s, d); \
+            break;                                                  \
+        default:                                                    \
+            (r) = PMIX_ERR_BAD_PARAM;                               \
+    }                                                               \
+} while(0)
+
+/**
+ * Unpacking conversion from a flexible representation to a
+ * signed integer value.
+ *
+ * type - C-type of a signed integer value
+ *
+ * val - flexible representation (uint64_t)
+ *
+ * ptr - pointer to a 64-bit output buffer for the upacked value
+ *
+ * (see a comment to `pmix_bfrops_pack_flex` for additional details)
+ */
+#define PMIX_BFROPS_UNPACK_FLEX_CONVERT_SIGNED(type, val, ptr)  \
+do {                                                            \
+    type __tbuf = 0;                                            \
+    uint64_t __tmp = val;                                       \
+    int sign = (__tmp) & 1;                                     \
+    __tmp >>= 1;                                                \
+    if (sign) {                                                 \
+        __tmp = ~__tmp;                                         \
+    }                                                           \
+    __tbuf = (type)__tmp;                                       \
+    memcpy(ptr, &__tbuf, sizeof(type));                         \
+} while (0)
+
+/**
+ * Unpacking conversion of a flexible representation value
+ * to an unsigned integer.
+ *
+ * type - C-type of unsigned integer value
+ *
+ * val - flexible representation value (uint64_t)
+ *
+ * ptr - pointer to a 64-bit output buffer for the upacked value
+ *
+ * (see a comment to `pmix_bfrops_pack_flex` for additional details)
+ */
+#define PMIX_BFROPS_UNPACK_FLEX_CONVERT_UNSIGNED(type, val, ptr)\
+do {                                                            \
+    type __tbuf = 0;                                            \
+    __tbuf = (type)val;                                         \
+    memcpy(ptr, &__tbuf, sizeof(type));                         \
+} while (0)
+
+
+/**
+ * Unpacking conversion of a flexible representation value
+ * to an integer.
+ *
+ * r - return status code
+ *
+ * t - type (pmix_data_type_t) of integer value, it is determines
+ *     which type of integer is converted
+ *
+ * s - flex-representation value (uin64_t)
+ *
+ * d - pointer to a 64-bit output buffer for the upacked value
+ *
+ * (see a comment to `pmix_bfrops_pack_flex` for additional details)
+ */
+#define PMIX_BFROPS_UNPACK_FLEX_CONVERT(r, t, s, d)                     \
+do {                                                                    \
+    (r) = PMIX_SUCCESS;                                                 \
+    switch (t) {                                                        \
+        case PMIX_INT16:                                                \
+            PMIX_BFROPS_UNPACK_FLEX_CONVERT_SIGNED(int16_t, s, d);      \
+            break;                                                      \
+        case PMIX_UINT16:                                               \
+            PMIX_BFROPS_UNPACK_FLEX_CONVERT_UNSIGNED(uint16_t, s, d);   \
+            break;                                                      \
+        case PMIX_INT:                                                  \
+        case PMIX_INT32:                                                \
+            PMIX_BFROPS_UNPACK_FLEX_CONVERT_SIGNED(int32_t, s, d);      \
+            break;                                                      \
+        case PMIX_UINT:                                                 \
+        case PMIX_UINT32:                                               \
+            PMIX_BFROPS_UNPACK_FLEX_CONVERT_UNSIGNED(uint32_t, s, d);   \
+            break;                                                      \
+        case PMIX_INT64:                                                \
+            PMIX_BFROPS_UNPACK_FLEX_CONVERT_SIGNED(int64_t, s, d);      \
+            break;                                                      \
+        case PMIX_SIZE:                                                 \
+            PMIX_BFROPS_UNPACK_FLEX_CONVERT_UNSIGNED(size_t, s, d);     \
+            break;                                                      \
+        case PMIX_UINT64:                                               \
+            PMIX_BFROPS_UNPACK_FLEX_CONVERT_UNSIGNED(uint64_t, s, d);   \
+            break;                                                      \
+        default:                                                        \
+            (r) = PMIX_ERR_BAD_PARAM;                                   \
+    }                                                                   \
+} while(0)
 
 /* Unpack generic size macros */
 #define PMIX_BFROP_UNPACK_SIZE_MISMATCH(reg_types, unpack_type, remote_type, ret)                     \
@@ -475,7 +703,11 @@ PMIX_EXPORT pmix_status_t pmix_bfrops_base_pack_envar(pmix_pointer_array_t *regt
 PMIX_EXPORT pmix_status_t pmix_bfrops_base_pack_coord(pmix_pointer_array_t *regtypes,
                                                       pmix_buffer_t *buffer, const void *src,
                                                       int32_t num_vals, pmix_data_type_t type);
-
+PMIX_EXPORT pmix_status_t pmix_bfrops_base_pack_int_flex(pmix_pointer_array_t *regtypes,
+                                                         pmix_buffer_t *buffer,
+                                                         const void *src,
+                                                         int32_t num_vals,
+                                                         pmix_data_type_t type);
 /*
 * "Standard" unpack functions
 */
@@ -609,6 +841,10 @@ PMIX_EXPORT pmix_status_t pmix_bfrops_base_unpack_envar(pmix_pointer_array_t *re
 PMIX_EXPORT pmix_status_t pmix_bfrops_base_unpack_coord(pmix_pointer_array_t *regtypes,
                                                         pmix_buffer_t *buffer, void *dest,
                                                         int32_t *num_vals, pmix_data_type_t type);
+PMIX_EXPORT pmix_status_t pmix_bfrops_base_unpack_int_flex(pmix_pointer_array_t *regtypes,
+                                                           pmix_buffer_t *buffer, void *dest,
+                                                           int32_t *num_vals,
+                                                           pmix_data_type_t type);
 /**** DEPRECATED ****/
 PMIX_EXPORT pmix_status_t pmix_bfrops_base_unpack_array(pmix_pointer_array_t *regtypes,
                                                         pmix_buffer_t *buffer, void *dest,
--- a/src/mca/bfrops/base/base.h
+++ b/src/mca/bfrops/base/base.h
@@ -14,6 +14,8 @@
  * Copyright (c) 2014-2019 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2019      Mellanox Technologies, Inc.
+ *                         All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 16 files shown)


src/mca/bfrops/base/bfrop_base_pack.c
```diff
@@ -336,7 +336,8 @@ pmix_status_t pmix_bfrops_base_pack_string(pmix_pointer_array_t *regtypes,
     for (i = 0; i < num_vals; ++i) {
         if (NULL == ssrc[i]) {  /* got zero-length string/NULL pointer - store NULL */
             len = 0;
-            if (PMIX_SUCCESS != (ret = pmix_bfrops_base_pack_int32(regtypes, buffer, &len, 1, PMIX_INT32))) {
+            PMIX_BFROPS_PACK_TYPE(ret, buffer, &len, 1, PMIX_INT32, regtypes);
```
karasevb: Move to the upper commit :
[ ](/karasevb)
` [bfrops: redefinition regtypes for internal usage](/pmix/pmix/pull/1130/commits/daef318771b1f537c49a0a63badeac5f1eae86f1)`

src/mca/bfrops/base/bfrop_base_pack.c
```diff
@@ -1070,3 +1070,25 @@ pmix_status_t pmix_bfrops_base_pack_coord(pmix_pointer_array_t *regtypes,
     }
     return PMIX_SUCCESS;
 }
+
+
+int pack_size(uint64_t size, uint8_t out_buf[9])
```
karasevb: Change to static

src/mca/bfrops/base/bfrop_base_unpack.c
```diff
@@ -1384,3 +1384,23 @@ pmix_status_t pmix_bfrops_base_unpack_coord(pmix_pointer_array_t *regtypes,
     }
     return PMIX_SUCCESS;
 }
+
+uint64_t unpack_size(uint8_t in_buf[])
```
karasevb: Change to static

src/mca/bfrops/base/bfrop_base_unpack.c
```diff
@@ -1385,22 +1385,65 @@ pmix_status_t pmix_bfrops_base_unpack_coord(pmix_pointer_array_t *regtypes,
...
+
+    PMIX_BFROPS_TYPE_SIZEOF(rc, type, val_size);
+    if (PMIX_SUCCESS != rc) {
+        PMIX_ERROR_LOG(rc);
+        return rc;
+    }
+    /* unpack the data */
+    for (i = 0; i < (*num_vals); ++i) {
+        unpack_sz = pmix_bfrop_unpack_flex(buffer->unpack_ptr, &tmp);
+        if ((uint32_t)unpack_sz > (val_size + 1)) {
```
karasevb: fix: compare against real val size

src/mca/bfrops/base/bfrop_base_pack.c
```diff
@@ -934,7 +934,7 @@ pmix_status_t pmix_bfrops_base_pack_query(pmix_pointer_array_t *regtypes,
     for (i=0; i < num_vals; i++) {
         /* pack the number of keys */
         nkeys = pmix_argv_count(pq[i].keys);
-        PMIX_BFROPS_PACK_TYPE(ret, buffer, &nkeys, 1, PMIX_UINT32, regtypes);
+        PMIX_BFROPS_PACK_TYPE(ret, buffer, &nkeys, 1, PMIX_INT32, regtypes);
```
artpol84: @rhc54 @jjhursey 
This issue was uncovered during a new flexible integer packing implementation.
The unpacking counterpart is expecting Signed integer.
In order to have an efficient packing for both signed and unsigned integers we pack them differently (because of the internal representation)

The side effect of this is that you have to use exactly the same type for packing and unpacking (signed and unsigned int's are not compatible anymore).
Is it ok with respect to the standard?

src/mca/bfrops/base/bfrop_base_unpack.c
```diff
@@ -1299,22 +1299,89 @@ pmix_status_t pmix_bfrops_base_unpack_coord(pmix_pointer_array_t *regtypes,
...
+    if (PMIX_UNLIKELY((buf_size-1) == idx &&
+                      (val & PMIX_BFROPS_FLEX_BASE7_CONT_FLAG))) {
         val = in_buf[idx++];
-        size = size + ((uint64_t)val << shift);
+        val_last = val;
+        value = value + ((uint64_t)val << shift);
+        shift_last = shift;
+    }
+    /* compute the most significant bit of val */
+    while (val_last != 0) {
```
karasevb: //TODO: implement a faster version

src/mca/bfrops/base/bfrop_base_unpack.c
```diff
@@ -1299,22 +1299,89 @@ pmix_status_t pmix_bfrops_base_unpack_coord(pmix_pointer_array_t *regtypes,
...
+        val_last >>= 1;
+        hi_bit++;
+    }
+    /* compute the real val size */
+    *out_val_size = (hi_bit + shift_last)/8 + !!((hi_bit + shift_last) & (8-1));
```
karasevb: (8-1) => ((1 << sizeof(char)) - 1)

src/mca/bfrops/base/bfrop_base_unpack.c
```diff
@@ -1299,22 +1299,89 @@ pmix_status_t pmix_bfrops_base_unpack_coord(pmix_pointer_array_t *regtypes,
...
+    if (PMIX_SUCCESS != rc) {
+        PMIX_ERROR_LOG(rc);
+        return rc;
+    }
+
+    /* unpack the data */
+    for (i = 0; i < (*num_vals); ++i) {
+        unpack_max_size = buffer->pack_ptr - buffer->unpack_ptr;
+        /* check remained buf size */
+        if (unpack_max_size > PMIX_BFROPS_FLEX_BASE7_MAX_BUF_SIZE) {
```
karasevb: Move this check into `pmix_bfrop_unpack_flex`

src/mca/bfrops/base/bfrop_base_pack.c
rhc54: I apologize for the delayed response - I had missed this question in my email. The standard actually requires that the pack and unpack have the same type. In the library, it will return an error if you pack an unsigned int and attempt to unpack it as an int - we check that the two pmix_data_type_t values match.

artpol84: @rhc54 great! Then we are ok here.
@karasevb is wrapping up my final comments. PRs will be ready for the community till tomorrow.

rhc54:
@artpol84 Is there an order in which this one and #1125 have to be committed?

artpol84:
No, they are independent of each other.

src/mca/bfrops/base/base.h
```diff
@@ -158,47 +160,209 @@ PMIX_EXPORT extern pmix_bfrops_globals_t pmix_bfrops_globals;
 #error Unsupported pid_t size!
 #endif
 
+#define PMIX_BFROPS_FLEX_BASE7_MAX_BUF_SIZE (SIZEOF_SIZE_T+1)
```
rhc54: I think you need to provide some explanation in the form of comments here - it isn't clear what you are doing in these macros, and why you are doing it.

artpol84: Agree.
@karasevb should we put a message from my commit here?
I think it explains the idea for positive numbers.
So please append description for negative ones and this should be complete enough.

karasevb: Done.

rhc54:
I'm sorry guys - maybe it's my old fossil brain, but I just don't understand this flexible packing thing, and the comments aren't helping explain it. What I need is for you to explain why you are taking even 16-bit values and appear to be converting them to 64-bit values.

The standard representation of a negative number is just an unsigned number with the highest bit set to indicate "negative". Your macros seem to replicate that outcome, only pushing the input into a 64-bit output value. I need to understand why we would want to do that - why not just pack (for example) a 16-bit negative value?

If it is a fixed size type (e.g., int16_t), then there is no question as to its size at both the source and receiving ends. For optimized buffers, we just pack that number in its native size and send it - the recipient is required to know the size and correctly unpack it. In other words, the recipient must unpack the buffer in the same order and with the same types as the source packed it. We require that the programmer ensure that ordering is preserved.

If it is a "generic" type like PMIX_INT, then there is always an ambiguity as to the size - there is no guarantee that the source and the destination will agree on that size. Thus, we send both the size of the type and the value, and then use the size on the receiving end to correctly convert the value to the local type. I can see how that might be shortened by using a byte to encode the source's size as it can only be one of a few set values (8, 16, 32, 64, or 128 bits) instead of sending the actual size value, but this PR doesn't seem to use that method.

For debug buffers, we send a tag identifying each of these value types - but that is solely for debug purposes. There is no value in trying to optimize packing of that information as it is never used in production. It is solely for debugging purposes so the programmer can get the pack/unpack sequences coordinated. I don't _think_ that is what you are addressing here, but I can't tell for certain.

Could you please explain what you are trying to accomplish and how this PR helps?

artpol84:
@rhc54 

**In short**
We were able to reduce the size of the modex message by a factor of 2 with this PR.

Now the details.

The important observation that I made and that lead to this work was that although types can accomodate a wide range of values, we in practice do not use the whole range. 
Here is the main description for the reference:
```
Typical representation of numbers is:
A[0]*B^0 + A[1]*B^1 + A[2]*B^2 + ... + A[n]*B^n
where B == 256 (one byte)

This encoding uses one bit of each byte for "continuation flag" so
integers are encoded with the same representation, but B = 128;
Continuation flag is used to reflect whether or not the next byte
contains next A[x] or this byte is the last one in the representation.
```

>I'm sorry guys - maybe it's my old fossil brain, but I just don't understand this flexible packing thing, and the comments aren't helping explain it. What I need is for you to explain why you are taking even 16-bit values and appear to be converting them to 64-bit values.

This is an intermediate step, we don't pack 64 bits. See an example below.

>The standard representation of a negative number is just an unsigned number with the highest bit set to indicate "negative".  Your macros seem to replicate that outcome, only pushing the input into a 64-bit output value.

This not correct. The standard representation of the negative values is "Two's complement" [link](https://en.wikipedia.org/wiki/Signed_number_representations#Two's_complement), while our flexible packing is using "Signed magnitude representation" [link](https://en.wikipedia.org/wiki/Signed_number_representations#Signed_magnitude_representation).


I figured 2 main questions:
1. How do we guarantee that the sizes for types are matching on 2 sides (i.e. size_t can be different)
2. Why do we need a special handling of the negative numbers.

Let start from the size_t and consider the case where system 1 has 8B size_t and system2 has 4B size_t.
System1 sends to system2

```C
size_t x = 1024; // Binary representation 0x0000 0000 0000 0400, 8B
```

So the typical representation is 
A[0]*B^0 + A[1]*B^1 + A[2]*B^2 + ... + A[n]*B^n = 0x04 * 256^1 + 0x0 * 256^0
A[0] = 0x04
A[1] = 0x00
A[2-7] = 0

Here is what flexible packing does
1. B[0] = 0x80 (1.000 0000, this byte contains 7 bits of payload, high bit set to 1 is a continuation flag that indicates that there will be one more byte in the representation
2. B[1] = 0x08 (0.000 0010, continuation flag is 0 as this byte is the last one. we have no payload bits

So instead of sending 8B over the wire:
```
original value = 0x0000 0000 0000 0400
```
We will send 2B:
```
wire: 0x0880 
```
And this value contains both the payload and the number of bits in the value.
Here is the breakdown of the 0x0880 value:
```
0x0880 = 0.000 00.10 1.000 0000
    0.000 00.10 = [cont flag = 0].[0s padding].[2 high bits of payload = 10]
    1.000 0000  = [cont flag = 0].[7 low bits of payload = 000 0000]
```
New packing relaxes the restriction of type size compatibility to the "fitability" which is not bad from my perspecive.
Now even if you have a mismatch, but the number still can fit 4B during unpack you will be fine. 
If it doesn't fit - you will be an overflow indication, because the bit-length is always encoded in the representation. 
I agree that this does change the behavior, but it doesn't breake correctness.

Now the negative number:
```C
short x = -3; // Binary representation 0xFFFD, 2B
```

So the typical representation is 
A[0]*B^0 + A[1]*B^1 = 0xFF * 256^1 + 0xFD * 256^0

Here is what flexible packing does
1. We split onto an abs value (0x3) and the sign(0x1) and re-represent the number in "Signed magnitude representation": (0x3 << 1) | 0x1 = 0x07 = 0x0000 011.1
2. Now we apply the algorithm described above to get flexible representation:
B[0] = 0x0.000 0.11.1 <continuation flag = 0>.<0s padding>.<abs value = 3>.<sign = 1>

So instead of sending 2B we will be sending 1B.

If this doesn't help, then for the sake of time saving lets have a call and go over this comment. I am confident that it pretty covers all the implementation details.

artpol84:
Here is where we check that type fits:
https://github.com/pmix/pmix/blob/6ca9b5c01402f1f33b1e09d1d2eaad009496427d/src/mca/bfrops/base/bfrop_base_unpack.c#L1381-L1383

rhc54:
Sorry, let me be clear. What I would like to see is you put that explanation into the code as a block comment. I'd like to avoid having such complex code in the code base without an explanation - people shouldn't have to search wiki's and PR threads to find out what it does.

I'm wondering if this should be in a framework somewhere for the sake of safety (so we fallback to the original method in case there turns out to be some cases that don't work but you haven't tried) and for alternative methods. Example: you could just test the magnitude of the value and use only the number of bits required to represent it. So an int less then 256 becomes a single byte, preceded by a byte that has a flag indicating how many bytes are used in the representation. The receiver knows it is looking for an int, unpacks the flag to see how many bytes were used to represent it, and converts.

Similar to what you are doing, but without the need to encode at the bit level. I'm not advocating for it, but the more I think about it, the more comfortable it would be to have a known-good backup.

artpol84:
With this approach, for small values, you will always be sending 2B instead of 1B as we do now. I strongly would like to avoid this.
Please point me to the place in the code where you'd like me to put this description.
We did a thorough testing of the packing itself in a standalone app. It was packing/unpacking integers from different ranges and comparing it against the original value. So I am confident in the correctness.
As a fallback we have bfrops v3 and if any bugs will be discovered - we will fix them.
So it seems to me that the only thing we really need to do is to copy the description I gave into a comment in the code.

We currently have sort of description here [link](https://github.com/pmix/pmix/blob/6ca9b5c01402f1f33b1e09d1d2eaad009496427d/src/mca/bfrops/base/bfrop_base_pack.c#L1033). So I can append an edited version there if you are OK with this.

rhc54:
Perhaps just adding a comment to the code that points to the explanation (maybe adding some of the above words) will suffice.

Like I said, I wasn't advocating the alternative approach. Just pointing out that there may be alternative approaches that people want to use and we shouldn't block them from doing so.

Unfortunately, falling back to bfrops v3 isn't an acceptable solution as there are datatypes in v4 that are not present in v3. It isn't a question of someone fixing things - it is an issue of something breaking that prevents users from being able to run. It would be good to ensure that v4 support can continue operating even if we hit problems with this scheme. I understand that you tested it, but that doesn't mean we won't hit an error in production - it is pretty common to do so with something new. It shouldn't be hard to move this into a new framework that has two components (simply packing/unpacking the values as we do today, and another that uses this base-256 scheme), and it protects us should something go wrong (e.g., big vs little endian machines).

artpol84:
So basically you’d like to see it as a bfrops5?

With respect to little/big endian I thought of that. It has to do with bytes ordering.
Because we separate value onto bites using local systems math operations I am sure that this encoding is endian agnostic.

artpol84:
The problem of having another component is the maintenance.
I know you tend to ignore components you not interested in. Who will ensure that all updates going to bfrops v4 are ported to v5?

Isn’t it better to do the good testing we normally do before the release? I believe there is a lot of changes that I’d consider complex that are going into PMIx v4 without facing such concerns.

Do we currently have successful experience with heterogeneous systems?
If so - why not to ask people to verify?
If not - I am sure we will face all sorts of other problems and not envision that this one will be

rhc54:
I'm not suggesting a bfrops 5 for the reason you identified - it doesn't help with the datatype definition problem. What I'm suggesting is that you create a new framework for encoding schemes, one component that just passes the value thru and another that converts it to/from base-256.

My point is solely that we can easily hit unanticipated problems when such schemes go into production. For example, consider the case that was reported where ds21 is broken - or the issue where ds12 is reporting errors. Both were tested prior to release, but are hitting problems in different environments. Not singling them out - everyone (myself included) writes code that winds up having a problem under some set of circumstances. It's good to have a backup when possible.

We always try to test before a release, but let's face it - we have a very small testing pool available to us. Our tests don't cover much of the environments found out there, though we all do our best. My concern over a backup isn't related to the complexity of this code, but rather that (a) we have a way to back it up if needed (lots of places in the code don't have that option), and (b) there are multiple ways of implementing it.

In this case, we have already identified three different ways of encoding the modex (the way we do it today, your way, and the alternative I hypothesized). We may also want to just compress the modex data before passing it up to the host - might want to do that regardless of whether or not we encoded it. This just feels like a framework to me as opposed to hard-coding it, and the framework needs to be separate from bfrops as it will need to apply to multiple bfrops components going forward.

I don't have a particular concern that can be addressed in a walk-thru, though I'm happy to discuss it. We don't have a call this week and I can't make the one next week, but we could do it the following one.

artpol84:
So how do you want to proceed with this?
1. Merge this one and later move into the framework?
2. Create a framework and then introduce new packing?
What is going to be a scope of the framework? Only integers (this is what we address currently)?

I'd prefer to merge this one as this PR address other things we definitely want to keep (like bfrops infrastructure refactoring) and then create a framework.
And I'd appreciate your help with establishing a framework as we don't have this experience. From what I see the interface that is required now is a pack/unpack functions for int8,16,32,64.

src/mca/bfrops/base/base.h
```diff
@@ -158,47 +160,306 @@ PMIX_EXPORT extern pmix_bfrops_globals_t pmix_bfrops_globals;
...
+/* Flexible packing */
+#define PMIX_BFROPS_FLEX_BASE7_MAX_BUF_SIZE (SIZEOF_SIZE_T+1)
+#define PMIX_BFROPS_FLEX_BASE7_MASK ((1<<7) - 1)
+#define PMIX_BFROPS_FLEX_BASE7_SHIFT 7
+#define PMIX_BFROPS_FLEX_BASE7_CONT_FLAG (1<<7)
+
+/**
+ * Packing conversion of a signed integer value to a flexible representation.
+ * The main idea is to split a signed negative value onto an absolute value
+ * and a sing bit stored in the special location.
```
jjhursey:
```suggestion
 * and a sign bit stored in the special location.
```

rhc54: I didn't know you could do that - neat!!

src/mca/bfrops/base/base.h
```diff
@@ -158,47 +160,306 @@ PMIX_EXPORT extern pmix_bfrops_globals_t pmix_bfrops_globals;
...
+    type __tbuf = 0;                                            \
+    uint64_t __tmp = val;                                       \
+    int sign = (__tmp) & 1;                                     \
+    __tmp >>= 1;                                                \
+    if (sign) {                                                 \
+        __tmp = ~__tmp;                                         \
+    }                                                           \
+    __tbuf = (type)__tmp;                                       \
+    memcpy(ptr, &__tbuf, sizeof(type));                         \
+} while (0)
```
jjhursey: (very minor suggestion) It would be nice if the `PMIX_BFROPS_PACK_FLEX_CONVERT_SIGNED ` and `PMIX_BFROPS_UNPACK_FLEX_CONVERT_SIGNED` were next to each other in the code (same with `UNSIGNED`). That way it's easier to see them together since they are paired.

artpol84: Sure,
@karasevb please do

jjhursey:
I like the cleanup of the bfrops code and the new encoding seems interesting. It might be good to separate the two changes (which is mostly already done in the commit breakdown). The cleanup seems good to go IMHO.

The encoding is interesting, and I don't see anything wrong with it from my review. I do have some questions for discussion based on the other discussion in this thread.

 * There is some performance analysis [here](https://github.com/pmix/pmix/issues/681#issuecomment-455986873), but it seems to be mostly focused on PR #1125 (though does have some details specific to this PR).
   - Do you have some (newer) specific numbers of modex was X without this PR and Y with this PR as it scales up - even at modest scales (32 or 64 nodes) now that PR #1125 has been merged?
 * I am nervous about the lack of ability to disable this feature and fall back on the old method. Not quite as much because of bugs being introduced (though that is a valid reason), but I'm concerned about encountering some situation where I don't want this encoding because I'm somehow packing my integers in a way that is unkind to this encoding.
   - So I see two options here: (1) define a framework for the encoding method and have 2 components, (2) have an adjustable threshold where under the threshold we use this encoding and over we use the traditional technique.
    - With (1) a user could pick the traditional component over the default encoded. I can't think of any 'other' way to encode at the moment, so I'm not 100% convinced a framework is the "right thing" to do, but it feels like an option.
    - With (2) a user could adjust the threshold down to 0. I'm not sure how much overhead that would add to pack/unpack.
    - Right now I'm leaning towards (2) and if there is something later that wants to add an additional encoding mechanism then they can take on the burden to create a framework.
 * The encoding feature seems like it might be interesting to apply to other places in the code than the modex operation to reduce how we exchange any sort of data. Is this change really only used by the modex or does it apply to other places in the code?
 * If we think about trying to backport this encoding to the v3.x branch (I'm not sure if folks are, but I'm just thinking out loud a bit) then, for cross-version compatibility, we would need an option to check if the integers are encoded or not. I'm not sure how to address that. The bfrops component is already part of the handshake, so a separate component might be an options here. ... I need to ponder that a bit more, if it is something we want to do.

src/mca/bfrops/base/base.h
jjhursey: A trick I learned at the MPI Forum. I'll show you sometime :)

artpol84:
@jjhursey, thanks for the review!

Some comments below:
> Do you have some (newer) specific numbers of modex was X without this PR and Y with this PR as it scales up - even at modest scales (32 or 64 nodes) now that PR #1125 has been merged?

By the [link](https://github.com/pmix/pmix/issues/681#issuecomment-455986873) you provided there are 3 tables describing 3 portions of the per-rank information.
#1125 only address rank meta-information (reduces it by 11B), but all other improvements are thanks to flexible packing.
In particular, it decreases:
* `rank header` by 11B;
* `rank info` by 9B; and
*`rank data` by 15B. NOTE in the rank data the payload is 50B (38B UCX endpoint + key string of 12B) and with flex. packing whole size if 55B, so only 5B of service overhead.

Overall the per-rank portion decreases from 106B down to 58B which is ~45% decrease.
And service overhead is 9B / 58B = 15% versus initial 52%.

@karasevb will provide scaling results later today.

> The encoding feature seems like it might be interesting to apply to other places in the code than the modex operation to reduce how we exchange any sort of data. Is this change really only used by the modex or does it apply to other places in the code?

I don't really see the opportunity elsewhere as you can't manipulate those numbers natively (have to pack/unpack). So it is best suitable for communication size reduction.

> I am nervous about the lack of ability to disable this feature and fall back on the old method. Not quite as much because of bugs being introduced (though that is a valid reason), but I'm concerned about encountering some situation where I don't want this encoding because I'm somehow packing my integers in a way that is unkind to this encoding.

Understood

> So I see two options here: (1) define a framework for the encoding method and have 2 components, (2) have an adjustable threshold where under the threshold we use this encoding and over we use the traditional technique.

I'm leaning towards the second one because I believe it will be easier to implement without any additional perf. penalties and code complications. 
As I said it would be more comfortable for me to merge the code the way it is and work on the framework in the additional PR. As this is a separate issue.
Again I would greatly appreciate help with framework establishment. We can take care of content.

> If we think about trying to backport this encoding to the v3.x branch (I'm not sure if folks are, but I'm just thinking out loud a bit) then, for cross-version compatibility, we would need an option to check if the integers are encoded or not. I'm not sure how to address that. The bfrops component is already part of the handshake, so a separate component might be an options here. ... I need to ponder that a bit more, if it is something we want to do.

We would like to have it in 3.x. I think having a bfrops32 component would be ok. This would isolate new changes in a modularized fashion + allowed to stress the differences of this port vs bfrops4 (where new datatypes are available).

rhc54:
I'm not really comfortable having this in a bfrops component - it isn't a buffer operation so much as an encoding step. Mind you, I'm not objecting to doing it. My concern here is that the code causes replication of bfrops components for no other reason than encoding the data.

I don't really have a strong preference about committing something and then cleaning it up. I guess I don't view it as a separate issue so much as getting the code architecture correct, but that's a detail. I definitely would like to see it moved to the framework prior to going into releases unless there is something that drives us to release sooner.

Could you please explain why we would backport this to a minor release? It seems a little odd for something like this to appear there - it's a pretty significant change. Is there something motivating that request vs waiting for 2-3 months for 4.0?

artpol84:
> I'm not really comfortable having this in a bfrops component - it isn't a buffer operation so much as an encoding step. Mind you, I'm not objecting to doing it. My concern here is that the code causes replication of bfrops components for no other reason than encoding the data.

To clarify, I was talking about a packing framework, not another bfrops component.

And I was hoping you can create one (framework) so we will move it there right away.

src/mca/bfrops/base/base.h
karasevb: Done with commit 5b423caaddabcb24337fc836b77e4704a7d438f4

artpol84:
Regarding 3.x:
Both IBM and we are a little behind in capturing new PMIx versions.
Speaking for myself - we might be interested in taking advantage of this optimization in a shorter term.

rhc54:
> To clarify, I was talking about a packing framework, not another bfrops component.
> 
> And I was hoping you can create one (framework) so we will move it there right away.

Oh, I see - thanks for clarifying. Sure, I can create one for you - I'll try to make it clear exactly how it is done so you'll be comfortable doing it in the future.

rhc54:
> Both IBM and we are a little behind in capturing new PMIx versions.
> Speaking for myself - we might be interested in taking advantage of this optimization in a shorter term.

Okay, let's take a look at it. I wasn't actually planning another release series in the 3.x group, but we can certainly consider it.

artpol84:
> Oh, I see - thanks for clarifying. Sure, I can create one for you - I'll try to make it clear exactly how it is done so you'll be comfortable doing it in the future.

Greatly appreciated!
But first let's wait till this one is mergeD
Or at least make sure it won't have merge conflicts

jjhursey:
I played around with this encoding this morning. I wanted to see for a set of datatypes targeted what ranges of integers will show a benefit. Below is what I found, in case it helps others.

Bytes in parentheses. For example `Same    (  2)` means `2 bytes` at this level.
A more detailed breakdown if you use the drop down arrow.

### Type: `short`

```
-----------------------   Signed (short) 2 bytes [-32768 : 32767]
 Benefit (  1):                       [-64, 63]
 Same    (  2):            [-8192, -65]       [64, 8191]
 Loss    (  3): [MIN, -8193]                           [8192, MAX]
----------------------- Unsigned (short) 2 bytes [0 : 65535]
 Benefit (  1): [0, 127]
 Same    (  2):        [128, 16383]
 Loss    (  3):                   [16384, MAX]
```

<details>
 <summary>Detailed breakdown</summary>

```
-----------------------   Signed (short) 2 bytes [-32768 : 32767]
 3 bytes :               -32768
 2 bytes :                -8192
 1 bytes :                  -64
 2 bytes :                   64
 3 bytes :                 8192
 3 bytes :                32767 : Search Limit
----------------------- Unsigned (short) 2 bytes [0 : 65535]
 1 bytes :                    0
 2 bytes :                  128
 3 bytes :                16384
 3 bytes :                65535 : Search Limit
```
</details>

### Type: `int`

```
-----------------------   Signed (  int) 4 bytes [-2147483648 : 2147483647]
 Benefit (1-3):                                      [-1048576, 1048575]
 Same    (  4):                 [-134217728, -1048577]                 [1048576, 134217727]
 Loss    (  5): [MIN, -134217729]                                                         [134217728, MAX]
----------------------- Unsigned (  int) 4 bytes [0 : 4294967295]
 Benefit (1-3): [0, 2097151]
 Same    (  4):            [2097152, 268435455]
 Loss    (  5):                               [268435456, MAX]
```

<details>
 <summary>Detailed breakdown</summary>

```
-----------------------   Signed (  int) 4 bytes [-2147483648 : 2147483647]
 5 bytes :          -2147483648
 4 bytes :           -134217728
 3 bytes :             -1048576
 2 bytes :                -8192
 1 bytes :                  -64
 2 bytes :                   64
 3 bytes :                 8192
 4 bytes :              1048576
 5 bytes :            134217728
 5 bytes :           2147483647 : Search Limit
----------------------- Unsigned (  int) 4 bytes [0 : 4294967295]
 1 bytes :                    0
 2 bytes :                  128
 3 bytes :                16384
 4 bytes :              2097152
 5 bytes :            268435456
 5 bytes :           4294967295 : Search Limit
```
</details>

### Type: `long`

```
-----------------------   Signed ( long) 8 bytes [-9223372036854775808 : 9223372036854775807]
 Benefit (1-7):                                                              [-281474976710656, 281474976710655]
 Same    (  8):                         [-36028797018963968, -281474976710657]                                 [281474976710656, 36028797018963967]
 Loss    (  9): [MIN, -36028797018963969]                                                                                                         [36028797018963968, MAX]
----------------------- Unsigned ( long) 8 bytes [0 : 18446744073709551615]
 Benefit (1-7): [0, 562949953421311]
 Same    (  8):                    [562949953421312, 72057594037927935]
 Loss    (  9):                                                       [72057594037927936, MAX]
```

<details>
 <summary>Detailed breakdown</summary>

```
-----------------------   Signed ( long) 8 bytes [-9223372036854775808 : 9223372036854775807]
 9 bytes : -9223372036854775808
 8 bytes :   -36028797018963968
 7 bytes :     -281474976710656
 6 bytes :       -2199023255552
 5 bytes :         -17179869184
 4 bytes :           -134217728
 3 bytes :             -1048576
 2 bytes :                -8192
 1 bytes :                  -64
 2 bytes :                   64
 3 bytes :                 8192
 4 bytes :              1048576
 5 bytes :            134217728
 6 bytes :          17179869184
 7 bytes :        2199023255552
 8 bytes :      281474976710656
 9 bytes :    36028797018963968
 9 bytes :  9223372036854775807 : Search Limit
----------------------- Unsigned ( long) 8 bytes [0 : 18446744073709551615]
 1 bytes :                    0
 2 bytes :                  128
 3 bytes :                16384
 4 bytes :              2097152
 5 bytes :            268435456
 6 bytes :          34359738368
 7 bytes :        4398046511104
 8 bytes :      562949953421312
 9 bytes :    72057594037927936
 9 bytes : 18446744073709551615 : Search Limit
```
</details>

rhc54:
@jjhursey Now that you have that test script, it would be interesting to see the results if you just compressed those values.

artpol84:
@jjhursey here is the optimization effect you were asking for
image

In this example, UCX endpoint size was 48B and 12B for the key string, so 60B per rank of payload.
This means that the overall size for 32 nodes is
* 32 x 1 x 60 = 1920B
* 32 x 28 x 60 =  53760

So the overhead of the current master (with namespace optimization) is
* 32 x 1PPN: 3872B - 1920B = 1952B (100%)
* 32 x 28PPN: 92000B - 53760B = 38240B (70%)

With this PR the situation will be the follwoing:
* 32 x 1PPN: 2304B - 1920B = 384B (20%)
* 32 x 28PPN: 61856B - 53760B = 8096B (15%)

**Coming soon ...**
We have optimized UCX endpoint to be 23B instead of 48B. For it the price of the overhead will be higher for both cases.
**Payload size**:
* 32 x 1 x (23 + 12) = 1120B
* 32 x 28 x (23 + 12) =  31360B
**Current master overhead**:
* 32 x 1PPN: 1952B / 1120B  = 174%
* 32 x 28PPN: 38240B / 31360B = 120%
**With this PR**:
* 32 x 1PPN: 384B / 1120B  = 34%
* 32 x 28PPN: 8096B / 31360B = 25%

artpol84:
> @jjhursey Now that you have that test script, it would be interesting to see the results if you just compressed those values.

Applying a compression algo like Lempel-Ziv to individual numbers won't buy you anything as it needs reasonable sample to adjust to the content.
It has to be a modex message if you want to get a representative result.
I believe that flex packing as a specialized compression should be a good complementary to a subsequent general-purpose compression afterwards.
Lempel-Ziv will handle those 0s nicely as well, but still will put some short statement instead of them while flex. packing eliminates them completely knowing the specifics of the problem.

I do not expect that PMIx packing will fall frequently in those inefficiency intervals often.
And there is not need for the scrip there, the definition of flex packing inefficiency is simple, you will get inefficient if you are using more than (TYPE_BITS - 2) bit numbers as flex packing is using 1 bit per byte for the service information.

artpol84:
I do expect that general-purpose compression will have some additional positive effect as at least key names usually have the same value, so you can eliminate it.

The problem is that at the modex time PMIx submits only a fraction (it's local portion) of the message, so there will be an additional overhead of service information like compressing tables embedded in each per-node contribution (which now is on the level of 2-3KB). So this will be a new type of overhead.

rhc54:
Obviously, one doesn't compress individual numbers - that wouldn't make any sense. My question was more about compressing each node-level contribution as that might provide a big enough block to gain something. Compression algos can't ferret out key names - the data is just treated like an opaque blob. The amount you gain depends on the compression algo vs the repetition frequency of the bit patterns. We were getting better than 10x on the modex in ORTE when we use it there on the aggregated blob - be interesting to see what we gain at the individual node level.

artpol84:
> Obviously, one doesn't compress individual numbers - that wouldn't make any sense.

I see, sorry. From the comment below it felt like you want to compress the numbers with a script similar to the one Josh created for the flexible packing:
> @jjhursey Now that you have that test script, it would be interesting to see the results if you just compressed those values.

> Compression algos can't ferret out key names - the data is just treated like an opaque blob. The amount you gain depends on the compression algo vs the repetition frequency of the bit patterns.

So because the key names are the same for each rank I have reasons to expect that compression algo will replace similar sequence of bits representing a key with a very short compression sequence.

> We were getting better than 10x on the modex in ORTE when we use it there on the aggregated blob - be interesting to see what we gain at the individual node level.

I think it will be interesting indeed.

artpol84:
One advantage of the flex. packing is that I expect it to be faster. Especially on the modex unpack time.

rhc54:
I just ran a quick test of compressing the equivalent of the full modex from 30,000 nodes running 24ppn (total of 720,000 procs). Each proc contributes a modex value consisting of the full pmix_key_t array and a payload of 48Bytes - this kinda matches our default operation. I don't have real numbers for the payload, so I inserted random values into each byte of the payload.

If I take that entire stream and compress it with zlib, it is the equivalent of every node passing its raw data to mpirun - this would be the size of the resulting blob broadcast out to the daemons:
```shell
NPROCS: 720000 FULL SIZE: 214 MBytes	Compressed Size: 33 MBytes
```
If I instead compress each node's contribution before sending it to mpirun, and then have mpirun simply aggregate and broadcast the resulting blob, it looks more like this:
```
NPROCS: 720000 FULL SIZE: 214 MBytes	Compressed Size: 19 MBytes
```
I have posted my program on the following gist: https://gist.github.com/rhc54/48e46b36a9b815f7e360e35f8c16e334. It would be good to run a similar simulation on the proposed flex-int packing scheme for comparison.

karasevb:
We tried to compress and compare modex sizes with our flexible packaging and without, we got the following results:

  | no optimization | nspace + flex optimization
-- | -- | --
size | 3281 | 1930
compressed size | 687 (link: [modex_dump_noopt.gz](https://github.com/pmix/pmix/files/2944371/modex_dump_noopt.gz)) | 659 (link: [modex_dump_opt.gz](https://github.com/pmix/pmix/files/2944372/modex_dump_opt.gz))

artpol84:
Indeed general purpose compression captures all things that flexible packing do + leverages higher-level relations between the data. I'm pretty sure that the "prefix" of each per-rank portion is the same and key names are the same as we discussed. This should be a sufficient redundancy.

The only concern about compression that I have now is the time to unpack the final blobe. I don't anticipate decompression to be slow. I expect it to be much faster than compression because it is a more straightforward process that doesn't require searching.

I'd like to do some timing. naive `time gunzip` showed 14ms for a one node blob. This is high, but I don't trust this way of measuring. We will take it next week.

So I'd like to step back with compression and suggest the following:
* Split this PR on 2
* First will contain commits with fixes and bfrops refactoring that I think is good anyway.
* The second part is questionable so will require additional discussion. It does improve the case where no zlib compression is available, so it has it's right to exist. On the other hand it' effectiveness is small (3%) when compression is used.

artpol84:
I gave it a thought over the night and I still think we need flexible packing.

From the measurements that @karasevb [did](https://github.com/pmix/pmix/pull/1130#issuecomment-470799845) we can conclude that we still have redundancy in the node contribution and we know we do.

The size of the node contribution with flexible packing (1930B) is comprised from 28 instances of:
* 48B UCX endpoint
* 12B of key name string
* 9B of service information

So total: `28 * (48 + 12 + 9) = 1932B`

There is a minor no overhead from the service information. The main 2 contributors are we have is
1. UCX endpoint size is not efficient
1. We store the key name for each per-rank info, and this is exactly same key "pml.ucx.4.0" in particular.

The first issue is allready addressed and in the next UCX release it's size will be reduced down to 22B (https://github.com/openucx/ucx/pull/3138, https://github.com/openucx/ucx/pull/3231, https://github.com/openucx/ucx/pull/3056).
And I assume that othere communication libraries have their address cormats in the binary form and optimized. if not - they can leverage general-purpose compression.

Fixing a second is a trivial protocol change similar to https://github.com/pmix/pmix/pull/1125, or a dictionary (as @bosilca approached it). It will be a simple change to the modex pack/unpack logic.

Now getting all together: `28 * 23 + 12 = 656B` which is exactly what zlib compression can give according to @karasevb measurements.

I am an advocate of having a smart and efficient protocol rather than using a brut force of GP compression.

I am still fine with having a GP compression option available in PMIx, I just don't want to use it for my customers.

artpol84:
I also don't want to introduce another dependency given that we know it is not required.

rhc54:
Let's stick with yesterday's proposed plan. Break this PR into two parts and commit the refactoring as that benefits everyone. Then let's hold off on the second half until we can get it into a framework as that part increasingly looks like it will be specific to a particular environment.

Meantime, I will write a "how-to" guide on the community web site detailing how to eliminate the modex as it really is no longer required for Infiniband devices. Hopefully can post it next week depending on free time during travel.

Just as a reminder: we already have a dependency on zlib. We use it to compress any string longer than a set limit and therefore link against it by default if zlib is found. I'm not advocating any particular solution for modex size as that operation is optional - just pointing out that compression isn't adding a new dependency.

artpol84:
Ok. I'd be extremely interested to see the howto that you mentioned.
It is still not obvious to me if we can avoid modex.

Do we have a howto on how to create the frameworks? It makes sense to open a second-part PR with new framework introduction.

Also we will work on key name elimination solution over the next weekend. I think it is also generic and will benefit all environments. 

>  Then let's hold off on the second half until we can get it into a framework as that part increasingly looks like it will be specific to a particular environment.

This is not fully true though it might seem that I'm talking about UCX only. In fact this is just because I have numbers for UCX. I'll check with MPICH guys how they represent the endpoints. I expect that it should also be small and binary packed. So what I said about UCX remains for other use cases.

artpol84:
I talked to @raffenet about how OFI represents it's endpoint. And they way I understood it is always 8B which is really small and falls into the "optimized ep" case.
So I expect that proposed solution will benefit this case as well.

raffenet:
Hmm, I may have been confusing my concepts. Looking at the manpage for `fi_getname`, it can (and probably is) much larger in some cases. https://ofiwg.github.io/libfabric/v1.7.0/man/fi_cm.3.html

rhc54:
yeah, it can be - and let's not forget that we have other fabrics besides IB to support.

artpol84:
> yeah, it can be - and let's not forget that we have other fabrics besides IB to support.

I guess OFI is fabric agnostic as well as UCX. Can you elaborate? IP address + port will be also a short endpoint description.

@raffenet can you tell how much larger? And more improtant is what is their internal representation. I guess it is natively binary and still in the order of 10s of bytes.

jjhursey:
I was getting a bit turned around in the conversation on this ticket so I spoke with @artpol84 and @rhc54 offline and here are my notes:

1. @artpol84 @karasevb will break this PR into two parts: (which there is no objection to bringing in once complete)
   a. Refactoring of the `bfrops` - no functional changes, just code cleanup.
   b. Introduce a new `pencode` (or some name) framework with two components. (i) current behavior, (ii) flexible-size encoding. @rhc54 said that he would help create the framework structure to help get that started.
2. @rhc54 will introduce an MCA parameter option to use `zlib` to compress the per-node modex.
3. Move the discussion of additional modex optimizations back to Issue #681 for further development.

Once the 2 PRs are ready for (1) I haven't heard any objection to bringing them in. I'm interested in experimenting with both (1) and (2) both separately and together. It's nice to have an option to enable/disable each at runtime so that the user can determine if they are helpful to their launch scenario/configuration.

karasevb:
This PR is superseded by #1152

PR #1132, opened 2019-02-27
test/test_server: do not invoke libevent directly

ggouaillardet:
but uses PMIx "wrappers" pmix_event_{add,del}() instead
of directly calling libevent event_{add,del}()


rhc54:
dang - thought i had gotten all of these. thx!

PR #1135, opened 2019-03-05
modex: avoid exchange unnecessary buffer when collect flag is not set

### LINKED ISSUE #143: Optimize empty fence
Currently if no data collection flag was specified we will still pack the collective type flag into the message body:
https://github.com/pmix/master/blob/master/src/server/pmix_server_ops.c#L561
this causes 5 bytes to be appended to the message (1 byte of data and 4 byte of bfrop type signature).

If we have relatively high number of nodes (~1024) it may cause total message size to grow up to of 5KB.

I think that we should provide RM with the empty message to exchange. And expect to receive the empty message at fence completion. If we don't - this is an error - somebody called fence with data collection and somebody without.

karasevb:
Closes #143

### DIFF
--- a/src/server/pmix_server_ops.c
+++ b/src/server/pmix_server_ops.c
@@ -3700,14 +3700,16 @@ static void _grpcbfunc(int sd, short argc, void *cbdata)
      * store it for us before releasing the group members */
     if (NULL != bo) {
         PMIX_CONSTRUCT(&xfer, pmix_buffer_t);
-        PMIX_LOAD_BUFFER(pmix_globals.mypeer, &xfer, bo->bytes, bo->size);
         PMIX_CONSTRUCT(&nslist, pmix_list_t);
-        // collect the pmix_namespace_t's of all local participants
+        /* Collect the nptr list with uniq GDS components of all local
+         * participants. It does not allow multiple storing to the
+         * same GDS if participants have mutual GDS. */
         PMIX_LIST_FOREACH(cd, &trk->local_cbs, pmix_server_caddy_t) {
             // see if we already have this nspace
             found = false;
             PMIX_LIST_FOREACH(nptr, &nslist, pmix_nspace_caddy_t) {
-                if (nptr->ns == cd->peer->nptr) {
+                if (0 == strcmp(nptr->ns->compat.gds->name,
+                            cd->peer->nptr->compat.gds->name)) {
                     found = true;
                     break;
                 }
@@ -3722,6 +3724,7 @@ static void _grpcbfunc(int sd, short argc, void *cbdata)
         }
 
         PMIX_LIST_FOREACH(nptr, &nslist, pmix_nspace_caddy_t) {
+            PMIX_LOAD_BUFFER(pmix_globals.mypeer, &xfer, bo->bytes, bo->size);
             PMIX_GDS_STORE_MODEX(ret, nptr->ns, &xfer, trk);
             if (PMIX_SUCCESS != ret) {
                 PMIX_ERROR_LOG(ret);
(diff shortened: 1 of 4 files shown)


src/server/pmix_server_ops.c
```diff
@@ -555,16 +555,16 @@ static pmix_status_t _collect_data(pmix_server_trkr_t *trk,
...
-    /* because the remote servers have to unpack things
-     * in chunks, we have to pack the bucket as a single
-     * byte object to allow remote unpack */
-    PMIX_UNLOAD_BUFFER(&bucket, bo.bytes, bo.size);
-    PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, buf,
-                     &bo, 1, PMIX_BYTE_OBJECT);
-    PMIX_BYTE_OBJECT_DESTRUCT(&bo);  // releases the data
-    if (PMIX_SUCCESS != rc) {
-        PMIX_ERROR_LOG(rc);
+	/* because the remote servers have to unpack things
```
karasevb: - Pack it in the debug mode.
- Make sure that buffer will be released correctly

karasevb:
@rhc54 @artpol84, please review

rhc54:
Something doesn't look right here. If we are in a non-debug build and the fence is not collecting data, then the packing routine will not pack the "ctype" byte in the modex bucket:
```c
#if PMIX_ENABLE_DEBUG
                         &ctype, 1, PMIX_BYTE);
#endif
```
Yet the unpack routine in "store_modex" appears to always expect that byte to be present. Isn't this a mismatch when running in optimized mode? I can see how you'd wind up with an empty bucket and so I guess you don't have a byte object in the message for this contributor - is that how you avoid getting an error? If so, then a comment to explain that would be helpful as otherwise it looks like a programming error - it took me quite a while to work though the logic to see how it might not generate an error.

karasevb:
If the collect flag is not set, we won't the go to the routine in "store_modex":
https://github.com/pmix/pmix/blob/14b6103dfc94d81dea9f5f21af43ed35eff9029c/src/server/pmix_server.c#L2316-L2319
Anyway, the unpack "store_modex" handles a empty buffer correctly:
https://github.com/pmix/pmix/blob/14b6103dfc94d81dea9f5f21af43ed35eff9029c/src/mca/gds/base/gds_base_fns.c#L215-L217

rhc54:
@karasevb Please see my committed comments to verify they are correct. Once done, please rebase and squash the commit so the signed-off checker passes.

karasevb:
@rhc54 Your comments are completely correct. Thanks.
Will update.

karasevb:
Current Jenkins failure goes from the check added with the commit: https://github.com/pmix/pmix/pull/1135/commits/46e778c250b3c623f2c0346a782c3b8709de2f63

The problem is on this cycle:
https://github.com/pmix/pmix/blob/4090e24a3ea622a401e362179c5d571a9e8f0f15/src/server/pmix_server.c#L2340-L2346

`PMIX_GDS_STORE_MODEX` invoked several times for multiple namespaces,  but from the first iteration  the modex bucket is completely read from the buffer and for the next iteration buffer has zero size. That leads to an error with on the check I had added.

jjhursey:
Per teleconf:
 * I believe that this PR is ready for review/merge.

artpol84:
Good to go

artpol84:

jjhursey:
@karasevb it looks like after merging PRs #1152 and #1149 we have a merge conflict here. Can you resolve it?

karasevb:
> @karasevb it looks like after merging PRs #1152 and #1149 we have a merge conflict here. Can you resolve it?

@jjhursey Done

PR #1142, opened 2019-03-07
Fix access to uninitialized info

dsolt:
If this is a tool that we already established info for, then
info variable is not set locally, but is in the peer structure.

Signed-off-by: David Solt <dsolt@us.ibm.com>


awlauria:
Alternatively, we can set `info = NULL` on initialization, and only set the peer if it is not NULL. I'm not sure which is more proper.

rhc54:
Just to clarify - it the tool was launched as a client, then the peer object already exists. In that code path, the "info" pointer isn't initialized and will cause a segfault. Using the peer->info pointer guarantees that it is always initialized before use.

jjhursey:
@dsolt can you PR this to the v3.x branch?

PR #1149, opened 2019-03-12
 modex: reduce modex size by replacing key names with indexes

karasevb:
This commit optimizes the modex size, avoiding duplication of key-names for key-value pairs. It generates the key-map with the unique names and packs key indexes instead of key-name strings. The key-map packaged separately and allows the remote side to determine the key-name by index when unpacking.

### DIFF
--- a/src/mca/gds/base/gds_base_fns.c
+++ b/src/mca/gds/base/gds_base_fns.c
@@ -98,7 +98,6 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
     pmix_buffer_t bkt;
     pmix_byte_object_t bo, bo2;
     int32_t cnt = 1;
-    char byte;
     pmix_collect_t ctype;
     bool have_ctype = false;
     pmix_server_trkr_t *trk = (pmix_server_trkr_t*)cbdata;
@@ -109,6 +108,8 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
     bool found;
     char  **kmap = NULL;
     uint32_t kmap_size;
+    pmix_gds_modex_key_fmt_t kmap_type;
+    pmix_gds_modex_blob_info_t blob_info_byte = 0;
 
     /* Loop over the enclosed byte object envelopes and
      * store them in our GDS module */
@@ -121,7 +122,7 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
         /* unpack the data collection flag */
         cnt = 1;
         PMIX_BFROPS_UNPACK(rc, pmix_globals.mypeer,
-                &bkt, &byte, &cnt, PMIX_BYTE);
+                &bkt, &blob_info_byte, &cnt, PMIX_BYTE);
         if (PMIX_ERR_UNPACK_READ_PAST_END_OF_BUFFER == rc) {
             /* no data was returned, so we are done with this blob */
             PMIX_DESTRUCT(&bkt);
@@ -134,9 +135,12 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
             goto exit;
         }
 
-        // Check that this blob was accumulated with the same data collection setting
+        /* Check that this blob was accumulated with
+         * the same data collection setting */
         if (have_ctype) {
-            if (ctype != (pmix_collect_t)byte) {
+            pmix_collect_t ctmp = PMIX_GDS_COLLECT_IS_SET(blob_info_byte) ?
+                        PMIX_COLLECT_YES : PMIX_COLLECT_NO;
+            if (ctype != ctmp) {
                 rc = PMIX_ERR_INVALID_ARG;
                 pbkt.base_ptr = NULL;
                 PMIX_ERROR_LOG(rc);
@@ -145,45 +149,51 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
             }
         }
         else {
-            ctype = (pmix_collect_t)byte;
+            ctype = PMIX_GDS_COLLECT_IS_SET(blob_info_byte) ?
+                        PMIX_COLLECT_YES : PMIX_COLLECT_NO;
             have_ctype = true;
         }
 
-        /* unpack the size of uniq keys names in the map */
-        cnt = 1;
-        PMIX_BFROPS_UNPACK(rc, pmix_globals.mypeer,
-                           &bkt, &kmap_size, &cnt, PMIX_UINT32);
-        if (PMIX_ERR_UNPACK_READ_PAST_END_OF_BUFFER == rc) {
-            rc = PMIX_SUCCESS;
-            PMIX_DESTRUCT(&bkt);
-            break;
-        } else if (PMIX_SUCCESS != rc) {
-            PMIX_ERROR_LOG(rc);
-            PMIX_DESTRUCT(&bkt);
-            break;
-        }
+        /* determine the key-map existing flag */
+        kmap_type = PMIX_GDS_KEYMAP_IS_SET(blob_info_byte) ?
+                    PMIX_MODEX_KEY_KEYMAP_FMT : PMIX_MODEX_KEY_NATIVE_FMT;
+        if (PMIX_MODEX_KEY_KEYMAP_FMT == kmap_type) {
+            /* unpack the size of uniq keys names in the map */
+            cnt = 1;
+            PMIX_BFROPS_UNPACK(rc, pmix_globals.mypeer,
+                               &bkt, &kmap_size, &cnt, PMIX_UINT32);
+            if (PMIX_ERR_UNPACK_READ_PAST_END_OF_BUFFER == rc) {
+                rc = PMIX_SUCCESS;
+                PMIX_DESTRUCT(&bkt);
+                break;
+            } else if (PMIX_SUCCESS != rc) {
+                PMIX_ERROR_LOG(rc);
+                PMIX_DESTRUCT(&bkt);
+                break;
+            }
 
-        /* init and unpack key names map, the position of the key name
-         * in the array determines the unique key index */
-        kmap = (char**)(calloc(kmap_size + 1, sizeof(char*)));
-        if (NULL == kmap) {
-            rc = PMIX_ERR_OUT_OF_RESOURCE;
-            PMIX_ERROR_LOG(rc);
-            goto exit;
-        }
-        cnt = kmap_size;
-        PMIX_BFROPS_UNPACK(rc, pmix_globals.mypeer, &bkt,
-                           kmap, &cnt, PMIX_STRING);
-        if (PMIX_SUCCESS != rc) {
-            PMIX_ERROR_LOG(rc);
-            PMIX_DESTRUCT(&bkt);
-            goto exit;
-        }
-        if (pmix_argv_count(kmap) != (int)kmap_size) {
-            rc = PMIX_ERR_UNPACK_FAILURE;
-            PMIX_ERROR_LOG(rc);
-            PMIX_DESTRUCT(&bkt);
-            goto exit;
+            /* init and unpack key names map, the position of the key name
+             * in the array determines the unique key index */
+            kmap = (char**)(calloc(kmap_size + 1, sizeof(char*)));
+            if (NULL == kmap) {
+                rc = PMIX_ERR_OUT_OF_RESOURCE;
+                PMIX_ERROR_LOG(rc);
+                goto exit;
+            }
+            cnt = kmap_size;
+            PMIX_BFROPS_UNPACK(rc, pmix_globals.mypeer, &bkt,
+                               kmap, &cnt, PMIX_STRING);
+            if (PMIX_SUCCESS != rc) {
+                PMIX_ERROR_LOG(rc);
+                PMIX_DESTRUCT(&bkt);
+                goto exit;
+            }
+            if (pmix_argv_count(kmap) != (int)kmap_size) {
+                rc = PMIX_ERR_UNPACK_FAILURE;
+                PMIX_ERROR_LOG(rc);
+                PMIX_DESTRUCT(&bkt);
+                goto exit;
+            }
         }
 
         /* unpack the enclosed blobs from the various peers */
@@ -236,7 +246,7 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
 
             /* call a specific GDS function to storing
              * part of the process data */
-            rc = cb_fn(ctx, &proc, kmap, &pbkt);
+            rc = cb_fn(ctx, &proc, kmap_type, kmap, &pbkt);
             if (PMIX_SUCCESS != rc) {
                 PMIX_ERROR_LOG(rc);
                 pbkt.base_ptr = NULL;
@@ -288,31 +298,44 @@ exit:
  *
  * kv - pmix key-value pair
  */
-pmix_status_t pmix_gds_base_modex_pack_kval(char ***kmap, pmix_buffer_t *buf,
+pmix_status_t pmix_gds_base_modex_pack_kval(pmix_gds_modex_key_fmt_t key_fmt,
+                                            pmix_buffer_t *buf, char ***kmap,
                                             pmix_kval_t *kv)
 {
     uint32_t key_idx;
-    char *tmp;
     pmix_status_t rc = PMIX_SUCCESS;
 
-    tmp = strdup(kv->key);
-    rc = pmix_argv_append_unique_idx(&key_idx, kmap, tmp, 0);
-    if (PMIX_SUCCESS != rc) {
+    if (PMIX_MODEX_KEY_KEYMAP_FMT == key_fmt) {
+        rc = pmix_argv_append_unique_idx((int*)&key_idx, kmap, kv->key, 0);
+        if (PMIX_SUCCESS != rc) {
+            PMIX_ERROR_LOG(rc);
+            return rc;
+        }
+        /* pack key-index */
+        PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, buf, &key_idx, 1, PMIX_UINT32);
+        if (PMIX_SUCCESS != rc) {
+            PMIX_ERROR_LOG(rc);
+            return rc;
+        }
+        /* pack key-value */
+        PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, buf, kv->value, 1, PMIX_VALUE);
+        if (PMIX_SUCCESS != rc) {
+            PMIX_ERROR_LOG(rc);
+            return rc;
+        }
+    } else if (PMIX_MODEX_KEY_NATIVE_FMT == key_fmt) {
+        PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, buf, kv, 1, PMIX_KVAL);
+        if (PMIX_SUCCESS != rc) {
+            PMIX_ERROR_LOG(rc);
+            return rc;
+        }
+    } else {
+        rc = PMIX_ERR_BAD_PARAM;
         PMIX_ERROR_LOG(rc);
-        free(tmp);
         return rc;
     }
-    /* pack key-index */
-    PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, buf, &key_idx, 1, PMIX_UINT32);
-    if (PMIX_SUCCESS != rc) {
-        return rc;
-    }
-    /* pack key-value */
-    PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, buf, kv->value, 1, PMIX_VALUE);
-    if (PMIX_SUCCESS != rc) {
-        return rc;
-    }
-    return rc;
+
+    return PMIX_SUCCESS;
 }
 
 /*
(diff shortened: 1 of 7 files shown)


artpol84:
This PR is based on the following proposal: https://github.com/pmix/pmix/pull/1130#issuecomment-470944302.
As it is right now it provides minor improvement as it replaces 12B key name strings with the 8B `size_t` index. However once flexible packing will be introduced this will significantly reduce the modex size.
It was verified by temporally changing `size_t` to `uint8_t`. With this change, the size of the per-node contribution for 28 PPN was decreased from 2875B to 2601B (274B).
Considering that
* key names (12B) are replaced with uint8_t index (1B) so we save 11B per process.
* We still store key name once (in the key map), so we expect to see an effect of 27xkey-name-saving size reduction.

274B / 11B ~ 25 (while 27 is expected). The difference comes from additional overhead caused by key map packing. This will be addressed by flexible packing as well.

test/pmix_test.c
```diff
@@ -64,6 +64,9 @@ int main(int argc, char **argv)
         exit(1);
     }
 
+    int delay = 0;
```
artpol84: @karasevb please remove the debugging leftovers

karasevb:
bot:retest

src/mca/gds/base/gds_base_fns.c
```diff
@@ -210,24 +250,111 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
...
+            break;
+        }
+    }
+    /* add unique key-name to the map */
+    if (false == found) {
+        tmp = strdup(kv->key);
+        pmix_value_array_append_item(kmap, &tmp);
+    }
+    /* pack key-index */
+    PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, buf, &key_idx, 1, PMIX_SIZE);
```
karasevb: PMIX_INT32

src/server/pmix_server_ops.c
```diff
@@ -557,47 +584,70 @@ static pmix_status_t _collect_data(pmix_server_trkr_t *trk,
...
+                    PMIX_DESTRUCT(&rank_blobs);
                     goto cleanup;
                 }
                 rel_rank += pcs.rank;
 
                 /* pack the relative rank */
-                PMIX_CONSTRUCT(&pbkt, pmix_buffer_t);
-                PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, &pbkt,
+                pbkt = PMIX_NEW(pmix_buffer_t);
+                PMIX_CONSTRUCT(pbkt, pmix_buffer_t);
```
rhc54: This isn't needed - PMIX_NEW already ran the constructors

src/server/pmix_server_ops.c
```diff
@@ -557,47 +584,70 @@ static pmix_status_t _collect_data(pmix_server_trkr_t *trk,
...
             PMIX_DESTRUCT(&cb);
         }
+        /* pack node part of modex to `bucket` */
+        /* pack the key names map for the remote server can
+         * use it to match key names by index */
+        size = pmix_value_array_get_size(&kmap);
+        if (0 < size) {
+            PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, &bucket,
+                             &size, 1, PMIX_SIZE);
+            PMIX_BFROPS_PACK(rc, pmix_globals.mypeer, &bucket,
```
rhc54: I think this is risky - the value_array class doesn't initialize its memory. You might be getting lucky and were given a memory location with zero's in it, but that is highly system dependent. If you weren't, then the array will not be NULL terminated and this will crash

src/server/pmix_server_ops.c
```diff
@@ -611,6 +661,12 @@ static pmix_status_t _collect_data(pmix_server_trkr_t *trk,
 
   cleanup:
     PMIX_DESTRUCT(&bucket);
+    if (kmap.array_size > 0) {
+        for (key_idx = 0; key_idx < kmap.array_size; key_idx++) {
+            free((PMIX_VALUE_ARRAY_GET_BASE(&kmap, char*))[key_idx]);
```
rhc54: This is not a good idea - if you want to free the individual strings, then you should loop thru the array itself, not bypass the accessor functions.

karasevb: Thanks for remarks. 
Keymap has been reworked using `pmix_argv_x`.

src/mca/gds/base/gds_base_fns.c
```diff
@@ -127,23 +129,70 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
...
+        cnt = kmap_size;
+        PMIX_BFROPS_UNPACK(rc, pmix_globals.mypeer, &bkt,
+                           kmap, &cnt, PMIX_STRING);
+        if (PMIX_SUCCESS != rc) {
+            PMIX_ERROR_LOG(rc);
+            PMIX_DESTRUCT(&bkt);
+            /* do not use `pmix_argv_free` here since `kmap`
+             * allocated preventively with the supposed size
+             * so pmix_argv doesn't know the size of the array */
+            free(kmap);
```
rhc54: Actually, since you allocated kmap with "calloc", you can and should use pmix_argv_free here. It will simply free all strings in the array until it hits NULL, which calloc guarantees by having initialized all kmap entries to zero.

src/mca/gds/base/gds_base_fns.c
```diff
@@ -127,23 +129,70 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
...
+            kmap = NULL;
+            goto exit;
+        }
+        if (pmix_argv_count(kmap) != (int)kmap_size) {
+            rc = PMIX_ERR_UNPACK_FAILURE;
+            PMIX_ERROR_LOG(rc);
+            PMIX_DESTRUCT(&bkt);
+            free(kmap);
```
rhc54: same here

src/mca/gds/base/gds_base_fns.c
```diff
@@ -210,24 +259,120 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
...
                 buff, &bo, &cnt, PMIX_BYTE_OBJECT);
     }
+
     if (PMIX_ERR_UNPACK_READ_PAST_END_OF_BUFFER == rc) {
         rc = PMIX_SUCCESS;
+    } else if (PMIX_SUCCESS != rc) {
+        PMIX_ERROR_LOG(rc);
+    }
+exit:
+    if (pmix_argv_count(kmap) > 0) {
```
rhc54: You don't need the extra call here - if kmap != NULL will suffice and save a little time

src/mca/gds/base/gds_base_fns.c
```diff
@@ -210,24 +259,120 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
...
+pmix_status_t pmix_gds_base_modex_pack_kval(char ***kmap, pmix_buffer_t *buf,
+                                            pmix_kval_t *kv)
+{
+    bool found = false;
+    uint32_t key_idx;
+    char *tmp;
+    pmix_status_t rc = PMIX_SUCCESS;
+    int size = pmix_argv_count(*kmap);
+
+    /* determine the uniq index of key */
```
rhc54: We have a pmix_argv_append_unique_nosize function that will do this for you. Only problem is that it returns "success" instead of the index where it put or found the string. We wouldn't want to modify that function as it gets used in various places, but maybe we should add another one (e.g., pmix_argv_append_unique_idx) that does the same logic but returns the index of the stored location?

jjhursey:
(let's give this new CI a try on a real branch)
bot:ibm:xver:retest

artpol84:
Do We have an existing use case for that?
How many keys they exchange?
Do they care about modex time?

artpol84:
> It has been common practice (at least, until now) to encode unique keys per process - typically with the process rank, but often with other information. In such use-cases, this optimization actually hurts, not helps.

This is an unoptimal practice for PMIx that has "rank" argument in the `PMIx_Get` operation. If the implementation keeps following that practice it doesn't care about performance and we should not try to optimize for it.
And it is trivial to change that implementation to support an optimal way of naming the keys.

Instead of optimizing for unoptimal cases we should provide the right description that emphasize that key name selection affects modex size..

artpol84:
Also even with this approach (of different key names) I don't see much of an issue with the proposed approach. With flexible packing we will get increase of the per-node contribution for just 1B per key (in extreme cases where you have many keys it'll be 2B) for the following reasons:
* Key names have to be in the modex anyhow so moving them into the dictionary doesn't change that.
* Only thing we are adding is the key index in the per-rank info blob. This index will be compressed with the flexible packing and (if the app doesn't submit too many keys) will occupy 1B per key.

ibm-ompi:

 Gist: https://gist.github.com/2f833c9a441874b69cab65d680af411d

src/mca/gds/base/gds_base_fns.c
karasevb: `pmix_argv_append_unique_idx` function added in the separate commit and updated to use here.

src/mca/gds/base/gds_base_fns.c
karasevb: Fixed

src/mca/gds/base/gds_base_fns.c
karasevb: Fixed

src/mca/gds/base/gds_base_fns.c
karasevb: Fixed

ibm-ompi:

 Gist: https://gist.github.com/0cabed5bcc30343f6cbfb1e6c26e3997

rhc54:
@karasevb Thanks for the fixes and for implementing the new argv function! I honestly wasn't implying that you should do that implementation - was just noting that it might prove useful. Much appreciated.

rhc54:
@artpol84 I think you might be looking at this from too limited a view. The point here is that PMIx is becoming more broadly adopted, both in terms of programming models/libraries and environments. Regardless of how we might personally view their methods, we need to respect their right to decide what is best for their overall environment. What we may perceive as obviously beneficial may not be viewed the same by others. This, unfortunately, means programming flexibility into the PMIx library.

We shouldn't impose a fixed code path that others might find contrary to their perceived benefit if there are ways of avoiding it. In this case, it isn't that hard to make creation of a key table something optional. Not every use-case is going to be willing to change their coding practice to push non-unique keys, not everyone is going to use the flexible integer mechanism, nor can we assume that only a limited number of keys will be "pushed". These are all assumptions that may not be true.

I know it takes extra effort - but I fear that is the cost of a general purpose community project. As we have found in other projects such as OMPI, we sometimes have to step back and add flexibility to support the broader community.

rhc54:
@karasevb Looks like this will need to be stashed in a component somewhere outside the base so it doesn't break cross-version compatibility. You might consider adding it as another API to the psquash framework.

artpol84:
@rhc54 I don’t see need in a yet one more component here.
There is no problem in automatic calculation of whether or not key map will be beneficial.
It can be calculated on a per-node basis and included as a bit field in the data collection byte.

artpol84:
> @karasevb Looks like this will need to be stashed in a component somewhere outside the base so it doesn't break cross-version compatibility

Can you clarify why you think that being in the base it breaks cross version compat?
The reason of CI failure needs to be investigated first.
Currently I don’t see architectural issues with it being in the base component.

This is a new CI and we need to make sure it is stable.

rhc54:
Given that Josh had tested it on a "neutral" PR and all passed, it seems that there is something about this PR that causes a problem. Looking at the code, the most likely reason is that this exists in the base where it will impact all code paths regardless of the version of the client or server. Putting it in a component may or may not be the solution - I was only offering some advice on a possible path forward. Please dial down the sensitivity.

artpol84:
> Putting it in a component may or may not be the solution - I was only offering some advice on a possible path forward. Please dial down the sensitivity.

I was just trying to see if you see a particular reason for the issue.

karasevb:
@jjhursey looking at the details of the output from cross-version seems something wrong with CI:
```From https://github.com/pmix/pmix-tests
   1c456e7..0932f2f  master     -> origin/master
error: Your local changes to the following files would be overwritten by merge:
	crossversion/xversion.py
Please, commit your changes or stash them before you can merge.
Aborting
Updating 1c456e7..0932f2f
```
Also my local cross-version test has not issues.

artpol84:
Thanks, @karasevb!

artpol84:
> There is no problem in automatic calculation of whether or not key map will be beneficial.
It can be calculated on a per-node basis and included as a bit field in the data collection byte.

@rhc54 I believe this suggestion should address your concerns in a flexible manner without unnneded introduction of a new framework.
It will be fairly easy to implement on top of this PR.

jjhursey:
I'm looking at the cross version CI issue now - it was not related to this PR, but in the structure of the self-updating mechanism inside the container. Should have it back up shortly.

jjhursey:
bot:ibm:xver:retest

jjhursey:
Cross version CI is fixed now (and passing on this PR). Sorry about the interruption.

karasevb:
@jjhursey Thanks!

jjhursey:
> This is an unoptimal practice for PMIx that has "rank" argument in the `PMIx_Get` operation. If the implementation keeps following that practice it doesn't care about performance and we should not try to optimize for it.
> And it is trivial to change that implementation to support an optimal way of naming the keys.

I don't fully agree with this statement.

Yes, it may be better for some (most?) use cases if each rank stores the same key name for each entry. In which case this PR will optimize the packing by reducing the redundant key information. However, there may be (are likely) other uses cases where they don't want to or can't make such a change. We shouldn't penalize those existing users.

For example, Rank X might store a key intended only for Rank Y (or a group of ranks identified by, say, some integer - a group id). PMIx allows Rank Y to ask for a key from Rank X via then rank argument, but it needs to further specify that it needs the 'key from Rank X for Rank/Group Y' which is encoded in the key name. In this scenario, this PR will increase the size of the modex over the existing practice.

So I want this optimization, but also the ability to not penalize the use case above. That's why I like this suggestion:

> There is no problem in automatic calculation of whether or not key map will be beneficial.
It can be calculated on a per-node basis and included as a bit field in the data collection byte.

So we perform the optimization only when we know it will help and not hurt. I think that is a good path forward for this PR.

Extra bonus points if there is a way to force with/without the optimization via an MCA parameter to reduce the calculate+check overhead for those that know they fall into one camp or the other.

jjhursey:
(Thinking a bit more over lunch) Since this might be something the application would programmatically want to decide ("I'm running with feature X, so I'll need to turn off/on the optimization") may be an attribute is a better approach than an MCA parameter.

artpol84:
Yeah, @karasevb will work on it tomorrow. To give some clarity on what I mean. 
We will extend the key map structure with the `key_count` field that will count a number of key occurrences:

| idx | key_name | key_count    | 
|-----| ---------- | ------------- |
| 0    | key1        | 2                  |
|1     | key2        | 1                  |

At the packing stepwe will calculate the precise amount of space each packing will occupy:
For the original approach (embedding key names):
```
(packed_size(key name)) * key_count
```
For the new approach:
```
(packed_size(key name)) + packed_size(idx) * key_count
```
SUM of those values by all the key map will represent the contribution of each method to the final size.

> Extra bonus points if there is a way to force with/without the optimization via an MCA parameter to reduce the calculate+check overhead for those that know they fall into one camp or the other.

Sure we can do that. Though I don't expect the overhead of this to be high.

artpol84:
>(Thinking a bit more over lunch) Since this might be something the application would programmatically want to decide ("I'm running with feature X, so I'll need to turn off/on the optimization") may be an attribute is a better approach than an MCA parameter.

Eather way works. Boris will implement the general ability to have alternative packing schemes and we can later see how we can force one or another if needed.

karasevb:
Added a new commit that implements selecting between old format and new (kmap) by evaluating the most effective size.
<details>
<summary>patch for debug output</summary>

```patch

diff --git a/src/mca/gds/base/gds_base_fns.c b/src/mca/gds/base/gds_base_fns.c
index 7a71967..d4d5f01 100644
--- a/src/mca/gds/base/gds_base_fns.c
+++ b/src/mca/gds/base/gds_base_fns.c
@@ -157,6 +157,11 @@ pmix_status_t pmix_gds_base_store_modex(struct pmix_namespace_t *nspace,
         /* determine the key-map existing flag */
         kmap_type = PMIX_GDS_KEYMAP_IS_SET(blob_info_byte) ?
                     PMIX_MODEX_KEY_KEYMAP_FMT : PMIX_MODEX_KEY_NATIVE_FMT;
+
+        pmix_output(0, "unpack key format: %s",
+                    kmap_type == PMIX_MODEX_KEY_KEYMAP_FMT ?
+                        "kmap" : "native");
+
         if (PMIX_MODEX_KEY_KEYMAP_FMT == kmap_type) {
             /* unpack the size of uniq keys names in the map */
             cnt = 1;
diff --git a/src/server/pmix_server_ops.c b/src/server/pmix_server_ops.c
index f73bc01..d830c37 100644
--- a/src/server/pmix_server_ops.c
+++ b/src/server/pmix_server_ops.c
@@ -631,6 +631,12 @@ static pmix_status_t _collect_data(pmix_server_trkr_t *trk,
                                 "key packing type %s",
                                 kmap_type == PMIX_MODEX_KEY_KEYMAP_FMT ?
                                     "kmap" : "native");
+
+            pmix_output(0, "kmap size: %lu, native size: %lu, key pack format: %s",
+                        key_fmt_size[PMIX_MODEX_KEY_KEYMAP_FMT],
+                        key_fmt_size[PMIX_MODEX_KEY_NATIVE_FMT],
+                        kmap_type == PMIX_MODEX_KEY_KEYMAP_FMT ?
+                            "kmap" : "native");
         }
         PMIX_CONSTRUCT(&rank_blobs, pmix_list_t);
```
</details>

Here is the test output (used attached patch), where 2 pmix_servers is started and they divide among himself 5 procs. The server (pid: 08052) that has 2 procs is selecting the old format for packing key names since most effective size is 46 for that format. The second server (pid: 08033) with 3 procs selects the “kmap” format, because it is more efficient (59 vs. 69 for old format).
```bash
user@blaster ~/pmix/master/test $ ./pmix_test -s 2 -n 5 --job-fence -c
[blaster:08052] kmap size: 47, native size: 46, key pack format: native
[blaster:08033] kmap size: 59, native size: 69, key pack format: kmap
[blaster:08033] unpack key format: native
[blaster:08052] unpack key format: native
[blaster:08052] unpack key format: kmap
[blaster:08033] unpack key format: kmap
==8033== test_server.c:server_finalize: Test finished OK!
```

jjhursey:
Per teleconf (March 21):
 * The automatic selection of whether or not to choose this optimization is in place [here](https://github.com/pmix/pmix/pull/1149/files#diff-cf506bd64a1184a18c5ad19dfdb8913dR627)
    - So if the `keymap` optimization would have actually made the packing larger, then the `native` (current) format is chosen. If it helps then the new `keymap` is selected.
    - As a result, this PR should never hurt 
 * There was a question in the conversation about (1) if we need to be able to force one mode or the other, and (2) how best to do so (MCA parameter or PMIx attribute).
   - I have convinced myself that an MCA parameter is probably not the best option since the app may want to decide programmatically and not at init time.
   - It was raised that we can add that attribute (or another mechanism) in a second PR with a dedicated discussion there. This seemed to be a fine path forward.
 * In response to @rhc54's concern about this optimization being too focused on OMPI, I asked @artpol84 to reach out to the MPICH team to see how this optimization might play with their usage model - as another consumer of the PMIx interface.
   - In particular, I'd like to know, before we merge, if they have any concerns based on their usage about the optimization given the automatic switch that @karasevb put in. If there are slight adjustments we can make then I would like to do it now, if possible.
   - I think that once we have feedback from MPICH then @rhc54's review concern may be addressed with that feedback and the addition of the automatic switch.
   - Any other consumers of PMIx that might be impacted by this should feel free to chime in on this PR.
 * Once we get feedback then we can review/merge. Target is to try to resolve this PR by next Thursday.

artpol84:
@raffenet
Checking the current usage of PMIx in MPICH I see that you are encoding a rank number in the key name:
https://github.com/pmodels/mpich/blob/54c1325a88568c8a55239cece35156b271812814/src/mpid/common/bc/mpidu_bc.c#L165

So we would like to clarify this question with you before merging this.
This PR provides an optimization to the PMIx in case every rank submits the key with the same names. Because `PMIx_Get` already has ([link](https://github.com/pmix/pmix/blob/master/include/pmix.h#L203:L205)) rank identification as part of the argument list (proc):
```C
PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc, const char key[],
                                   const pmix_info_t info[], size_t ninfo,
                                   pmix_value_t **val);
```
Additional identification using a key name is not required anymore.
Avoiding this allows to significantly reduce per-node contribution after this PR is merged.
So basically my recomendation would be to change the key name from "bc-<rank>" to be just "bc" and request this key with the proper `proc` argument.

If, however, you'd like to stick to the current implementation the final commit 8cfbac9eae45185e09962914b898ef6c1abbbf74 that @karasevb provided will detect that all key names are different and proposed optimizations hurts the footprint. In this case the old way of encoding the modex message will be used.
So in any case you should not be negatively affected by this, and can benefit if will decide to introduce the slight change I described above.

artpol84:
Also I see that you have a `root_only` where only one rank per node submits information. In this case what is proposed in this PR will not be used (based on automatic calculation of the sutability of the approach, logic can be found here: 8cfbac9eae45185e09962914b898ef6c1abbbf74)

raffenet:
> Additional identification using a key name is not required anymore.
> Avoiding this allows to significantly reduce per-node contribution after this PR is merged.
> So basically my recomendation would be to change the key name from "bc-" to be just "bc" and request this key with the proper `proc` argument.

I don't see why we can't do this. I'll create a PR for MPICH to remove the rank from the string and test it out.

artpol84:
@raffenet Thanks!
@jjhursey I think we have the feedback we were looking for.

src/server/pmix_server_ops.c
```diff
@@ -506,32 +522,122 @@ static void fence_timeout(int sd, short args, void *cbdata)
...
+                        kname_size * key_count[i];
+                key_fmt_size[PMIX_MODEX_KEY_KEYMAP_FMT] =
+                        kname_size + key_count[i]*kidx_size;
+            }
+            PMIX_RELEASE(key_count_array);
+
+            /* select the most efficient key-name pack format */
+            kmap_type = key_fmt_size[PMIX_MODEX_KEY_NATIVE_FMT] >
+                        key_fmt_size[PMIX_MODEX_KEY_KEYMAP_FMT] ?
+                        PMIX_MODEX_KEY_KEYMAP_FMT : PMIX_MODEX_KEY_NATIVE_FMT;
```
jjhursey: Thanks for adding the automatic switch here!

raffenet:
> > Additional identification using a key name is not required anymore.
> > Avoiding this allows to significantly reduce per-node contribution after this PR is merged.
> > So basically my recomendation would be to change the key name from "bc-" to be just "bc" and request this key with the proper `proc` argument.
> 
> I don't see why we can't do this. I'll create a PR for MPICH to remove the rank from the string and test it out.

Removed rank from key string in https://github.com/pmodels/mpich/commit/566d0f7ae040adb815372b28578775661c17e567.

artpol84:
Thanks, @raffenet! That was fast!

artpol84:

PR #1152, opened 2019-03-15
Flexible integer packing with `psquash` framework in `bfrops` v4

### LINKED ISSUE #681: Modex size analysis

karasevb:
- Introduced the new `psquash` framework (Thanks @jjhursey: https://github.com/pmix/pmix/issues/681#issuecomment-472149066)
- Added flexible packing for `psquash:flex128` component (Thanks @artpol84: https://github.com/pmix/pmix/pull/1130#issuecomment-470141719)
- `psquash:native`  component implements the native `bfrops` integer packing functions.
- `bfrops` v4 uses `psquash` framework for integer packing.

Refs #1130 
Closes #681

### DIFF
--- a/src/mca/psquash/base/base.h
+++ b/src/mca/psquash/base/base.h
@@ -1,0 +1,35 @@
+/* -*- C -*-
+ *
+ * Copyright (c) 2019      IBM Corporation.  All rights reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ *
+ */
+#ifndef PMIX_PSQUASH_BASE_H_
+#define PMIX_PSQUASH_BASE_H_
+
+#include <src/include/pmix_config.h>
+
+#ifdef HAVE_STRING_H
+#include <string.h>
+#endif
+
+#include "src/class/pmix_pointer_array.h"
+#include "src/mca/mca.h"
+#include "src/mca/base/pmix_mca_base_framework.h"
+
+#include "src/mca/psquash/psquash.h"
+
+
+BEGIN_C_DECLS
+
+PMIX_EXPORT extern pmix_mca_base_framework_t pmix_psquash_base_framework;
+
+PMIX_EXPORT pmix_status_t pmix_psquash_base_select(void);
+
+END_C_DECLS
+
+#endif
--- a/src/mca/psquash/base/base.h
+++ b/src/mca/psquash/base/base.h
@@ -1,6 +1,9 @@
 /* -*- C -*-
  *
  * Copyright (c) 2019      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2019      Mellanox Technologies, Inc.
+ *                         All rights reserved.
+ *
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -25,6 +28,41 @@
 
 
 BEGIN_C_DECLS
+
+/**
+ * Sizeof by PMIx type integer values.
+ *
+ * r - return status code
+ * t - type (pmix_data_type_t) of integer value
+ * s - size of type in bytes
+ * (see a comment to `pmix_bfrops_pack_flex` for additional details)
+ */
+#define PMIX_SQUASH_TYPE_SIZEOF(r, t, s)                    \
+do {                                                        \
+    (r) = PMIX_SUCCESS;                                     \
+    switch (t) {                                            \
+        case PMIX_INT16:                                    \
+        case PMIX_UINT16:                                   \
+            (s) = SIZEOF_SHORT;                             \
+            break;                                          \
+        case PMIX_INT:                                      \
+        case PMIX_INT32:                                    \
+        case PMIX_UINT:                                     \
+        case PMIX_UINT32:                                   \
+            (s) = SIZEOF_INT;                               \
+            break;                                          \
+        case PMIX_INT64:                                    \
+        case PMIX_UINT64:                                   \
+            (s) = SIZEOF_LONG;                              \
+            break;                                          \
+        case PMIX_SIZE:                                     \
+            (s) = SIZEOF_SIZE_T;                            \
+            break;                                          \
+        default:                                            \
+            (r) = PMIX_ERR_BAD_PARAM;                       \
+    }                                                       \
+} while (0)
+
 
 PMIX_EXPORT extern pmix_mca_base_framework_t pmix_psquash_base_framework;
 
--- a/src/mca/psquash/psquash.h
+++ b/src/mca/psquash/psquash.h
@@ -1,6 +1,8 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2019      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2019      Mellanox Technologies, Inc.
+ *                         All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -42,11 +44,13 @@ typedef pmix_status_t (*pmix_psquash_base_module_init_fn_t)(void);
 typedef void (*pmix_psquash_base_module_finalize_fn_t)(void);
 
 /**
- * Maximum size of the buffer passed to encode/decode for integers.
+ *  Maximum size of the type.
  *
- * Define a constant max here to avoid dynamic memory allocation in the call path.
+ * type - Type (PMIX_SIZE, PMIX_INT to PMIX_UINT64)
+ * size - size of the type
  */
-#define PMIX_PSQUASH_INT_MAX_BUF_SIZE (SIZEOF_SIZE_T+1)
+typedef pmix_status_t (*pmix_psquash_get_max_size_fn_t) (pmix_data_type_t type,
+                                                         size_t *size);
 
 /**
  * Encode a basic integer type into a contiguous destination buffer.
@@ -54,12 +58,12 @@ typedef void (*pmix_psquash_base_module_finalize_fn_t)(void);
  * type     - Type of the 'src' pointer (PMIX_SIZE, PMIX_INT to PMIX_UINT64)
  * src      - pointer to a single basic integer type
  * dest     - pointer to buffer to store data
- * dest_len - length, in bytes, of the dest buffer
+ * dst_len  - pointer to the packed size of dest, in bytes
  */
+
 typedef pmix_status_t (*pmix_psquash_encode_int_fn_t) (pmix_data_type_t type,
-                                                       void *src,
-                                                       uint8_t dest[PMIX_PSQUASH_INT_MAX_BUF_SIZE],
-                                                       uint8_t *dest_len);
+                                                       void *src, void *dest,
+                                                       size_t *dst_len);
 
 /**
  * Decode a basic a contiguous destination buffer into a basic integer type.
@@ -68,11 +72,11 @@ typedef pmix_status_t (*pmix_psquash_encode_int_fn_t) (pmix_data_type_t type,
  * src      - pointer to buffer where data was stored
  * src_len  - length, in bytes, of the src buffer
  * dest     - pointer to a single basic integer type
+ * dst_len  - pointer to the unpacked size of dest, in bytes
  */
 typedef pmix_status_t (*pmix_psquash_decode_int_fn_t) (pmix_data_type_t type,
-                                                       uint8_t src[PMIX_PSQUASH_INT_MAX_BUF_SIZE],
-                                                       uint8_t src_len,
-                                                       void *dest);
+                                                       void *src, size_t src_len,
+                                                       void *dest, size_t *dst_len);
 
 /**
  * Base structure for a PSQUASH module
@@ -83,6 +87,8 @@ typedef struct {
     /** init/finalize */
     pmix_psquash_base_module_init_fn_t     init;
     pmix_psquash_base_module_finalize_fn_t finalize;
+
+    pmix_psquash_get_max_size_fn_t         get_max_size;
 
     /** Integer compression */
     pmix_psquash_encode_int_fn_t           encode_int;
--- a/src/mca/psquash/psquash.h
+++ b/src/mca/psquash/psquash.h
@@ -83,6 +83,8 @@ typedef pmix_status_t (*pmix_psquash_decode_int_fn_t) (pmix_data_type_t type,
  */
 typedef struct {
     const char *name;
+    /* flag of necessity to pack the type of value */
+    bool int_type_is_encoded;
 
     /** init/finalize */
     pmix_psquash_base_module_init_fn_t     init;
--- a/src/mca/psquash/psquash.h
+++ b/src/mca/psquash/psquash.h
@@ -1,0 +1,112 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
+/*
+ * Copyright (c) 2019      IBM Corporation.  All rights reserved.
+ * $COPYRIGHT$
+ *
+ * Additional copyrights may follow
+ *
+ * $HEADER$
+ */
+
+/**
+ * @file
+ *
+ * This interface is for the encoding/decoding of basic types and the
+ * compression/decompression of larger blobs of data (i.e., modex).
+ *
+ * Available plugins may be defined at runtime via the typical MCA parameter
+ * syntax.
+ */
+
+#ifndef PMIX_PSQUASH_H
+#define PMIX_PSQUASH_H
+
+#include <src/include/pmix_config.h>
+
+#include "src/mca/mca.h"
+#include "src/mca/base/pmix_mca_base_var.h"
+#include "src/mca/base/pmix_mca_base_framework.h"
+
+BEGIN_C_DECLS
+
+/******    MODULE DEFINITION    ******/
+
+/**
+ * Initialize the module
+ */
+typedef pmix_status_t (*pmix_psquash_base_module_init_fn_t)(void);
+
+/**
+ * Finalize the module
+ */
+typedef void (*pmix_psquash_base_module_finalize_fn_t)(void);
+
+/**
+ * Maximum size of the buffer passed to encode/decode for integers.
+ *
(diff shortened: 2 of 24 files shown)


karasevb:
bot:retest

rhc54:
Sorry about the errant review - it was posted in the incorrect window 😦

artpol84:
@rhc5;
I’m not sure I understand how flex256 hurts here. 
One of the important use of it is to eliminate extra 0s in the service information. We do not intend to touch user data.
So it is not clear to me why you need to disable it for some jobs.
Can you elaborate?

artpol84:
To re-iterate: UCX endpoint is a blob and all we need from this PR is to reduce the amount of service information that is going to be the same for all jobs/namespaces.

rhc54:
I'm afraid I don't know what you mean by "service information". The only thing in OMPI's modex is endpoint info. Can you clarify?

Either way, I think you may have missed my point. We need to step away from thinking of "modex" in terms of how OMPI uses it, or even how MPI does. Other programming models may well exchange information of a completely different nature than what we do. We need to respect that possibility and provide flexibility where possible.

In this case, as with the modex key-table, it isn't that hard for us to do so. It means we need a multi-select framework (which we have multiple examples of already in PMIx) and an attribute indicating which component this particular job/nspace wishes to use. This is pretty simple to provide and leaves open the option for other environments to still use PMIx according to their own preferences.

rhc54:
Actually, I think we can implement this even more simply by utilizing the PMIX_COLLECTIVE_ALGO attribute. You could pass that into the specific fence operation, giving it a value that indicates you want flex256 encoding for this particular fence. This way you could contain tracking of the psquash component to the fence tracker object instead of having to include it in the client connection protocol.

artpol84:
> I'm afraid I don't know what you mean by "service information". The only thing in OMPI's modex is endpoint info. Can you clarify?

Service information is the meta information that bfrops is introducing to describe user data.
I think you are missing what I’m trying to say as well.

Can you provide an exact concern about flexible packing? In what cases it will not work?

artpol84:
> Other programming models may well exchange information of a completely different nature than what we do. We need to respect that possibility and provide flexibility where possible.

Bfrops meta information will be the same regardless.

artpol84:
Flexible encoding does not break things so your concern about possible ephemeral programming model seems to be to weak to enforce additional amount of work that will be on us.
As we don’t break anything and benefit existing use cases I think we can defer solving this to a time when the modex size problem will be encountered (if it ever will, though I doubt that)

rhc54:
Okay, you are now starting to finally communicate what I originally asked: what are you trying to accomplish. It now appears that it is not reduction of the modex data itself, but instead compaction of the bfrops description information. In the case of an optimized system, that consists of the number of items being stored for each entry - the data type is not stored in such cases. So we apparently are talking about a different way to encode an int32 that describes how many elements of this type were stored. Is that what you are proposing here?

It would have been helpful to have that clarified earlier in this thread, but that is now a separate issue. I'll give some thought as to what that might mean now that we have some clarity as to the target. I'm puzzled by your last comment, however - if the modex size problem doesn't yet exist, then why are we even talking about this?

Please understand: you may not recognize or understand uses beyond OMPI - but that doesn't mean they don't exist or deserve equal respect. Checking the fence call to see if a collective algo was specified and then using the plugin it directs is a pretty minimal change. Sadly, we are likely to spend more time arguing about it than implementation requires. Going forward, we need to become more sensitive to hard-coding OMPI optimizations and recognize that other models/uses need to be considered, even if it means extra work to ensure adequate flexibility exists.

artpol84:
> Okay, you are now starting to finally communicate what I originally asked: what are you trying to accomplish. It now appears that it is not reduction of the modex data itself, but instead compaction of the bfrops description information. In the case of an optimized system, that consists of the number of items being stored for each entry - the data type is not stored in such cases. So we apparently are talking about a different way to encode an int32 that describes how many elements of this type were stored. Is that what you are proposing here?

This is very disappointing to hear that "I'm only now starting to communicate". I was explaining this in a very detailed level for a couple of months. Please see my main (most detailed comments).
* https://github.com/pmix/pmix/issues/681#issuecomment-367920982
* https://github.com/pmix/pmix/issues/681#issuecomment-368249149
* https://github.com/pmix/pmix/issues/681#issuecomment-455986873
* https://github.com/pmix/pmix/pull/1130#issuecomment-470339443

It is hard to build an efficient discussion when the other side is not carefully following the responses.
Maybe it is better to discuss this over the phone.

> Checking the fence call to see if a collective algo was specified and then using the plugin it directs is a pretty minimal change. Sadly, we are likely to spend more time arguing about it than implementation requires. Going forward, we need to become more sensitive to hard-coding OMPI optimizations and recognize that other models/uses need to be considered, even if it means extra work to ensure adequate flexibility exists.

I disagree that this is a minor thing. This has an implication on the usage of this feature. Basically, as you describe, it will be disabled by default. While I believe that in most common cases it will be beneficial.

jjhursey:
It is unfortunate that we all have been talking past each other on this topic. I must admit that it did take me a while to understand that this is focused on the bfrops data structure surrounding each rank's modex blob and not the blob itself. I'm happy that we have reached an understanding of the intention of this PR.

Now that I'm seeing the code change and the discussion around it, I'm wondering (again) if a framework is really necessary for this if it is only going to be used by the bfops framework. I still want the ability to enable/disable this at runtime. But could that best be done with an MCA parameter instead of a full framework?

I'm not suggesting another implementation - you all have put in a lot of effort into two versions of this already. Just a bit more discussion. We should definitely talk about this PR and PR #1149 on the teleconf this week.

I'd have to think if we want to enable/disable this per-namespace or not. Since it's hidden from the end user I'm leaning towards not - leaving it globally on or off.

artpol84:
@jjhursey note, though we are targeting service information, this change affects all bfrop packing functionality.
This means that if a user is using PMIx data types he will be influenced by this change as well (which is not necessarily bad IMO).

@rhc54's point was that flex packing doesn't help if user provides a blob. This is not correct because bfrops service information is still being compressed.

jjhursey:
Per teleconf (March 21):
 * Adding the framework did help make the code cleaner/clearer so it is good to keep (vs an MCA parameter), and the framework clearly separates this activity from that of the broader bfrops framework activity. So from a code clarity stance, a framework here is a good thing to do.
 * We can merge this in without the compression extension to the `psquash` framework. @jjhursey will investigate that as a separate activity - no need to hold up this PR.
 * @rhc54 's earlier review seems to have been addressed by the later conversation clarifying that this optimization focuses on the data structure around the modex (and maybe the modex content if it uses integers). I've dismissed the earlier review to prevent holding this change for that reason. If there are other reasons to hold this PR please re-review.
 * The WIP flag is still on this PR - @karasevb are there more changes needed for this PR before final review?
 * If there are no objections to this PR then I think we can do the final review and, barring any issues, merge to `master`.

src/mca/psquash/base/base.h
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Sizeof by PMIx type integer values.
+ *
+ * r - return status code
+ * t - type (pmix_data_type_t) of integer value
+ * s - size of type in bytes
+ * (see a comment to `pmix_bfrops_pack_flex` for additional details)
+ */
+#define PMIX_SQUASH_TYPE_SIZEOF(r, t, s)                    \
```
jjhursey: Question: Should this macro be part of `bfrops` instead of `psquash`?

I originally put it here since the flex component was the primary user, but looking at it now I think it might be better associated with `bfrops`. I'm fine either way.

src/mca/psquash/psquash.h
```diff
@@ -0,0 +1,123 @@
...
+typedef pmix_status_t (*pmix_psquash_decode_int_fn_t) (pmix_data_type_t type,
+                                                       void *src, size_t src_len,
+                                                       void *dest, size_t *dst_len);
+
+/**
+ * Base structure for a PSQUASH module
+ */
+typedef struct {
+    const char *name;
+    /* flag of necessity to pack the type of value */
```
jjhursey: I don't understand what this flag is communicating. From the bfrops code I _think_ it means this.
```suggestion
    /* flag indicating if the type is encoded within the value, otherwise, it is necessary to further pack the type with the value. */
```
Is this rewording accurate?

If so then `flex128` has `int_type_is_encoded=false` and `native` has it set to `true`. And it's only when it's `true` that the `bfrops` component encodes the type. That seems to be the opposite of the meaning above.

I would expect `flex128` to set `int_type_is_encoded=true` (since the type **is** encoded in the value), and `native` to set ``int_type_is_encoded=false` (since the type **is not** encoded in the value).

src/mca/psquash/psquash.h
```diff
@@ -0,0 +1,123 @@
...
+    pmix_psquash_base_module_finalize_fn_t finalize;
+
+    pmix_psquash_get_max_size_fn_t         get_max_size;
+
+    /** Integer compression */
+    pmix_psquash_encode_int_fn_t           encode_int;
+    pmix_psquash_decode_int_fn_t           decode_int;
+
+    /** Blob compression (i.e., modex) */
+    // JJH Todo
```
jjhursey: You can remove this todo comment and the one above (lines 99 and 100). I'll follow up on it in a later PR.

src/mca/psquash/psquash.h
artpol84: @karasevb I agree with @jjhursey.
the name is confusing. Lets fix it as proposed.

src/mca/psquash/psquash.h
karasevb: Done with the latest force-pushed commit

src/mca/psquash/base/base.h
karasevb: I moved this macro because it is used only for `psquash` and is not used for `bfrops`, so there is no need to keep it in `bfrops`.

karasevb:
Ready to go.

src/mca/psquash/base/base.h
jjhursey: That makes sense. Thanks

artpol84:

artpol84:
@rhc54 @jjhursey 
I recently discovered that flex128 packing is a well-known practice: 
https://en.m.wikipedia.org/wiki/Variable-length_quantity
In particular, our exact implementation matched Google Buffers zigzag encoding idea:
https://developers.google.com/protocol-buffers/docs/encoding#varints

This should not be a problem as it is BSD-licensed:
https://opensource.google.com/projects/protobuf

artpol84:
Not surprising as the idea was quite simple.

artpol84:
But the point is that as this encoding is widely used - it is safe to use it in PMIx

PR #1156, opened 2019-03-21
gds/ds21: update the pointer of `seg_hdr` when segment re-attached

karasevb:
When re-attaching the lock-segment, the `seg_hdr` pointer was not
updated, which caused to reference on incorrect data and damage.

Signed-off-by: Boris Karasev <karasev.b@gmail.com>

Issue #1114, opened 2019-02-11
Strange occasional segfault in pmix_gds_ds21_lock_init


akesandgren:
Thank you for taking the time to submit an issue!

### What version of the PMIx Reference Library are you using? (e.g., v1.0, v2.1, git master @ hash, etc.)

PMIx 2.2.1 release

### Describe how PMIx was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

From source

### Please describe the system on which you are running

* Operating system/version: Ubuntu 16.04
* Computer hardware: AMD Bulldozer
* Network type: Mellanox ConnectX-2

-----------------------------

## Details of the problem

I am seeing intermittent segfaults in PMIx v2.2.1 in gds_ds21_lock_pthread.c on line 239 according to gdb.
This is strange since both lock_item and seg_hdr look sane when inspecting the core file.
Line in question is:
```
lock_item->num_locks = seg_hdr->num_locks;
```

OpenMPI used is the 3.1.1 release.

The stacktrace is:
```
#0  0x00002b6cebfe193c in pmix_gds_ds21_lock_init (ctx=0x30ac268, 
    base_path=0x30ac530 "/var/spool/slurmd/pmix.11651139.0//pmix_dstor_ds21_337243", name=0x7ffc7e914d1d "slurm.pmix.11651139.0", local_size=1, uid=2008, 
    setuid=false) at gds_ds21_lock_pthread.c:239
#1  0x00002b6cebfef098 in pmix_common_dstor_init (
    ds_name=0x2b6cebfe24e0 "ds21", info=0x0, ninfo=0, 
    lock_cb=0x2b6cebfe5200 <pmix_ds21_lock_module>, 
    file_cb=0x2b6cebfe53a0 <pmix_ds21_file_module>) at dstore_base.c:1727
#2  0x00002b6cebfe0af9 in ds21_init (info=0x0, ninfo=0) at gds_ds21_base.c:35
#3  0x00002b6cebcf2db1 in pmix_gds_base_select (info=0x0, ninfo=0)
    at base/gds_base_select.c:86
#4  0x00002b6cebcc3a1c in pmix_rte_init (type=1, info=0x0, ninfo=0, 
    cbfunc=0x2b6cebc8a73d <pmix_client_notify_recv>) at runtime/pmix_init.c:308
#5  0x00002b6cebc8be50 in PMIx_Init (proc=0x2b6cebc478e0 <my_proc>, info=0x0, 
[...]

PR #1159, opened 2019-03-22
v3.1/gds/ds21: update the pointer of `seg_hdr` when segment re-attached

karasevb:
When re-attaching the lock-segment, the `seg_hdr` pointer was not
updated, which caused to reference on incorrect data and damage.

(cherry picked from commit ef2575f3ac21a3261da16d827fe2efd27b46151c)

Corresponds to #1156


artpol84:
Ref: #1114

PR #1160, opened 2019-03-22
v2.2/gds/ds21: update the pointer of `seg_hdr` when segment re-attached

karasevb:
When re-attaching the lock-segment, the `seg_hdr` pointer was not
updated, which caused to reference on incorrect data and damage.

(cherry picked from commit ef2575f3ac21a3261da16d827fe2efd27b46151c)

Corresponds to #1156


artpol84:
Ref: #1114

karasevb:
@rhc54 Could you merge this too?

rhc54:
Will do, though I don't have a schedule for the next 2.2.x release.

artpol84:
Why not release to fix the bug?
Shouldn’t require much additional testing

rhc54:
We don't generally release after a single bug fix, but rather stage things out or else we drive the distros batty with updates. Let's see what else has come in, what else shows up in the near future, etc.

PR #1167, opened 2019-03-26
server: plug a memory leak in _collect_data()

ggouaillardet:
This was identified by Coverity with CID 337282


rhc54:
Thanks! Just FYI: I'm prepping a 3.1.3 release so any checking you can make on it would be appreciated.

PR #1168, opened 2019-03-26
util/argv: remove the useless 'overwrite' parameter

ggouaillardet:
Remove the unused and useless 'overwrite' parameter from
pmix_status_t pmix_argv_append_unique_idx() and
pmix_argv_append_unique_nosize()

In order to maintain backward compatibility, the 'overwrite' parameter
was not removed from the public PMIX_ARGV_APPEND_UNIQUE() macro
but is now ignored.

Note pmix_argv_append_unique_nosize() was updated accordingly in pmix_extend.h

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -1,8 +1,8 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
  * Copyright (c) 2013-2019 Intel, Inc.  All rights reserved.
- * Copyright (c) 2016-2018 Research Organization for Information Science
- *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2016-2019 Research Organization for Information Science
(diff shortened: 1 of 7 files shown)


ggouaillardet:
@rhc54 be sure to note `include/pmix_extend.h` is modified in this PR. If this is not acceptable, then I will revamp it.
I did not modify the prototype of the `PMIX_ARGV_APPEND_UNIQUE` macro, and I guess you will be fine with that.

rhc54:
Actually, we are free to do whatever we like with that macro as it doesn't appear in the current v3.1 standard. I would go ahead and correct it. We can decide whether or not to add those macros to the standard as a separate question.

ggouaillardet:
I corrected the macro.

If you are fine with  `pmix_argv_append_unique_nosize()` being updated in `include/pmix_extend.h`, then you can merge this PR.

rhc54:
Thanks!

PR #1170, opened 2019-03-26
bfrops: fixed coverity: `Using uninitialized value "*tmpbuf"`

karasevb:
CID 1444191
CID 1444190
CID 1444187
CID 1444185
CID 1444180
CID 1444178
CID 1444175
CID 1444173


jjhursey:
Reviewing the current Coverity report it looks like https://github.com/pmix/pmix/pull/1155 is targeted at this set of errors too. I like having the calloc to be double sure, so I'm ok with taking this too.

PR #1171, opened 2019-03-26
psquash/flex128: Fix mismatch prototype

jjhursey:
The `master` branch will not build on my Mac because of the conflicting function signatures.

```
$$ gcc --version
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 10.0.0 (clang-1000.10.44.4)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

```
psquash_flex128.c:316:15: error: conflicting types for 'flex_pack_integer'
static size_t flex_pack_integer(size_t val,
              ^
psquash_flex128.c:220:15: note: previous declaration is here
static size_t flex_pack_integer(uint64_t val,
              ^
psquash_flex128.c:342:15: error: conflicting types for 'flex_unpack_integer'
static size_t flex_unpack_integer(const uint8_t in_buf[], size_t buf_size,
              ^
psquash_flex128.c:223:15: note: previous declaration is here
static size_t flex_unpack_integer(const uint8_t in_buf[], size_t buf_size,
              ^
```

### DIFF
--- a/src/mca/psquash/flex128/psquash_flex128.c
+++ b/src/mca/psquash/flex128/psquash_flex128.c
@@ -217,11 +217,11 @@ static pmix_status_t flex128_decode_int(pmix_data_type_t type, void *src,
                                         size_t src_len, void *dest,
                                         size_t *dst_size);
 
-static size_t flex_pack_integer(uint64_t val,
+static size_t flex_pack_integer(size_t val,
                                 uint8_t out_buf[FLEX_BASE7_MAX_BUF_SIZE]);
 
 static size_t flex_unpack_integer(const uint8_t in_buf[], size_t buf_size,
-                                  uint64_t *out_val, size_t *out_val_size);
+                                  size_t *out_val, size_t *out_val_size);
 
 pmix_psquash_base_module_t pmix_flex128_module = {
     .name = "flex128",
@@ -279,7 +279,7 @@ static pmix_status_t flex128_decode_int(pmix_data_type_t type, void *src,
                                         size_t src_len, void *dest, size_t *dst_size)
 {
     pmix_status_t rc = PMIX_SUCCESS;
-    uint64_t tmp;
+    size_t tmp;
     size_t val_size, unpack_val_size;
 
     PMIX_SQUASH_TYPE_SIZEOF(rc, type, val_size);


rhc54:
Hmmm...I think you want to go the other way. The internals of those functions both treat the value in question as ```size_t```. I think the function signature should match that designation. Changing the prototype to ```uint64_t``` only works if ```size_t``` is actually 64-bits

jjhursey:
I tried it the other way, and the compiler started complaining like this:
```
psquash_flex128.c:290:51: warning: incompatible pointer types passing 'uint64_t *' (aka 'unsigned long long *') to parameter of type
      'size_t *' (aka 'unsigned long *') [-Wincompatible-pointer-types]
    *dst_size = flex_unpack_integer(src, src_len, &tmp, &unpack_val_size);
                                                  ^~~~
psquash_flex128.c:224:43: note: passing argument to parameter 'out_val' here
                                  size_t *out_val, size_t *out_val_size);
                                          ^
1 warning generated.
```
So I err'ed on the side of picking the more specific type.

I think we might want to wait for @artpol84 / @karasevb to review if there is a subtle reason to go one way or the other.

rhc54:
Yeah, the problem is that there is inconsistency as you drop down the layers of calls. Someone needs to scrub that stack. My concern is that we not assume that ```size_t``` is equivalent in size to ```uint64_t``` as there are systems where this isn't true - and the number of such systems will likely grow over time.

artpol84:
We want to have size_t as the parameter, not uint64
For flexible packing all we need is the data type with maximum bit width. And size_t gives us that flexibility automatically. We don’t want to use uint64 on 32bit systems as it would be slow or won’t work at all.

So I agree with @rhc54 we want to go the other way.
It seems like in this case that “tmp” value in the error has to change it’s type to size_t as well.

artpol84:
https://github.com/pmix/pmix/blob/master/src/mca/psquash/flex128/psquash_flex128.c#L282

This guy is wrong. Has to be size_t.

Sorry about that- I guess gcc was fine with it.

jjhursey:
Ok. Let me try and flip it around. 1 sec

rhc54:
Yeah, it is weird - it seems that it passes CI and builds fine in a lot of environments, but the Mac plus particular minimum compiler version generates the error. Odd - technically correct, but odd.

Thanks for addressing it!

jjhursey:
Alright. I just force pushed an update. It builds fine on my Mac now. Take a look and let me know what you think.

artpol84:
Thank you!

rhc54:
kewl - thx!

PR #1175, opened 2019-03-27
v3.x/modex: reduce modex size by replacing key names with indexes 

karasevb:
Corresponds to #1149 
Includes fixes: #1168 

Depends on #1174.

PR #1173, opened 2019-03-27
v3.x/bfrops: redefinition regtypes for internal usage

karasevb:
Corresponds to #1144.
Includes fixes: #1170, #1161, #1155

Issue #1169, opened 2019-03-26
Porting MODEX optimizations to v3.x series


karasevb:
Since the MODEX optimizations (#1152, #1149, #1125) are merged to master, we would like to port those to the nearest release branch. 
What are the steps to open a new branch v3.2? 
Port of all PRs is ready, once new branch appears I'll open PRs against it.

@rhc54 @jjhursey @artpol84 

PRs which should be ported:
- [x]  bfrops: redefinition regtypes for internal usage #1144 (done: https://github.com/pmix/pmix/pull/1173)
- [x] modex: optimization of rank identification size reduction #1125 (done: #1174)
- [x] modex: reduce modex size by replacing key names with indexes #1149 (done: #1175)
- [x] Flexible integer packing with `psquash` framework in `bfrops` v4 #1152 (done: #1176)

jjhursey:
For now, just PR them against the `v3.1` branch (which would be used as the base for a `v3.2`), then we can change the base of the PR on GitHub once a `v3.2` branch is created.

artpol84:
Thanks!
@karasevb, please proceed. And pick a today’s fix from @jjhursey for MAC: https://github.com/pmix/pmix/pull/1171

jjhursey:
Probably also want PR #1168 which is a bit of cleanup for the argv.c file related to PR #1149.

karasevb:
Since bfrops v4 (which has been modified) is not present in PMIx v3, and new features require to add with new bfrops, so the question is: which version of bfrops required to introduce for PMIx v3? Bfrops v31 or v32?

karasevb:
I’ve currently added bfrops v32, but if it’s not OK, I’ll change the version.

jjhursey:
For PR #1176 I think it's ok for now to keep it at v32 as we discuss the best path forward there. I need to think about it a bit more. Let's make sure to discuss it on the call this week.

Since these optimizations are targeted at v3.2, it shouldn't hold up the pending v3.1.x bug fix release.

jjhursey:
Per teleconf:
 * Split these into two sets targeted at two different release points divided by the cross-version impact.
 * 3.1.4
   - modex: optimization of rank identification size reduction #1125 (done: #1174)
   - modex: reduce modex size by replacing key names with indexes #1149 (done: #1175)
 * 3.2
   - bfrops: redefinition regtypes for internal usage #1144 (done: #1173)
[...]

PR #1176, opened 2019-03-27
v3.x/Flexible integer packing with psquash framework in bfrops 

karasevb:
Corresponds to #1152 
Includes fixes #1171 

Depends on #1173, #1175

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -4,7 +4,7 @@
  * Copyright (c) 2016-2018 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2016      IBM Corporation.  All rights reserved.
- * Copyright (c) 2016-2018 Mellanox Technologies, Inc.
+ * Copyright (c) 2016-2019 Mellanox Technologies, Inc.
  *                         All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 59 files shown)


jjhursey:
Ref issue #1169

jjhursey:
From this week's teleconf:
 * Cross version test script enhancement request (separate from this PR)
   - https://github.com/pmix/pmix-tests/issues/4
 * Creating a `bfrops/v32` component seems like the correct path
 * During the client/server handshake we need to make sure both sides are using the same `psquash` component since the normal traffic between the client and server packs/unpacks data so both sides need to be using the same `bfrops` and `psquash` components.
   - Consider the scenario where the client is using (bfrops/v32 + psquash/flex128) and the server is using (bfrops/v32 + psquash/native). Now the encoding mismatch will cause a conflict.
   - The suggestion was that if the server picks the `bfrops/v32` component then the client is forced to use the matching `psquash` component (probably `flex128`). This is a short term solution.
      - The user can always fall back to `bfops/v31` if they hit any issues with the `psquash/flex128` component (or framework).
   - Longer term we need the `psquash` framework to be multi-select and allow the handshake to pick the most compatible version.
      - I created https://github.com/pmix/pmix/issues/1184 to track that enhancement request.

artpol84:
@karasevb lets discuss this

jjhursey:
Any progress on this discussion?

ibm-ompi:

 Gist: https://gist.github.com/b8acfba24f374a0f4dd0531a418de00b

rhc54:
@karasevb @artpol84 Please retarget v3.2 branch

ibm-ompi:

 Gist: https://gist.github.com/9ff00f280f7ed12b1477ad57d221c3b6

karasevb:
Closed by #1423

Issue #1189, opened 2019-04-02
Event race condition between registration and cached delivery


jjhursey:
Thank you for taking the time to submit an issue!

### What version of the PMIx Reference Library are you using? (e.g., v1.0, v2.1, git master @ hash, etc.)

v3.1 branch [HEAD](https://github.com/pmix/pmix/commit/b3e93da2a5419b585f25716b56cc6273f56d25c4) with Open MPI v4.0.x (approximate HEAD of branch)

### Describe how PMIx was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

git clone of v3.1 branch

### Please describe the system on which you are running

* Operating system/version: RHEL 7.5
* Computer hardware: ppc64le
* Network type: TCP

-----------------------------

## Details of the problem

I'm working a bug with Open MPI's debugger support through `mpirun -tv ...` which uses PMIx events instead of the MPIR_debug_gate [here](https://github.com/open-mpi/ompi/blob/v4.0.x/ompi/mca/rte/orte/rte_orte_module.c#L164-L199) to release the application process from MPI_Init. I can reproduce with 2 processes on a single node (`mpirun -tv -np 2 ./hello`)

I've noticed that there is a race condition between when the PMIx client receives the local `pmix_hdlr_reg_cbfunc_t` call and when it starts receiving events on the `pmix_notification_fnt_t` handler.

Open MPI assumes that the registration callback is called before any notification callback. If that is not true then an event is lost as it falls out of [this loop](https://github.com/pmix/pmix/blob/v3.1/src/event/pmix_event_registration.c#L850-L851) since the index does not match. The index is resolved in the registration callback [here](https://github.com/open-mpi/ompi/blob/v4.0.x/opal/mca/pmix/pmix3x/pmix3x.c#L1259).

[...]

PR #1190, opened 2019-04-02
Ensure any cached notifications arrive after reg completes

rhc54:
We want the register_events callback function to be executed prior to
delivery of any corresponding events. Ensure that happens by injecting
the registration into an event that won't fire until after the server
has notified the client of completion of the registration.

### DIFF
--- a/src/server/pmix_server_ops.c
+++ b/src/server/pmix_server_ops.c
@@ -1836,29 +1836,194 @@ pmix_status_t pmix_server_connect(pmix_server_caddy_t *cd,
     return rc;
 }
 
+static void _check_cached_events(int sd, short args, void *cbdata)
+{
+    pmix_setup_caddy_t *scd = (pmix_setup_caddy_t*)cbdata;
+    pmix_notify_caddy_t *cd;
+    pmix_range_trkr_t rngtrk;
+    pmix_proc_t proc;
+    int i;
+    size_t k, n;
+    bool found, matched;
+    pmix_buffer_t *relay;
+    pmix_status_t ret = PMIX_SUCCESS;
+    pmix_cmd_t cmd = PMIX_NOTIFY_CMD;
+
+    /* check if any matching notifications have been cached */
+    rngtrk.procs = NULL;
+    rngtrk.nprocs = 0;
+    for (i=0; i < pmix_globals.max_events; i++) {
+        pmix_hotel_knock(&pmix_globals.notifications, i, (void**)&cd);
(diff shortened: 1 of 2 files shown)


rhc54:
Fixes #1189

rhc54:
@karasevb Mellanox failed due to lack of disk space. Trying again

:bot:mellanox:retest

rhc54:
bot:mellanox:retest

jjhursey:
Thanks! I'll give this patch a try against my test case this morning.

jjhursey:
I confirmed that this fixed the original problem.

rhc54:
@jjhursey I'm assuming you would like this in v3.1.3, yes?

src/server/pmix_server_ops.c
```diff
@@ -2056,47 +2221,68 @@ pmix_status_t pmix_server_register_events(pmix_peer_t *peer,
...
+        scd = PMIX_NEW(pmix_setup_caddy_t);
+        PMIX_RETAIN(peer);
+        scd->peer = peer;
+        scd->codes = codes;
+        scd->ncodes = ncodes;
+        scd->procs = affected;
+        scd->nprocs = naffected;
+        scd->opcbfunc = NULL;
+        scd->cbdata = NULL;
+        PMIX_THREADSHIFT(scd, _check_cached_events);
```
jjhursey: The `DEBUGGER-RELEASE` case goes here since it's not an environmental event. So here we thread shift the `_check_cached_events` and return `PMIX_OPERATION_SUCCEEDED`. Does the return of this function then trigger the message back to the client to acknowledge the event (and thus trigger the client side registration callback)? I'm just not seeing where that acknowledgment is triggered - even with the old code.

jjhursey:
Yes, please. I applied this PR to the v3.1.3 branch for testing and it worked great.

I think I follow the code except I can't seem to find where the registration message goes back to the client. Just to complete my understanding of the patch.

src/server/pmix_server_ops.c
rhc54: Yes - returning operation_succeeded to the pmix_server/switchyard causes the switchyard to return "success" to the client. Since the switchyard is operating inside an event, the message to the client gets queued prior to the threadshift being executed. Thus, the client gets the registration callback followed by the results of ```_check_cached_events```

jjhursey: Perfect. Thanks!

PR #1191, opened 2019-04-03
configury: use PMIX_ENABLE_DLOPEN_SUPPORT instead of enable_dlopen

ggouaillardet:
The $enable_dlopen variable might be set to "unknown" by libtool.m4.
In order to cope with that, use $PMIX_ENABLE_DLOPEN_SUPPORT that is
set before $enable_dlopen might be overwritten.
Also add a test to make sure PMIX_ENABLE_DLOPEN_SUPPORT was not set
after $enable_dlopen was overwritten.

Refs. pmix/pmix#1121

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -901,6 +901,10 @@ AC_DEFUN([PMIX_DEFINE_ARGS],[
                         [Whether build should attempt to use dlopen (or
                          similar) to dynamically load components.
                          (default: enabled)])])
+    AS_IF([test "$enable_dlopen" = "unknown"],
+          [AC_MSG_WARN([enable_dlopen variable has been overwritten by configure])
+           AC_MSG_WARN([This is an internal error that should be reported to PMIx developers])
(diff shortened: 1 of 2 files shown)


config/pmix.m4
```diff
@@ -901,6 +901,10 @@ AC_DEFUN([PMIX_DEFINE_ARGS],[
                         [Whether build should attempt to use dlopen (or
                          similar) to dynamically load components.
                          (default: enabled)])])
+    AS_IF([test "$enable_dlopen" = "unknown"],
```
rhc54: Won't this error out in the OMPI case where we are embedded?

ggouaillardet: Nope, even if PMIx is embedded, it's `configure` is a different process (not like `hwloc` integration view). And I already embedded the patched PMIx and had no issue.

rhc54: sounds good - thx!

PR #1192, opened 2019-04-03
configury: use PMIX_ENABLE_DLOPEN_SUPPORT instead of enable_dlopen

rhc54:
The $enable_dlopen variable might be set to "unknown" by libtool.m4.
In order to cope with that, use $PMIX_ENABLE_DLOPEN_SUPPORT that is
set before $enable_dlopen might be overwritten.
Also add a test to make sure PMIX_ENABLE_DLOPEN_SUPPORT was not set
after $enable_dlopen was overwritten.

Refs. pmix/pmix#1121

(cherry picked from commit 7e4917c915e9234ab5530cb51d1ada5f509f6749)


rhc54:
I reviewed what Gilles did

PR #1193, opened 2019-04-03
Resolve get of proc-specific job-level info from another nspace

### LINKED ISSUE #1172: PMIx_Get of local process rank data failing

## Background information

This issue is similar to issue 1168 but is not the same.   In that case if I set
 export PMIX_MCA_gds=hash 
the problem goes away, but this problem is slightly different.  See below for details.
That ticket is about when the rank is PMIX_RANK_WILDCARD, but this issue is about 
using specific ranks.

3.1

git clone

* Operating system/version: Linux
* Computer hardware: IBM Power
* Network type: TCP

-----------------------------

## Details of the problem

In this case I am trying to do:
proc.rank = 0;
strcpy(proc.nspace, argv[1]);
rc = PMIx_Get(&proc, PMIX_HOSTNAME, NULL, 0, &val);

In this case proc is known to be a running on the local node. (when it is not I see a call to server_dmodex_req coming up to the server).  The PMIx_Get call hangs, but no server_dmodex_req call is coming up to the server.   If I do the same get call on the process itself, it works ok and even if I do it on another rank in the same namespace running locally, but when it is a different namespace and a specific rank (not PMIX_RANK_WILDCARD), then it hangs.  

I can reproduce this using PMIx_Spawn under mpirun/orted as well.

rhc54:
The current code focused solely on retrieving modex info for procs where
the "get" request goes up to the PMIx server as there was an assumption
that job-level information would already be available. However, when
asking for information on a different nspace, the job-level info may be
available at the server, but not to the requesting process.

We had a code path to return the job-level info from the other nspace,
but that didn't include the proc-specific portion of the job-level info.
In other words, we returned things like PMIX_UNIV_SIZE but did not
return PMIX_HOSTNAME for each proc in the other nspace. This change
fixes that gap.

As a note for the future: it would be simpler if we just provided a
"hook" to the other proc's shmem region. However, that would require
some additional code to avoid leaving a proc with a stale handle to the
other nspace's shmem region once that nspace goes away.

Fixes #1172 

### DIFF
--- a/examples/client.c
+++ b/examples/client.c
@@ -254,6 +254,9 @@ int main(int argc, char **argv)
         fprintf(stderr, "Client ns %s rank %d: PMIx_Commit failed: %d\n", myproc.nspace, myproc.rank, rc);
         goto done;
     }
+    if (0 == myproc.rank) {
+        sleep(2);
+    }
 
     /* call fence to synchronize with our peers - instruct
      * the fence operation to collect and return all "put"
--- a/examples/dynamic.c
+++ b/examples/dynamic.c
@@ -49,9 +49,7 @@ int main(int argc, char **argv)
     char nsp2[PMIX_MAX_NSLEN+1];
     pmix_app_t *app;
     char hostname[1024], dir[1024];
-    pmix_proc_t *peers;
-    size_t npeers, ntmp=0;
-    char *nodelist;
+    size_t ntmp=0;
 
     if (0 > gethostname(hostname, sizeof(hostname))) {
         exit(1);
--- a/examples/dynamic.c
+++ b/examples/dynamic.c
@@ -71,14 +71,14 @@ int main(int argc, char **argv)
     (void)strncpy(proc.nspace, myproc.nspace, PMIX_MAX_NSLEN);
     proc.rank = PMIX_RANK_WILDCARD;
 
-    /* get our universe size */
-    if (PMIX_SUCCESS != (rc = PMIx_Get(&proc, PMIX_UNIV_SIZE, NULL, 0, &val))) {
-        fprintf(stderr, "Client ns %s rank %d: PMIx_Get universe size failed: %d\n", myproc.nspace, myproc.rank, rc);
+    /* get our job size */
+    if (PMIX_SUCCESS != (rc = PMIx_Get(&proc, PMIX_JOB_SIZE, NULL, 0, &val))) {
+        fprintf(stderr, "Client ns %s rank %d: PMIx_Get job size failed: %d\n", myproc.nspace, myproc.rank, rc);
         goto done;
     }
     nprocs = val->data.uint32;
     PMIX_VALUE_RELEASE(val);
-    fprintf(stderr, "Client %s:%d universe size %d\n", myproc.nspace, myproc.rank, nprocs);
+    fprintf(stderr, "Client %s:%d job size %d\n", myproc.nspace, myproc.rank, nprocs);
 
     /* call fence to sync */
     (void)strncpy(proc.nspace, myproc.nspace, PMIX_MAX_NSLEN);
@@ -103,13 +103,6 @@ int main(int argc, char **argv)
         app->env = (char**)malloc(2 * sizeof(char*));
         app->env[0] = strdup("PMIX_ENV_VALUE=3");
         app->env[1] = NULL;
-        PMIX_INFO_CREATE(app->info, 2);
-        (void)strncpy(app->info[0].key, "DARTH", PMIX_MAX_KEYLEN);
-        app->info[0].value.type = PMIX_INT8;
-        app->info[0].value.data.int8 = 12;
-        (void)strncpy(app->info[1].key, "VADER", PMIX_MAX_KEYLEN);
-        app->info[1].value.type = PMIX_DOUBLE;
-        app->info[1].value.data.dval = 12.34;
 
         fprintf(stderr, "Client ns %s rank %d: calling PMIx_Spawn\n", myproc.nspace, myproc.rank);
         if (PMIX_SUCCESS != (rc = PMIx_Spawn(NULL, 0, app, 1, nsp2))) {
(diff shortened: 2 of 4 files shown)


rhc54:
Fixes #1172

rhc54:
@jjhursey @dsolt This is a fairly significant change in PMIx_Get behavior so we should probably test it pretty carefully to ensure it doesn't break other corner-cases - e.g., modex data request for procs in another nspace.

rhc54:
@jjhursey @dsolt I added Dave's reproducer to the pmix-tests repo.

dsolt:
I am working on doing my own tests with this now and then I will review the code.... shouldn't take long.

rhc54:
bot:mellanox:retest

dsolt:
This seems to work for me when the spawn'd process is running on the same host as the parent.  However, if the spawned process is running on a different host than its parent, I still see a hang.  I'm looking into it to make sure its not the result of something strange on server side...

dsolt:
I should point out that for my particular use case,  it is actually the child processes (that go spawned to a different host than its parent), that calls PMIx_Get on job-level data of its parent (PMIX_NPROC_OFFSET for example).  That call generates a server_modex_req call up to  the RM.  When the server calls the callback, the PMIx_Get completes and the child process goes on toe call PMIx_Get again, but this time on rank 0 of the parent namespace.  This does not generate any further calls up to the RM and appears to hang.

rhc54:
Okay, I'll give that use-case a try - I was only testing the case described in #1172

rhc54:
Just to be clear: does the node that eventually hosts the child process know about the parent job?

dsolt:
Hmmm.   In my case I realize it does not.   However, it is strange because it did successfully do a PMIx_Get on job-level data for that unknown namespace.   (I do have my #1120 fix cherry-picked in as well, maybe that is allowing it to get this far).

rhc54:
I won't be able to replicate this as PRRTE doesn't work that way - not that I'm saying one is right and the other wrong, just noting that I don't have an environment where I can replicate the situation. However, I might be able to figure this out by code inspection and give you something you can test.

rhc54:
@dsolt I believe I tracked this down and fixed it:

> If someone requests job-level info from a new nspace, we will
"know" the nspace when they request anything else - but we won't know
anything about their procs if we don't locally have at least one of
them. Thus, we cannot wait for the procs to "register" with us as that
may never happen.

Please give it a try and let me know

dsolt:
This is getting close :) and I'm painfully learning a lot in the process.  What I'm seeing is that my second client is doing a PMIx_Get on pmix.hname on the first client who is executing on the same node.  I see it checking that it doesnt have this data and sending the request up to the server.  The server sends back a reply which seems to have all the namespace information for the first client (all the things like pmix.univ_size of anything that isn't rank specific) and then it is also sending back all rank data that was published by individual ranks.  However it doesn't seem to pull down job data for individual ranks.  So it is not seeing pmix.hname come back, though it does see other keys that that rank explicitly published with PMIx_Put().   I'm not sure the naming conventions for these different types of data, but there are 1) thinks like pmix.univ

rhc54:
Hmmm...my bad. It does indeed make sense. I thought I had captured those, but guess I missed them - let me look again. They get stored in a different hash table and so I have to ensure I am properly covering all the spots.

rhc54:
@dsolt I fixed this so we pickup the job-level data for each proc in addition to the global job-level data. Please give it a try and let me know if it works for you.

As things stand, you need to pass two dmodex requests to the local server in order to get all the info you might want when asking for modex data on a proc from an unknown nspace:

1. a request for rank=wildcard to get all the job-level data, including for every proc in the job
1. a request for "remote" data for the specific proc you were asking about

I was searching for a way to avoid making the two requests, but I can't find one. The problem is that the PMIx_server_dmodex_req function doesn't include an "info" array, and so you only have the "proc" ID you can pass in. Not enough info to know whether or not the remote requestor already has job-level info and just needs modex info for a given proc.

We have a few options here if we want to improve things:

1. the requesting server knows that it lacks job-level info for this nspace, so it could automatically generate two calls to the host (one for rank=wildcard and the second for the rank in question)
1. the requesting host could track the requests and generate either one or two calls, as required
1. the receiving host could track the requests and execute PMIx_server_dmodex_request multiple times, if needed
1. the receiving PMIx server library could track the requests and return job-level info along with the rank of interest

Any thoughts?

dsolt:
This still didn't work :) , but I think I know why.  In my case, the 2nd process is being spawned by the first and during PMIx_Connect_nb call by the parent it is getting the child's job data, but that code path isn't bringing back the rank-specific job data.  Then when the PMIx_Get happens, it doesn't think it needs to ask for it.  So, I think I just need to make sure your changes are applied to the PMIx_Connect_nb path (maybe I need to appy the changes you made in pmix_server_get to pmix_connect_get?).   As for your question on how to handle the double request issue....

rhc54:
Ah...indeed! Let me take a look and see if any other channels exist where that could happen. I'll add the fix for this one here.

rhc54:
On the double request issue: Option 1 guarantees that a server only ever gets one copy of the remote nspace info. In Option 4, you can get multiple copies sent to you if you ask for data from more than one remote node (as each would have a separate record of not having sent you anything).

The negative to Option 1 is that two requests flow over the wire, but I'm thinking that isn't too big a deal as this isn't a very common occurrence. So I'd lean towards Option 1 myself, and agree that we should keep it in the CL itself.

I'll include it here and let's see how it behaves.

rhc54:
@dsolt Please give this a try - should include all the fixes discussed above.

dsolt:
This appears to work for a single host where the conditions exist as described in the original issue related to this PR.  When the child and parent are on different hosts, I see the child send a dmodex request for PMIX_RANK_WILDCARD and then another dmodex for the specific rank, and neither response contains the data I am looking for.  (I am running with PMIX_MCA_gds=hash).   I think this problem though is related to the bigger issue of our implementation not calling PMIx_server_register_nspace when PMIx_Connect calls server_connect in the RM.  

Going back to my tool example, a tool might ask, "I need to know what hostname rank 3 is running on" (and thus do a PMIx_Get on some namespace (that it is not connected to) on rank 3 with PMIX_HOSTNAME.   That would seem to lead to this failure as well, but that is probably out of the scope of this PR.   I think it is also out of the scope of PR #1120 as well though, since that one was just about making sure that when a dmodex reply comes back, we know how to handle it.  It would not address the problem that the dmodex reply would not contain this per-rank registered data, data that the rank put itself with PMIx_Put().

Anyhow, unless you believe any of that issue could/should be addressed here, I think I'm good with this PR and ready to approve.  I have looked over the code quite a bit, but was going to spend a few more minutes reviewing before I hit approve.

rhc54:
Just to be clear: what data are you looking for?

rhc54:
Think I found the problem - it looks to me like there is a mismatch between how the data was packed and how it is being unpacked, leading to it being "dropped". I may have to create a way to replicate this scenario to more efficiently resolve it.

dsolt:
In this case one process is doing a PMIx_Get() of PMIX_HOSTNAME of another rank in another namespace.

dsolt:
Ok.  If you want to do a screen share sometime I can step you through the code and show you what I'm seeing.

rhc54:
Check this latest try and see if that catches it. If it still isn't working right, then let's do the screen share.

dsolt:
Back to my last scenario.... the server where the parent is running see's two requests coming from the client related to the remote namespace. The first is for PMIX_RANK_WILDCARD.  The second is for the specific rank 0.   During the first request, a remote request is made for all job data and that request comes back and I can verify that it contains both general namespace data and the per-rank job data.  I I can verify that this data is then parsed and stored correctly on the parent's server.   However, at that point it looks for outstanding requests and it sees that its client has an outstanding request on PMIX_RANK_WILDCARD.    At pmix_server_get.c:720 it attempts to satisfy this request, but on line 730 it calls PMIX_GDS_FETCH_KV.   This call does not know to return all the per-rank data, so it only returns wildcard data.   Later another request comes up for rank 0 data and again it calls pmix_server_get and it comes into this satisfy_request code and again comes to down into this code on line 730, but it also does not know to send back rank data.   The comment even seems right:

717         /* if they are asking about a rank from an nspace different
718          * from their own, or they gave a rank of "wildcard", then
719          * include a copy of the job-level info */

But it is not responding with the per-rank blob.   I'm trying to figure out how I can get per-rank data passed down to the client.  I tried using get_job_data() function from here, but I am not packing things correctly.   Still looking into it...

rhc54:
Okay, I can fix it. Probably need to adjust the unpacking side in the client.

rhc54:
I see the problem - looks to me like it may be trying to insert the data into a shared memory region instead of sending it to the client's hash module. I'll have to try and trace this to see where we intercept to make the change.

dsolt:
Ok, I'm free after 3pm CST if you want to see me step through.

rhc54:
@karasevb Something appears out-of-sync here with Jenkins as a simple "make check" passes all tests on my local machine. Is there some kind of parameter or other variation involved in the Jenkins script?

rhc54:
@karasevb Would it be possible to "cat" the contents of test/test-suites.log upon a failure of "make check"? I don't see any way to view that file from the Jenkins console and it would help to see the actual errors.

rhc54:
@karasevb @artpol84 I'll try to poke into this a bit, but I'm actually seeing a bunch of errors being reported into the ".log" files of the individual tests - and yet the server consistently is reporting "OK", thus mistakenly causing "make check" to report all tests as "passed". It would appear that either the exit status of the test client itself isn't getting updated with the error, or the server isn't picking it up.

karasevb:
bot:mellanox:retest

ibm-ompi:

 Gist: https://gist.github.com/0234965fb51a8a6b50813c659b23ee16

ibm-ompi:

 Gist: https://gist.github.com/3d863bface451312c0d5db7a735618ce

jjhursey:
bot:ibm:xver:retest

ibm-ompi:

 Gist: https://gist.github.com/b406d6896468583dd3b839284c228e74

ibm-ompi:

 Gist: https://gist.github.com/479841281124b94781325ccf990e9c2b

rhc54:
@jjhursey This is a good one:
```
fatal: unable to access 'https://github.com/pmix/pmix-tests/': Could not resolve host: github.com; Unknown error
```

ibm-ompi:

 Gist: https://gist.github.com/ce4c64fc3cab7d981a8e4f6af06fe0d5

jjhursey:
Yeah, I've seen that a few times now. I've also seen GH struggle with resolving requests (displaying their unicorn page) in the past day or so. So it is hard to pinpoint if it's a network issue on IBM's side or not. I have a little script that pings GH every 15 min and it is seeing the same thing - GH failing to handle requests. It's usually resolved within a few minutes - so retrying is the best solution I have at the moment.

ibm-ompi:

 Gist: https://gist.github.com/d5226ec5c57c1dcf200061f6b2b812fe

ibm-ompi:

 Gist: https://gist.github.com/8670f52335fdbcaee70a5e31a3cb480f

rhc54:
@dsolt This might fix #1120 - probably need to test that use-case and see.

artpol84:
@rhc54

> @karasevb Would it be possible to "cat" the contents of test/test-suites.log upon a failure of "make check"? I don't see any way to view that file from the Jenkins console and it would help to see the actual errors

Is this still required?

rhc54:
No, I think we are okay - thanks for checking!

PR #1194, opened 2019-04-03
Ensure any cached notifications arrive after reg completes

rhc54:
We want the register_events callback function to be executed prior to
delivery of any corresponding events. Ensure that happens by injecting
the registration into an event that won't fire until after the server
has notified the client of completion of the registration.

(cherry picked from commit 2b185bf1bf9d6eb9e0030c48544eacf742336442)


jjhursey:
Ref Issue #1189

PR #1197, opened 2019-04-04
Implement "make check" in the tests subdir

rhc54:
Add a script that runs the Jenkins tests and returns a non-zero status
if any of them fail.


rhc54:
Refs #1027

rhc54:
Shouldn't be required - the build system automatically makes the tests (along with the library) when you execute ```make``` at the top-level directory

ax3l:
Oh nice, didn't know that. thanks!

PR #1206, opened 2019-04-12
dist: Clean up handling of --disable-dlopen in dist process

bwbarrett:
"make check" can only succeed if the plugins have been installed or
"--disable-dlopen" has been passed to configure.  The first can't
happen in "make distcheck", so force the second for all invocations
of "make distcheck" by setting AM_DISTCHECK_CONFIGURE_FLAGS in
Makefile.am.

Clean up the handling of AM_DISTCHECK_CONFIGURE_FLAGS and
DISTCHECK_CONFIGURE_FLAGS in make_dist_tarball.  First, with the
above fix, passing --disable-dlopen is no longer required.  Second,
users (and scripts) should set DISTCHECK_CONFIGURE_FLAGS, not
AM_DISTCHECK_CONFIGURE_FLAGS.  Automake will combine the two options
when running "make distcheck", so the right things will happen with
this configuration.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -47,6 +47,8 @@ Copyright (c) 2012      The University of Wisconsin-La Crosse. All rights
                         reserved.
 Copyright (c) 2013-2019 Intel, Inc. All rights reserved.
 Copyright (c) 2011-2014 NVIDIA Corporation.  All rights reserved.
+Copyright (c) 2019      Amazon.com, Inc. or its affiliates.  All Rights
+                        reserved.
 
 $COPYRIGHT$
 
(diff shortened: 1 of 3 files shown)


bwbarrett:
This should backport cleanly to any of the other branches that you've backported your "make check" cleanup patches to.

rhc54:
@bwbarrett This is failing with an error:
```shell
make: unrecognized option '--with-libevent=/var/lib/jenkins/jobs/gh-pmix-pr/workspace/jenkins/libevent-2.0.22-stable/install
```
It looks like a configure argument is incorrectly being passed to `make`. Not entirely sure I can spot where that is happening.

rhc54:
I've got a fix in final test - will push soon

Issue #1204, opened 2019-04-12
"make check" fails when dstore is disabled


rhc54:
We seem to have unintentionally built a dependency on dstore into the code base. When running `make check` with `PMIX_MCA_gds=hash`, I get the following test failures:
```shell
./pmix_test -n 4 --job-fence -c 2>&1
==82964== ERROR [test_fence.c:392:test_job_fence]: smoky_nspace-0:3: PMIx_Get failed: -17 from smoky_nspace-0:0, key key-f101:0
==82962== ERROR [test_fence.c:392:test_job_fence]: smoky_nspace-0:1: PMIx_Get failed: -17 from smoky_nspace-0:0, key key-f101:0
==82963== ERROR [test_fence.c:392:test_job_fence]: smoky_nspace-0:2: PMIx_Get failed: -17 from smoky_nspace-0:0, key key-f101:0
==82964== ERROR [test_fence.c:392:test_job_fence]: smoky_nspace-0:3: PMIx_Get failed: -17 from smoky_nspace-0:0, key key-f101:1
==82962== ERROR [test_fence.c:392:test_job_fence]: smoky_nspace-0:1: PMIx_Get failed: -17 from smoky_nspace-0:0, key key-f101:1
==82963== ERROR [test_fence.c:392:test_job_fence]: smoky_nspace-0:2: PMIx_Get failed: -17 from smoky_nspace-0:0, key key-f101:1
==82964== ERROR [test_fence.c:392:test_job_fence]: smoky_nspace-0:3: PMIx_Get failed: -17 from smoky_nspace-0:0, key key-f101:2
==82962== ERROR [test_fence.c:392:test_job_fence]: smoky_nspace-0:1: PMIx_Get failed: -17 from smoky_nspace-0:0, key key-f101:2
==82961== ERROR [test_fence.c:392:test_job_fence]: smoky_nspace-0:0: PMIx_Get failed: -17 from smoky_nspace-0:1, key key-f101:0
==82963== ERROR [test_fence.c:392:test_job_fence]: smoky_nspace-0:2: PMIx_Get failed: -17 from smoky_nspace-0:0, key key-f101:2
==82964== ERROR [test_fence.c:428:test_job_fence]: smoky_nspace-0:3 [ERROR]: PMIx_Get returned -17 instead of not_found
==82962== ERROR [test_fence.c:428:test_job_fence]: smoky_nspace-0:1 [ERROR]: PMIx_Get returned -17 instead of not_found
[...]

Issue #1198, opened 2019-04-04
"make check" shows two failing tests


rhc54:
I wrote a script ```test/run_tests.pl``` that is used by ```make check``` to execute a set of tests in the ```test``` directory. I lifted the tests themselves directly from the ```contrib/pmix-jenkins.sh``` script, but I don't know if that is currently what Mellanox Jenkins is using or not.

When I run the script, I see the following output:

```shell
./pmix_test -s 2 -n 2 --job-fence 2>&1
==30632== test_server.c:server_finalize: Test finished OK!

CODE 0
./pmix_test -n 4 --job-fence -c 2>&1
==30675== test_server.c:server_finalize: Test finished OK!

CODE 0
./pmix_test -n 4 --job-fence 2>&1
==30745== test_server.c:server_finalize: Test finished OK!

CODE 0
./pmix_test -n 4 --ns-dist 3:1 --fence "[db | 0:0-2;1:0]" 2>&1
==30816== test_server.c:server_finalize: Test finished OK!

CODE 0
./pmix_test -n 4 --ns-dist 3:1 --fence "[db | 0:;1:3]" 2>&1
==30901== ERROR [test_fence.c:190:test_fence]: smoky_nspace-0:0: PMIx_Get failed: -17 from smoky_nspace-1:3, key key-f0:0
==30901== ERROR [test_fence.c:192:test_fence]: smoky_nspace-0:0: PMIx_Get failed (-1) from smoky_nspace-1:3
==30901== ERROR [pmix_client.c:146:main]: smoky_nspace-0:0 Fence test failed: -1
==30902== ERROR [test_fence.c:190:test_fence]: smoky_nspace-0:1: PMIx_Get failed: -17 from smoky_nspace-1:3, key key-f0:0
==30902== ERROR [test_fence.c:192:test_fence]: smoky_nspace-0:1: PMIx_Get failed (-1) from smoky_nspace-1:3
==30903== ERROR [test_fence.c:190:test_fence]: smoky_nspace-0:2: PMIx_Get failed: -17 from smoky_nspace-1:3, key key-f0:0
==30903== ERROR [test_fence.c:192:test_fence]: smoky_nspace-0:2: PMIx_Get failed (-1) from smoky_nspace-1:3
==30902== ERROR [pmix_client.c:146:main]: smoky_nspace-0:1 Fence test failed: -1
==30903== ERROR [pmix_client.c:146:main]: smoky_nspace-0:2 Fence test failed: -1
[...]

PR #1207, opened 2019-04-13
Fix the tests in the "test" subdir and silence a warning

rhc54:


rhc54:
Fixes #1204

rhc54:
Fixes #1198

artpol84:
@rhc54 thank you very much for tracking this down.
I am sorry for being silent - I was on vacation past 2 weeks.
A couple of follow-up questions:
* Have you had a chance to identify the difference between CI and `make check`? 
  * In particular why 
https://github.com/pmix/pmix/pull/1207/files#diff-45e5beb2251b9d715cf51150bb055a1fR155
 wasn't triggered in CI and did trigger in `make check`?
  * The fact that we were not tracking `PMIX_ERR_PROC_ENTRY_NOT_FOUND` probably points out that we are not testing `hash` with our CI. I thought we do.
* Does it makes sense to change dstore to return `PMIX_ERR_PROC_ENTRY_NOT_FOUND` in the appropriate places to keep all GDS consistent?

rhc54:
@artpol84 No problem - figured you were probably out 😄 I think the problem is that CI is looking at the exit code of the server, and I found that the server wasn't returning an error on one of the PMIx_Get variations. I was seeing PMIX_ERROR_LOG reports being output, which caught my attention and led me to find that return code mistake. When I changed that, I uncovered the need to update the test.

I only found the `hash` issue by setting `PMIX_MCA_gds=hash` in the environment. Otherwise, the `hash` component is only used to stage the job info and isn't involved in retrieving it. Testing that component on its own is something we probably should add as a separate test case, using one of the tests that failed before.

If it isn't too much effort, it probably does make sense to ensure the `hash` and `dstore` components return consistent values. Might be worth reviewing the `hash` as well as it may not be consistent itself.

Running `make check` in Jenkins would make sense to me, but adding a test to `make check` takes more effort than adding one to Jenkins. I didn't have time to cleanup the `make check` logic - I'm sure someone could do so and make that process more efficient as well. If that gets done, then the two methods for adding a test would be pretty equivalent and just using `make check` would seem reasonable.

artpol84:
I noticed that pmix_test has a parameter to select GDS component:
```
$ ./pmix_test -h
usage: pmix_test [-h] [-e foo] [-b] [-c] [-nb]
	-n       provides information about the job size (for checking purposes)
	-e foo   use foo as test client
	-v       verbose output
	-t <>    set timeout
	-o out   redirect clients logs to file out.<rank>
...
	--gds <external gds name>           set GDS module "--gds hash|ds12", default is hash
```

But I don't see it being used in jenkins scripts. We need to update them. @karasevb please take care of that

PR #1214, opened 2019-04-17
Travis: -Werror

ax3l:
Try to keep the code base warning free by failing CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,9 @@ cache:
 
 env:
   global:
-    # TODO: -Werror
-    # TODOL -fsanitize=address -fsanitize=undefined
-    - CFLAGS: "-Wall -Wextra"
-    - CXXFLAGS: "-Wall -Wextra"
+    # TODO: -fsanitize=address -fsanitize=undefined
+    - CFLAGS="-Wall -Wextra -Werror -Wno-unused-function"
+    - CXXFLAGS="-Wall -Wextra -Werror -Wno-unused-function"
 
 addons:
   apt:


.travis.yml
```diff
@@ -8,8 +8,8 @@ env:
   global:
     # TODO: -Werror
     # TODOL -fsanitize=address -fsanitize=undefined
-    - CFLAGS: "-Wall -Wextra"
-    - CXXFLAGS: "-Wall -Wextra"
+    - CFLAGS: "-Wall -Wextra -Werror"
```
ax3l: seams to be ignored by the build scripts since I can see warnings here: https://travis-ci.org/pmix/pmix/builds/521290397

rhc54: We shouldn't be ignoring it - I'll have to look. However, we cannot do anything about the warning you are seeing there as it is an artifact of flex, not us. We've had to endure it since the beginning. IIRC, somebody (@ggouaillardet ? ) came up with a way of silencing it once - but I'm not sure what might have brought it back, assuming my recollection is correct.

jjhursey: The "best" solution to the flex warning that I have seen is to not pass the `-Werror` to that specific part of the build. Not great, but it's something.

ax3l: @rhc54 did you find a way to pass my CFLAGS to the build process? I am trying to add address sanitizers to your CI so OpenMPI applications can be used with less suppression files in the future.

rhc54: I did - unfortunately, it causes the build to fail in many places as the picky mode generates warnings over the slightest thing (e.g., a function with a parameter that doesn't get used in that specific case because we have multiple variations of that function in our plugin system). I'm having to work thru the code base to get things to co-exist with it. Can't make it a high priority right now, so this may take a while to complete. Will try to periodically post progress on it.

ax3l: Ok, but can you tell my why exporting `CFLAGS` (as in the CI script) does not influence the build? How did you trigger `-Werror` to be properly applied?

ax3l: Urgh, autotools does not play well with `-Werror` either: https://lists.gnu.org/archive/html/bug-autoconf/2017-01/msg00008.html

ax3l:
Will open a new PR with further updates.

PR #1215, opened 2019-04-18
Add a new configure option specifically to enable non-global dlopen

rhc54:
Add --enable-nonglobal-dlopen to control adding libpmix to every MCA component. Default it to enabled.


ggouaillardet:
I think `--disable-nonglobal-dlopen` should be forced when `--enable-embedded-mode` is used but `--with-devel-headers` is not.
The rationale is it is incorrect to have a dynamic component depend on a static `libpmix.a`.
And in my understanding, this scenario can only occur in these conditions.

PR #1216, opened 2019-04-19
Default to disable-nonglobal-dlopen in embedded mode

rhc54:
Refs https://github.com/pmix/pmix/pull/1215#issuecomment-484771286


rhc54:
@ggouaillardet I think your comment is a valid point. I modified it a bit, however, to only disable nonglobal dlopen in that situation IF the user didn't specifically direct otherwise.

PR #1218, opened 2019-04-20
Fix case of multiple launcher calls in job script

### LINKED ISSUE #1211: Using a launcher to launch a script which invokes further launch commands

## Background information

master from git clone on power systems, TCP

-----------------------------

## Details of the problem

When processes are created by a launcher, that launcher is directed to call PMIx_server_setup_fork() to setup an environment and PMIx_server_register_client() to inform the convenience library of the process being created.  We do this so that the process being created can call PMIx_Init and have in its environment information on how to connect to the pmix server.   However, it is often the case that the thing being launched is not a PMIx client, but instead is a script containing commands to launch further applications.   For example:

`
% cat myjob

#!/bin/sh
mylauncher --np 16 a.out
mylauncher --np 64 b.out

% mylauncher --np 1 myjob`

I expect that many implementations of a launching system would implement mylauncher using a call to PMIx_tool_init().   It seems to me that these tools should not be influenced by their environment, but that they should always connect in the same way.  They are tools and not clients and shouldn't be looking in their environment, which is probably not intended for them.  An easy solution to this is to have mylauncher unset all relevant PMIX environment variables that are not meant for it.  If this is the best way, note that it is not a maintenance free approach since someone might introduce another connection method with new environment variables tomorrow, making the current list of unset's insufficient.  The other approach seems to be that PMIx_tool_init does not look in the environment for how to connect.  You have to pass in info's if you want to connect to the pmix server in a specific way.   It seems like this is the better approach.  What do others think?

rhc54:
When a launcher is invoked multiple times in a job script, each
invocation will have the same "client". The job-level info and tracking
objects are released after each invocation, so we have to recreate them.

Fixes #1211

### DIFF
--- a/src/include/pmix_globals.h
+++ b/src/include/pmix_globals.h
@@ -121,6 +121,9 @@ typedef uint8_t pmix_cmd_t;
 /* provide a "pretty-print" function for cmds */
 const char* pmix_command_string(pmix_cmd_t cmd);
 
+/* provide a hook to init tool data */
+PMIX_EXPORT extern pmix_status_t pmix_tool_init_info(void);
+
 /* define a set of flags to direct collection
  * of data during operations */
 typedef enum {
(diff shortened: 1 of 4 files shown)


rhc54:
@dsolt Please see if this fixes the problem in your environment - seems to solve it in mine.

PR #1222, opened 2019-04-22
Fix case of multiple launcher calls in job script

### LINKED ISSUE #1211: Using a launcher to launch a script which invokes further launch commands

rhc54:
When a launcher is invoked multiple times in a job script, each
invocation will have the same "client". The job-level info and tracking
objects are released after each invocation, so we have to recreate them.

Fixes #1211

(cherry picked from commit 04fd3e7735dca00850fca1e97449fa85836440cd)

### DIFF
--- a/src/include/pmix_globals.h
+++ b/src/include/pmix_globals.h
@@ -110,6 +110,9 @@ typedef uint8_t pmix_cmd_t;
 /* provide a "pretty-print" function for cmds */
 const char* pmix_command_string(pmix_cmd_t cmd);
 
+/* provide a hook to init tool data */
+PMIX_EXPORT extern pmix_status_t pmix_tool_init_info(void);
+
 /* define a set of flags to direct collection
  * of data during operations */
 typedef enum {
--- a/src/mca/psec/native/psec_native.c
+++ b/src/mca/psec/native/psec_native.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015-2018 Intel, Inc. All rights reserved.
+ * Copyright (c) 2015-2019 Intel, Inc.  All rights reserved.
  * Copyright (c) 2016      IBM Corporation.  All rights reserved.
  * Copyright (c) 2017      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
@@ -25,7 +25,7 @@
 #include "src/util/error.h"
 #include "src/util/output.h"
 
-#include "src/mca/psec/psec.h"
+#include "src/mca/psec/base/base.h"
 #include "psec_native.h"
 
 static pmix_status_t native_init(void);
@@ -49,14 +49,14 @@ pmix_psec_module_t pmix_native_module = {
 
 static pmix_status_t native_init(void)
 {
-    pmix_output_verbose(2, pmix_globals.debug_output,
+    pmix_output_verbose(2, pmix_psec_base_framework.framework_output,
                         "psec: native init");
     return PMIX_SUCCESS;
 }
 
 static void native_finalize(void)
 {
-    pmix_output_verbose(2, pmix_globals.debug_output,
+    pmix_output_verbose(2, pmix_psec_base_framework.framework_output,
                         "psec: native finalize");
 }
 
@@ -167,7 +167,7 @@ static pmix_status_t validate_cred(struct pmix_peer_t *peer,
     size_t n, m;
     uint32_t u32;
 
-    pmix_output_verbose(2, pmix_globals.debug_output,
+    pmix_output_verbose(2, pmix_psec_base_framework.framework_output,
                         "psec: native validate_cred %s",
                         (NULL == cred) ? "NULL" : "NON-NULL");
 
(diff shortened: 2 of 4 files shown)


jjhursey:
@dsolt can you confirm that this PR addresses the issue that you were seeing?

dsolt:
I get this when I run: 

Program received signal SIGSEGV, Segmentation fault.
0x000010000081468c in pmix_pointer_array_get_item (table=0x3fffffff8a18, element_index=15)
    at /smpi_tmp/dsolt/pmix/pmix-master/src/class/pmix_pointer_array.h:131
131         p = table->addr[element_index];
Missing separate debuginfos, use: debuginfo-install glibc-2.17-222.el7.ppc64le zlib-1.2.7-17.el7.ppc64le
(gdb) print element_index
$1 = 15
(gdb) print table->addr
$2 = (void **) 0x80

I will investigate, but I wanted to provide some status before our weekly meeting

dsolt:
#0  0x000010000081468c in pmix_pointer_array_get_item (table=0x3fffffff8a18, element_index=15)
    at /smpi_tmp/dsolt/pmix/pmix-master/src/class/pmix_pointer_array.h:131
#1  0x000010000081709c in pmix4_bfrops_base_pack_sizet (regtypes=0x3fffffff8a18, buffer=0x3fffffff8a88, 
    src=0x1, num_vals=4, type=35464) at bfrop_pmix4.c:611
#2  0x000010000028fd7c in pmix_bfrops_base_pack_buffer (
    regtypes=0x1000008301d8 <mca_bfrops_v4_component+264>, buffer=0x3fffffff8a18, src=0x3fffffff8a88, 
    num_vals=1, type=4) at base/bfrop_base_pack.c:90
#3  0x000010000028fbe8 in pmix_bfrops_base_pack (regtypes=0x1000008301d8 <mca_bfrops_v4_component+264>, 
    buffer=0x3fffffff8a18, src=0x3fffffff8a88, num_vals=1, type=4) at base/bfrop_base_pack.c:61
#4  0x00001000008167d8 in pmix4_pack (buffer=0x3fffffff8a18, src=0x3fffffff8a88, num_vals=1, type=4)
    at bfrop_pmix4.c:476
#5  0x0000100000b261d8 in send_connect_ack (sd=13, myflag=0x3fffffff8b3c "\003", iptr=0x100b68c0, niptr=2)
    at ptl_tcp.c:1175
#6  0x0000100000b25a64 in try_connect (uri=0x100bed60 "tcp4://127.0.0.1:45133", sd=0x3fffffff8ec4, 
    iptr=0x100b68c0, niptr=2) at ptl_tcp.c:1031
#7  0x0000100000b2770c in df_search (dirname=0x100a9e30 "/tmp/jsm.c712f6n01.9484596/1234", 
    prefix=0x100b66f0 "pmix.c712f6n01.tool", info=0x100b68c0, ninfo=2, sd=0x3fffffff8ec4, 
    nspace=0x3fffffff8eb8, rank=0x3fffffff8ec0, uri=0x3fffffff8ea0) at ptl_tcp.c:1491
#8  0x0000100000b24234 in connect_to_peer (peer=0x100b5ff0, info=0x0, ninfo=0) at ptl_tcp.c:679
#9  0x00001000002c3e48 in pmix_ptl_base_connect_to_peer (peer=0x100b5ff0, info=0x0, ninfo=0)
    at base/ptl_base_stubs.c:114
#10 0x0000100000248868 in PMIx_tool_init (proc=0x1004222c <my_pmix_proc>, info=0x0, ninfo=0)
    at tool/pmix_tool.c:587

dsolt:
Oops, I had some v4 pmix/.so files laying around that it was trying to pick up.   So, I tested this without our workaround of unsetting everything in the environment and this worked for me!

rhc54:
woohoo! Thanks

PR #1230, opened 2019-04-30
SPEC: allow splitting RPM in pmix and pmix-compat packages

kcgthb:
Introduces modification in the SPEC file to allow separate packaging of backward-compatibility PMI-1 and PMI-2 libraries and header files. 

This has been discussed in https://bugs.schedmd.com/show_bug.cgi?id=4511 and the idea is to allow sites to install a combination of PMIx (libpmix.so) and backward-compatibility PMI-1 (libpmi.so) and PMI2 (libpmi2.so) from either the PMIx project or Slurm, without conflicting with each other.

Right now, the PMIx 3.x RPM installs:
* `libpmi.so -> libpmi.so.1`
* `libpm2.so -> libpm2.so.1`
* `libpmix.so ->  libpmix.so.2`

and the `slurm-libpmi` RPM installs:
* `libpmi.so -> libpmi.so.0`
* `libpm2.so -> libpm2.so.0`

So the symlinks conflict but point to different versions of the same libraries. Meaning the the PMIx RPM cannot be installed in conjunction with `slurm-libpmi`, which is an issue for applications that are compiled against `libpmi{,2}.so.0`.

So this PR proposes to enable a flag within the SPEC file (`build_all_in_one_rpm`) which was already defined but not used, to permit splitting the PMIx package in `pmix` and `pmix-compat` when built with:
```
$ rpmbuild --define "build_all_in_one_rpm 0" -bb pmix.spec
```
Without setting that flag, nothing changes compared to the existing behavior: a monolithic RPM is built, which provides all the `pmi{,2,x}.so` libs.

With that flag defined:
* `pmix-compat` provides:
```
/usr/include/pmi.h
/usr/include/pmi2.h
/usr/lib64/libpmi.la
/usr/lib64/libpmi.so
/usr/lib64/libpmi.so.1
/usr/lib64/libpmi.so.1.0.0
/usr/lib64/libpmi2.la
/usr/lib64/libpmi2.so
/usr/lib64/libpmi2.so.1
/usr/lib64/libpmi2.so.1.0.0
```
* `pmix` provides `libpmix.*` and all the rest of the current `pmix` RPM contents.

This should allow sites to use PMI-1 and PMI-2 compatibility libs from Slurm alongside PMIx libraries from upstream.

### DIFF
--- a/contrib/pmix.spec
+++ b/contrib/pmix.spec
@@ -231,6 +231,22 @@ scalability.
 
 This RPM contains all the tools necessary to compile and link against PMIx.
 
+# if build_all_in_one_rpm = 0, build split packages
+%if !%{build_all_in_one_rpm}
+%package libpmi
+Summary: PMI-1 and PMI-2 compatibility libraries
+Requires: %{name}%{?_isa} = %{version}-%{release}
+Conflicts: slurm-libpmi
+
+%description libpmi
+The %{name}-libpmi package contains libpmi and libpmi2 libraries that provide
+the respective APIs and a copy of the PMIx library – each API is translated
+into its PMIx equivalent. This is especially targeted at apps/libs that are
+hardcoded to dlopen “libpmi” or “libpmi2”.
+This package conflicts sith slurm-libpmi, which provides its own, incompatible 
+versions of libpmi.so and libpmi2.so.
+%endif
+
 #############################################################################
 #
 # Prepatory Section
@@ -346,6 +362,10 @@ export CFLAGS CXXFLAGS FCFLAGS
 # We've had cases of config.log being left in the installation tree.
 # We don't need that in an RPM.
 find $RPM_BUILD_ROOT -name config.log -exec rm -f {} \;
+
+# If we build separate RPMs, then move the libpmi.* and libpmi2.* compat libs 
+# out of the way
+find $RPM_BUILD_ROOT -name 'libpmi.' | xargs rm -f
 
 # First, the [optional] modulefile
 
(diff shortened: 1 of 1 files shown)


contrib/pmix.spec
```diff
@@ -491,13 +511,31 @@ test "x$RPM_BUILD_ROOT" != "x" && rm -rf $RPM_BUILD_ROOT
...
 %doc README INSTALL LICENSE
 
+# if building separate RPMs, split the compat libs
+%if !%{build_all_in_one_rpm}
+%exclude %{_libdir}/libpmi.*
+%exclude %{_libdir}/libpmi2.*
+%exclude %{_includedir}/pmi.*
+%exclude %{_includedir}/pmi2.*
+
+%files compat
```
wickberg: I'd suggest naming this package "libpmi", as that's exactly what's in the package, and this would line up with what Slurm has been shipping as a Conflict in our slurm.spec file for the past couple of releases.

%package libpmi
Summary: Slurm\'s implementation of the pmi libraries
Group: System Environment/Base
Requires: %{name}%{?_isa} = %{version}-%{release}
Conflicts: pmix-libpmi
%description libpmi
Slurm\'s version of libpmi. For systems using Slurm, this version
is preferred over the compatibility libraries shipped by the PMIx project.

kcgthb: Good point! I'll let the PMIx packagers / devs chime in and decide.

rhc54: Sounds okay with me! Thanks so much for doing this! I do, however, need you to sign off the commit with the usual Linux signature:
```shell
Signed-off-by: My name <email address>
```

kcgthb: All right! 
The additional RPM package has been renamed to `pmix-libpmi`, and I signed the commit off.

rhc54: Thanks again!!

rhc54:
Failures have nothing to do with this one. We will investigate separately.

kcgthb:
Thanks for merging!

PR #1235, opened 2019-05-04
Fix node-local Fence detection

artpol84:


artpol84:
@rhc54 @jjhursey I think this is what was causing the issues with the fence I mentioned on the call this week.

rhc54:
@artpol84 Is this a problem in v3.1.x as well? Just wondering if this needs to go to the release branches given the detection issue (i.e., direct modex hid it).

artpol84:
I will check.
I thought this local fence optimization only in master.

artpol84:
I don't see the local tracker check in v3.1:
https://github.com/pmix/pmix/blob/v3.1/src/server/pmix_server_ops.c#L460

rhc54:
ok, thanks for checking - just wanted to make sure we aren't missing something there.

PR #1237, opened 2019-05-08
Avoid double-free of tracker

### LINKED ISSUE #1236: pmix server aborts when returning PMIX_OPERATION_SUCCEEDED in server_fencenb_cb()
v3.1.3rc1

Compiled from the release candidate tarball using
configure --enable-debug CFLAGS=-DHAVE_SYSLOG

Operating system/version: Fedora release 29
Computer hardware: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
Network type: none (local only)

-----------------------------

## Details of the problem

Using the pmix server library, after standard version 3.1 section 11.2.5 it should be fine to return PMIX_OPERATION_SUCCEEDED immediately and not to call cbfunc in pmix_server_module_t's fence_nb function pointer function of type pmix_server_fencenb_fn_t. If I do this, the process aborts with the following backtrace:
```
#0  0x00007ff4fb9b357f in raise () at /lib64/libc.so.6
#1  0x00007ff4fb99d895 in abort () at /lib64/libc.so.6
#2  0x00007ff4fb99d769 in _nl_load_domain.cold.0 () at /lib64/libc.so.6
#3  0x00007ff4fb9aba26 in .annobin_assert.c_end () at /lib64/libc.so.6
#4  0x00007ff4fa867766 in server_switchyard (peer=0x7ff4f4017500, tag=103, buf=0x7ff4fa709b40)
    at ../../src/server/pmix_server.c:3308
#5  0x00007ff4fa86a3db in pmix_server_message_handler (pr=0x7ff4f4017500, hdr=0x7ff4f4015a78, buf=0x7ff4fa709b40, cbdata=0x0)
    at ../../src/server/pmix_server.c:3536
#6  0x00007ff4fa8eff96 in pmix_ptl_base_process_msg (fd=-1, flags=4, cbdata=0x7ff4f40159a0)
    at ../../../../src/mca/ptl/base/ptl_base_sendrecv.c:797
#7  0x00007ff4fa7b5dc5 in event_process_active_single_queue () at /lib64/libevent-2.1.so.6
#8  0x00007ff4fa7b67c7 in event_base_loop () at /lib64/libevent-2.1.so.6
#9  0x00007ff4fa88cfbf in progress_engine (obj=0x1c23c40) at ../../src/runtime/pmix_progress_threads.c:109
#10 0x00007ff4fb64558e in start_thread () at /lib64/libpthread.so.0
#11 0x00007ff4fba78683 in clone () at /lib64/libc.so.6
```

Debug output:
```
pmix[21430]: SWITCHYARD for pspmix_0x000053b1[0:21425]_21492:0:16
pmix[21430]: recvd pmix cmd FENCE from pspmix_0x000053b1[0:21425]_21492:0
pmix[21430]: recvd FENCE
pmix[21430]: recvd fence from pspmix_0x000053b1[0:21425]_21492:0 with 1 procs
pmix[21430]: get_tracker called with 1 procs
pmix[21430]: fence complete
systemd[1]: Started Process Core Dump (PID 21475/UID 0).
psid[20382]: Received SIGCHLD for pid 21430 (0x53b6) with status 0 after signal 6
systemd-coredump[21476]: Process 21430 (psid) of user 0 dumped core.
                                                   Stack trace of thread 21433:
                                                   #0  0x00007ff4fb9b357f n/a (/usr/lib64/libc-2.28.so)
```

The program that I used to test this is the dmodex example coming with pmix. The first fence in there is working since for the data exchange, the host server needs to do something and thus later calls cbfunc, but the second fence called without data produces the described abort after PMIX_OPERATION_SUCCEEDED is returned directly.

rhc54:
Fixes #1236 

### DIFF
--- a/src/server/pmix_server_ops.c
+++ b/src/server/pmix_server_ops.c
@@ -1457,10 +1457,29 @@ pmix_status_t pmix_server_disconnect(pmix_server_caddy_t *cd,
         pmix_list_get_size(&trk->local_cbs) == trk->nlocal) {
         trk->host_called = true;
         rc = pmix_host_server.disconnect(trk->pcs, trk->npcs, trk->info, trk->ninfo, cbfunc, trk);
-        if (PMIX_SUCCESS != rc) {
-            /* remove this contributor from the list - they will be notified
-             * by the switchyard */
+        if (PMIX_SUCCESS != rc && PMIX_OPERATION_SUCCEEDED != rc) {
+            /* clear the caddy from this tracker so it can be
+             * released upon return - the switchyard will send an
+             * error to this caller, and so the op completion
+             * function doesn't need to do so */
             pmix_list_remove_item(&trk->local_cbs, &cd->super);
+            cd->trk = NULL;
+            /* we need to ensure that all other local participants don't
+             * just hang waiting for the error return, so execute
+             * the op completion function - it threadshifts the call
+             * prior to processing, so it is okay to call it directly
+             * from here */
+            trk->host_called = false; // the host will not be calling us back
+            cbfunc(rc, trk);
+        } else if (PMIX_OPERATION_SUCCEEDED == rc) {
+            /* the operation was atomically completed and the host will
+             * not be calling us back - ensure we notify all participants.
+             * the cbfunc thread-shifts the call prior to processing,
+             * so it is okay to call it directly from here */
+            trk->host_called = false; // the host will not be calling us back
+            cbfunc(PMIX_SUCCESS, trk);
+            /* ensure that the switchyard doesn't release the caddy */
+            rc = PMIX_SUCCESS;
         }
     } else {
         rc = PMIX_SUCCESS;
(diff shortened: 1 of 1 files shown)


src/server/pmix_server_ops.c
```diff
@@ -730,8 +730,10 @@ pmix_status_t pmix_server_fence(pmix_server_caddy_t *cd,
                                        trk->info, trk->ninfo,
                                        data, sz, trk->modexcbfunc, trk);
         if (PMIX_SUCCESS != rc) {
+            pmix_list_remove_item(&trk->local_cbs, &cd->super);
```
jjhursey: Do we need this line in `master` as well [here](https://github.com/pmix/pmix/blob/master/src/server/pmix_server_ops.c#L1036-L1039)?

rhc54: There is an alternate solution currently in master, but I think this approach is better - will have to bring it in, remove the current solution, and verify it is still okay.

krempel-pt:
@rhc54 I can confirm that the abort is gone with the new patch. Thanks.

jjhursey:
Thank you @krempel-pt. We will merge this in and roll a v3.1.3rc3 with the fix - probably later today.

krempel-pt:
@jjhursey Perhaps you should wait with 3.1.3rc3 until #1236 is completely fixed, seems that the abort was not the only problem at this corner.

jjhursey:
Yeah, I just saw the comment over there. We'll hold off until that issue is resolved. We have another outstanding issue that may gate rc3 as well.

rhc54:
Yeah, I see the problem. I'm on vacation for a few more days, traveling back home for three of them, but will try to poke on this as time permits.

rhc54:
@krempel-pt Pretty sure this should get it - please give it a try.

krempel-pt:
@rhc54 Yeah, it's working now, thank you.

PR #1239, opened 2019-05-09
Define host/network conversion for size_t

### LINKED ISSUE #1238: Type mismatch in PTL header field portability transformation
The datatype for `nbytes` field in the header is `size_t`:
https://github.com/pmix/pmix/blob/master/src/mca/ptl/ptl_types.h#L113

However `htonl` function is used that only address uint32_t
https://github.com/pmix/pmix/blob/master/src/mca/ptl/ptl_types.h#L279

rhc54:
Use the appropriate size for the given configuration

Fixes #1238 

### DIFF
--- a/src/include/types.h
+++ b/src/include/types.h
@@ -144,6 +144,19 @@ static inline uint64_t pmix_ntoh64(uint64_t val)
 #endif
 }
 
+/* Convert size_t value from host to network byte order and back */
+#if SIZEOF_SIZE_T == 4
+
+#define pmix_htonsizet(x) htonl(x)
+#define pmix_ntohsizet(x) ntohl(x)
+
+#elif SIZEOF_SIZE_T == 8
+
+#define pmix_htonsizet(x) pmix_hton64(x)
+#define pmix_ntohsizet(x) pmix_ntoh64(x)
+
+#endif
+
 
 /**
  * Convert between a local representation of pointer and a 64 bits value.
--- a/src/mca/ptl/base/ptl_base_sendrecv.c
+++ b/src/mca/ptl/base/ptl_base_sendrecv.c
@@ -264,8 +264,8 @@ static pmix_status_t send_msg(int sd, pmix_ptl_send_t *msg)
     iov[0].iov_len = msg->sdbytes;
     if (!msg->hdr_sent && NULL != msg->data) {
         iov[1].iov_base = msg->data->base_ptr;
-        iov[1].iov_len = ntohl(msg->hdr.nbytes);
-        remain += ntohl(msg->hdr.nbytes);
+        iov[1].iov_len = pmix_ntohsizet(msg->hdr.nbytes);
+        remain += pmix_ntohsizet(msg->hdr.nbytes);
         iov_count = 2;
     } else {
         iov_count = 1;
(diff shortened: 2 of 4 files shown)


artpol84:
@rhc54
This might break backwards compatibility

rhc54:
It passed all the xversion tests, both in CI and on my Mac. Do you have some reason to believe it would break it?

artpol84:
Because the old version would use ntohl and work with the first 4 bytes of the field while new version will consider all 8 bytes.

For example, lets consider 8-byte nbytes field in the header and size = 4.
* old way: 0x0000 0004 0000 0000 (only 4 bytes are filled and the rest is untouched)
* new way: 0x0000 0000 0000 0004  

So if the old code will pack it and the new code will unpack the following value will be obtained:
0x0004 0000 0000 instead of 4

If vice-versa: the old code will obtain 0 as it wouldn't look into the higher portion of the data.

It is very strange that cross-version haven't caught it. I'm pretty sure it will be that way.

rhc54:
The 8-byte converter simply converts each 32-bit piece separately - it doesn't intermix them. So if you take a value that is in the lower 4-bytes, it will convert into the lower 4-bytes of the result We then convert those lower 4-bytes back on the remote end.

Bottom line: as long as the message size was under 32-bits in size, it should always resolve to the same value. A message that large is feasible but highly unlikely for PMIx.

artpol84:
If 64-byte version works the way you describe it is erroneous. 
Here is an example. Say you work with the value `0x1122 3344 5566 77888`
Here is it's representation in both endianess:
* `BE: 1122 3344 5566 7788`
* `LE: 8877 6655 4433 2211`

Let's assume you pack on the BE machine and unpack on LE. On BE machine converter essentially won't do anything as network byte order is BE as well. But on LE machine it'll break the number:
* `ntohl(1122 3344) =  4433 2211`
* `ntohl(5566 7788) = 8877 6655`
Concatenating those we get `4433 2211 8877 6655` and should get `8877 6655 4433 2211`

rhc54:
Okay, let me try to be more precise.  The exact algorithm does the following:
```
    r.l[0] = htonl(w.l[1]);
    r.l[1] = htonl(w.l[0]);
```
In other words, we do the 32-bit ```htonl``` on each 4-byte piece of the 64-bit number. We don't intermix those results, but we do swap the order of the two 4-byte pieces.

Hence, when you use a straight `ntohl` on the backend, you convert what was the lowest 4-bytes of the sender's value. Thus, unless someone sends more than 2**32 bytes of data, it won't matter.

artpol84:
That is correct. Unless you stay inside 4-bytes you are fine.
My point was more general. I'm saying that `pmix_hton64()` function is incorrect and will not work as expected for values larger than 2^32 if anywhere in PMIx such values will appear.

rhc54:
I think it is more correct to say that the matching of `pmix_hton64` with `ntohl` will not work for large values - the function itself is correct. Regardless, if anyone is sending messages that large inside of PMIx, then they deserve to have it fail. No excuse for a message of such size in this system 😄

artpol84:
Sure,
But this function may be used for some other purposes that are valid. And in that case, one might hit the problem.

artpol84:
> I think it is more correct to say that the matching of pmix_hton64 with ntohl will not work for large values - the function itself is correct.
This is my point- the algorithm you provided is incorrect in the sense that itwill not work between little- and big-endian machines

artpol84:
Seems like I’ve missed switching of 4-byte values:

artpol84:
Follow up on the call: the header size determination:
https://github.com/pmix/pmix/blob/master/src/mca/ptl/base/ptl_base_sendrecv.c#L506

PR #1240, opened 2019-05-10
PTL: Fix connection establishment protocol (with refactoring)

artpol84:
On the server side the folowing messages are sent to the client during connection:
* (4B) Security verification
* (4B) Status
* (4B) Server-local index of the peer

client only receives 2 messgaes:
* (4B) Status
* (4B) Server-local index of the peer

For unknown reason in most of the time extra 4 bytes were consumed by the client.
But sometimes these extra bytes were prepending a valid message shifting tag and
size away and PMIx_Init hangs were observed.

### DIFF
--- a/src/mca/psec/psec.h
+++ b/src/mca/psec/psec.h
@@ -5,6 +5,8 @@
  *
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
+ * Copyright (c) 2019      Mellanox Technologies, Inc.
+ *                         All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -155,22 +157,32 @@ PMIX_EXPORT pmix_psec_module_t* pmix_psec_base_assign_module(const char *options
                 pmix_output_verbose(2, pmix_globals.debug_output,                                           \
                                     "credential validated");                                                \
             }                                                                                               \
-            /* send them the result */                                                                      \
-            if (PMIX_SUCCESS != (_r = pmix_ptl_base_send_blocking((p)->sd, (char*)&(_r), sizeof(int)))) {   \
-                PMIX_ERROR_LOG(_r);                                                                         \
-            }                                                                                               \
             (r) = _r;                                                                                       \
+        } else if (NULL != (p)->nptr->compat.psec->server_handshake) {                                      \
+            /* request the handshake if the security mode calls for it */                                   \
+            pmix_output_verbose(2, pmix_globals.debug_output,                                               \
+                                "requesting handshake");                                                    \
+            _r = PMIX_ERR_READY_FOR_HANDSHAKE;                                                              \
+            (r) = _r;                                                                                       \
+        } else {                                                                                            \
+            /* this is not allowed */                                                                       \
+            (r) = PMIX_ERR_NOT_SUPPORTED;                                                                   \
+        }                                                                                                   \
+    } while(0)
+
+
+#define PMIX_PSEC_SERVER_HANDSHAKE(r, p, d, nd, in, nin, c)                                                        \
+    do {                                                                                                    \
+        int _r;                                                                                             \
+        /* if a credential is available, then check it */                                                   \
+        if (NULL != (p)->nptr->compat.psec->validate_cred) {                                                \
+            (r) = PMIX_SUCCESS;                                                                             \
         } else if (NULL != (p)->nptr->compat.psec->server_handshake) {                                      \
             /* execute the handshake if the security mode calls for it */                                   \
             pmix_output_verbose(2, pmix_globals.debug_output,                                               \
                                 "executing handshake");                                                     \
-            _r = PMIX_ERR_READY_FOR_HANDSHAKE;                                                              \
-            if (PMIX_SUCCESS != (_r = pmix_ptl_base_send_blocking((p)->sd, (char*)&(_r), sizeof(int)))) {   \
+            if (PMIX_SUCCESS != (_r = p->nptr->compat.psec->server_handshake((p)->sd))) {                   \
                 PMIX_ERROR_LOG(_r);                                                                         \
-            } else {                                                                                        \
-                if (PMIX_SUCCESS != (_r = p->nptr->compat.psec->server_handshake((p)->sd))) {               \
-                    PMIX_ERROR_LOG(_r);                                                                     \
-                }                                                                                           \
             }                                                                                               \
             (r) = _r;                                                                                       \
         } else {                                                                                            \
@@ -178,6 +190,7 @@ PMIX_EXPORT pmix_psec_module_t* pmix_psec_base_assign_module(const char *options
             (r) = PMIX_ERR_NOT_SUPPORTED;                                                                   \
         }                                                                                                   \
     } while(0)
+
 
 /****    COMPONENT STRUCTURE DEFINITION    ****/
 
--- a/src/mca/ptl/tcp/ptl_tcp.c
+++ b/src/mca/ptl/tcp/ptl_tcp.c
@@ -15,6 +15,8 @@
  *                         reserved.
  * Copyright (c) 2013-2019 Intel, Inc.  All rights reserved.
  * Copyright (c) 2018      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2019      Mellanox Technologies, Inc.
+ *                         All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -1249,17 +1251,24 @@ static pmix_status_t recv_connect_ack(int sd, uint8_t myflag)
         return rc;
     }
     reply = ntohl(u32);
+    /* see if they want us to do the handshake */
+    if (PMIX_ERR_READY_FOR_HANDSHAKE == reply) {
+        PMIX_PSEC_CLIENT_HANDSHAKE(rc, pmix_client_globals.myserver, sd);
+        if (PMIX_SUCCESS != rc) {
+            return rc;
+        }
+        /* Receive the final status of the handshake */
+        rc = pmix_ptl_base_recv_blocking(sd, (char*)&u32, sizeof(uint32_t));
+        if (PMIX_SUCCESS != rc) {
+            return rc;
+        }
+        reply = ntohl(u32);
+    }
+    if (PMIX_SUCCESS != reply) {
+        return reply;
+    }
 
     if (0 == myflag) {
-        /* see if they want us to do the handshake */
-        if (PMIX_ERR_READY_FOR_HANDSHAKE == reply) {
-            PMIX_PSEC_CLIENT_HANDSHAKE(rc, pmix_client_globals.myserver, sd);
-            if (PMIX_SUCCESS != rc) {
-                return rc;
-            }
-        } else if (PMIX_SUCCESS != reply) {
-            return reply;
-        }
         pmix_output_verbose(2, pmix_ptl_base_framework.framework_output,
                             "pmix: RECV CONNECT CONFIRMATION");
 
@@ -1269,11 +1278,7 @@ static pmix_status_t recv_connect_ack(int sd, uint8_t myflag)
             return rc;
         }
         pmix_globals.pindex = ntohl(u32);
-    } else {  // we are a tool
-        /* if the status indicates an error, then we are done */
-        if (PMIX_SUCCESS != reply) {
-            return reply;
-        }
+    } else { // we are a tool
         /* if we needed an identifier, recv it */
         if (3 == myflag || 6 == myflag) {
             /* first the nspace */
@@ -1315,21 +1320,6 @@ static pmix_status_t recv_connect_ack(int sd, uint8_t myflag)
                             pmix_globals.myid.nspace, pmix_globals.myid.rank,
                             pmix_client_globals.myserver->info->pname.nspace,
                             pmix_client_globals.myserver->info->pname.rank);
-
-        /* get the returned status from the security handshake */
-        pmix_ptl_base_recv_blocking(sd, (char*)&reply, sizeof(pmix_status_t));
-        if (PMIX_SUCCESS != reply) {
-            /* see if they want us to do the handshake */
-            if (PMIX_ERR_READY_FOR_HANDSHAKE == reply) {
-                PMIX_PSEC_CLIENT_HANDSHAKE(reply, pmix_client_globals.myserver, sd);
-                if (PMIX_SUCCESS != reply) {
-                    return reply;
-                }
-                /* if the handshake succeeded, then fall thru to the next step */
-            } else {
-                return reply;
-            }
-        }
     }
 
     if (sockopt) {
--- a/src/mca/ptl/tcp/ptl_tcp_component.c
+++ b/src/mca/ptl/tcp/ptl_tcp_component.c
@@ -16,6 +16,8 @@
  * Copyright (c) 2017-2018 Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2018-2019 IBM Corporation.  All rights reserved.
+ * Copyright (c) 2019      Mellanox Technologies, Inc.
+ *                         All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -985,7 +987,7 @@ static void connection_handler(int sd, short args, void *cbdata)
     pmix_ptl_hdr_t hdr;
     pmix_peer_t *peer;
     pmix_rank_t rank=0;
-    pmix_status_t rc;
+    pmix_status_t rc, reply;
     char *msg, *mg, *version;
     char *sec, *bfrops, *gds;
     pmix_bfrop_buffer_type_t bftype;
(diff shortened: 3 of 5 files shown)


artpol84:
@jjhursey this issue I mentioned on the call.
I still see another issue but it is less frequent and happens in finalize.

artpol84:
@rhc54 I spent time debugging it and want to work on the fix further.
Please do not close my PRs.
You can provide your thoughts but I’d like to have a fix

artpol84:
I will address the cases you mentioned and will update PR tomorrow

artpol84:
Btw for handshake it should work fine I think.
Only absence of security module might be an issue

rhc54:
No problem - happy to let you resolve it if you like. The problem is that the handshake and "no-module" paths both require that the final resolution be returned to the client. The correct patch is probably to simply remove the code that returned status in the macro for the "non-handshake" path, letting the code in the main function return the result for all three paths.

rhc54:
If you feel it would help, you are more than welcome to replace the security handshake macro with the actual code - nothing sacred about the macro. The intent is that the server always makes final determination of the outcome and communicates it to the client, regardless of the security mechanism in use.

src/mca/psec/psec.h
```diff
@@ -155,22 +155,32 @@ PMIX_EXPORT pmix_psec_module_t* pmix_psec_base_assign_module(const char *options
                 pmix_output_verbose(2, pmix_globals.debug_output,                                           \
                                     "credential validated");                                                \
             }                                                                                               \
-            /* send them the result */                                                                      \
-            if (PMIX_SUCCESS != (_r = pmix_ptl_base_send_blocking((p)->sd, (char*)&(_r), sizeof(int)))) {   \
```
artpol84: @rhc54 note, that here a htonl conversion was missing.

src/mca/ptl/tcp/ptl_tcp.c
```diff
@@ -1249,17 +1249,24 @@ static pmix_status_t recv_connect_ack(int sd, uint8_t myflag)
         return rc;
     }
     reply = ntohl(u32);
```
artpol84: while here ntohl conversion is present.
It was working because the representation of 0x0 is the same in both LE and BE.

src/mca/ptl/tcp/ptl_tcp_component.c
```diff
@@ -1746,68 +1754,9 @@ static void process_cbfunc(int sd, short args, void *cbdata)
     /* acquire the object */
     PMIX_ACQUIRE_OBJECT(cd);
 
-    /* send this status so they don't hang */
```
artpol84: This was moved down to the place where a `peer` is fully initialized.

src/mca/ptl/tcp/ptl_tcp.c
```diff
@@ -1315,21 +1318,6 @@ static pmix_status_t recv_connect_ack(int sd, uint8_t myflag)
                             pmix_globals.myid.nspace, pmix_globals.myid.rank,
                             pmix_client_globals.myserver->info->pname.nspace,
                             pmix_client_globals.myserver->info->pname.rank);
-
```
artpol84: IIUC we don't have psec modules with handshake so far.
So this change should be fine.

One clarification:
It seems to me that this code is not quite correct:
* Status message is received above (generic code for both client and server)
* but here you do another receive before and not after client's handshake.
If I understand the right way would be to perform the handshake and only afterward exchange with messages.

src/mca/ptl/tcp/ptl_tcp_component.c
```diff
@@ -1896,16 +1845,78 @@ static void process_cbfunc(int sd, short args, void *cbdata)
...
     cred.size = pnd->len;
-    PMIX_PSEC_VALIDATE_CONNECTION(rc, peer, NULL, 0, NULL, NULL, &cred);
-    if (PMIX_SUCCESS != rc) {
-        pmix_output_verbose(2, pmix_ptl_base_framework.framework_output,
-                            "validation of tool credentials failed: %s",
-                            PMIx_Error_string(rc));
-        PMIX_RELEASE(peer);
-        pmix_list_remove_item(&pmix_server_globals.nspaces, &nptr->super);
-        PMIX_RELEASE(nptr);  // will release the info object
+    PMIX_PSEC_VALIDATE_CONNECTION(reply, peer, NULL, 0, NULL, NULL, &cred);
```
artpol84: As I haven't found any psec component that supports handshake, I think it is ok to move this above the subsequent sends.
And it is more logical to have authentification performed as the very first step.

src/mca/ptl/tcp/ptl_tcp_component.c
```diff
@@ -1896,16 +1845,78 @@ static void process_cbfunc(int sd, short args, void *cbdata)
...
         CLOSE_THE_SOCKET(pnd->sd);
-        goto done;
+        PMIX_RELEASE(pnd->peer);
+        PMIX_RELEASE(pnd);
+        PMIX_RELEASE(cd);
+        return;
+    }
+
+    /* if we got an identifier, send it back to the tool */
+    if (pnd->need_id) {
```
artpol84: Also, here
https://github.com/pmix/pmix/pull/1240/files#diff-2f3e46bf48df10ec8101693fba191ef3R1283
I don't see that this case is expected in any way.
Only 2 receives are performed instead of 4.

artpol84:
@rhc54, what would be a good way to check that tools communication is not broken?

src/mca/ptl/tcp/ptl_tcp_component.c
rhc54: Some of us have components that implement handshakes (e.g., openssl) outside of the core code due to security requirements of specific customers. Let's make sure we don't break that support.

src/mca/ptl/tcp/ptl_tcp.c
rhc54: The handshake itself exchanges messages as part of its procedure. I'd have to look at the broader code to understand the situation.

src/mca/ptl/tcp/ptl_tcp.c
rhc54: Probably also because I'm unaware of anyone actually using mixed endian clusters any more. Still, we try to maintain it just as good practice.

src/mca/ptl/tcp/ptl_tcp_component.c
rhc54: Again, I'd have to look at the broader context to understand the scenario. Will do so upon return from vacation.

rhc54:
> @rhc54, what would be a good way to check that tools communication is not broken?

Easiest would be to switch to the `test/simple` directory and run `./simptest -e ./simptool`.

src/mca/ptl/tcp/ptl_tcp.c
artpol84: But here if one side performs htonl, but another side doesn't it will lead to an issue even on homogeneous cluster

src/mca/ptl/tcp/ptl_tcp.c
artpol84: So let me rephrase:
Does it make sense to send the message before server has completed handshake?
IMO it doesn't, but it seems to be here in this sequence.

src/mca/ptl/tcp/ptl_tcp_component.c
artpol84: Sure.
How can we do that?

src/mca/ptl/tcp/ptl_tcp.c
artpol84: Bringing the whole piece in question here:
```C
        pmix_ptl_base_recv_blocking(sd, (char*)&reply, sizeof(pmix_status_t));	
        if (PMIX_SUCCESS != reply) {	
            /* see if they want us to do the handshake */	
            if (PMIX_ERR_READY_FOR_HANDSHAKE == reply) {	
                PMIX_PSEC_CLIENT_HANDSHAKE(reply, pmix_client_globals.myserver, sd);	
                if (PMIX_SUCCESS != reply) {	
                    return reply;	
                }	
                /* if the handshake succeeded, then fall thru to the next step */	
            } else {	
                return reply;	
            }	
        }
```

artpol84: The original status was already received above.
And after this code piece I don't see any additional exchange to confirm that handshake was successful

artpol84:
```
$ ./simptest -e ./simptool
Testing version 4.0.0a1
[artpol-VirtualBox:03736] SERVER: ERRHANDLER REGISTRATION CALLBACK CALLED WITH STATUS 0, ref=0
Collecting inventory
[artpol-VirtualBox:03736] INVENTORY RECEIVED
Inventory collected: 0
[artpol-VirtualBox:03736] INVENTORY READY FOR DELIVERY
[artpol-VirtualBox:03736] SERVER: TOOL CONNECT
[artpol-VirtualBox:03752] Tool ns foobar rank 0: Running
[artpol-VirtualBox:03736] SERVER: QUERY
[artpol-VirtualBox:03736] 	Key: foobar
[artpol-VirtualBox:03736] 	Key: spastic
[artpol-VirtualBox:03752] Client ns foobar rank 0: Finalizing
[artpol-VirtualBox:03736] SERVER: FINALIZED foobar:0 WAKEUP 1
	pmix.pgm.model:	PMIX
	pmix.mdl.name:	test
	pmix.evname:	SIMPTEST-MODEL
Test finished OK!
```

artpol84:
Looks like working.
We should implement the same thing for the test suite so we will cover tools as well.

rhc54:
> Looks like working.
> We should implement the same thing for the test suite so we will cover tools as well.

Sure, I can do that easily enough. Good idea.

src/mca/ptl/tcp/ptl_tcp_component.c
rhc54: We used to have an openssl component but we removed it awhile ago and now provide it only upon customer request. It will take me some time to get it released again to the public repo. Meantime, I can check it when I get back next week.

artpol84:
I was thinking about the handshake codepath and I don't think it would work at all in the current code base for the following reasons.

* Server would send a non-converted code for PMIX_ERR_READY_FOR_HANDSHAKE [here](https://github.com/pmix/pmix/blob/master/src/mca/psec/psec.h#L167:L168). This will be value `-14` = `0xFFFFFFF1`
* Client would receive it with conversion [here](https://github.com/pmix/pmix/blob/master/src/mca/ptl/tcp/ptl_tcp.c#L1251). This would result in `ntohl(0xFFFFFFF1) = 0x1FFFFFFF` which is positive and not equal to original -14.

Bottom line this sort of contributes to my statement that we are not breaking backward compatibility with tools as it was simply not working with the handshake.

> We used to have an openssl component but we removed it awhile ago and now provide it only upon customer request. It will take me some time to get it released again to the public repo. Meantime, I can check it when I get back next week.

This would be helpful because my original perception of the handshake logic was that server only needs to communicate that it is waiting for the handshake to begin and that afterwards client-server communication (handshake) will happen over a separate channel and no additional confirmation would be requires. So I was surprised by the following:

> The problem is that the handshake and "no-module" paths both require that the final resolution be returned to the client.

As I interpret that as in the handshake sequence:
* Server sends PMIX_ERR_READY_FOR_HANDSHAKE
* They do handshake
* Another message is required to confirm that server has accepted the connection.

artpol84:
@rhc54, as you know we have some urgency with 3.1.3 release.
If you can propose any way to speed up verification of this PR would be great.
Maybe you can send me OpenSSL component and I’ll test it myself today.

artpol84:
@rhc54?

rhc54:
Sorry - I returned from vacation last night and am only just beginning to find the bottom of my work inbox, let alone finding time to deal with this issue. I should be able to make time tomorrow.

>  Maybe you can send me OpenSSL component and I’ll test it myself today.

Somehow, I don't recall "email it to your competitor" being one of the options in my corporate software release manual 😄

artpol84:
Ok, thanks.
Regarding OpenSSL - it wasn’t clear to me that it is a private component.

artpol84:
I decided to summarize my observations to communicate my current understanding:

# If `validate_cred` psec is used
## For regular client connection & PTL/tcp (if all previous steps are successful)
1. Validate credentials [srv]
1. **Communicate** validation status [srv + cli]
1. **Communicate** the PMIX_SUCCESS status [srv, **cli doesn't expect - bug**]
1. **Communicate** client's array index [srv + cli]

### Comments:
* Currently, step (3) is not handled by the client. This is the reason for this PR.
* step (2) seems to be exhaustive as the status of the validation can be communicated on the step (3) which is always there.
* in PTL/usock step(3) is missing.

## For tools connection & PTL/tcp (if all previous steps are successful)
1. **Communicate** the current status (will PMIX_SUCCESS) [srv, cli]
1. **Communicate** `pmix_globals.myid.nspace`
1. **Communicate** `pmix_globals.myid.rank`
1. Validate credentials [srv]
1. **Communicate** validation status [srv + cli]

### Comments:
* The sequence of status (1) and validation status (3) exchanges is opposite compared to the client connection (steps (2) and (3)). This is confusing.

# If `handshake` psec is used
## For regular client connection & PTL/tcp (if all previous steps are successful)
1. **Communicate** validation status == READY_FOR_HANDSHAKE [srv + cli]
1. Perform handshake
1. **Communicate** the the status of handshake [srv, **cli doesn't expect - bug**]
1. **Communicate** client's array index [srv + cli]

### Comments:
* Given that this is a handshake, do we really need to communicate the status of it? Or client/server are capable to agree over this handshake channel?
* As PMIX_PSEC_VALIDATE_CONNECTION (server) doesn't use `htonl` to transform PMIX_ERR_READY_FOR_HANDSHAKE [[link](https://github.com/pmix/pmix/blob/master/src/mca/psec/psec.h#L167)] and the client performs `ntohl` [[link](https://github.com/pmix/pmix/blob/master/src/mca/ptl/tcp/ptl_tcp.c#L1251)], I don't think that this codepath is working based on the discussion here https://github.com/pmix/pmix/pull/1240#issuecomment-491475003.

## For tools connection & PTL/tcp (if all previous steps are successful)
1. **Communicate** the current status (will PMIX_SUCCESS) [srv, cli]
1. **Communicate** `pmix_globals.myid.nspace`
1. **Communicate** `pmix_globals.myid.rank`
1. **Communicate** validation status == READY_FOR_HANDSHAKE [srv + cli]
1. Perform handshake

### Comments:
* In this scenario the result of the handshake is not being communicated. At least there is no communication after the VALIDATION macro [[link](https://github.com/pmix/pmix/blob/master/src/mca/ptl/tcp/ptl_tcp_component.c#L1896:L1909)].

## Changes in this PR:
* PSEC: Validate credentials **OR** for handshake return PMIX_ERR_READY_FOR_HANDSHAKE [srv]
* **Communicate** PSEC status [srv + cli]
* if Handshake mode: 
  * perform handshake; 
  * **Communicate** handshake status (See the question about requirement to communicate this above). [srv + cli]
For clients:
* **Communicate** client's array index [srv + cli]

For tools:
* **Communicate** `pmix_globals.myid.nspace`
* **Communicate** `pmix_globals.myid.rank`

artpol84:
I asked @karasevb to check this PR with PTL/usock. He found some unrelated bug and will push the fix into this PR in few hours..

karasevb:
I tested by `make check` with ptl/usock, all iters (>1000) were successful. 
The first commit (06ef19e) should cherry-picked to branches, and the second one intended only for the master branch.

artpol84:
Thanks, @karasevb !

src/mca/ptl/usock/ptl_usock.c
```diff
@@ -566,7 +566,7 @@ void pmix_usock_send_handler(int sd, short flags, void *cbdata)
     pmix_peer_t *peer = (pmix_peer_t*)cbdata;
     pmix_ptl_send_t *msg = peer->send_msg;
     pmix_status_t rc;
-    uint32_t nbytes;
+    size_t nbytes;
```
artpol84: @rhc54 note, that effectively this change is breaking backward compatibility with previous versions as we discussed in https://github.com/pmix/pmix/pull/1239.

artpol84: I mean not this one, but the one that caused it in #1239

rhc54: I'm looking at this PR now - I am not convinced we want this last commit as we cannot backport it to the release branches to make them work. Will have to ponder and look at alternatives.

jjhursey:
Adding the tool check to the Cross-version CI might be a good thing to do. Would it be as simple as all permutations of versions `A` and `B`:
```
cd $HOME/pmix-A/test/simple
./simptest -e $HOME/pmix-B/test/simple/simptool
```
If so I can work on adding that to the testing setup.

rhc54:
> Adding the tool check to the Cross-version CI might be a good thing to do. Would it be as simple as all permutations of versions `A` and `B`:
> 
> ```
> cd $HOME/pmix-A/test/simple
> ./simptest -e $HOME/pmix-B/test/simple/simptool
> ```
> 
> If so I can work on adding that to the testing setup.

Yep, that would do it! Don't include v1.2.5, obviously.

artpol84:
This PR breaks tool compatibility in the way I described in https://github.com/pmix/pmix/pull/1240#issuecomment-492542968.
The rationale was the following:
We need to fix each branch anyway because of (a) the bug in connection protocol and (b) the fact that server sends unconverted handshake status and clients are using ntohl to interpret it.
Given that we had to touch every branch and recommend everyone to update I believe that it makes sense to refactor the protocol that is being fixed anyway.

I can restore the tools protocol if absolutely needed, but believe its not that important for the reasons above.

artpol84:
Also @rhc54, could you confirm that https://github.com/pmix/pmix/pull/1240#issuecomment-492542968 is making sense to you.
I extracted all my concerns/rationale here. So you can only consider this comment out of the whole discussion.

artpol84:
@rhc54 any progress on this?

rhc54:
I'm working on it as time permits - we can discuss it on today's call.

jjhursey:
Per teleconf:
 - Ralph working on fixing the Tool connection portion of this PR (will post a PR when ready)
 - Prefer to not touch the `ptl/usock` component since it's only there to support the `v1.2` series. We should avoid making changes to `ptl/usock` if at all possible.
 - Needs more investigation and review

artpol84:
bot:retest

ibm-ompi:

 Gist: https://gist.github.com/97adb2838d0127ac4fabf8f03c6a5c2f

artpol84:
@rhc54, note that this PR is mentioned in the release notes [here](https://github.com/pmix/pmix/releases/tag/v3.1.3rc3):
Changes since v3.1.2rc2
PR #1237: Avoid double-free of collective tracker
PR #1237: Ensure all participants are notified of fence complete
PR #1237: Ensure all participants are notified of connect and disconnect complete
PR #1250: Fix PMIx_server_finalize hang (rare)
PR #1271: PTL/usock doesn't support tools
PR #1240: Fix the PTL connection establishment protocol <---- !!!!!
PR #1280: Fix tool connection in psec/handshake mode

But I'm actually going to close it as we went ahead with an alternative solution: So last two lines of the release notes should both reference #1280.

PR #1242, opened 2019-05-10
PTL: Fix connection establishment protocol

artpol84:
On the server side the folowing messages are sent to the client during connection:
* (4B) Security verification
* (4B) Status
* (4B) Server-local index of the peer

client only receives 2 messgaes:
* (4B) Status
* (4B) Server-local index of the peer

For unknown reason in most of the time extra 4 bytes were consumed by the client.
But sometimes these extra bytes were prepending a valid message shifting tag and
size away and PMIx_Init hangs were observed.


artpol84:
@jjhursey we need this in RC.

PR #1253, opened 2019-05-15
Preserve user-provided CFLAGS to protect autotools

rhc54:
Autotools expect to perform compiler tests without additional CFLAGS. If
the user provides CFLAGS, this can interfere - e.g., providing -Werror
will cause autotools to abort every time it tests for some
functionality. So capture the user-provided flags and restore them after
configure is done running tests.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -19,7 +19,7 @@
 # Copyright (c) 2012      Oracle and/or its affiliates.  All rights reserved.
 # Copyright (c) 2013      Mellanox Technologies, Inc.
 #                         All rights reserved.
-# Copyright (c) 2014-2018 Intel, Inc.  All rights reserved.
+# Copyright (c) 2014-2019 Intel, Inc.  All rights reserved.
 # Copyright (c) 2016      IBM Corporation.  All rights reserved.
 # Copyright (c) 2016-2018 Research Organization for Information Science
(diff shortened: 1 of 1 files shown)


rhc54:
@ax3l I separated out the protection for user-provided CFLAGS and am committing it by itself. As stated in #1214, there are numerous errors encountered when -Werror is added. I have worked thru some of them and will continue plugging them over time.

ax3l:
I found an alternative solution yesterday, but enabling `-Werror` [only on the builds](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=138750).

Just for the record:
```diff
diff --git a/configure.ac b/configure.ac
index 1adb2877..d0cca42d 100644
--- a/configure.ac
+++ b/configure.ac
@@ -233,6 +233,17 @@ AC_MSG_CHECKING([final LIBS])
 AC_MSG_RESULT([$LIBS])
 
 ####################################################################
+# -Werror for CI scripts
+####################################################################
+
+AC_ARG_ENABLE(werror,
+    AC_HELP_STRING([--enable-werror],
+                   [Treat compiler warnings as errors]),
+[
+    CFLAGS="$CFLAGS -Werror"
+])
+
+####################################################################
 # Version information
 ####################################################################
```

PR #1255, opened 2019-05-16
Travis: Clang and Undefined Behavior Sanitizer

ax3l:
Adds Clang compile routines and `-fsantize=undefined` checks to Travis-CI.

Fix small warnings in sign compare and sometimes uninitialized variables.

Over the next development steps, we should try to remove the amount of `-Wno-...` flags provided to CI by fixing the warnings.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -248,6 +248,17 @@ AC_MSG_CHECKING([final LIBS])
 AC_MSG_RESULT([$LIBS])
 
 ####################################################################
+# -Werror for CI scripts
+####################################################################
+
+AC_ARG_ENABLE(werror,
+    AC_HELP_STRING([--enable-werror],
+                   [Treat compiler warnings as errors]),
+[
+    CFLAGS="$CFLAGS -Werror"
+])
+
+####################################################################
 # Version information
 ####################################################################
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,7 @@ dist: xenial
 cache:
   apt: true
 
-env:
-  global:
-    # TODO: -Werror
-    # TODOL -fsanitize=address -fsanitize=undefined
-    - CFLAGS: "-Wall -Wextra"
-    - CXXFLAGS: "-Wall -Wextra"
+language: c
 
 addons:
   apt:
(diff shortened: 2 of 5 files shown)


configure.ac
```diff
@@ -232,6 +232,17 @@ AC_MSG_RESULT([$LDFLAGS])
 AC_MSG_CHECKING([final LIBS])
 AC_MSG_RESULT([$LIBS])
 
+####################################################################
+# -Werror for CI scripts
```
ax3l: @rhc54 according to [this source](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=138750), providing such a flag is probably a clean solution to enable `-Werror` (related to #1253)

rhc54: I really disagree with that source. Think of it from a packager's perspective - it means that every package would be defining "special" CFLAG options that have to be provided in a package-unique manner. You'd go nuts trying to track them all.

If you dig a bit into the autotools, you find that there is actually no reason for any user-provided CFLAGs to be applied to the configure tests. Those tests are purely looking at capabilities and existence of various features. User-provided flags are targeted solely at the build. Better to just capture them and apply them where they belong.

rhc54:
I'll try to push what I have done so we avoid a lot of duplicated effort 😄

configure.ac
ax3l: I fully agree on the packager perspective since I wear that hat as well from time to time :) Nevertheless, I thought a `-Werror` flag for CI is good enough, package managers do not use that flag.

As I understood autotools, all tests are quite lightweight and therefore unfortunately not isolated from the actual build (flags)... Nevertheless, let's merge the other open branch and then rebase this PR and see how far we get :)

rhc54: My intention is to default `-Werror -Wextra` whenever we build from a Git repo - i.e., a developer's build - to avoid allowing breakage to re-enter the code base. Otherwise, we'll just be chasing our tails.

My main concern really was that I hate to add custom variations to expected behavior. If a user sets CFLAGs, it would be unexpected to find that everything failed unless they used the "backdoor"...plus we'd have to open backdoors for every flag that caused a problem in autotools (the list is longer than just `-Werror`).

ax3l: I would not make `-Werror` the default on a developer machine as it might slip into shipped packages which breaks things. Instead, just make `-Werror` the default in CI and it will just enforce what warnings a developer saw with `-Wall` already and achieve the same effect :)

rhc54: I'm afraid that `-Wall` isn't showing most of the warnings that are causing problems here, at least not on my Mac or on the Linux systems (Ubuntu, CentOS, and SLES) that I test. Our current build is totally warning free on my systems.

ax3l: Sorry, I mean `-Wall -Wextra` of course.

rhc54: Yeah, but sadly many of our developers will just ignore the warning as "trivial" or "too picky" and commit. I'd really rather not play whack-a-mole with the CI. If we really are going to require that `-Werror -Wextra` pass before commit, then we should just make that the default for developer builds.

ax3l:
But aren't all contributions going through PRs and thus the CI?

rhc54:
Yes - but like I said, if we are going to require they pass that test, then we should build that way by default. It makes no sense to say "hey, go ahead and build randomly and we'll reject you upon test". Let's just build to what we test.

rhc54:
@jjhursey I put a comment on the xversion CI test as it is somehow passing now for this PR. I don't see any commits to pmix-tests that might have fixed it, but we may need to check master and friends again to see what is going on.

configure.ac
ax3l: Ah, so not all contributions have to pass the CI before being merged via a PR?

rhc54: No, that's not the policy. However, when CI becomes too onerous it loses credibility and is seen as more of the problem than the cure. If someone spends a ton of time prepping a commit, it generates strong negative reactions to have CI apply some unexpected barrier. Like I said, if this is a gating barrier, then you build to that barrier - no point is served by imposing it at the end. Seriously, could we just please move on?

ax3l: No problem, I just didn't understand if all commits go via PRs or not. Did not intend to nag, if it came across as such :)

PR #1271, opened 2019-05-19
PTL/usock: `usock` doesn't support Tools, fail query for this case.

artpol84:
Port this portion from 6bbbb4bf0 in master.


artpol84:
Because it was missing it looks like v3.1 is passing cross-version test while it shouldn't.
Communication goes through USOCK component.

ibm-ompi:

 Gist: https://gist.github.com/4149895039ab7aa10c7156ef8cb4d826

rhc54:
Thanks for tracking this down! We probably should update the xversion test to more accurately reflect what combinations really should support tools - it may be that tool xversion support isn't as pervasive in the older versions (and doesn't need to be as adoption has come later).

PR #1265, opened 2019-05-17
Travis: Address Sanitizer

ax3l:
Enable address sanitizers in Travis CI to avoid memory leaks sneaking in during development.

PR #1275, opened 2019-05-20
IPv6 IF Read: Buffer Overflow

ax3l:
Fix a buffer overflow on some systems on which `IF_NAMESIZE` is defined to 10 but we read 20 chars in `fscanf`.

First seen in #1265.

Note: it's probably better to modify the `fscanf` line instead, e.g. by pre-processor math and pre-processor stringification.

### DIFF
--- a/src/mca/pif/linux_ipv6/pif_linux_ipv6.c
+++ b/src/mca/pif/linux_ipv6/pif_linux_ipv6.c
@@ -80,7 +80,7 @@ static int if_linux_ipv6_open(void)
 {
     FILE *f;
     if ((f = fopen("/proc/net/if_inet6", "r"))) {
-        char ifname[IF_NAMESIZE];
+        char ifname[21];  // note: IF_NAMESIZE might be too small, e.g. 10;
         unsigned int idx, pfxlen, scope, dadstat;
         struct in6_addr a6;
         int iter;


rhc54:
Good catch - thanks!

ax3l:
You are welcome :)
I am not 100% happy with my proposed fix, because it leaves the corner-case open if `IF_NAMESIZE` is very large... Maybe something for a refactoring.

orivej:
> Fix a buffer overflow on some systems on which IF_NAMESIZE is defined to 10 but we read 20 chars in fscanf.

According to `[32, 48) 'ifname'` in https://github.com/pmix/pmix/pull/1265#issuecomment-493385772 `IF_NAMESIZE` is 16, not 10. I have checked that it is 16 on [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/gnu/net/if.h;h=303d092739f91254dcfd4f1dfdf7e60ce0bcdf85;hb=HEAD#l31) and [musl](http://git.musl-libc.org/cgit/musl/tree/include/net/if.h?h=v1.1.22#n10).

ax3l:
Then another include must pull it als well, defined to 10 on Debian and Ubuntu.

Furthermore 16 is also less than 20 :)

orivej:
I'm sure that the code with this patch works, but the comment is misleading because `IF_NAMESIZE` can't be 10. `IF_NAMESIZE` has been defined to 16 in glibc since its [introduction](https://sourceware.org/git/?p=glibc.git;a=commit;f=sysdeps/gnu/net/if.h;h=84d7b1ea9ea6bc3b23578d11bf157a9c0f81821f) in 2001. It is 16 on Ubuntu 16.04.

ax3l:
I don't know where the define came from, but I checked it on my system and it was 10 there. Can double check again if this helps to pinpoint it down further, seams to be defined at multiple places in headers.

orivej:
Please do. (You can see where the defines come from by supplying `-dD -E` to the compiler command line.)

ax3l:
Double checked again with gcc 6.3.0 on Debian: defined to `16` indeed from the `<net/if.h>` include in the file.

Same for clang 6.0 builds. I will update the comment in a PR, thanks for catching this!

PR #1264, opened 2019-05-17
PTL: Fix the connection establishment protocol (minimalistic)

artpol84:
Avoid sending extra message in the client connection case.
Is an alternative to:
https://github.com/pmix/pmix/pull/1240

Signed-off-by: Artem Polyakov <artpol84@gmail.com>

PR #1277, opened 2019-05-21
psec/dummy handshake

karasevb:
This commit adds a psec dummy component intended to check the PTL
handshake scenario. The configuration option `--enable-dummy-handshake`
allows to build this component, it is disabled by default.

Refs: https://github.com/pmix/pmix/pull/1264#issuecomment-494012521

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -1270,6 +1270,22 @@ fi
 AC_DEFINE_UNQUOTED([PMIX_ENABLE_PTY_SUPPORT], [$PMIX_ENABLE_PTY_SUPPORT],
                    [Whether user wants PTY support or not])
 
+#
+# psec/dummy_handshake
+#
+
+AC_MSG_CHECKING([if want build psec/dummy_handshake])
+AC_ARG_ENABLE(dummy-handshake,
+              AC_HELP_STRING([--enable-dummy-handshake],
+                             [Enables psec dummy component intended to check the PTL handshake scenario (default: disabled)]))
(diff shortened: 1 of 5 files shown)


karasevb:
Updated with the force-pushed commit, @artpol84 please review.

artpol84:
This component isn't built by default.
It is expected though that we will turn it on for CI tests and run with this component forced to verify that the handshake path is working.

We were using this component to verify #1264

rhc54:
This is fine - but begs a question as to whether we should similarly disable/enable build of the "none" component. Something to ponder. Meantime, thx!

artpol84:
@rhc54 couple of comments:

First, I suggest to port this to other branches so xversion will be able to run with it. If you are OK we will open PRs.

Second, the main reason I made this component disabled by default is that otherwise, the sneaky client may force our "dummy" component during the handshake and workaround the security barrier as our this component is permissive.
Is the same scenario possible with "none" component?
When thinking about this I guess what is happening is that the server side actually selects the psec component and enforces clients to obey. Still having a permissive security component in the installation is risky as who knows how the intruder could make server to select it.

rhc54:
Agreed - please do the PR. It is indeed possible for a user to force the "none" component so long as the server allows it to be used. Enabling it does, however, require an active action by the server as the component won't allow itself to be opened without an MCA param being set. Might be good enough - worth a chat.

artpol84:
@karasevb Please port this to the release branches, please.
Up to v2.0

PR #1280, opened 2019-05-23
Fix/v3.1/cli connect v2

artpol84:


artpol84:
I will ask @karasevb to run xversion manually to confirm that this is working fine

PR #1289, opened 2019-05-28
Avoid output_verbose overhead when it won't print

rhc54:
Don't render the output if the verbosity is below the trigger level and
nothing will be output


ibm-ompi:

 Gist: https://gist.github.com/a00d429397a87007e07ba64108aae75b

PR #1293, opened 2019-06-02
Remove arbitrary limit on linux_ipv6 if_inet6 interface name length

orivej:
Address sanitizer has reported potential stack buffer overflow in the OpenMPI code base (https://github.com/open-mpi/ompi/blob/v4.0.1/opal/mca/if/linux_ipv6/if_linux_ipv6.c#L106). I have found that this issues was already fixed in pmix by #1275, but it added a misleading [comment](https://github.com/pmix/pmix/pull/1275/files#diff-23cc7c5cd65a9ece0cb57339957fb087R83) ([explanation](https://github.com/pmix/pmix/pull/1275#issuecomment-497997915)) so I could not copy it verbatim. Besides, the numbers 21 and 20 used in the current `pif_linux_ipv6.c` do not correspond to the Linux limit on network interface names, which is 15:
https://elixir.bootlin.com/linux/v5.1/source/include/linux/netdevice.h#L1790
https://elixir.bootlin.com/linux/v5.1/source/include/uapi/linux/if.h#L33

The possible fixes include:
1. Hardcode current limits (16 byte buffer, 15 byte name length).
2. Generate correct scanf format string based on `IF_NAMESIZE` in runtime.
3. Generate off-by-one scanf format string based on `IF_NAMESIZE` using the preprocessor.
4. Let scanf allocate the buffer of the right size. (Implemented in this PR.)

The first option has the smallest diff, but it hardcodes two constants.

The second has the largest diff.

The third looks like this:
```c
char ifname[IF_NAMESIZE+1];
scanf("… %x %"STRINGIFY(IF_NAMESIZE)"s\n");
strncpy(intf->if_name, ifname, IF_NAMESIZE+1);
```
This would work in the current pmix codebase which has (surprisingly) increased `pmix_pif_t.if_name` size from `IF_NAMESIZE` to `IF_NAMESIZE+1` in 1e85a0fd58a8aceb2bba5072d8bfa2d00f160081, but it would be incorrect in the OpenMPI codebase which has not.

The forth option involves a dynamic memory allocation, but since it happens at initialization it seems OK.

### DIFF
--- a/src/mca/pif/linux_ipv6/pif_linux_ipv6.c
+++ b/src/mca/pif/linux_ipv6/pif_linux_ipv6.c
@@ -80,20 +80,24 @@ static int if_linux_ipv6_open(void)
 {
     FILE *f;
     if ((f = fopen("/proc/net/if_inet6", "r"))) {
-        char ifname[21];  // note: IF_NAMESIZE might be too small, e.g. 10;
+        char ifname[IF_NAMESIZE];
+        char *ifnameptr;
         unsigned int idx, pfxlen, scope, dadstat;
         struct in6_addr a6;
         int iter;
         uint32_t flag;
         unsigned int addrbyte[16];
 
-        while (fscanf(f, "%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x %x %x %x %x %20s\n",
+        while (fscanf(f, "%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x %x %x %x %x %ms\n",
                       &addrbyte[0], &addrbyte[1], &addrbyte[2], &addrbyte[3],
                       &addrbyte[4], &addrbyte[5], &addrbyte[6], &addrbyte[7],
                       &addrbyte[8], &addrbyte[9], &addrbyte[10], &addrbyte[11],
                       &addrbyte[12], &addrbyte[13], &addrbyte[14], &addrbyte[15],
-                      &idx, &pfxlen, &scope, &dadstat, ifname) != EOF) {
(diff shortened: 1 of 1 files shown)


ibm-ompi:

 Gist: https://gist.github.com/58df8e5770c9f77c77214a6de097cdc0

orivej:
gcc `-Werror` [check](https://travis-ci.org/pmix/pmix/builds/540303722) on Travis has failed with `ISO C does not support the 'm' scanf flag [-Werror=format=]`…

rhc54:
OMPI will be "fixed" once we officially release PMIx v3.1.3 and update OMPI. While we appreciate the contribution, I think the fixed larger size approach is just fine here - I see no reason to make it more complicated. As you saw, using more complex approaches leads to added problems trying to deal with all the various compiler support issues.

orivej:
Note that this issue is present not only in pmix bundled with ompi, but also in the `if_linux_ipv6.c` linked above. Is it going to be updated with pmix in ompi, or does this require a separate commit to ompi?

rhc54:
OMPI has a strict policy of only modifying the embedded PMIx from official PMIx release tarballs. Separate commits directly into OMPI are not allowed as it gets the two code bases out of sync. The next OMPI release is 4.0.2 scheduled for Sept '19.

orivej:
OMPI has two copies of the code in question:
- https://github.com/open-mpi/ompi/blob/v4.0.1/opal/mca/if/linux_ipv6/if_linux_ipv6.c#L106
- https://github.com/open-mpi/ompi/blob/v4.0.1/opal/mca/pmix/pmix3x/pmix/src/mca/pif/linux_ipv6/pif_linux_ipv6.c#L90
The second is an embedded PMIx. Is the first one too?

rhc54:
Nope - but that isn't something we can do anything about. Belongs to them.

PR #1294, opened 2019-06-04
Tools that don't connect do not need to provide ID

rhc54:
If no nspace/rank is given, then can still register attributes and satisy some simple requests


ibm-ompi:

 Gist: https://gist.github.com/e308fd8d06e64ab2c0381c7ac46ad514

PR #1303, opened 2019-06-05
Allow jobs to not specify proc-level info

rhc54:
There is no requirement that a job specify proc-level info just in
order to run. So don't error out if none is available


ibm-ompi:

 Gist: https://gist.github.com/a81b1dd54b933c5650d0afa8ea9aa223

rhc54:
Travis actually finished successfully - the checkmark update was lost due to a race condition

PR #1319, opened 2019-06-14
Fix schedular example in python bindings

dsikich:
The setup_application call only requires two arguments, and was
being passed three. This fix removes the appinfo from the call,
and stores the returned list from setup_application in appinfo.


ibm-ompi:

 Gist: https://gist.github.com/9da8e363ddbe5141343f63f07acee8e2

PR #1321, opened 2019-06-15
Prototype inline implementation of convenience macros

### LINKED ISSUE #1312: Naked memory function (malloc, calloc, free) in pmix_common.h
## Background information

MPICH has a memory tracing facility by tracking memory functions such as `malloc`, `free`. This is currently done through macros, and requires all memory functions inside MPICH be wrapped. Currently, this option is incompatible with PMIX as its API header files contains naked `malloc` and `free`.

xref: https://github.com/pmodels/mpich/issues/3832

At the moment, I was building from git master @35de87e1aad7b245f35fa57a6c10a8e92e2763e4

`git clone` into custom folder.

* Operating system/version: linux/ubuntu 18.04

-----------------------------

## Details of the problem

While pondering on work around solutions -- all nasty -- I can't help thinking that this is PMIx api issue and it is easy to fix here. For example, how about turn these macros -- `_CREATE`, `_FREE`, `_DESTRUCT` into inline functions? Of course, personally, I would like to see them as real functions but we may have different view on what is performance critical. Is it possible to make it a build option (inline vs. normal functions)? To the least, if they need be in macros, are they complete? That is, for all memory that call `free` in the macros, are they all allocated by the macros? As long as they are balanced, then it won't necessarily break memory tracing facilities. If that is the case, one solution maybe use `pmix_malloc`, `pmix_free` etc. in stead of naked calls, and they can be redefined to whatever.

rhc54:
Closes #1312

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -1093,6 +1093,23 @@ typedef uint8_t pmix_coord_system_t;
 #define PMIX_COORD_CYLINDRICAL  0x02
 
 
+/* define some "hooks" external libraries can use to
+ * intercept memory allocation/release operations */
+static inline void* pmix_malloc(size_t n)
+{
+    return malloc(n);
+}
+
+static inline void pmix_free(void *m)
+{
+    free(m);
+}
+
+static inline void* pmix_calloc(size_t n, size_t m)
+{
+    return calloc(n, m);
+}
+
 /* declare a convenience macro for checking keys */
 #define PMIX_CHECK_KEY(a, b) \
     (0 == strncmp((a)->key, (b), PMIX_MAX_KEYLEN))
@@ -1134,7 +1151,7 @@ typedef struct pmix_byte_object {
 
 #define PMIX_BYTE_OBJECT_CREATE(m, n)   \
     do {                                \
-        (m) = (pmix_byte_object_t*)malloc((n) * sizeof(pmix_byte_object_t));   \
+        (m) = (pmix_byte_object_t*)pmix_malloc((n) * sizeof(pmix_byte_object_t));   \
         if (NULL != (m)) {                                                     \
             memset((m), 0, (n)*sizeof(pmix_byte_object_t));                    \
         }                                                                      \
@@ -1149,7 +1166,7 @@ typedef struct pmix_byte_object {
 #define PMIX_BYTE_OBJECT_DESTRUCT(m)    \
     do {                                \
         if (NULL != (m)->bytes) {       \
-            free((m)->bytes);           \
+            pmix_free((m)->bytes);           \
         }                               \
     } while(0)
 
@@ -1159,10 +1176,10 @@ typedef struct pmix_byte_object {
         if (NULL != (m)) {                      \
             for (_bon=0; _bon < n; _bon++) {    \
                 if (NULL != (m)[_bon].bytes) {  \
-                    free((m)[_bon].bytes);      \
+                    pmix_free((m)[_bon].bytes);      \
                 }                               \
             }                                   \
-            free((m));                          \
+            pmix_free((m));                          \
             (m) = NULL;                         \
         }                                       \
     } while(0)
@@ -1193,7 +1210,7 @@ typedef struct {
 
 #define PMIX_ENVAR_CREATE(m, n)                                     \
     do {                                                            \
-        (m) = (pmix_envar_t*)calloc((n) , sizeof(pmix_envar_t));    \
+        (m) = (pmix_envar_t*)pmix_calloc((n) , sizeof(pmix_envar_t));    \
     } while (0)
 #define PMIX_ENVAR_FREE(m, n)                       \
     do {                                            \
@@ -1202,7 +1219,7 @@ typedef struct {
             for (_ek=0; _ek < (n); _ek++) {         \
                PMIX_ENVAR_DESTRUCT(&(m)[_ek]);      \
             }                                       \
-            free((m));                              \
+            pmix_free((m));                              \
         }                                           \
     } while (0)
 #define PMIX_ENVAR_CONSTRUCT(m)        \
@@ -1214,11 +1231,11 @@ typedef struct {
 #define PMIX_ENVAR_DESTRUCT(m)         \
     do {                               \
         if (NULL != (m)->envar) {      \
-            free((m)->envar);          \
+            pmix_free((m)->envar);          \
             (m)->envar = NULL;         \
         }                              \
         if (NULL != (m)->value) {      \
-            free((m)->value);          \
+            pmix_free((m)->value);          \
             (m)->value = NULL;         \
         }                              \
     } while(0)
@@ -1252,14 +1269,14 @@ typedef struct pmix_data_buffer {
 } pmix_data_buffer_t;
 #define PMIX_DATA_BUFFER_CREATE(m)                                          \
     do {                                                                    \
-        (m) = (pmix_data_buffer_t*)calloc(1, sizeof(pmix_data_buffer_t));   \
+        (m) = (pmix_data_buffer_t*)pmix_calloc(1, sizeof(pmix_data_buffer_t));   \
     } while (0)
 #define PMIX_DATA_BUFFER_RELEASE(m)             \
     do {                                        \
         if (NULL != (m)->base_ptr) {            \
-            free((m)->base_ptr);                \
+            pmix_free((m)->base_ptr);                \
         }                                       \
-        free((m));                              \
+        pmix_free((m));                              \
         (m) = NULL;                             \
     } while (0)
 #define PMIX_DATA_BUFFER_CONSTRUCT(m)       \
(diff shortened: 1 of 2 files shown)


ibm-ompi:

 Gist: https://gist.github.com/87f4756f8a6b9eb390a3902ca2ea1e84

rhc54:
bot:ibm:retest

hzhou:
FWIW, a minimal patch could be simply to create wrapper inline functions `pmix_malloc`, `pmix_free`, `pmix_calloc`, and replace all `malloc`, `free`, and `calloc`.
```
static inline void *pmix_malloc(size_t n) {
    return malloc(n);
}

static inline void pmix_free(void *m) {
    free(m);
}

static inline void *pmix_calloc(size_t n, size_t m) {
    return calloc(n, m);
}
```
, and
```
#define PMIX_BYTE_OBJECT_CREATE(m, n)   \
    do {                                \
        (m) = (pmix_byte_object_t*)pmix_malloc((n) * sizeof(pmix_byte_object_t));   \
        if (NULL != (m)) {                                                     \
            memset((m), 0, (n)*sizeof(pmix_byte_object_t));                    \
        }                                                                      \
    } while(0)
```

rhc54:
Unfortunately, it isn't quite that simple. We allowed the macros to utilize a few internal functions via the pmix_extend.h header. In order to fully resolve your situation, we have to elevate those functions to be defined as `static inline` in pmix_common.h as they contain malloc/free calls.

Your suggestion is fine for the base malloc/free/calloc functions, and I am making use of it - just is the minor part of the problem.

Not that big a deal - just tedious 😄

ibm-ompi:

 Gist: https://gist.github.com/3dff2ebf304fe10143048bc4e1086af7

ibm-ompi:

 Gist: https://gist.github.com/227571cde7d0b30e87068021bef01072

ibm-ompi:

 Gist: https://gist.github.com/b078caa3b6a3347abd73caba49658bb4

ibm-ompi:

 Gist: https://gist.github.com/69f4c3b79a5533c16dea5a634b796419

ibm-ompi:

 Gist: https://gist.github.com/ba0bcd907ca3fca99447a344345ac6d2

ibm-ompi:

 Gist: https://gist.github.com/a59612fc23a276ccfc4c1219b147453c

ibm-ompi:

 Gist: https://gist.github.com/ba6e206902a88358ac5ae3458a4a90c9

ibm-ompi:

 Gist: https://gist.github.com/b52e15076a0adf3d7d975b84840a7c24

hzhou:
> Unfortunately, it isn't quite that simple. We allowed the macros to utilize a few internal functions via the pmix_extend.h header. In order to fully resolve your situation, we have to elevate those functions to be defined as `static inline` in pmix_common.h as they contain malloc/free calls.

We don't need intercept all `malloc/free` calls. Third-party dependency libraries are allowed to use `malloc/free` internally and it is the respective maintainer's responsibility to make sure not to leak memory or provide the facility to guard such possibility. We are only having the problem with naked calls exposed through macros. Macros gets expanded at usage time and become indistinguishable that whose memroy calls these belong. So as long as the memory functions are inside functions such as `pmix_malloc`, they won't interfere with our own memory functions.

Is your solution to expose all memory functions used in `pmix` to `pmix_malloc`, `pmix_calloc` and `pmix_free`? This will provide a possibility for us to trace `pmix` memroy functions, which is fine, but certainly is beyond our current concerns.

rhc54:
> We don't need intercept all malloc/free calls, only those made by us. 

Understood - the problem is that the macros in question directly call functions that call malloc/free. If I don't bring them up to be visible, then you won't see the malloc/free you are executing.

> Is your solution to expose all memory functions used in pmix to pmix_malloc, pmix_calloc and pmix_free?

Not really, though a number will become visible anyway. It is a simple global search/replace to rename the functions to their `pmix_` equivalents. However, I believe your ability to see them down in the library will depend upon visibility.

hzhou:
>>    We don't need intercept all malloc/free calls, only those made by us.

> Understood - the problem is that the macros in question directly call functions that call malloc/free. If I don't bring them up to be visible, then you won't see the malloc/free you are executing.

This is the part that confuses me. We don't want to see the `malloc/free` that is used by `pmix`; i.e. we would like you to hide the visibility from us. Even when we are calling `malloc/free` indirectly with macros in `pmix_common.h`, from our point of view, it is `PMIx` calling them, It is our responsibility to call `PMIx` API correctly, but we don't want to track how PMIx calls `malloc` etc. internally. The wrapper functions such as `pmix_malloc` hides them from us, so we are good.

hzhou:
From `PMIx` point of view, it probably makes sense to reroute all memory functions throughout `PMIx` to use `pmix_malloc` and friends. If the performance is not a critical concern, then we can simply turn these wrapper functions (`pmix_malloc` etc. ) into real functions, and expose headers, e.g. `pmix_mem.h` through out `PMIx`.

rhc54:
Let's see if I can explain this to clarify the confusion. Many of the macros call functions such as `pmix_value_xfer`, which in turn call `malloc` inside them. Thus, if you invoke the macro, _you_ are calling `malloc`, not the PMIx library.

`pmix_value_xfer` and friends reside down inside the PMIx library - they are not defined as `inline` functions in the header. I could surface them as `static inline` functions in the header, and that might actually be more technically correct, but I think I won't do that for now. Instead, I'll simply replace `malloc` and friends with their `pmix_` prefixed equivalents as that should suffice for now.

The issue you will face is that we also call those functions down inside the library. Thus, you are going to wind up seeing calls by the library to `malloc` and friends regardless - it cannot be avoided without us creating duplicate code (one inside the library and one outside in the headers).

hzhou:
`pmix_value_xfer` is a regular function defined inside `PMIx`, right? That should be fine. It is commonly allowed and assumed that an API function may internally call `malloc` and its friends, and typically we don't want to see it surface into a header file.  In fact, our original complaint is `PMIX_BYTE_OBJECT_CREATE` etc. expose their internal memory calls resulting interferences with our own memory calls.

My understanding is: ideally, an API header should only contain declarations, not code. Sometimes, we leak code into headers for performance reasons -- macros and inline functions. The leaking, in this case, is an unfortunate side-effect that we should strive to control.

hzhou:
> Thus, you are going to wind up seeing calls by the library to malloc and friends regardless - it cannot be avoided without us creating duplicate code (one inside the library and one outside in the headers).

If the call is inside a regular function inside PMIx, we won't see the calls to malloc etc unless we trap it with debugger.

hzhou:
I think the confusion is probably due to me not explaining our memory tracer sufficiently. Let me try that:

In MPICH, we wrap all our memory function calls using a set of macros. E.g. we use `MPL_malloc` instead of `malloc`. Normally, `MPL_malloc` is defined as `malloc`. When we configure under `--enable-g=mem`, we define `MPL_malloc` into `MPL_trmalloc`, which runs additional bookkeeping. Since naked memory call won't be accounted, we define (under `--enable-g=mem`) `malloc` into an error string to make sure naked memory call won't get accidentally used. 

When PMIx header that contains naked memory call is included and used, this will cause error under `--enable-g=mem`. It is OK for PMIx to call memory functions inside its implementation since our memory tracing macros won't be able to touch them. It is also desirable that we don't touch third-party memory calls since it is not our responsibility to guard their memory leaks.

ibm-ompi:

 Gist: https://gist.github.com/88f92ac46b349f76da0ebc63bf8b310f

ibm-ompi:

 Gist: https://gist.github.com/e500d6bf495d6c0e34deb1f879d5a4f6

rhc54:
Sigh - to be honest, this is why I usually prefer to have someone asking for a feature provide the patch. I was assuming you actually wanted the number of times the application itself directly calls `malloc` to match the number of times it calls `free`. The change I've now provided will not do that as the `malloc` being called by the macro (via a function such as `pmix_value_xfer`) will not be counted, while the corresponding `free` that will be called by the release macro _will_ be seen and therefore counted. Thus, you are guaranteed to see a mismatch in your count.

However, I do agree that your "detector" macro will pass. I will leave the problem of the mismatched count to you 😄

rhc54:
@jjhursey We are getting errors from the cross-version tester - the error report indicates that the docker memory is overflowing. My guess is that the build and install directory trees aren't getting cleaned up. Can you please take a look when you get a chance?

hzhou:
@rhc54  Thanks Ralph. I am not familiar with PMIx at all, therefore not comfortable providing patches yet. Your discussion on `pmix_value_xfer` is very illuminating. Maybe next time I will try contributing directly. Cheers :)

jjhursey:
Thanks for the heads up on CI. It looks like docker built up a bunch of exited containers on that node. I've cleaned it out now and am setting up some tools to try to keep it clean. Things should be working now.

PR #1342, opened 2019-07-05
Fix if_linux_ipv6_open interface filter

orivej:
`if_linux_ipv6_open` should add interfaces with global scope to use them to reach any network host.

This ports the fix from https://github.com/open-mpi/ompi/pull/4930


orivej:
After this PR I would like to port https://github.com/open-mpi/ompi/pull/6797

PR #1343, opened 2019-07-06
remove unnecessary libtool init for c++

pkovacs:
Adding a patch to remove the libtool  language init for C++ which is not needed to build pmix.   There is an unneeded check for g++ otherwise and this forces downstream packagers to pull in g++ as a superfluous build dependency.


ibm-ompi:

PR #1344, opened 2019-07-06
remove unnecessary libtool init for c++

pkovacs:

Adding a patch to remove the libtool language init for C++ which is not needed to build pmix. There is an unneeded check for g++ otherwise and this forces downstream packagers to pull in g++ as a superfluous build dependency.


pkovacs:
Yes, I read PR #1180 and noticed this last vestige was still there.  Thanks for accepting.

PR #1346, opened 2019-07-08
fix incorrect pointer casts/deref

pkovacs:
fix incorrect pointer casts (major); 
adjust test timeout code (minor);
correct english (trivial)

### DIFF
--- a/test/test_server.c
+++ b/test/test_server.c
@@ -187,7 +187,7 @@ static void server_unpack_procs(char *buf, size_t size)
     char *nspace;
 
     while ((size_t)(ptr - buf) < size) {
-        ns_count = (size_t)*ptr;
+        ns_count = *(size_t *)ptr;
         ptr += sizeof(size_t);
 
         for (i = 0; i < ns_count; i++) {
@@ -201,10 +201,10 @@ static void server_unpack_procs(char *buf, size_t size)
             nspace = ptr;
             ptr += PMIX_MAX_NSLEN+1;
 
-            ntasks = (size_t)*ptr;
+            ntasks = *(size_t *)ptr;
             ptr += sizeof(size_t);
 
-            ltasks = (size_t)*ptr;
+            ltasks = *(size_t *)ptr;
             ptr += sizeof(size_t);
 
             PMIX_LIST_FOREACH(tmp, server_nspace, server_nspace_t) {
@@ -226,7 +226,7 @@ static void server_unpack_procs(char *buf, size_t size)
             }
             size_t i;
             for (i = 0; i < ltasks; i++) {
-                int rank = (int)*ptr;
+                int rank = *(int *)ptr;
                 ptr += sizeof(int);
                 if (ns_item->task_map[rank] >= 0) {
                     continue;
@@ -651,13 +651,13 @@ static void server_unpack_dmdx(char *buf, int *sender, pmix_proc_t *proc)
 {
     char *ptr = buf;
 
-    *sender = (int)*ptr;
+    *sender = *(int *)ptr;
     ptr += sizeof(int);
 
     memcpy(proc->nspace, ptr, PMIX_MAX_NSLEN +1);
     ptr += PMIX_MAX_NSLEN +1;
 
-    proc->rank = (int)*ptr;
+    proc->rank = *(int *)ptr;
     ptr += sizeof(int);
 }
 


test/pmix_test.c
```diff
@@ -141,12 +139,12 @@ int main(int argc, char **argv)
...
         // To avoid test hang we want to interrupt the loop each 0.1s
-        double test_current;
+        struct timeval time_now, time_elapsed;
 
         // check if we exceed the max time
-        gettimeofday(&tv, NULL);
-        test_current = tv.tv_sec + 1E-6*tv.tv_usec;
-        if( (test_current - test_start) > params.timeout ){
+        gettimeofday(&time_now, NULL);
+        timersub(&time_now, &time_start, &time_elapsed);
```
artpol84: Could you explain what is achieved by this change?
It seems to be exactly the same thing?
If not - can you provide more details about what is wrong with original code.
In my understanding arithmetic should work the same for both big and little endian.

test/test_server.c
```diff
@@ -187,7 +187,7 @@ static void server_unpack_procs(char *buf, size_t size)
     char *nspace;
 
     while ((size_t)(ptr - buf) < size) {
-        ns_count = (size_t)*ptr;
```
artpol84: Good catch, thanks!

test/pmix_test.c
rhc54: @artpol84 It is just a more correct way of subtracting two timeval structs:
```
timersub() subtracts the time value in b from the time value in a, and places the result in the timeval pointed to by res. The result is normalized such that res->tv_usec has a value in the range 0 to 999,999. 
```

artpol84: I wouldn’t say it’s more correct. It’s a different one.
So I tend to keep the code unless it’s wrong (like fixes for the type casting)

pkovacs: Code clarity > clever code every single day and twice on Sunday.   The change was also motivated by trying to correct the timing problem I still see on test14.  All other tests are passing now.

pkovacs:
I didn't mean to undo my change by resolving the issue.  I guess that battle is lost.

rhc54:
> I didn't mean to undo my change by resolving the issue. I guess that battle is lost.

It still looks okay to me. All "resolve" does is collapse that discussion for the viewer - has no impact on anything else.

pkovacs:
@rhc54 Here's what I am still wrestling with on test14.  Note the difference between 14 and 15 on that fence is the -c flag.   15 terminates smoothly everytime and 14 times out every time.   The -c flag adds a collect data request to the fence, 14 omits it.  That one difference causes big endian to thrash around in bfrops until it times out.   The -c flag causes normal termination.  All other arches are fine -- just big endian is the problem.

rhc54:
@pkovacs Hmmm...try putting `PMIX_MCA_bfrops=v3` in your environment and see if that works.

pkovacs:
It's running v3 for sure.  I have debug on bfrops up to 100, but I'll try what you suggest.   The only difference is that last clause in pack_fence(), the clause that adds PMIX_INFO is not run on 14.   Without it, it never terminates,

test/pmix_test.c
artpol84: I don’t see this code to have more clarity. Original way is pretty typical as well.

rhc54:
My bad - I forgot you are working on v3.1.3 instead of master. My suggestion will have no affect. I'll take a look.

pkovacs:
Test 14 fails on s390x below and my ppc64 vm.   This is the last issue to clear the package.

```
=====================================
   pmix 3.1.3: test/test-suite.log
=====================================
# TOTAL: 16
# PASS:  15
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
.. contents:: :depth: 2
FAIL: run_tests14.pl
====================
./pmix_test -s 2 -n 2 --job-fence 2>&1
==42487== ERROR [pmix_test.c:154:main]: Test exited by a timeout!
==42487== ERROR [cli_stages.c:159:cli_cleanup]: rank 0 with state 1 unexpectedly terminated.
==42487== ERROR [pmix_test.c:160:main]: Test was aborted!
==42501== ERROR [pmix_test.c:154:main]: Test exited by a timeout!
==42501== ERROR [cli_stages.c:159:cli_cleanup]: rank 1 with state 1 unexpectedly terminated.
==42501== ERROR [pmix_test.c:160:main]: Test was aborted!
==42501== ERROR [server_callbacks.c:87:finalized]: cannot find rank 1
==42487== ERROR [server_callbacks.c:87:finalized]: cannot find rank 0
==42487== test_server.c:server_finalize: Test finished OK!
CODE 256
```

rhc54:
I don't immediately see anything wrong in the base code, which makes me wonder if the problem lies in the test code where the two servers unpack each other's fence contribution. I'm not familiar with that code - @artpol84 can you take a look?

pkovacs:
I'll keep looking.  Once I get onto a problem my nature is to stay on it.

ibm-ompi:

ibm-ompi:

ibm-ompi:

pkovacs:
I removed all but the pointer problems from this PR.   I need to trigger the CI again because I force pushed the signed version in the middle of the tests.  I hope I can do that by closing and reopening this PR.

I've looked some more at the last remaining problem: `run_tests14.pl`.  While I don't have a definite answer, I can say the following:

- pmix_client.c calls `test_job_fence` in test_fence.c and cpu usage rises immediately to 100% and stays there until the timer kills the processes.
- `test_job_fence` uses `PMIx_Put` to write a few keys and then, for each server, `PMIx_Get` retrieves those keys. 
- The `PMIx_Get` is implemented as a `GET` macro with blocking and non-blocking code selected as desired. 
- The non-blocking `GET` and `FENCE` macros use spinlock style looping to acquire their data. 

It may also simply be the case that these big endian vms running on little endian hypervisors just don't have enough computing power to handle these spinlocks with the way the linux schedulers now work. 

There are no segfaults anymore -- things are just pegged at 100% until the timer kills them.

I would have to profile the code to see what it's doing.  That's where I'm at right now with this.

rhc54:
@pkovacs You can retrigger the CI using (adding spaces so it doesn't trigger) "bot : retest" in a comment. You can even target the CI to re-execute like "bot : ibm : retest"

pkovacs:
bot:ibm:retest

pkovacs:
This is ready to go.

rhc54:
Thanks @pkovacs ! Let me know if you figure out what is going on with that remaining test.

PR #1356, opened 2019-07-13
fix another pointer cast/deref in test suite

pkovacs:


pkovacs:
See my comment here: https://github.com/pmix/pmix/issues/1345#issuecomment-511094634

rhc54:
Thanks again!!

PR #1369, opened 2019-07-21
Fix legacy support for PMI-1

rhc54:
Legacy interfaces cannot pass the proc whose data they are requesting.
Instead, legacy apps publish keys that have the proc's rank embedded in
them, thus making them globally unique. When requesting the data, the
requestor can only pass PMIX_RANK_UNDEF to PMIx_Get since it has no
other information.

In that scenario, we have to search data from all ranks (since it will
have been stored by rank when it was published) to find the unique key.
Since the datastore components cannot handle PMIX_RANK_UNDEF, we must
also ensure that all data operations involving PMIX_RANK_UNDEF flow
thru the GDS hash component

Also, fix the pmi_client test program - any keys "put" by a process must
be globally unique

### DIFF
--- a/src/client/pmi1.c
+++ b/src/client/pmi1.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2014-2018 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2014-2019 Intel, Inc.  All rights reserved.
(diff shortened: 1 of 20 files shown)


rhc54:
@jjhursey Someone else will need to fix PMI-2 - there appears to be lots of breakage there.

ibm-ompi:

 Gist: https://gist.github.com/0037274c32925f6ab23aa068a428ab26

rhc54:
bot:ibm:xver:retest

PR #1371, opened 2019-07-22
Fix legacy support for PMI-1 and Cleanup handling of requests for different nspaces

rhc54:


jjhursey:
Question on this PR: You mentioned that the dstore components don't support the search pattern necessary for PMI1 (since PMI1 doesn't specify the process information). Does this PR exclude those components from running or does a user need to explicitly set `PMIX_MCA_gds=hash`?

Did we open a ticket to extend the dstore component search algorithm to support the 'unknown' process mode?

rhc54:
> Question on this PR: You mentioned that the dstore components don't support the search pattern necessary for PMI1 (since PMI1 doesn't specify the process information). Does this PR exclude those components from running or does a user need to explicitly set PMIX_MCA_gds=hash?

No - we check for the rank=undef case and only let the hash component process such requests. We still use dstore for the job-level (i.e., non-modex) info.

> Did we open a ticket to extend the dstore component search algorithm to support the 'unknown' process mode?

No - I'm not sure there is any value in doing so. I don't believe it will be any more efficient than what the hash provides.

PR #1374, opened 2019-07-24
Final updates before release

rhc54:


hppritcha:
Commit 2422fe85df4a98774e8c6e55a230bda991798f92 broke spawn support at NERSC on the 4.0.2 and later releases.  The OMPI spawnee processes are unable to retrieve modex keys from spawner process owing to what appears to be the inability to ```pmix_get``` from processes in a different namespace.
@dsolt not sure if this is what you're seeing or not.  You can check by either trying to revert the commits in this PR or using PMIx 3.1.3 release.

rhc54:
I take it you mean "comm_spawn" support? I'm surprised as OMPI itself distributes all such keys between the two namespaces, and so PMIx doesn't have to do anything.

I can check that comm_spawn is working on master.

hppritcha:
All I know is git bisect shows if I revert commit 2422fe8 the OMPI spawnee processes can retrieve the OB1 BTL keys from the spawner and the customer's application works again.  They only hit this problem when trying to move from OMPI 4.0.1 to 4.0.2.

rhc54:
Hmm...MTT is failing on NERSC because the tests cannot find `mpirun` for some reason. Might be worth fixing when you get a chance.

All other MTT's are showing no problems with OMPI v4.0.x branch for comm_spawn (except for loop_spawn in one case). Is it possible this has been fixed for v4.0.3? Or is this perhaps some artifact associated with running on a Cray?

hppritcha:
okay, let's approach this a different way.  What was 2422fe8 suppose to be fixing?

rhc54:
The problem was quite different from the one @dsolt was addressing. In his case, JSM's daemons don't know about every job that is being spawned. Thus, in his case, he has to address the problem of how to get the nspace info in the first place.

The problem I was addressing in that commit was different in that the PRRTE daemon always has knowledge of every job being spawned, even when it won't have any local clients for that job. Thus, the daemon has the job-level info for the "other" nspace, but that info isn't in the dstore shared memory that the client is using (as that shmem is dedicated to the client's own nspace). So we needed to detect that situation and pass the info down (via hash component) to the requesting client as it isn't currently possible to tell that client "here is another dstore shmem that holds the other nspace info".

Thus, the info should be there at the daemon - the question is: why isn't it being passed down to the clients? I'll have to investigate that one.

Issue #1375, opened 2019-07-25
Fastpath "get" is truncating keys


rhc54:
I've been seeing something that had me puzzled - whenever we did a modex to obtain data and passed the results down to a process, we wound up turning around and asking the server for the data anyway. In digging a little deeper, I finally discovered that the `fastpath` method for obtaining values is returning truncated keys.

You can see this easily by running `./simptest` in the `test/simple` directory:
```
$ ./simptest
Testing version 4.0.0a1
Collecting inventory
Inventory collected: 0
[Ralphs-iMac-2.local:39098] Client ns foobar rank 0: Running
[Ralphs-iMac-2.local:39098] CLIENT SERVER URI: pmix-server.39084;tcp4://127.0.0.1:55902
simpclient EVENT HANDLER REGISTRATION RETURN STATUS 0, ref=0
foobar:0 Model event handler called with status -147(PMIX MODEL DECLARED)
foobar:0	pmix.pgm.model:	TEST
foobar:0	pmix.mdl.name:	PMIX
foobar:0	pmix.evname:	SIMPCLIENT-MODEL
[Ralphs-iMac-2.local:39098] Client: ERRHANDLER REGISTRATION CALLBACK CALLED WITH STATUS 0, ref=1
[Ralphs-iMac-2.local:39098] Client foobar:0 universe size 1
[Ralphs-iMac-2.local:39098] Client ns foobar rank 0 cnt 0: PMIx_Get foobar-0-local-0 returned correct
[Ralphs-iMac-2.local:39098] Client ns foobar rank 0 testing internal modex blob
[Ralphs-iMac-2.local:39098] 	Key: fo
[Ralphs-iMac-2.local:39098] 	Key: pm
[Ralphs-iMac-2.local:39098] 	Key: pm
[Ralphs-iMac-2.local:39098] 	Key: pm
[Ralphs-iMac-2.local:39098] 	Key: pm
[Ralphs-iMac-2.local:39098] 	Key: pm
[Ralphs-iMac-2.local:39098] Client ns foobar rank 0: Finalizing
Client ns foobar rank 0:PMIx_Finalize successfully completed
SIMPTEST: Model event handler called with status -147(PMIX MODEL DECLARED)
	pmix.pgm.model:	PMIX
	pmix.mdl.name:	test
	pmix.evname:	SIMPTEST-MODEL
Test finished OK!
```
[...]

PR #1377, opened 2019-07-26
Skip fastpath/dstore for NULL keys

rhc54:
Refs #1375

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -101,7 +101,7 @@ PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc,
                         (NULL == key) ? "NULL" : key);
 
     /* try to get data directly, without threadshift */
-    if (PMIX_RANK_UNDEF != proc->rank) {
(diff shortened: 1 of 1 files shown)


ibm-ompi:

 Gist: https://gist.github.com/585a34bd9ffe7a4fd1738894e63a3c9d

artpol84:
@rhc54
Can you provide some insights about what was the problem. I kind of see what it could be, but would like to know for sure.

artpol84:
Is it in context of https://github.com/pmix/pmix/pull/1371?

rhc54:
Actually, I referenced the issue where it can be seen (#1375) most easily. It impacts the case where PMIx_Get is used to "pull" all the info for a given rank, which causes an array of pmix_info_t structs to be returned. I haven't tried queries that returned other types of values that might involve a key, so I don't know if the dstore is always truncating returned keys or only doing so for the array case.

artpol84:
Ok, I found that one as well. We will take a look.

jjhursey:
Will need to backport to the v3.x branch (maybe the v2.x branch).

rhc54:
Already in 3.1.4 - needs to be added to v2.2.3

PR #1378, opened 2019-07-26
Fix problems in Log, skip fastpath for NULL keys, update simptests

rhc54:


ibm-ompi:

 Gist: https://gist.github.com/7af1078c6acf76a4085b1377500af94c

rhc54:
@jjhursey Fascinating - first it marked all 9 tests as having passed. Then it went back and changed the marking for the `make check` test to `failed` due to an infrastructure problem (said it couldn't build an unchanged branch). So it seems to have thought it all passed, and then changed its mind!

jjhursey:
Odd. From Jenkins I only see the one run that failed and generated the Gist that was posted here. I'll try to find some cycles in the next week to improve the output of that test when it encounters a build failure.

I understand the frustration. I'm sorry that the IBM CI has been having network connectivity issues (though I think this is broader than just IBM as I see it with other systems too). And being behind a firewall it makes it difficult for others to see what Jenkins thinks happened. Neither are within my realm of control. If someone else wants to host the xversion CI I can provide instructions. All of the code to do so is publicly available on GitHub, and it runs in a Docker container. So if you have a public Jenkins and a host with Docker installed it shouldn't take long to get set up.

rhc54:
I'm not complaining - just noting when it has problems. I don't think it is IBM-specific, though I agree that is hard to tell from the outside. This behavior was especially puzzling as the PR was marked as all 9 tests passed, and then just as I was about to tap "commit" a few minutes later, it changed the one test to "failed" and modified the report to 8 passed/1 failed. Hadn't seen that one before!

PR #1382, opened 2019-08-02
 dstore: fixed truncate key-names while restoring

### LINKED ISSUE #1375: Fastpath "get" is truncating keys

karasevb:
The key name length was determined incorrectly, which led to truncation when copying the name.

Fixes #1375

### DIFF
--- a/src/mca/common/dstore/dstore_base.c
+++ b/src/mca/common/dstore/dstore_base.c
@@ -2157,9 +2157,10 @@ static pmix_status_t _dstore_fetch(pmix_common_dstore_ctx_t *ds_ctx,
                     break;
                 }
             } else if (NULL == key) {
+                char *kname_ptr = PMIX_DS_KNAME_PTR(ds_ctx, addr);
                 PMIX_OUTPUT_VERBOSE((10, pmix_gds_base_framework.framework_output,
                             "%s:%d:%s: for rank %s:%u, found target key %s",
-                            __FILE__, __LINE__, __func__, nspace, cur_rank, PMIX_DS_KNAME_PTR(ds_ctx, addr)));
+                            __FILE__, __LINE__, __func__, nspace, cur_rank, kname_ptr));
 
                 uint8_t *data_ptr = PMIX_DS_DATA_PTR(ds_ctx, addr);
                 size_t data_size = PMIX_DS_DATA_SIZE(ds_ctx, addr, data_ptr);
@@ -2173,8 +2174,8 @@ static pmix_status_t _dstore_fetch(pmix_common_dstore_ctx_t *ds_ctx,
                     PMIX_ERROR_LOG(rc);
                     goto done;
                 }
-                pmix_strncpy(info[kval_cnt - 1].key, PMIX_DS_KNAME_PTR(ds_ctx, addr),
-                        PMIX_DS_KNAME_LEN(ds_ctx, addr));
+                pmix_strncpy(info[kval_cnt - 1].key, kname_ptr,
+                        PMIX_DS_KNAME_LEN(ds_ctx, kname_ptr));
                 pmix_value_xfer(&info[kval_cnt - 1].value, &val);
                 PMIX_VALUE_DESTRUCT(&val);
                 buffer.base_ptr = NULL;


ibm-ompi:

jjhursey:
bot:ibm:gnu:retest

artpol84:
@karasevb do we need to backport it to v3.x? I assume we do.

karasevb:
@artpol84 Yes, I will cherry-pick when this PR is merged.

rhc54:
@karasevb Could you please confirm that the data returned when `NULL == key` is the same when the rank is other than `PMIX_RANK_UNDEF`? I suspect it is, but I'd like to see that confirmed as there were some issues about proc-specific job-level data. I think I handled those in the `hash` component in a way that ensured `dstore` would "do the right thing" when we transferred the job-level data across - but best to check.

karasevb:
@rhc54 Currently we have a restriction for `PMIX_RANK_UNDEF` and `key == NULL` in dstore: https://github.com/pmix/pmix/blob/master/src/mca/common/dstore/dstore_base.c#L1957-L1962. I'll eliminate this limitation, but this will require a small redesign for `dstor_fetch`. Do you want these changes to be in this PR?

rhc54:
I'm not sure it is worth doing, frankly, as `PMIX_RANK_UNDEF` means the user is coming thru the PMI-1/2 interfaces and we have to search the entire space anyway - might as well just do that with the `hash` component. I was more interested in the case of an actual rank value when the key is `NULL`. We should return all info about that rank (job-level and proc-level) plus anything the rank has pushed via `PMIx_Put`.

rhc54:
@karasevb Let me try to be a little clearer. This PR removes the check for `NULL == key`, which means that we will pass that use-case down to the dstore components. It isn't clear to me that there is any benefit in doing so as it would require you to add logic to the dstore to collect data from the job-level info, the proc-specific job-level info, plus anything the proc contributed via modex.

This represents a somewhat unusual operation that only the daemon typically would execute (and only when someone asks for info about a namespace other than their own). Hence, it isn't clear to me that there is any benefit in modifying dstore to support it.

If we don't want to do that modification, then we should leave the `NULL == key` check. Make sense?

karasevb:
@rhc54 thanks for the clarification! 
I agree that the check for `NULL == key` should be left. A commit that reverts this will be removed.
Regarding your first question - I checked which keys returned for actual rank and `key == NULL`: these keys are the same for `hash` and `dstor`.

rhc54:
Excellent - thanks!

karasevb:
@rhc54  PR is ready to go.

PR #1387, opened 2019-08-05
Check for EINVAL return from posix_fallocate

rhc54:


rhc54:
@alarcher Can you verify this is fixed here as well?

ibm-ompi:

ibm-ompi:

 Gist: https://gist.github.com/d8b14ea0727fdeaaab25c628d6c01cc5

jjhursey:
The IBM CI was having some network issues. These network issues are not being displayed in the logs due to an issue in the xversion.py tester script running in quiet mode. I updated the script so it should (1) show the build errors if/when they occur and (2) if a 'git' operation times out because of a network issue then it will try to continue on with what it has to test with if possible. Hopefully, that helps make it a bit more robust.

bot:ibm:retest

rhc54:
@artpol84 I think this is a pretty simple review - the bug reporter has already confirmed it solved the problem on zfs. Would you have a few minutes to take a peek so we can get 3.1.4 out?

alarcher:
@rhc54 Sorry I missed the second message. This works fine. Thank you.

PR #1391, opened 2019-08-06
Update python bindings to use error codes

dsikich:
Currently the python bindings use exceptions,
and these changes update them to return PMIX
error codes instead. This was the newly agreed
upon method going forward.

### DIFF
--- a/bindings/python/pmix.pxi
+++ b/bindings/python/pmix.pxi
@@ -70,7 +70,7 @@ cdef int pmix_load_darray(pmix_data_array_t *array, mytype, mylist:list):
     if PMIX_INFO == mytype:
         array[0].array = PyMem_Malloc(mysize * sizeof(pmix_info_t))
         if not array[0].array:
-            raise MemoryError()
+            return PMIX_ERR_NOMEM
         n = 0
         infoptr = <pmix_info_t*>array[0].array
         for item in mylist:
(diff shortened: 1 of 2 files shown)


dsikich:
@rhc54 yeah I agree it should really be a debug/verbose print instead of just print

rhc54:
@dsikich Check this out: https://inventwithpython.com/blog/2012/04/06/stop-using-print-for-debugging-a-5-minute-quickstart-guide-to-pythons-logging-module/

Basically:

The logging is better than printing because:

    It's easy to put a timestamp in each message, which is very handy.
    You can have different levels of urgency for messages, and filter out
         less urgent messages.
    When you want to later find/remove log messages, you won't get them
         confused for real print() calls.
    If you just print to a log file, it's easy to leave the log function calls in
         and just ignore them when you don't need them. (You don't have to
         constantly pull out print() calls.)

Using print is for coders with too much time on their hands. Use logging instead. Also, learn to use the Python debugger to debug bugs and Pylint to prevent bugs and make your code readable.

To print log messages to the screen, copy and paste this code:

    import logging
    logging.basicConfig(level=logging.DEBUG, format='%(asctime)s - %(levelname)s - %(message)s')
    logging.debug('This is a log message.')

dsikich:
@rhc54 yeah, using the logger seems like what we want. I can update to this in the near future.

PR #1405, opened 2019-08-20
Fix xnspace dmodex and add verbose debug

rhc54:
When we perform a direct modex request for information on a different
nspace, we need to store that info in the hash table. Otherwise, dstore
will simply ignore it. Add some helpful verbose debug as well.


rhc54:
@dsolt This is the branch I mentioned earlier today

PR #1430, opened 2019-09-04
Add python wrapper for PMIx_Lookup

dsikich:
@rhc54 Can you check that I am calling PMIx_Lookup correctly? When the pdata returns from PMIx_Lookup I get a "-47", "function not supported" error. 

### DIFF
--- a/bindings/python/client.py
+++ b/bindings/python/client.py
@@ -37,9 +37,16 @@ def main():
     info = [{'key': 'ARBITRARY', 'flags': PMIX_INFO_REQD, 'value':10, 'val_type':PMIX_INT}]
     rc = foo.fence(procs, info)
     print("Fence should be not supported", rc)
-    info = [{'key': 'ARBITRARY', 'flags': None, 'value':10, 'val_type':PMIX_INT}]
+    info = [{'key': 'ARBITRARY2', 'value':15, 'val_type':PMIX_INT}]
     rc = foo.publish(info)
     print("Publish result: ", rc)
+    procs = []
+    info  = []
+    rc = foo.fence(procs, info)
+    print("Fence result", rc)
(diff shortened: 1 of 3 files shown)


rhc54:
You will indeed get "not supported" because the `server.py` program doesn't provide the server module function to implement it. Frankly, you should be getting the same error for the `publish` command as nothing was provided to support it either.

You can see that is the case when looking at the stub functions that provide the interface between the python server and the PMIx server library - e.g., the function behind "publish":
```python
cdef int publish(const pmix_proc_t *proc,
                 const pmix_info_t info[], size_t ninfo,
                 pmix_op_cbfunc_t cbfunc, void *cbdata) with gil:
    keys = pmixservermodule.keys()
    if 'publish' in keys:
        args = {}
        myprocs = []
        ilist = []
        pmix_unload_procs(proc, 1, myprocs)
        args['proc'] = myprocs[0]
        if NULL != info:
            pmix_unload_info(info, ninfo, ilist)
            args['info'] = ilist
        rc = pmixservermodule['publish'](args)
    else:
        return PMIX_ERR_NOT_SUPPORTED
```
Since `server.py` does not provide a "publish" module, this function will always return "not supported" to the client.

That said, your code looks fine - it simply can't yield any other response until `server.py` gets extended.

dsikich:
@rhc54 Ah, ok. Do you want to wait to merge this code until server.py gets extended? Or should I submit this as a regular PR now?

rhc54:
Why don't you add the server.py code to this PR and then remove the "draft" status - we should probably commit them together.

dsikich:
@rhc54 sounds good

dsikich:
@rhc54 I'm in the process of adding the server.py code for lookup and publish so I am closing this one and will open a new PR with this code in it.

PR #1432, opened 2019-09-06
Return the correct status from PMIx_Publish

### LINKED ISSUE #1431: pmix_server_message_handler() may lose return status
I was  trying to get Stony Brook SHMEM to work with Slurm, but found that Slurm doesn’t support the PMIx_Publish() or PMIx_Lookup() calls. It look me a while to figure this out, because PMIx_Publish() was returning rc=0, even though the dummy Slurm server function _publish_fn() was returning PMIX_ERR_NOT_SUPPORTED. (Interesting to note that the PMIx_Lookup() call did properly return PMIX_ERR_NOT_SUPPORTED.)

It appears to me that pmix_server_mesage_handler() is losing the return code, probably at the PMIX_BFROPS_PACK which is currently at line 3647 in the master version of pmix_server.c.

I found this with Slurm 18.08.7 and a release candidate of PMIx 3.1.3, gitd3ccd83, built and installed locally.

System: CentOS 7.6, x86_64 hardware, Mellanox OFED 4.5 on Mellanox ConnectX-3 IB gear.

rhc54:
Fixes #1431 

Thanks to @dsikich, @artpol84, and @ariebs for all independently reporting this problem! Must have been a good week for PMIx_Publish.

### DIFF
--- a/src/client/pmix_client_pub.c
+++ b/src/client/pmix_client_pub.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2014-2018 Intel, Inc. All rights reserved.
+ * Copyright (c) 2014-2019 Intel, Inc.  All rights reserved.
  * Copyright (c) 2014-2019 Research Organization for Information Science
  *                         and Technology (RIST).  All rights reserved.
  * Copyright (c) 2014      Artem Y. Polyakov <artpol84@gmail.com>.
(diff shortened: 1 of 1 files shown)


src/client/pmix_client_pub.c
```diff
@@ -538,11 +538,12 @@ static void wait_cbfunc(struct pmix_peer_t *pr,
                        buf, &ret, &cnt, PMIX_STATUS);
     if (PMIX_SUCCESS != rc) {
         PMIX_ERROR_LOG(rc);
+        ret = rc;
     }
 
   report:
     if (NULL != cb->cbfunc.opfn) {
-        cb->cbfunc.opfn(rc, cb->cbdata);
+        cb->cbfunc.opfn(ret, cb->cbdata);
```
jjhursey: See comment on PR #1433 - the earlier error cases are no longer wired to the error mechanism correctly.

PR #1433, opened 2019-09-06
Return the correct status from PMIx_Publish

rhc54:


src/client/pmix_client_pub.c
```diff
@@ -539,11 +539,12 @@ static void wait_cbfunc(struct pmix_peer_t *pr,
                        buf, &ret, &cnt, PMIX_STATUS);
     if (PMIX_SUCCESS != rc) {
         PMIX_ERROR_LOG(rc);
+        ret = rc;
     }
 
   report:
     if (NULL != cb->cbfunc.opfn) {
-        cb->cbfunc.opfn(rc, cb->cbdata);
+        cb->cbfunc.opfn(ret, cb->cbdata);
```
jjhursey: So `ret` is the error coming from the server-side and we want to report that instead of the `rc` which is `PMIX_SUCCESS`. If `rc` isn't `PMIX_SUCCESS` then we report back `rc`. This makes sense to me (it took me a minute - which is why I'm typing this out).

However, the `PMIX_ERR_BAD_PARAM` and `PMIX_ERR_UNREACH` are no longer wired-in correctly since they only set `rc` and not `ret`. So in those two earlier cases `ret` is used without being initialized. I think those two earlier calls also need a `ret = rc` in them.

PR #1438, opened 2019-09-11
Fix python generate_regex function

dsikich:
The python generate regex function was calcuating the
length of the returned regex from PMIx_generate_regex
incorrectly. It was also indexing into regex with the
incorrect values. This was caused by the recent changes
to what is returned by the PMIx_generate_regex function.

Also, code was never entering this block:

```c
if "pmix" == regex[:4]:
```

Even when it should have, so I think the second commit should fix that, where I convert
the byte literal to ascii. 

### DIFF
--- a/bindings/python/pmix.pyx
+++ b/bindings/python/pmix.pyx
@@ -708,7 +708,7 @@ cdef class PMIxServer(PMIxClient):
             ba = bytearray(regex)
         else:
             # extract the length of the blob
-            length = int(&regex[6]) + 6 + len(&regex[6])
+            length = 5 + len(&regex[5])
             ba = bytearray(length)
             pyregex = <bytes> regex[:length]
             index = 0
@@ -728,7 +728,7 @@ cdef class PMIxServer(PMIxClient):
             ba = bytearray(ppn)
         else:
             # extract the length of the blob
-            length = int(&ppn[6]) + 6 + len(&ppn[6])
+            length = 4 + len(&ppn[4])
             ba = bytearray(length)
             index = 0
             pyppn = <bytes> ppn[:length]
--- a/bindings/python/pmix.pyx
+++ b/bindings/python/pmix.pyx
@@ -704,17 +704,27 @@ cdef class PMIxServer(PMIxClient):
         else:
             pyhosts = hosts
         rc = PMIx_generate_regex(pyhosts, &regex)
-        if "pmix" == regex[:4]:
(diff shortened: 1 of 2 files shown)


rhc54:
I'm afraid this isn't quite correct. The regex consists of a prefix followed by a ':' plus a NULL terminator (to end the string) and then the payload, which can include arbitrary characters (including NULLs) anywhere inside it. The current code assumes that the payload is a NULL-terminated string - it therefore will truncate the payload if that payload includes a NULL byte.

If the regex has a "pmix" prefix, that indicates it is a "native" regex and the payload therefore consists of a NULL-terminated string - i.e., it looks like:
```
pmix:<0><string>
```
You can't just set:
```python
ba = bytearray(regex)
```
because the bytearray function will truncate the regex at the end of the prefix (which is NULL terminated). You need to pull the entire regex (the two concatenated strings) into the bytearray.

If it has a "blob" prefix the regex (and the ppn is just another regex) looks like this:
```
blob:<0>1234<0><bytearray>
```
where `<0>` is a NULL terminator. The "1234" string tells you the length of the enclosed bytearray. You therefore need to set the size of the output bytearray to include the prefix, the size string, and the entire input bytearray.

Finally, there is a third case where this is no prefix at all - i.e., you will not find a ':' in the string. This means that the regex is simply a string and can be directly copied across as a bytearray.

dsikich:
@rhc54 ok, I think what was actually happening was that even if regex[:4] was "pmix" it was in byte form so it was never entering the if statement when it was a "native" regex. So it would go into the else block and parse the regex incorrectly in that case.

dsikich:
@rhc54 Is there still changes I need to make to this PR? I update the length calculations, and I updated the pmix string comparison since even if the byte literal was 'pmix' it was still failing the comparison without the conversion to ascii and never going into the if statement.

rhc54:
Yes - please see my prior note. The parsing of the returned value from "PMIx_generate_regex" is incorrect and needs to be updated.

dsikich:
@rhc54 ok, I pushed an update. Should PMIx_generate_ppn return a regex to the ppn variable? It looks like it is not returning a regex currently.

rhc54:
Both "generate_regex" and "generate_ppn" can return a simple string:

> Finally, there is a third case where this is no prefix at all - i.e., you will not find a ':' in the string. This means that the regex is simply a string and can be directly copied across as a bytearray.

If the regex would be longer than the simple string, then we return the string - and there will be no prefix.

PR #1440, opened 2019-09-12
Fix 'get' of data for unknown namespace

rhc54:
Data from an unknown namespace is stored on the server's hash tables, so
look for it there when trying to resolve the request.


gpaulsen:
Looks like this (or some of this) was cherry-picked back to v3.1 branch here:
https://github.com/openpmix/openpmix/commit/75189a592730c025dc5291cceff19eca4dd64104

PR #1449, opened 2019-09-16
Continue prototyping Python bindings

rhc54:


rhc54:
@dsikich FYI - this covers most of the remaining bindings. Not tested, so consider these just prototypes

PR #1455, opened 2019-09-19
Minor touchup to the python bindings

rhc54:

### DIFF
--- a/bindings/python/pmix.pyx
+++ b/bindings/python/pmix.pyx
@@ -76,6 +76,7 @@ cdef class PMIxClient:
         memset(self.myproc.nspace, 0, sizeof(self.myproc.nspace))
         self.myproc.rank = <uint32_t>PMIX_RANK_UNDEF
         myhdlrs = []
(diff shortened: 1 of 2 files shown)


rhc54:
@jjhursey This is really weird...and disturbing. The status window showed as "all 10 tests passed", which seemed odd given your earlier note. I checked all the tests and verified that all showed green, so I went ahead an committed it - and then the status changed to show the "make check" cross-version test as having failed!

Any idea how that can possibly happen?

jjhursey:
Huh that is odd. I've not seen that before. Let me look at the logs to see if it sent success before it sent failure.

jjhursey:
@rhc54 I think I may have found the issue. The "postprocessing" script handles weird Jenkins failure cases for us. In one scenario it would send back "success" then have Jenkins send "failed". it's 1-2 seconds apart, but depending on when GH shows the update you could have seen it during that window. I've fixed the script so that it won't do that any more.

PR #1458, opened 2019-09-25
add python wrappers and server host functions

dsikich:
Server and client functions for publish, unpublish, and lookup. Also uses the thread shift example to call the lookup callback, but I think that maybe needs some work in terms of how to generalize it for the rest of the callback functions. 

  - Add client and server module functions
      publish, unpublish, and lookup
    - Add initial thread shifting with timer
      based on example in PR #1457 using
      lookup callback function
    - Add load and unload functions for pdata
      struct

### DIFF
--- a/bindings/python/client.py
+++ b/bindings/python/client.py
@@ -25,7 +25,6 @@ def main():
     procs = []
     info = []
     rc = foo.fence(procs, info)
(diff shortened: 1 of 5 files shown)


rhc54:
The cross-version check is failing for an unrelated matter that I still need to resolve.

PR #1467, opened 2019-10-01
add python query_info client wrapper

dsikich:
@rhc54 Getting segfault in pmix_query.c when no qualifiers are specified. I want to verify I am running query info correctly in the python client. I believe it should return a function not supported message if it is being called correctly. 

### DIFF
--- a/bindings/python/pmix.pxi
+++ b/bindings/python/pmix.pxi
@@ -903,6 +903,57 @@ cdef void pmix_free_pdata(pmix_pdata_t *array, size_t sz):
         n += 1
     PyMem_Free(array)
 
+# Convert a dictionary of key-value pairs into an
+# array of pmix_query_t structs
+#
+# @array [INPUT]
+#        - malloc'd array of pmix_query_t structs
+#
+# @pyq [INPUT]
+#          - a list of dictionaries, where each
+#            dictionary has a list of keys, and a
+#            list of qualifers
+#
+cdef int pmix_load_queries(pmix_query_t *array, pyq:list):
+    n = 0
+    for q in pyq:
+        nstrings = len(q['keys'])
+        if 0 < nstrings:
+            array[n].keys = <char **> PyMem_Malloc(nstrings * sizeof(char*))
+            if not array[n].keys:
+                return PMIX_ERR_NOMEM
+        # load keys from dictionary
+        n = 0
+        for k in q['keys']:
+            #keystr = str(k).encode('ascii')
+            array[n].keys[n] = strdup(k)
+            n += 1
+        array[n].nqual = len(q['qualifiers'])
+        rc = pmix_alloc_info(&(array[n].qualifiers), &(array[n].nqual), q['qualifiers'])
+        n += 1
+    if PMIX_SUCCESS != rc:
+        return rc
+    return PMIX_SUCCESS
+
+cdef int pmix_unload_queries(const pmix_query_t *queries, size_t nqueries, ilist:list):
+    cdef char* kystr
+    cdef size_t n = 0
+    keylist = []
+    qualist = []
+    query = {}
+    while n < nqueries:
+        n = 0
+        while NULL != queries.keys[n]:
+            mykey = str(queries.keys[n])
+            keylist.append(mykey)
+            n += 1
+        pmix_unload_info(queries[n].qualifiers, queries[n].nqual, qualist)
+        query['keys']       = keylist
+        query['qualifiers'] = qualist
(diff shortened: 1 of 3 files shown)


bindings/python/pmix.pxi
```diff
@@ -903,6 +903,57 @@ cdef void pmix_free_pdata(pmix_pdata_t *array, size_t sz):
...
+#
+# @pyq [INPUT]
+#          - a list of dictionaries, where each
+#            dictionary has a list of keys, and a
+#            list of qualifers
+#
+cdef int pmix_load_queries(pmix_query_t *array, pyq:list):
+    n = 0
+    for q in pyq:
+        nstrings = len(q['keys'])
```
rhc54: Perhaps use the pmix_load_argv function instead of doing it manually here?

bindings/python/pmix.pxi
```diff
@@ -903,6 +903,57 @@ cdef void pmix_free_pdata(pmix_pdata_t *array, size_t sz):
...
+            if not array[n].keys:
+                return PMIX_ERR_NOMEM
+        # load keys from dictionary
+        n = 0
+        for k in q['keys']:
+            #keystr = str(k).encode('ascii')
+            array[n].keys[n] = strdup(k)
+            n += 1
+        array[n].nqual = len(q['qualifiers'])
+        rc = pmix_alloc_info(&(array[n].qualifiers), &(array[n].nqual), q['qualifiers'])
```
rhc54: I'm confused here. Why compute array[n].nqual and then call alloc_info with the address of that value? It appears that the function assigns the value, so no point in computing it first.

BTW: this is the location of the problem. Reuse of 'n' means that array[n] is now pointing into some random memory location.

bindings/python/pmix.pxi
```diff
@@ -903,6 +903,57 @@ cdef void pmix_free_pdata(pmix_pdata_t *array, size_t sz):
...
+cdef int pmix_load_queries(pmix_query_t *array, pyq:list):
+    n = 0
+    for q in pyq:
+        nstrings = len(q['keys'])
+        if 0 < nstrings:
+            array[n].keys = <char **> PyMem_Malloc(nstrings * sizeof(char*))
+            if not array[n].keys:
+                return PMIX_ERR_NOMEM
+        # load keys from dictionary
+        n = 0
```
rhc54: Looks like you are reusing 'n' inside of a loop that is indexing 'n' - so I suspect the routine is overwriting array[n]? I think you need a second counter here.

bindings/python/pmix.pxi
```diff
@@ -903,6 +903,57 @@ cdef void pmix_free_pdata(pmix_pdata_t *array, size_t sz):
...
+    return PMIX_SUCCESS
+
+cdef int pmix_unload_queries(const pmix_query_t *queries, size_t nqueries, ilist:list):
+    cdef char* kystr
+    cdef size_t n = 0
+    keylist = []
+    qualist = []
+    query = {}
+    while n < nqueries:
+        n = 0
```
rhc54: Ouch - reusing 'n' again here.

bindings/python/pmix.pyx
```diff
@@ -617,10 +620,62 @@ cdef class PMIxClient:
...
+                queries = <pmix_query_t*> PyMem_Malloc(nqueries * sizeof(pmix_query_t))
+                if not queries:
+                    return PMIX_ERR_NOMEM
+            rc = pmix_load_queries(queries, pyq)
+            print("RC LOAD QUERIES: ", self.error_string(rc))
+            if PMIX_SUCCESS != rc:
+                n = 0
+                while n < nqueries:
+                    pmix_free_info(queries[n].qualifiers, queries[n].nqual)
+                    n = 0
```
rhc54: Reusing 'n' here

bindings/python/pmix.pyx
```diff
@@ -617,10 +620,62 @@ cdef class PMIxClient:
...
+        if PMIX_SUCCESS == rc:
+            rc = pmix_unload_queries(queries, nqueries,  pyresults)
+            # free results info structs
+            pmix_free_info(results, nresults)
+        # free memory for query structs
+        if 0 < nqueries:
+            n = 0
+            while n < nqueries:
+                pmix_free_info(queries[n].qualifiers, queries[n].nqual)
+                n = 0
```
rhc54: Reusing 'n' here

dsikich:
@rhc54 thanks for the feedback, I will close this draft and submit a new one when fixed.

PR #1468, opened 2019-10-01
Test blocking form of PMIx_Query_info

rhc54:


rhc54:
Hmmm...somehow, I don't think this is a problem with our code:
```
checking for gcc... pgcc
checking whether the C compiler works... no
configure: error: in `/smpi_dev/mpiczar/jenkins/workspace/pmix_public_pr_master_pgi/pmix-src':
configure: error: C compiler cannot create executables
See `config.log' for more details
```
@jjhursey Any ideas?

jjhursey:
yeah there was a file system blip that it probably got caught in.

PR #1470, opened 2019-10-03
add python query_info client wrapper

dsikich:
Add query client wrapper, and server
module function. This also adds NULL
terminator to pmix_load_argv, and
the alloc_info function was removed.


dsikich:
@rhc54 it is not letting me assign you as a reviewer? Not sure why. Also I need to go back and look at the alloc_info function since I didn't get to test it yet after the NULL terminator fix was added to pmix_load_argv. If it does work properly, I will add it to another PR.

PR #1476, opened 2019-10-08
Cache local peers in deprecated form for older versions, try to fix the "tests", continue workaround of dstore

rhc54:
**Provide some handy definitions for comparing release versions.** 
Update the dbase store_job_info function to check for an earlier version and translate the node info array, extracting and storing just the local peers for each node.

**Try to fix the "test" area**
The tests do not correctly compute the node and proc maps when running
with multiple servers - they only compute those values for the local
server. This wasn't detected in the past because we weren't using that
info. However, now that we are correctly tracking info by node, this
causes the tests to become confused and fail.

I was unable to get the tests to correctly compute the pattern of proc
assignments for multiple servers, so just tell the tests to only run one
server until someone with more knowledge of this code can fix the
problems.

**Continue to work around the dstore's limitations**
There are additional places where we need to get node or app-specific
info, so test for those keys and redirect to the hash component when
found. Note that this is impacting overall latency as these checks must
be done in the critical path prior to looking at the dstore.

Allow the server to pass its hostname to be used by its clients so that
a test can establish a "fake" hostname.

### DIFF
--- a/src/client/pmix_client.c
+++ b/src/client/pmix_client.c
@@ -237,14 +237,6 @@ static void job_data(struct pmix_peer_t *pr,
                             pmix_client_globals.myserver,
                             nspace, buf);
 
-    /* if anything is left in the buffer, let the
-     * internal hash component have it - e.g., we
-     * may have been passed node and app info for
-     * our own nspace */
-    PMIX_GDS_STORE_JOB_INFO(cb->status,
-                            pmix_globals.mypeer,
-                            nspace, buf);
-
     free(nspace);
     cb->status = PMIX_SUCCESS;
     PMIX_POST_OBJECT(cb);
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -86,7 +86,9 @@ PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc,
 {
     pmix_cb_t *cb;
     pmix_status_t rc;
-    size_t n;
+    size_t n, nfo;
+    pmix_proc_t p;
+    pmix_info_t nodeinfo, *iptr;
 
     PMIX_ACQUIRE_THREAD(&pmix_global_lock);
 
@@ -101,7 +103,37 @@ PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc,
                         (NULL == proc) ? "NULL" : PMIX_NAME_PRINT(proc),
                         (NULL == key) ? "NULL" : key);
 
+    memcpy(&p, proc, sizeof(pmix_proc_t));
+    iptr = (pmix_info_t*)info;
+    nfo = ninfo;
+
     if (PMIX_RANK_UNDEF == proc->rank || NULL == key) {
+        goto doget;
+    }
+    /* if they are asking about a node-level piece of info,
+     * then the rank must be UNDEF */
+    if (pmix_check_node_info(key)) {
+        p.rank = PMIX_RANK_UNDEF;
+        /* see if they told us to get node info */
+        if (NULL == info) {
+            /* guess not - better do it */
+            PMIX_INFO_LOAD(&nodeinfo, PMIX_NODE_INFO, NULL, PMIX_BOOL);
+            iptr = &nodeinfo;
+            nfo = 1;
+        }
+        goto doget;
+    }
+    /* if they are asking about an app-level piece of info,
+     * then the rank must be UNDEF */
+    if (pmix_check_app_info(key)) {
+        p.rank = PMIX_RANK_UNDEF;
+        /* see if they told us to get app info */
+        if (NULL == info) {
+            /* guess not - better do it */
+            PMIX_INFO_LOAD(&nodeinfo, PMIX_APP_INFO, NULL, PMIX_BOOL);
+            iptr = &nodeinfo;
+            nfo = 1;
+        }
         goto doget;
     }
 
@@ -112,10 +144,10 @@ PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc,
             PMIX_CHECK_KEY(info, PMIX_SESSION_INFO)) {
             goto doget;
         }
-
     }
+
     /* try to get data directly, without threadshift */
-    if (PMIX_SUCCESS == (rc = _getfn_fastpath(proc, key, info, ninfo, val))) {
+    if (PMIX_SUCCESS == (rc = _getfn_fastpath(&p, key, iptr, nfo, val))) {
         goto done;
     }
 
@@ -124,7 +156,7 @@ PMIX_EXPORT pmix_status_t PMIx_Get(const pmix_proc_t *proc,
      * recv routine so we know which callback to use when
      * the return message is recvd */
     cb = PMIX_NEW(pmix_cb_t);
-    if (PMIX_SUCCESS != (rc = PMIx_Get_nb(proc, key, info, ninfo, _value_cbfunc, cb))) {
+    if (PMIX_SUCCESS != (rc = PMIx_Get_nb(&p, key, iptr, nfo, _value_cbfunc, cb))) {
         PMIX_RELEASE(cb);
         return rc;
     }
@@ -576,6 +608,7 @@ static void _getnbfn(int fd, short flags, void *cbdata)
     pmix_proc_t proc;
     bool optional = false;
     bool immediate = false;
+    bool internal_only = false;
     struct timeval tv;
     pmix_query_caddy_t *cd;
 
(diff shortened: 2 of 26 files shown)


rhc54:
@jjhursey Any idea on why the PGI compiler keeps reporting "compiler cannot build executables"?

jjhursey:
Thanks for the heads up. Since CI no longer posts comments I don't get any notice when CI fails like this.
It looks like someone moved the PGI compiler on the box. Let me see if I can fix.

jjhursey:
bot:ibm:pgi:retest

jjhursey:
There I got it back on track. Sorry for the glitch.

jjhursey:
I'm seeing CI hangs in the make check cross version CI. I just happened to notice that it's hanging on the CI for this PR. I'm trying to get it cleared out now, but a couple of the hanging test cases:
```
12:07:11 ======================================================================
12:07:11 Server : master -> Client:    xv2
12:07:11 -----> : Run Client
12:07:11     Client: /home/pmixer/scratch/pmix-xv2/test/simple/simpclient
12:07:11 Server : /home/pmixer/scratch/pmix-master/test/simple
12:07:11 Command: cd /home/pmixer/scratch/pmix-master/test/simple ; ./simptest -n 2 -e /home/pmixer/scratch/pmix-xv2/test/simple/simpclient
14:17:51 Status : -15 ***FAILED***
14:17:51 -----> : Test output
14:17:51 Testing version 4.0.0a1
14:17:51 Collecting inventory
14:17:51 Inventory collected: 0
14:17:51 [64398cc18d90:60357] Client ns foobar rank 0: Running
14:17:51 [64398cc18d90:60357] CLIENT SERVER URI: pmix-server.60354;tcp4://127.0.0.1:54728
14:17:51 foobar:0 Model event handler called with status -147(PMIX MODEL DECLARED)
14:17:51 foobar:0	pmix.pgm.model:	TEST
14:17:51 foobar:0	pmix.mdl.name:	PMIX
14:17:51 foobar:0	pmix.evname:	SIMPCLIENT-MODEL
14:17:51 [64398cc18d90:60358] Client ns foobar rank 1: Running
14:17:51 [64398cc18d90:60358] CLIENT SERVER URI: pmix-server.60354;tcp4://127.0.0.1:54728
14:17:51 [64398cc18d90:60357] Client foobar:0 job size 2
14:17:51 foobar:1 Model event handler called with status -147(PMIX MODEL DECLARED)
14:17:51 foobar:1	pmix.pgm.model:	TEST
14:17:51 foobar:1	pmix.mdl.name:	PMIX
14:17:51 foobar:1	pmix.evname:	SIMPCLIENT-MODEL
14:17:51 [64398cc18d90:60358] Client foobar:1 job size 2
14:17:51 [64398cc18d90:60358] Client ns foobar rank 1: PMIx_Get local peers failed: NOT-FOUND
14:17:51 [64398cc18d90:60357] Client ns foobar rank 0: PMIx_Get local peers failed: NOT-FOUND
14:17:51 [64398cc18d90:60358] Client ns foobar rank 1: Finalizing
14:17:51 [64398cc18d90:60357] Client ns foobar rank 0: Finalizing
14:17:51 
14:17:51 ======================================================================
```

```
14:17:52 ======================================================================
14:17:52 Server :   v3.1 -> Client:    xv2
14:17:52 -----> : Run Client
14:17:52     Client: /home/pmixer/scratch/pmix-xv2/test/simple/simpclient
14:17:52 Server : /home/pmixer/scratch/pmix-v3.1/test/simple
14:17:52 Command: cd /home/pmixer/scratch/pmix-v3.1/test/simple ; ./simptest -n 2 -e /home/pmixer/scratch/pmix-xv2/test/simple/simpclient
14:19:20 Status : -15 ***FAILED***
14:19:20 -----> : Test output
14:19:20 Testing version 3.1.5rc1
14:19:20 Collecting inventory
14:19:20 Inventory collected: 0
14:19:20 [64398cc18d90:60433] Client ns foobar rank 1: Running
14:19:20 [64398cc18d90:60433] CLIENT SERVER URI: pmix-server.60429;tcp4://127.0.0.1:60737
14:19:20 [64398cc18d90:60432] Client ns foobar rank 0: Running
14:19:20 [64398cc18d90:60432] CLIENT SERVER URI: pmix-server.60429;tcp4://127.0.0.1:60737
14:19:20 foobar:1 Model event handler called with status -147(PMIX MODEL DECLARED)
14:19:20 foobar:1	pmix.pgm.model:	TEST
14:19:20 foobar:1	pmix.mdl.name:	PMIX
14:19:20 foobar:1	pmix.evname:	SIMPCLIENT-MODEL
14:19:20 foobar:0 Model event handler called with status -147(PMIX MODEL DECLARED)
14:19:20 foobar:0	pmix.pgm.model:	TEST
14:19:20 foobar:0	pmix.mdl.name:	PMIX
14:19:20 foobar:0	pmix.evname:	SIMPCLIENT-MODEL
14:19:20 [64398cc18d90:60433] Client foobar:1 job size 2
14:19:20 [64398cc18d90:60432] Client foobar:0 job size 2
14:19:20 [64398cc18d90:60433] Client ns foobar rank 1: PMIx_Get local peers failed: NOT-FOUND
14:19:20 [64398cc18d90:60432] Client ns foobar rank 0: PMIx_Get local peers failed: NOT-FOUND
14:19:20 [64398cc18d90:60433] Client ns foobar rank 1: Finalizing
14:19:20 [64398cc18d90:60432] Client ns foobar rank 0: Finalizing
14:19:20 
14:19:20 ======================================================================
```

jjhursey:
Here is the final gist:
 * https://gist.github.com/ibm-ompi/d91e15e16b9fa59756af75fe15064304

rhc54:
Yeah, I know - this was a checkpoint commit that passes "make check". The remaining problems are in the tests themselves where we attempt to "fake" having multiple nodes - but do it incorrectly. I'll be taking a look at that problem as I poke some more at it over the next few weeks.

artpol84:
@rhc54, I know it's been a while since this went in, but I'd like to confirm the cases where dstore is bypassed with this change.
It seems to deal with PMIx versions, so does this only enabled if the versions are different? Or for the same version, we will also bypass the dstore while accessing job-level info?
The reason I'm asking is that for dstore we have a fastpath, while for hash we don't. So I'm trying to understand what kind of data will be affected by this change.

rhc54:
Actually, it has little to do with the PMIx version. We check to see if the server is earlier than 3.1.5 simply because the functionality otherwise executed (i.e., when the server is 3.1.5 or 4.x) isn't supported by earlier versions of the server. If someone is trying to access things like node-level info while working against an earlier server version, they are simply going to get an error.

dstore is effectively disabled when the client requests node-level, app-level, or session-level information. In such cases, the dstore doesn't currently handle the corresponding data arrays and will fail to return that info. There is a new "latency" introduced by the need to bypass dstore, however, as we have to check the PMIx_Get request in all cases. Thus, even if they aren't asking for node/app/session-level info, we still have to check.

Note that I also added a check in the dstore code itself that tries to alleviate the situation for earlier clients by transferring at least the node-local info to the dstore. Thus, if the client asks for node-level info for their own local node, they can get an answer - it will still fail, however, if they ask for the info from a different node. I do translate the info about what peers are on each node into the old way of indexing it (i.e., using the hostname as the key) in that case so that "resolve_peers" can still work.

Likewise, if they are asking for info on a key that is not associated with a proc (i.e., came in via the PMI-1/2 interfaces), or are asking for a NULL key (thereby indicating they want all the data returned), we also have to disable the dstore to properly handle the request. Fortunately, those checks are pretty quick - but they still do introduce some latency.

IIRC, I raised the latency impact concern earlier and asked that someone check to see if it hurts the Spectrum MPI case, and if so by how much. I don't recall hearing back, so it is possible that hasn't been checked yet. I think it would be worth at least a quick check to see how much it hurt them.

artpol84:
> IIRC, I raised the latency impact concern earlier and asked that someone check to see if it hurts the Spectrum MPI case, and if so by how much. I don't recall hearing back, so it is possible that hasn't been checked yet. I think it would be worth at least a quick check to see how much it hurt them.

Yes, it needs to be checked:

Looking at this from the other angle, what prevents dstore from handling node-level, app-level and session-level info? I believe we discussed that, but I have forgotten. 
Is it an architectural limitation? If so - is it PMIx standard level or dstore level?
If not - is it just a lack of support in the dstore itself?
I think

rhc54:
Just lack of support in dstore itself - you need to be able to handle the job/node/app/session info array keys. I saw no architectural limitation in dstore that would get in the way. I just didn't want to make such substantial changes in that code as I don't know it well enough. If/when you folks do update it, I expect we should be able to remove those checks and restore the unprotected fastpath.

artpol84:
Cool. Will take care of this after the holiday break

PR #1479, opened 2019-10-14
Pyeventhandler

dsikich:
The GIL was causing the pyeventhandler to lock up. Occasionally on some test runs the user event handler is not invoked, or the Register_event_handler (blocking) call from the C API returns a -1. So I'm still working on what is going on there.

### DIFF
--- a/bindings/python/pmix.pxi
+++ b/bindings/python/pmix.pxi
@@ -75,6 +75,16 @@ ctypedef struct pmix_pyshift_query_t:
     pmix_info_cbfunc_t query
     void *cbdata
 
+ctypedef struct pmix_pyshift_event_handler_t:
+    char *op
+    pmix_status_t status
+    pmix_info_t *results
+    size_t nresults
+    pmix_op_cbfunc_t cbfunc
+    void *thiscbdata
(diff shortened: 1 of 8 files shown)


dsikich:
The evhdlr _registration_id is always zero. Sometimes the user event handler is invoked, but only when its refid is also zero.

rhc54:
@dsikich Please see #1480 - should resolve the issue

dsikich:
@rhc54 Hmm I ran the python tests and now the user event handler never gets invoked

dsikich:
@rhc54 after looking into it. The user event handler is added to myhdlrs in register_event_handler, and then when it goes to run pyeventhandler the myhdlrs list is empty? Not sure if it would try to run pyeventhandler before the handler is added to the myhdlrs list.

rhc54:
Be sure you have "global myhdlrs" in both pyeventhandler and register_event - otherwise, they are treated as local variables

dsikich:
@rhc54 Yeah, I noticed it didn't, but I tested after adding 'global', and the list is still empty when it enters pyeventhandler, but in register_event_handler it is added

PR #1482, opened 2019-10-15
Python: Cache events if evhdlr not found

rhc54:
There can be a race condition between registration and recording the
event handler in the myevhdlr list. If we lose that race, then just
cache the event in a timeshift and try again.


rhc54:
@dsikich Should fix the race condition problem

rhc54:
bot:ibm:xver:retest

rhc54:
bot:ibm:xvercheck:retest

dsikich:
@rhc54 there was a simple bug using the rc before it was set in the 'if not found' part of pyeventhandler. After I fixed that it is still not calling pyeventhandler it seems, not sure why yet.

PR #1483, opened 2019-10-16
Clarify tests to show that Python evhdlrs are being invoked

rhc54:
The client.py program was finalizing before the event handler had a
chance to actually be invoked. A better solution would be to "hold"
client.py until the handler was invoked and had returned using the
available handshake protocols in the event handler callback, but this
suffices to illustrate the situation


dsikich:
@rhc54 thanks, that certainly helps see what is going on.

PR #1490, opened 2019-10-17
python: add correct return from directmodex

dsikich:
Add correct return to directmodex server module function,
and allocate memory for strings in caddy objects for both
directmodex and fencenb.

### DIFF
--- a/bindings/python/pmix.pxi
+++ b/bindings/python/pmix.pxi
@@ -89,6 +89,16 @@ ctypedef struct pmix_pyshift_event_handler_t:
     void *notification_cbdata
     pmix_event_notification_cbfunc_fn_t event_handler
 
+ctypedef struct pmix_pyshift_dmodex_t:
+    char *op
+    pmix_modex_cbfunc_t modex
+    pmix_status_t status
+    const char *data
+    size_t ndata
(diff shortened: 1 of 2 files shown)


dsikich:
@rhc54 I'm currently going through the python portion of the v4 standard draft to make sure the return signatures on the server module functions match what is in there.. I don't know if there is a good way to test directmodex since I think it will require multiple servers?

rhc54:
@dsikich You can simulate a direct modex behavior by having the server.py module function simply provide a data payload that contains some piece of info that your client.py can "get".

dsikich:
@rhc54 is the dmodex_fn in test/simple/simptest.c an example of what you mean? I translated something similar to python but it locks up. The dmodex_request function in pmix.pyx passes a dmodx_cbfunc to PMIx_server_dmodex_request, which looks like a non-blocking call. It looks like the dmodex_request is blocking with the wait call until the rc is returned? I'm trying to get a better understanding of the purpose of dmodx_cbfunc so I can figure out why the call to dmodex_request from the client is hanging. It looks like dmodx_request is just waiting until dmodx_cbfunc sets the flag for it to continue, which seems right to me.

PR #1506, opened 2019-10-22
python: add callbacks and client wrappers

dsikich:
1. Add toolconnected and allocate callbacks, invocations,
and return correct data from their servermodule functions

2. Add client wrapper for PMIx_Process_monitor

3. Add client wrapper for PMIx_Get_credential and server cb to server module function

4. Add client wrapper for PMIx_Validate_credential and server cb to server module function

### DIFF
--- a/bindings/python/pmix.pxi
+++ b/bindings/python/pmix.pxi
@@ -80,9 +80,20 @@ ctypedef struct pmix_pyshift_t:
     pmix_event_notification_cbfunc_fn_t event_handler
     pmix_tool_connection_cbfunc_t toolconnected
     pmix_credential_cbfunc_t getcredential
+    pmix_validation_cbfunc_t validationcredential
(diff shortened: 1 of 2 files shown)


dsikich:
@rhc54 in particular can you check I called the allocate callback correctly? I used NULL for the last two parameters since I based it off other calls to pmix_op_cbfunc_t cbfunc.

dsikich:
@rhc54 I added a client wrapper for PMIx_Process_monitor. I noticed that in the v4 draft standard that the server does not return any data other than the return code? But yet the client function does return a results array? So not sure which one to go with? This is also true for the get credential and validate credential functions. They have results arrays but the v4 standard doesn't have return data for them. I'm also wondering about the group client functions, as I cannot find those in the v4 standard either? I've been going through to make sure all of the functions are in the python bindings and match the v4 draft for python.

PR #1507, opened 2019-10-27
Update the "make check" tests

rhc54:
Use SIGCHLD detection to determine that a child has exited as it is more reliable. Track the returned exit codes to help determine success or failure of the overall tests. Minor cleanup of the exit status for pmix_client.

### DIFF
--- a/test/cli_stages.c
+++ b/test/cli_stages.c
@@ -17,6 +17,7 @@
 cli_info_t *cli_info = NULL;
 int cli_info_cnt = 0;
 bool test_abort = false;
+bool test_complete = false;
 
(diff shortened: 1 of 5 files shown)


test/cli_stages.c
```diff
@@ -17,6 +17,7 @@
 cli_info_t *cli_info = NULL;
 int cli_info_cnt = 0;
 bool test_abort = false;
+bool test_complete = false;
```
artpol84: @rhc54 IIUC, `test_complete` should be `volatile` to prevent it to be cached in the register. SIGCHLD callback will be invoked in the service thread and the poll happens in the main thread.

PR #1509, opened 2019-10-28
Python add wrappers and fix return codes

dsikich:
1. fix client wrapper return info for get_credential and monitor
2. Add wrappers for iof_pull and iof_push
3. Add server wrapper for register_attributes function


dsikich:
The v4 draft standard has iof_push and iof_pull as part of the Tool class, they already had server stub functions but no wrapper in PMIx.pyx. It is not clear to me if these belong in the tool or client class, because the standard has them in both? Also not clear whether iof_deregister is needed for Tool class?

rhc54:
I would put all the IOF functions in the tool class, including "deregister_iof" as a tool might want to collect output for awhile and then perhaps switch to some other source. I think tools are the primary consumer of that feature.

dsikich:
@rhc54 I don't see PMIx_IOF_Deregister in the server interface in include/pmix_server.h? I see the function pointers for iof_pull and push_stdin: 
``` C
1291         self.myserver.iof_pull = <pmix_server_iof_fn_t>iofpull
1292         self.myserver.push_stdin = <pmix_server_stdin_fn_t>pushstdin
```
So I'm not sure which function pointer to call for the IOF_Deregister?

PR #1512, opened 2019-10-30
Fix tool and query support tests

rhc54:
Signed-off-by: Ralph Castain <rhc@pmix.org>

PR #1514, opened 2019-10-30
CI: enable MLNX CI using Github Actions

amaslenn:
Original testing string was here:
https://github.com/mellanox-hpc/jenkins_scripts/blob/master/jenkins/pmix/pmix_jenkins.sh

This scipt became an `entrypoint.sh` for MLNX Actions with the following
key differences:
1. Removed COverity related code. PMIx uses official scan for that.
2. Removed Jenkins specific code. Variables with `jenkins` in name were
not renamed to keep files comparable.
3. `make check` is now part of the build

### DIFF
--- a/.github/actions/mlnx/Dockerfile
+++ b/.github/actions/mlnx/Dockerfile
@@ -1,0 +1,9 @@
+FROM centos:7.6.1810
+
+RUN yum install -y perl perl-Data-Dumper \
+    automake libtool flex make bzip2 git which rpm-build libevent-devel
+
+COPY entrypoint.sh /entrypoint.sh
+
+ENTRYPOINT ["/entrypoint.sh"]
+
--- a/.github/actions/mlnx/action.yml
+++ b/.github/actions/mlnx/action.yml
@@ -1,0 +1,12 @@
+name: 'MLNX CI for PMIx'
+description: 'Build & test PMIx'
+inputs:
+  target:
+    description: 'build/srcrpm/test'
+    required: true
+    default: 'build'
+runs:
+  using: 'docker'
+  image: 'Dockerfile'
+  args:
+    - ${{ inputs.target }}
--- a/.github/actions/mlnx/entrypoint.sh
+++ b/.github/actions/mlnx/entrypoint.sh
@@ -1,0 +1,424 @@
+#!/bin/bash -xeE
+
+rel_path=$(dirname $0)
+abs_path=$(readlink -f $rel_path)
+
+if [ "$1" = "build" ]; then
+    jenkins_test_build="yes"
+    jenkins_test_check="no"
+    jenkins_test_src_rpm="no"
+elif [ "$1" = "srcrpm" ]; then
+    jenkins_test_build="no"
+    jenkins_test_check="no"
+    jenkins_test_src_rpm="yes"
+elif [ "$1" = "test" ]; then
+    jenkins_test_build="no"
+    jenkins_test_check="yes"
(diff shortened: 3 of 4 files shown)


amaslenn:
cc @karasevb

amaslenn:
Could admins please enable Github Actions in the repo settings? cc @mike-dubman

artpol84:
@jjhursey, could you help with this?
This is a replacement for our Jenkins CI.

amaslenn:
> Why did you delete the link between `make_tarball` and `make_dist_tarball`? These are intentionally supposed to be linked to each other to avoid drift. Is there some intended difference?

My bad, I changed unrelated files. Will update.

amaslenn:
It should be good now. Don't know what is wrong with `IBM CI (Cross-version: make check)`.

Results for the new check could be found here (see the most recent one): https://github.com/amaslenn/openpmix/actions.
Two tests (`direct` and `full modex`) are failing on `master` but passed for `v3.1.4`.

rhc54:
@amaslenn Thanks! Yes, those two tests are not correct and will always fail on master and v3.1.5. The problem is that the tests incorrectly pass a node and proc map that only contains data for the local node - both should contain data for the entire job. We got away with it before because PMIx wasn't tracking node-specific data. Now that the tracking has been correctly implemented, the tests need to be corrected.

I tried to take a shot at it but couldn't get the proc assignment pattern quite correct. For now, I have the tests locked to just one server. I suspect you have those two tests using multiple servers in your test script, yes? You'll either need to set it to one server or someone will need to fix the test server code.

jjhursey:
Is this something that needs to be cherry-picked to each of the release branches that need Mellanox CI testing? If so, are you worried about a drift between the release branches?

The `make check` issue is that there is a break on master that is causing one of the tests to faili with a ton of output. Too much for GitHub to handle so it is rejecting the update to the Gist after testing. I made a note [here](https://github.com/openpmix/openpmix/pull/1512#issuecomment-548075454) about the error. It is not related to this PR.

amaslenn:
@rhc54 new MLNX CI (a Github Action) runs inside a docker container, so it is a single node test.

@jjhursey wherever these new files exist, CI will be enabled there. Please let me know which braches should I update with this change once this one merged.

Github Actions will be generally available mid of November. Prior that @openpmix org should sign up for a beta program to enable this check.

karasevb:
>  I suspect you have those two tests using multiple servers in your test script, yes?
That's right, there is a simulation of two PMIx servers. I'll see why it fails for `master`.

jjhursey:
Thanks @karasevb for taking a look. We probably need a central place for tracking the fixes related to `make check` - I created Issue #1515 to help coordinate the work left to do.

jjhursey:
I don't have permission to enable GitHub Actions for this project it seems. But we will work on that.

The `master` and `v3.1` branches are the two active at the moment. So you will want to cherry-pick this back to the `v3.1` branch.

My only concern with this type of change is the potential for drift between the versions in the various branches over time. I am wondering if we can leverage the `pmix-tests` repo to hold the CI script portion of this, then have the Docker container pull the latest version of the script for `pmix-tests` when it does it's testing. Something like that would make it easier to update the CI across all branches with a single change. I don't know if that is possible with GitHub Actions or not.

rhc54:
GitHub Actions is a pre-beta new feature for GitHub, and so openpmix had to be "signed up" for their test program before it could be enabled. I've done that now and will commit this PR so it can start operating.

PR #1519, opened 2019-11-01
Correctly xfer node-level data arrays for pre-v3.1.5 clients

rhc54:
Earlier clients are looking for node info as a data array with the name of the node as the key.


rhc54:
bot:ibm:xvercheck:retest

PR #1533, opened 2019-11-04
Repair job-fence tests

rhc54:
No such thing as "remote" procs as all the testing is done on a single
node.

### DIFF
--- a/test/test_fence.c
+++ b/test/test_fence.c
@@ -63,7 +63,7 @@ int test_fence(test_params params, char *my_nspace, pmix_rank_t my_rank)
     pmix_proc_t *pcs;
     bool participate;
     int fence_num = 0;
-    char *sval;
+    char sval[500];
     int put_ind;
 
     if (NULL != params.noise) {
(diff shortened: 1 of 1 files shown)


artpol84:
@rhc54 this is not the right thing to do.
we are actually using this test with Slurm and for slurm there are remote processes!

artpol84:
I see that `get_local_peers` is still in the master. I'd like to get some context on where this functionality was removed, where - not and what was the reason.

v2.2 is not really a branch we care of, so I don't mind this being removed from v2.2 only.
But we do use pmix_client binary to run with Slurm in our testing.

rhc54:
That's fine - I noted in the comments that we were constraining things because the tests were not correctly computing the process map, but were getting away with it due to compensating errors. I'd have to look at it again to see where it was removed.

Sadly, the community does have to care about v2.2, at least for a while longer due to some issues in compatibility with existing Slurm installations. Whatever solution eventually deployed is going to have to account for it.

artpol84:
v3.1 has it: https://github.com/openpmix/openpmix/blob/v3.1/test/test_fence.c#L368

I'm basically voting for getting this functionality back everywhere. Especially if we will use multi-server. In this case, though we will work with 1 node, there will be 2 virtual nodes.

rhc54:
I'm unaware of any objection to doing so - I took it as far as I could given a limited understanding of what the code was attempting to do. Feel free to modify as required. It would also be good to comment the code a little more thoroughly while you folks are working on it in case others need to dig into it again in the future.

PR #1535, opened 2019-11-05
python: add missing wrappers and iof hdler

dsikich:
- add tool iof_deliver function
    - add server collect_inventory function
    - add server deliver_inventory function
    - move iof functions to tool class
    - add pyiofhandler function for iof_pull


dsikich:
@rhc54 the iof_deregister function is commented out in pmix.pyx because I get a runtime error that it cannot find the symbol when I run the python tests even though it compiles. I'm not sure if I need to add a "stub" function for this call, and I can find the server module functions for iof pull/push but not deregister, so not sure.

PR #1549, opened 2019-11-15
Fix memory leak as explained in issue 1548

dsolt:
Conflicts:
	src/mca/common/dstore/dstore_base.c


rhc54:
@dsolt Can you please sign the commit?

jjhursey:
PR'ed to v3.1 series in PR #1554

artpol84:
@karasevb we need to double-check that removal of the in-use check doesn't break the original design.

Issue #1515, opened 2019-10-31
"make check" Cross Version Failures


jjhursey:
The `IBM CI (Cross-version: make check)` test is failing now that it is doing more serious checking. There have been a number of fixes to the unit tests, and the codebase to address the issue. We are not quite done yet. I'm creating this issue to give us a central point to coordinate our activities.

Here is the same output:
```
======================================================================
Server : master -> Client:   v3.1
-----> : Run simptest Make Check
Make Check: /home/pmixer/scratch/pmix-v3.1/test/pmix_client
Server : /home/pmixer/scratch/pmix-master/test
Command: cd /home/pmixer/scratch/pmix-master/test ; timeout --preserve-status -k 35 30 ./pmix_test -n 4 --ns-dist 3:1 --fen
ce "[db | 0:0-2;1:0]" -e /home/pmixer/scratch/pmix-v3.1/test/pmix_client
Status : 0
-----> : Test output (Not shown)
...
-----> : Run simptest Make Check
Make Check: /home/pmixer/scratch/pmix-v3.1/test/pmix_client
Server : /home/pmixer/scratch/pmix-master/test
Command: cd /home/pmixer/scratch/pmix-master/test ; timeout --preserve-status -k 35 30 ./pmix_test -n 5 --test-resolve-peers --ns-dist "1:2:2" -e /home/pmixer/scratch/pmix-v3.1/test/pmix_client
Status : 143 ***FAILED***
-----> : Test output
[93bb8b136d4a:32192] PMIX ERROR: NOT-FOUND in file gds_hash.c at line 2995
==32192== ERROR [test_resolve_peers.c:27:resolve_nspace]: smoky_nspace-1:1: Resolve peers test failed: rc = -46
[93bb8b136d4a:32190] PMIX ERROR: NOT-FOUND in file gds_hash.c at line 2995
==32190== ERROR [test_resolve_peers.c:27:resolve_nspace]: smoky_nspace-0:0: Resolve peers test failed: rc = -46
[93bb8b136d4a:32191] PMIX ERROR: NOT-FOUND in file gds_hash.c at line 2995
==32191== ERROR [test_resolve_peers.c:27:resolve_nspace]: smoky_nspace-1:0: Resolve peers test failed: rc = -46
==32187== [1572442800.611971] ERROR [cli_stages.c:208:errhandler]:  PMIX server event handler with status = -25
[...]

PR #1552, opened 2019-11-22
Fixes multiserver test mode

karasevb:
- Fixes `nodeid` condition while finding nodeinfo item (see description in commit message)
- Fixes multiserver test mode

### DIFF
--- a/test/test_server.c
+++ b/test/test_server.c
@@ -22,6 +22,7 @@
 
 #include "pmix_server.h"
 #include "src/include/pmix_globals.h"
+#include "src/util/error.h"
 
 #include "test_server.h"
 #include "test_common.h"
@@ -138,10 +139,11 @@ static void set_namespace(int local_size, int univ_size,
     size_t ninfo;
     pmix_info_t *info;
     ninfo = 8;
-    char *regex, *ppn;
+    char *regex, *ppn, *tmp;
     char *ranks = NULL, **nodes = NULL;
     char **rks=NULL;
     int i;
+    int rc;
 
     PMIX_INFO_CREATE(info, ninfo);
     pmix_strncpy(info[0].key, PMIX_UNIV_SIZE, PMIX_MAX_KEYLEN);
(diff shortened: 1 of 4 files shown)


karasevb:
The cross-version issue should be fixed with PR #1553

test/test_server.c
```diff
@@ -212,7 +220,7 @@ static void set_namespace(int local_size, int univ_size,
     }
     PMIx_generate_ppn(ranks, &ppn);
     free(ranks);
-    PMIX_INFO_LOAD(&info[5], PMIX_PROC_MAP, ppn, PMIX_REGEX);
+    PMIX_INFO_LOAD(&info[5], PMIX_PROC_MAP, ppn, PMIX_STRING);
```
rhc54: I believe this needs to be ```PMIX_REGEX``` as ```ppn``` was generated by ```PMIx_generate_ppn``` and may be a binary blob instead of a string.

test/test_server.c
```diff
@@ -165,22 +167,28 @@ static void set_namespace(int local_size, int univ_size,
...
+    if (NULL != nodes) {
+        tmp = pmix_argv_join(nodes, ',');
+        pmix_argv_free(nodes);
+        nodes = NULL;
+        if (PMIX_SUCCESS != (rc = PMIx_generate_regex(tmp, &regex) )) {
+            PMIX_ERROR_LOG(rc);
+            return;
+        }
+        free(tmp);
+        PMIX_INFO_LOAD(&info[4], PMIX_NODE_MAP, regex, PMIX_STRING);
```
rhc54: Ditto here - needs to be a ```PMIX_REGEX```

karasevb: Done.

test/test_server.c
karasevb: Fixed with separate commit.

artpol84:
Given that multi-server is fixed, can we "enable multi-server dimension" as suggested here:
https://github.com/openpmix/openpmix/issues/1515#issuecomment-552019248 ?

rhc54:
I'm not sure this actually fixed it, but we can try and see if it did. I have no objection to turning that back on so long as it can pass.

artpol84:
The target is to make it work. Lets enable and see if it still fails, then @karasevb will investigate further.

rhc54:
Sounds fine - feel free to do so.

jjhursey:
Does this need to be ported to the v3.1 branch?

artpol84:
Yes, see https://github.com/openpmix/openpmix/pull/1565.
Boris is working on this now

PR #1560, opened 2019-12-06
configury: Allow to override build date+host+user

bmwiedemann:
Allow to override build date+host+user with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

Also switch to UTC to be independent of timezone.
Also use ISO 8601 date format to be easier to parse.

Note: This date call is designed to work with different flavors
of date (GNU, BSD and others).
If only GNU (Linux) support is needed, the patch can be simplified.

This PR was done while working on reproducible builds for openSUSE.

For reproducible builds, it would also be a good solution to just drop one or both of these date+user+host to avoid the added complexity.

### DIFF
--- a/config/pmix_functions.m4
+++ b/config/pmix_functions.m4
@@ -94,9 +94,19 @@ EOF
 # Save some stats about this build
 #
 
-PMIX_CONFIGURE_USER="`whoami`"
-PMIX_CONFIGURE_HOST="`(hostname || uname -n) 2> /dev/null | sed 1q`"
-PMIX_CONFIGURE_DATE="`date`"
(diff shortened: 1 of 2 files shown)


rhc54:
Thanks!

jjhursey:
This should be PR'ed to the v3.1 branch. We can get it in the pipeline for v3.1.5

jjhursey:
Ah Ralph beat me to it :) PR #1561

PR #1562, opened 2019-12-07
python: bug fixes for pmix_load_value

dsikich:
Fix for most bugs in pmix_load_value that were uncovered
with the cython tests.

### DIFF
--- a/bindings/python/pmix.pxi
+++ b/bindings/python/pmix.pxi
@@ -703,12 +703,8 @@ cdef int pmix_load_value(pmix_value_t *value, val:dict):
         value[0].data.dval = double_val
     # TODO: need a way to verify usable timevals passed in?
     elif val['val_type'] == PMIX_TIMEVAL:
-        if isinstance(val['value'], tuple):
-            value[0].data.tv.tv_sec  = val['value'][0]
-            value[0].data.tv.tv_usec = val['value'][1]
-        else:
-            value[0].data.tv.tv_sec = val['value']
-            value[0].data.tv.tv_usec = 0
+        value[0].data.tv.tv_sec  = val['value']['sec']
+        value[0].data.tv.tv_usec = val['value']['usec']
     elif val['val_type'] == PMIX_TIME:
         value[0].data.time = val['val_type']
     elif val['val_type'] == PMIX_STATUS:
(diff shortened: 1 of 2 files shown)


dsikich:
@rhc54 I left two TOTO's in the pmix_load_value function as the python bytearray() function doesn't seem to convert to what the PMIX_BYTE_OBJECT would expect. And the type conversion table doesn't say what a PMIX_REGEX is? That one is missing as well.

rhc54:
I'm not sure I understand what you are doing with the bytearray. The byte object expects the "bytes" field to point to an array of bytes, and the "size" field to contain the number of bytes in that array. Your test seems to be declaring a bytearray of size 1 - perhaps technically correct but not realistic. You may be hitting something in python that has issues with your declaration.

The PMIX_REGEX type doesn't cleanly convert to a specific python type - you have to look at the first few bytes of the value to determine how it is represented. I'm not sure what happened to the code in there, but I used to have code that looked at those initial bytes (up to the colon) and made the right translation. For example, if the value starts with ```blob```, then you know the next field (up to the next colon) contains the number of bytes and the rest of the value contains the bytes themselves. If the value starts with ```pmix```, then the value is a string and can be treated as such. Perhaps you mistakenly removed that code?

dsikich:
@rhc54 where was the PMIX_REGEX code? I looked back to commits as far back as July, and didn't see anything in the pmix_load_value function at least: https://github.com/openpmix/openpmix/blob/7e40284db0b97d9ea9884404b1254493be4ae581/bindings/python/pmix.pxi

rhc54:
I honestly don't remember - could be it was in a branch that didn't get committed. However, you can see the way to handle it in the "pack" code in the preg components. For example, the compress component uses this:
```c
    if (0 != strncmp(input, "blob", 4)) {
        return PMIX_ERR_TAKE_NEXT_OPTION;
    }

    /* extract the size */
    slen = strtoul(&input[6], NULL, 10) + 6 + strlen(&input[6]);

    /* ensure the buffer has enough space */
    ptr = pmix_bfrop_buffer_extend(buffer, slen);
    if (NULL == ptr) {
        return PMIX_ERR_NOMEM;
    }

    /* xfer the data */
    memcpy(ptr, input, slen);
    buffer->bytes_used += slen;
    buffer->pack_ptr += slen;
```
You would have to follow the same logic for converting it to Python:

- check the initial 4 characters for the type
- if "native" then treat as a string
- if "blob" then get the size, malloc the space, and move the data portion of the regex into the bytearray.

Should be pretty straightforward

dsikich:
@rhc54 so this is similar to generate_regex in pmix.pyx?

```Python
    def generate_regex(self, hosts):
        cdef char *regex;
        if isinstance(hosts, str):
            pyhosts = hosts.encode('ascii')
        else:
            pyhosts = hosts
        rc = PMIx_generate_regex(pyhosts, &regex)
        if "pmix" == regex[:4].decode("ascii"):
            # remove null characters
            if b'\x00' in regex:
                regex.replace(b'\x00', '')
            ba = bytearray(regex)
        elif "blob" == regex[:4].decode("ascii"):
            sz_str    = len(regex)
            sz_prefix = 5
            # extract length of bytearray
            regex.split(b'\x00')
            len_bytearray = regex[1]
            length = len(len_bytearray) + sz_prefix + sz_str
            ba = bytearray(length)
            pyregex = <bytes> regex[:length]
            index = 0
            while index < length:
                ba[index] = pyregex[index]
                index += 1
        else:
            # last case with no ':' in string
            ba = bytearray(regex)
        return (rc, ba)
```

rhc54:
Aha - that is the code I was remembering. Yep, that is what you want to mimic.

PR #1564, opened 2019-12-11
Simplify tracker initialization

artpol84:
After reviewing #1471 I think that there is no need in `ns_local` anymore.
The whole point behind it was to verify that in case of a wildcard all processes in the namespace are located on the local node (on the per namespace basis).

What the code does now - it is filtering all the local ranks from the fence signature, and then this value is compared to the total process count.
I agree this is a cleaner way, so let's get rid of `ns_local`.

### DIFF
--- a/src/server/pmix_server_ops.c
+++ b/src/server/pmix_server_ops.c
@@ -390,7 +390,6 @@ static pmix_server_trkr_t* new_tracker(char *id, pmix_proc_t *procs,
     pmix_server_trkr_t *trk;
     size_t i;
     bool all_def;
-    pmix_rank_t ns_local = 0;
     pmix_namespace_t *nptr, *ns;
     pmix_rank_info_t *info;
     pmix_nspace_caddy_t *nm;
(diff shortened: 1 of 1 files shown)


artpol84:
@awlauria @rhc54

artpol84:
```
Run ***FAILED***: master -> v3.1  [-s 2 -n 2 --job-fence -e ]
Run ***FAILED***: master -> v3.1  [-s 2 -n 2 --job-fence -c -e ]
Run ***FAILED***: v3.1 -> v3.1  [-s 2 -n 2 --job-fence -e ]
Run ***FAILED***: v3.1 -> v3.1  [-s 2 -n 2 --job-fence -c -e ]
Run ***FAILED***: v3.1 -> v2.2  [-s 2 -n 2 --job-fence -e ]
Run ***FAILED***: v3.1 -> simplify_tracker  [-s 2 -n 2 --job-fence -e ]
Run ***FAILED***: simplify_tracker -> v3.1  [-s 2 -n 2 --job-fence -e ]
Run ***FAILED***: simplify_tracker -> v3.1  [-s 2 -n 2 --job-fence -c -e ]
```
Looks like https://github.com/openpmix/openpmix/pull/1552 needs to be ported to v3.1. Will try now.

awlauria:
Looks like good cleanup to me.

jjhursey:
Does this need to be PR'ed to v3.1?

artpol84:
@jjhursey no, as far as I recall, my initial fix only went to the master

PR #1568, opened 2019-12-13
Cleanup example use of pmix_value_t

### LINKED ISSUE #1133: Question about use of pmix_value_t

## Background information

Question about use of pmix_value_t

github master @ 13baa77fe2ac3a1cd70f05729875a4a6e1e85de4

git-clone

* Operating system/version: CentOS 7
* Computer hardware: x86_64
* Network type: IB

-----------------------------

## Details of the problem

This is actually a question.  In e.g. `examples/client.c`, there is code like this

```C
     95 int main(int argc, char **argv)
     96 {
     97     int rc;
     98     pmix_value_t value;
     99     pmix_value_t *val = &value;
```

Is line 99 pointing to value in line 98 necessary, or could I just use an uninitialized pointer in line 99?  A few tests have worked, but this could easily "just appear to work" accidentally.

jjhursey:
* Fixes #1133
 * Remove `pmix_value_t *val = &value` where not meaningful.
 * Cleanup a few places where `val` was not released, and a couple
   of other minor memory leaks.

### DIFF
--- a/examples/alloc.c
+++ b/examples/alloc.c
@@ -15,6 +15,7 @@
  * Copyright (c) 2011      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2013-2019 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Mellanox Technologies, Inc.  All rights reserved.
+ * Copyright (c) 2019      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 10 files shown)


jjhursey:
The `make check` failure is expected (outside the scope of this PR). However, the normal `Cross-version` failure I did not expect. This PR shouldn't impact that, but I was not aware that it was failing as well.

rhc54:
Odd - no, that test was passing previously.

rhc54:
Looks like something borked master and v3.1 for these tests - they are even failing internally in those branches.

rhc54:
bot:ibm:retest

PR #1569, opened 2019-12-14
python: add missing data conversions for types

dsikich:
- add missing data conversion function for PMIX_REGEX
    - add tests for pmix_value_unload on all type values


dsikich:
@rhc54 the only thing missing from pmix load/unload now should be a pmix_unload_darray function, which I plan to write next.

Issue #1548, opened 2019-11-15
Dstore leaking 8MB of memory when a namespace exits while another namespace is running


dsolt:
Thank you for taking the time to submit an issue!

### What version of the PMIx Reference Library are you using? (e.g., v1.0, v2.1, git master @ hash, etc.)

3.1.3 (e6837057034e7f22910d66ebdc590389b874f4f2)

### Describe how PMIx was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

git clone, built both debug or non-debug

### Please describe the system on which you are running

* Operating system/version:
* Computer hardware:
* Network type:

-----------------------------

## Details of the problem

If we submit a job that registers a namespace:

jsrun --np 1 /hello_world_sleep.x 10 (an MPI Hello_world program that sleeps for 10 seconds before calling MPI_Finalize())
sleep 5
jsrun --np 1 /hello_world_sleep.x 10 

Wait for both job steps to exit.
Notice the total resident memory used by the PMIx server will have increased by 8MB.
This can be repeated with continual memory use increase.

THe problem appears to be coming from here:

```
-    /* A lot of nspaces may be using same session info
-     * session record can only be deleted once all references are gone */
-    if (!in_use) {
```

in src/mca/common/dstore/dstore_base.c

This checks to see if other nspaces are running:

```
for (map_idx = 0; map_idx < size; map_idx++){
       if (ns_map[map_idx].in_use &&
                       (ns_map[map_idx].data.tbl_idx == ns_map_data->tbl_idx)) {
           if (0 == strcmp(ns_map[map_idx].data.name, nspace)) {
            _esh_session_map_clean(ds_ctx, &ns_map[map_idx]);
               continue;
          }
          in_use++;
     }
}
````

I don't believe we can do this.  Once a namespace is deregistered, I think we have to deregister it.  It would be great of other namespaces could query information about namespaces that have exited, but its just not practical to keep the storage of every namespace, possibly indefinitely since you can have a job step queue with thousands of jobs that will prevent memory from being cleaned up.   Even if we fixed it so that once all namespaces were gone, we cleaned up all the namespaces at that point, that would not be sufficient.  A queue of thousands of jobs steps would eat of GB of memory since it would not reach the point where no namespaces are running until the queue was empty.

jjhursey:
I believe this was resolved by the PRs below (please reopen if there is more work to do)
 - `master` PR #1549 
 - `v3.1` PR #1554

PR #1573, opened 2019-12-18
Fix/master/dstore session leak

artpol84:
Ref: #1548 
Alternative to: #1549

### DIFF
--- a/src/mca/common/dstore/dstore_base.c
+++ b/src/mca/common/dstore/dstore_base.c
@@ -823,6 +823,8 @@ static ns_track_elem_t *_get_track_elem_for_namespace(pmix_common_dstore_ctx_t *
 {
     ns_track_elem_t *new_elem = NULL;
     size_t size = pmix_value_array_get_size(ds_ctx->ns_track_array);
+    ns_track_elem_t *ns_trk;
+    size_t i;
 
     PMIX_OUTPUT_VERBOSE((10, pmix_gds_base_framework.framework_output,
                          "%s:%d:%s: nspace %s",
@@ -838,16 +840,27 @@ static ns_track_elem_t *_get_track_elem_for_namespace(pmix_common_dstore_ctx_t *
         return pmix_value_array_get_item(ds_ctx->ns_track_array, ns_map->track_idx);
     }
 
+    /* Try to find an empty tracker structure */
+    ns_trk = PMIX_VALUE_ARRAY_GET_BASE(ds_ctx->ns_track_array, ns_track_elem_t);
+    for (i = 0; i < size; i++) {
+        ns_track_elem_t *trk = ns_trk + i;
+        if (!trk->in_use) {
+            new_elem = trk;
+        }
+    }
+    /* If we failed - allocate a new tracker */
+    if (NULL == new_elem) {
+        if (NULL == (new_elem = pmix_value_array_get_item(ds_ctx->ns_track_array, size))) {
+            return NULL;
+        }
+    }
+
     /* create shared memory regions for this namespace and store its info locally
      * to operate with address and detach/unlink afterwards. */
-    if (NULL == (new_elem = pmix_value_array_get_item(ds_ctx->ns_track_array, size))) {
-        return NULL;
-    }
     PMIX_CONSTRUCT(new_elem, ns_track_elem_t);
     pmix_strncpy(new_elem->ns_map.name, ns_map->name, sizeof(new_elem->ns_map.name)-1);
     /* save latest track idx to info of nspace */
     ns_map->track_idx = size;
-
     return new_elem;
 }
 
@@ -2468,6 +2481,20 @@ PMIX_EXPORT pmix_status_t pmix_common_dstor_del_nspace(pmix_common_dstore_ctx_t 
         if (ns_map[map_idx].in_use &&
                         (ns_map[map_idx].data.tbl_idx == ns_map_data->tbl_idx)) {
             if (0 == strcmp(ns_map[map_idx].data.name, nspace)) {
+                /* Unmap corresponding memory regions and stop tracking this namespace */
+                size_t nst_size = pmix_value_array_get_size(ds_ctx->ns_track_array);
+                if (nst_size && (dstor_track_idx >= 0)) {
+                    if((dstor_track_idx + 1) > (int)nst_size) {
+                        rc = PMIX_ERR_VALUE_OUT_OF_BOUNDS;
+                        PMIX_ERROR_LOG(rc);
+                        goto exit;
+                    }
+                    trk = pmix_value_array_get_item(ds_ctx->ns_track_array, dstor_track_idx);
+                    if (true == trk->in_use) {
+                        PMIX_DESTRUCT(trk);
+                    }
+                }
+                /* Cleanup the mapping structure */
                 _esh_session_map_clean(ds_ctx, &ns_map[map_idx]);
                 continue;
             } else {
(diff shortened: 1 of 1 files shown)


artpol84:
@rhc54 @jjhursey 
We had a debugging session with @dsolt today and this is what came out as an alternative fix to the issue.
@dsolt is testing the PR in his environment now and will update here once done.

rhc54:
Looking at this PR, I think there is a misunderstanding that needs to be addressed - might just be a case of misnaming something, or perhaps a problem in the code. Let's quickly review the available information and how the PMIx server library gets it.

There are five classes of information in the system:
- session-level info that describes the user's overall allocation. These values have nothing to do with a particular namespace/job, but might include the number of nodes allocated, slots allocated on each node, etc. Passing of session-level info is done thru the corresponding info array and _must_ include the session ID.

- namespace/job-level info that includes (for example) the nodes this job is using (which might be a subset of the allocation), number of procs in the job, etc. Passed via the corresponding info array to separate it from any other info, or can simply be passed as individual values.

- app-level info that includes the actual executable to be run, number of procs in that app (which would be the same as the number of procs in the job if only one app is in it), starting rank within the app (which won't always be zero if there are multiple apps in the job), etc. Passed via the corresponding info array and _must_ include the app ID it there are multiple apps in the namespace.

- proc-level info such as the rank (global, local, node), binding locality, etc. Passed via the corresponding info array and _must_ include the global rank.

- node-level info that includes the number of procs on that node, number of namespaces that have procs executing on the node, etc. This is done via the corresponding info array and _must_ include either the hostname, the numerical host ID, or both.

We currently only have one method by which the host RM can pass any of this information down to the PMIx server library - and that is ```PMIx_server_register_nspace```. True to its name, this API is called whenever a new namespace/job is being registered with the server library. The info passed will mostly consist of job, app, and proc-level info - however, there are info keys by which the host can include session and node-level info at the same time. Any such info included in the registration should have some relation to the namespace being registered. If someone registers several namespaces and includes session info in each registration, then that session info either updates or extends the existing session info, assuming matching session IDs. Same for node info.

It is certainly true that session-level info may support multiple namespaces. If provided by the host, session-level info needs to be stored separately from the namespace info so that it can be deleted once all namespaces within that session have been deregistered. We don't currently have a method for the host to "deregister" session-level info.

Namespace/job-level info and the accompanying proc-level info must be deleted immediately once ```deregister_nspace``` is called. It is the responsibility of the RM to determine when the namespace is no longer viable (i.e., all procs have terminated). PMIx must not interfere with that decision.

We haven't really dealt with the question of longevity of node-level info. I believe the assumption was that it would be deleted along with the session-level info since the nodes would be associated with a session. However, that may not be completely accurate for all scenarios and likely requires a little more thought. Might need a "deregister" function for node-level info.

My concern here is that this PR seems to imply that "session" information might actually be getting confused with "namespace" information. We do need to be careful here or the RM is going to consume a lot of memory in the PMIx server library. The rules are rather simple:

- If the RM has determined that a namespace is no longer required and therefore calls ```deregister_nspace```, the namespace-level info (which includes the associated proc-level info) _must_ be deleted.

- If session-level info was provided during some namespace registration, then that info must also be deleted once the last namespace associated with that session ID has been ```deregistered```. 

- For now, I'd suggest deleting the node-level info when all sessions which had processes on those nodes have been deleted.

Hope that helps to clarify things

artpol84:
@rhc54, you are right. As we agreed on the call the dostor's understanding of the session differs from the one in the PMIx standard. The reason is that sessions in the dstore were introduced more than 3 years ago (1a633af4eec6a9471b0b497282c882767176a6d0) and since then a lot of clarifications were made to the semantics of PMIx.

Back then, the session only meant a dstore-internal notion of namespaces that should be cross-accessible by processes in them. dstore session doesn't represent any information (i.e. it doesn't store PMIx session data). The motivation behind the dstore session was my understanding of the intention to use a single PMIx server as part of the persistent RM daemon that serves many allocations and possibly multiple users simultaneously. I'm pretty sure there were discussions like that, or it was my misunderstanding. As we talked on the call, this is not considered as a use case these days, however, cross-access between namepaces is relevant still.

The dstore relies on the "job user ID" (https://github.com/openpmix/openpmix/commit/1a633af4eec6a9471b0b497282c882767176a6d0#diff-ef98725cc50c8a568d8e5622d505e776R1021) to determine if this is an existing session or a new one. `jobuid` might be passed explicitly by RM through the info keys, otherwise, servers UID will be used.

@dsolt, based on this observation, I believe that if you have long-running jobs, the session will remain as long as those jobs are active. I'd suggest having another debugging session to check this scenario.

On the other hand, the original fix (#1549) was removing the session even if it was referenced by other namespaces which IMO is data corruption.

* The short-term solution would be to make sure that all namespace-related items are removed and unmapped (I suspect they currently remain in case for cross-access).
* As the long-term solution, I've opened an issue suggesting modernization of the dstore that includes addressing this issue. Ref: #1574

jjhursey:
Per teleconf discussion, I'm putting a WIP flag on this:
 * Dave is still working on verifying this patch
   * Dave/Artem will have another debugging session to investigate the fix further.
 * Ralph's clarification on the life of the data should be captured in the standard document if it is not already. That really helped me, and highlighted a couple of API calls that might need to be added for the session/node level information.
 * Discussion on the meaning of connectedness that we need to continue once folks are back in Jan. There was some uncertainty on the call about expected behavior which needs to be investigated (both in code and in the standard doc) for clarification. Note that Section 6.3 of the standard has information on this that may answer the questions below.
 * Below I'll try to capture the scenarios from the call (feel free to expand/clarify in the comments).
   * **Scenario 1**: Two independent jobs
     * Job A and Job B both have their own namespaces.
     * Can Job A access information about Job B without being connected?
       - If we think of Job A as a debugger/monitor for Job B
     * What happens if Job A asks for information about Job B after it has terminated
       - It should return "not found", right?
     * What if Job A is on a different node than Job B?
       - Should PMIx ask the RM to deliver the information (via `dmodex`?)?
       - Then is the RM is responsible for calling `deregister_nspace` for Job B on both nodes so dstore knows to clean up the information?
     * What if Job A is on the same node as Job B?
        - Does the server get the same upcall even if dstore can find the information locall?
   * **Scenario 2**: Two jobs that are connected
     - While connected they share namespace information.
     - The RM is responsible for tracking connectedness so PMIx must trust the RM to only call `deregister_nspace` once the namespace is no longer connected to any other job in the PMIx session.
     - If one job crashes then the RM should not `deregister_nspace` for that job until the other connected jobs have completed. Right?
   * **Scenario 3**: Two jobs that were once connected, but are no longer
     - Jobs were connected for a time, but have since disconnected.
     - At this point is there any responsibly for keeping Job B's data after it completes just in case Job A might want it in the future? I'm thinking no since this was a scenario 2 and is now a scenario 1.

jjhursey:
Is this PR ready to have the WIP label removed? I'd like to talk about it some more on the Thursday teleconf before we merge to better understand the adjustment to the deletion model.

dsolt:
I did find an additional issue not found in my testing over the break.  My testing referenced above just was looking for leaks, but did not do any real puts/gets to verify shared segments could be found.  When using real codes (MPI in this case) the 2nd code would fail because it could not find the shared memory index in an array.  This was easily diagnosed as a pre-existing bug that was never hit in the past because entries were never deleted from the array in the past.  Now that they are, we find the index for new entries was set to be 1 past the previous array size instead of using the actual location of the first free entry in the array.  This issue was easily fixed and we verified code is working and not leaking now.

jjhursey:
Please PR to the v3.1 branch.

Issue #1577, opened 2020-01-01
pthread_mutexattr_setpshared() is not available on all platforms


geowiwi:
## Background information

- Do not try to set pthread_rwlockattr_setpshared with PTHREAD_PROCESS_SHARED when it is not available.
- Do not try to use pthread_mutexattr_setpshared() when it is not available.

### What version of the PMIx Reference Library are you using? (e.g., v1.0, v2.1, git master @ hash, etc.)

PMIx 3.1.4, embedded with Open MPI 4.0.2

### Describe how PMIx was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

OpenBSD 6.6-current Open MPI 4.0.2 distribution package with embedded PMIx 3.1.4

### Please describe the system on which you are running

* Operating system/version: OpenBSD 6.6-current
* Computer hardware: amd64
* Network type: Ethernet

-----------------------------

```shell
----------------------------------------------------------------------------------------------
openbsd-1$ mpirun -np 2 -hostfile hlist
/home/martin/bin/test plist

orted:/usr/local/lib/pmix/mca_gds_ds21.so: undefined symbol
'pthread_mutexattr_setpshared'
ld.so: orted: lazy binding failed!
Killed
--------------------------------------------------------------------------
ORTE has lost communication with a remote daemon.

  HNP daemon   : [[62306,0],0] on node openbsd-1
  Remote daemon: [[62306,0],1] on node openbsd-2

This is usually due to either a failure of the TCP network
connection to the node, or possibly an internal failure of
the daemon itself. We cannot recover from this failure, and
therefore will terminate the job.
```

PMIx sets HAVE_PTHREAD_SHARED/HAVE_PTHREAD_MUTEXATTR_SETPSHARED accordingly but does not use the appropriate defines.

A possible fix is in the OpenBSD-current portstree (unified diff against the OpenMPI sources with embedded PMIx 3.1.4, but should also apply to newer versions):
```shell
$OpenBSD: patch-opal_mca_pmix_pmix3x_pmix_src_mca_gds_ds12_gds_ds12_lock_pthread_c,v 1.1 2019/12/22 15:47:36 martin Exp $

Index: opal/mca/pmix/pmix3x/pmix/src/mca/gds/ds12/gds_ds12_lock_pthread.c
--- opal/mca/pmix/pmix3x/pmix/src/mca/gds/ds12/gds_ds12_lock_pthread.c.orig
[...]

PR #1580, opened 2020-01-04
Correct the gds/dstore configure logic

rhc54:
After digging deeper into two reports of error log outputs from dstore
components that were built but could not run, it turned out that there
were a few compensating errors in the dstore configure code. First, the
use of AC_EGREP_HEADER was incorrect as it always failed - that function
looks for an exact match of the output from "egrep" and the provided
pattern. Instead, egrep was outputting the entire line that contained
the provided string, resulting in it always failing.

Someone clearly realized things were failing as the result of that test
was never being used. This led to error logs being emitted by ds21 when
attempting to run on systems that actually didn't support that specific
option.

What needed to happen was to correctly test for the existence of the
option and the corresponding function, and then to disable the build of
the ds21 component if either were not found. A similar test was required
for the ds21 component, albeit for a different function.

The revised logic breaks things apart a little more so it is clearer
what precisely is and isn't available, and adds configure.m4 code for
each of the dsN components. The ds12 component builds if either the
flock is available OR rwlock_setpshared and the PROCESS_SHARED attribute
are available. The ds21 component builds if mutexattr_setpshared and
PROCESS_SHARED are available.

This corrects the change originally committed in PR #1578.

Refs #1577
Refs #1578
Refs https://www.mail-archive.com/devel@lists.open-mpi.org/msg21000.html.

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -17,7 +17,7 @@ dnl Copyright (c) 2009      Los Alamos National Security, LLC.  All rights
 dnl                         reserved.
 dnl Copyright (c) 2009-2011 Oak Ridge National Labs.  All rights reserved.
 dnl Copyright (c) 2011-2013 NVIDIA Corporation.  All rights reserved.
-dnl Copyright (c) 2013-2019 Intel, Inc.  All rights reserved.
+dnl Copyright (c) 2013-2020 Intel, Inc.  All rights reserved.
 dnl Copyright (c) 2015-2019 Research Organization for Information Science
 dnl                         and Technology (RIST).  All rights reserved.
 dnl Copyright (c) 2016      Mellanox Technologies, Inc.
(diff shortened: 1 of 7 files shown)


rhc54:
@karasevb @artpol84 I believe the Mellanox tester is actually incorrect. In one configuration, you disable pthread locks, which should (from what I see in the code) cause ds21 to fail as it requires the presence of pthread locking support. The ds12 component, in contrast, has its own dedicated fcntl_lock code.

I believe this may have been overlooked due to the errors in the lock configuration code that allowed ds21 to still build even when --disable-pthlck was provided in the configuration cmd line. Now that the code has been corrected, the error is visible.

If you want ds21 to still operate in that configuration, then I believe you will need to add fcntl_lock code to that component.

rhc54:
bot:mellanox:retest

rhc54:
hmmm...retest doesn't appear to be working. Will test via another PR

jjhursey:
Does this need to be cherry-picked to the v3.1 branch? It seems like it does.

rhc54:
Huh - guess I forgot to do so! I'll post it now

PR #1581, opened 2020-01-06
Enable cycling of init/finalize calls

### LINKED ISSUE #1558: Problem with multiple PMIx_Init / Finalize call sequence
I observe what seems to be a bug in a program, which uses OpenMPI and explicit PMIx calls. I use OpenMPI 4.0.2 with external PMIx 3.1.4. Consider the following pseudo-code:

```
PMIx_Init(...)

[...]

PMIx_Finalize(...)

MPI_Init(...)
```
The above sequence results in OpenMPI failure with the following message:
```
OPAL ERROR: Error in file ext3x_client.c at line 112
*** An error occurred in MPI_Init_thread
*** on a NULL communicator
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
```
The problem disappears if I move the `PMIx_Finalize` after the `MPI_Init` call. 

From the standard I understood that multiple `PMIx_Init` calls are allowed, as long as followed by `PMIx_Finalize`, so it would seem the above is a bug.

Could you please comment? Thanks!

rhc54:
The PMIx library needs to be able to support multiple calls to
PMIx_Init/PMIx_Finalize as an app might cycle thru these calls prior to
entering a PMIx-enabled library. We could set a rule prohibiting calling
init if the library has been completely finalized (i.e., if the number
of calls to finalize equals the number of calls to init, thus driving
the refcount to zero and causing the library to fully close down), but
that seems overly restrictive.

Fixes #1558 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,6 +162,7 @@ test/simple/simpcoord
 test/python/run_sched.sh
 test/python/run_server.sh
 test/simple/pmitest
+test/simple/simpcycle
 
 # coverity
 cov-int
--- a/src/common/pmix_attributes.c
+++ b/src/common/pmix_attributes.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2014-2019 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2014-2020 Intel, Inc.  All rights reserved.
  * Copyright (c) 2016      Mellanox Technologies, Inc.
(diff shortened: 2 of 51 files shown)


ggouaillardet:
@rhc54 I could not add some more commits to this PR, i created #1582 
feel free to squash and/or rewrite my commits.

the second `PMIx_Init()` still fails, but on the server side, in `connection_handler()` from `ptl/tcp` at line 1589, we have `found = false`.

PR #1587, opened 2020-01-09
Correct the gds/dstore configure logic

rhc54:
After digging deeper into two reports of error log outputs from dstore
components that were built but could not run, it turned out that there
were a few compensating errors in the dstore configure code. First, the
use of AC_EGREP_HEADER was incorrect as it always failed - that function
looks for an exact match of the output from "egrep" and the provided
pattern. Instead, egrep was outputting the entire line that contained
the provided string, resulting in it always failing.

Someone clearly realized things were failing as the result of that test
was never being used. This led to error logs being emitted by ds21 when
attempting to run on systems that actually didn't support that specific
option.

What needed to happen was to correctly test for the existence of the
option and the corresponding function, and then to disable the build of
the ds21 component if either were not found. A similar test was required
for the ds21 component, albeit for a different function.

The revised logic breaks things apart a little more so it is clearer
what precisely is and isn't available, and adds configure.m4 code for
each of the dsN components. The ds12 component builds if either the
flock is available OR rwlock_setpshared and the PROCESS_SHARED attribute
are available. The ds21 component builds if mutexattr_setpshared and
PROCESS_SHARED are available.

This corrects the change originally committed in PR #1578.

Refs #1577
Refs https://www.mail-archive.com/devel@lists.open-mpi.org/msg21000.html.

(cherry picked from commit bf2a49eb78d53581764b73e7e3dbeb990b2b2c6b)

### DIFF
--- a/.github/actions/mlnx/entrypoint.sh
+++ b/.github/actions/mlnx/entrypoint.sh
@@ -398,10 +398,6 @@ if [ "$jenkins_test_check" = "yes" ]; then
         echo "----dstore/flock----" >> $run_tap
 
         gds_list="hash ds12,hash"
-        # check the existence of ds21 component
-        if [ "$has_gds_ds21" -gt 0 ]; then
-            gds_list="$gds_list ds21,hash"
-        fi
 
         for gds in $gds_list; do
             echo "Checking with $gds:" >> $run_tap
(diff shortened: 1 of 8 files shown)


jjhursey:
Fixes https://github.com/openpmix/openpmix/pull/1579 code as well (that's the v3.1 version of the "Protect pthread setpshared calls" PR)

PR #1588, opened 2020-01-16
Fix bugs in gds/hash and preg/native

rhc54:
Ensure we capture the number of procs in the nspace tracker when
registering an nspace so we don't lose proc-specific data.

Check the proc/node map types for string vs regex

Ensure the "pmix" regex string retains the NULL terminator


rhc54:
Cross-version is failing for:
```text
master  -->  PR
v3.1       -->  PR
```
This is expected as the bugs are in both master and v3.1 branches, and are being exposed by the update to simpclient in this PR.

PR #1589, opened 2020-01-16
Fix bugs in gds/hash and preg/native

rhc54:
Ensure we capture the number of procs in the nspace tracker when
registering an nspace so we don't lose proc-specific data.

Check the proc/node map types for string vs regex (skipped for this cherry-pick)

Ensure the "pmix" regex string retains the NULL terminator (skipped for this cherry-pick)

(cherry picked from commit 76c9a347a11d811febd60e289cd846f22da30683)


rhc54:
As before, cross-version is failing because v3.1 doesn't include the fix in this PR.

PR #1591, opened 2020-01-16
Fix/v3.1/dstore session leak

artpol84:


artpol84:
Ref: #1548
Cherry-pick: #1573

Issue #1595, opened 2020-01-20
new_tracker always returns a tracker with trk->local set to false


dsolt:
git master, server/pmix_server_ops.c
also seems to be an issue in other branches such as 3.1.xx

dsolt:
the problem is different in 3.1 branch.  trk->local is still not calculated correctly, but for different reasons.  there trk->local can be set to true in one place, but it is only tested on one of the namespaces.  so if there are multiple namespaces, only the last one will be checked.  if it is all local, then the whole tracker is set to true even though only one namespace was local.

rhc54:
@dsolt I am truly puzzled by your title on this issue. The "new_tracker" function is supposed to return a new tracker, and we always initialize trackers to being "non-local" (i.e., trk->local is initialized to false). So is the true issue here that the local flag isn't being correctly computed later on?

If so, could you please change the issue title to correctly reflect the actual problem? This one was confusing.

rhc54:
Hmmm...I think I may see what you meant. The logic in "new_tracker" sets the "local" flag to false if any participating proc isn't local. However, that logic requires that the "local" flag be initialized to true. I can correct that one easily enough.

rhc54:
@dsolt Please see if the referenced PRs fixed what you were observing.

jjhursey:
@dsolt does PR #1620 fix the issue for you on the v3.1 series? I'd like to get this fix in for the upcoming v3.1.5.

jjhursey:
Is this ready to close?

jjhursey:
Per teleconf this issue is ok to close now.

PR #1598, opened 2020-01-21
Properly initialize new trackers

rhc54:
Refs #1595

### DIFF
--- a/src/server/pmix_server_ops.c
+++ b/src/server/pmix_server_ops.c
@@ -1,6 +1,6 @@
 /* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
 /*
- * Copyright (c) 2014-2019 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2014-2020 Intel, Inc.  All rights reserved.
  * Copyright (c) 2014-2018 Research Organization for Information Science
  *                         and Technology (RIST).  All rights reserved.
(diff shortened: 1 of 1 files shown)


src/server/pmix_server_ops.c
```diff
@@ -454,26 +450,21 @@ static pmix_server_trkr_t* new_tracker(char *id, pmix_proc_t *procs,
...
-        ns_local = 0;
         PMIX_LIST_FOREACH(info, &nptr->ranks, pmix_rank_info_t) {
             if (procs[i].rank == info->pname.rank ||
                 PMIX_RANK_WILDCARD == procs[i].rank) {
                     pmix_output_verbose(5, pmix_server_globals.base_output,
                                         "adding local proc %s.%d to tracker",
                                         info->pname.nspace, info->pname.rank);
                 /* track the count */
-                ns_local++;
+                trk->nlocal++;
```
dsolt: I don't think this will correctly handle the case where PMIX_RANK_WILDCARD is used for the procs[i].rank, but only some of those ranks are local.  In that case, it won't trigger line 437 and it willl go into this 'if' code.   Do we need to do a full count as in:

https://github.com/openpmix/openpmix/pull/1602
?

rhc54: I don't think either one is correct. Let me ponder this a bit more.

rhc54: Please see #1603  - I believe it resolves the problems and hopefully all the comments will help you to follow the code. Once we converge on it, I'll update this one.

PR #1602, opened 2020-01-22
Fix issue 1598

dsolt:
Set the correct value of local for new trackers.

### DIFF
--- a/src/server/pmix_server_ops.c
+++ b/src/server/pmix_server_ops.c
@@ -7,7 +7,7 @@
  *                         All rights reserved.
  * Copyright (c) 2016-2019 Mellanox Technologies, Inc.
  *                         All rights reserved.
- * Copyright (c) 2016-2019 IBM Corporation.  All rights reserved.
+ * Copyright (c) 2016-2020 IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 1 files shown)


dsolt:
WIP, will sign off if we want to use this.

artpol84:
Interestingly, I recall that recently there was a fix in that area. Was it incorrect?

rhc54:
@artpol84 No, the fix was correct - this was a proposal from @dsolt that we noted was incorrect, as was the one I had originally proposed. This PR should be closed.

artpol84:
But why the code seems different that at the time of https://github.com/openpmix/openpmix/pull/1564?

artpol84:
In particular, the following is missing:

```C
    if (trk->nlocal == nptr->nprocs) {
        trk->local = true;
    }
```

artpol84:
It was there at the time:
https://github.com/openpmix/openpmix/pull/1564/files#diff-cf506bd64a1184a18c5ad19dfdb8913dL506

rhc54:
Because @dsolt was looking at the original file. My original proposal was done in parallel. He asked which one was correct, and the conclusion was "neither". So I started over again - took a few days to work thru it all and straighten it out, but it now passes @jjhursey's multi-node CI for PRRTE. We had been routinely failing it before this go-around, plus we discovered problems when working on the dmodex issue (which I am still working to finalize).

PR #1603, opened 2020-01-23
Re-address the collective tracker problem

rhc54:
There are several problems needing to be addressed. First, there exists a very small race condition window between registration of nspaces and/or clients and receipt of a collective operation request from a local client. We were trying to do too much in the register_client area, resulting in some potential area to miscount the number of local participants. Sadly, we weren't looking for this race condition at all in the register_nspace code. It could be that this race shouldn't happen, but I decided to protect against it just to be safe.

Second, the code in the "new_tracker" area wasn't correct. Cleanly separating out the various possible scenarios helps clarify the logic and avoid some of the problems.

I added quite a few comments to the code to help people understand the flow.

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -1169,10 +1169,12 @@ static inline void* pmix_calloc(size_t n, size_t m)
     }while(0)
 
 /* define a convenience macro for loading nspaces */
-#define PMIX_LOAD_NSPACE(a, b)                          \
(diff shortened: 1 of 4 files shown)


rhc54:
Once this passes CI (I removed some cruft that inadvertently got in), I'm going to take this for now (we can always revise it later) as the master is broken right now for multi-node fence.

Issue #1586, opened 2020-01-09
Obtaining asynchronous keys


rhc54:
Ref #1427 
Ref  https://github.com/open-mpi/ompi/issues/6982

A user is trying to use PMIx in a way we hadn't anticipated, but should probably be supported. Here is a summary of the problem:

* procs do a global exchange of some basic set of key-value pairs

* some process (call it proc-A) requests a specific key-value pair from target process (proc-B) that is not included in that global exchange.

* The host daemon (call it host-1) does the correct thing and queries the remote host (host-2) of proc-B for the information

* host-2 calls PMIx_server_dmodex_request to request info for proc-B. Note, however, that the API doesn't allow specification of the key that is being sought - it only requests ALL data posted by proc-B prior to the request

* the PMIx server library sees that proc-B has posted data and returns that blob

* host-2 returns the blob to host-1, which delivers it to its PMIx server library. Note that the blob at this point only includes the same data that was in the global exchange because proc-B hasn't posted the new data yet! The server library notifies the client library in proc-A, which checks for the requested key and returns NOT_FOUND

* proc-B then posts a new key, which happens to be the one proc-A is looking for - but it is too late.

What the user would like to have happen is for host-2 to wait until proc-B posts the desired key, and then respond to the direct modex request. There are several possible solutions that immediately came to mind. I will post each as a separate comment below so people can "emoji-vote" or directly comment on them separately.

Just as a reminder:
* Hooray or Rocket: I support this so strongly that I want to be an advocate for it
* Heart: I think this is an ideal solution
* Thumbs up: I’d be happy with this solution
* Confused: I’d rather we not do this, but I can tolerate it
* Thumbs down: I’d be actively unhappy, and may even consider other technologies instead
* Eyes: You must be kidding, let's not do this one

rhc54:
Solution 1: Don't support this behavior

Would require clarification in the Standard, but no further work in the library

rhc54:
Solution 2: do the full blob exchange. If the PMIx client fails to find the requested piece of info in that response, then have it follow-up with a "one piece of data" request. When host-2 receives that request, it knows to pursue it with a PMIx_Get call instead of PMIx_server_dmodex_request.

Pretty easy to implement, but would result in slow behavior. Still, the Standard doesn't guarantee anything about speed-of-response!

No changes required to the Standard. Would require changes to OpenPMIx and PRRTE.

rhc54:
[...]

PR #1607, opened 2020-01-26
Allow user to request cache refresh

rhc54:
Define a new PMIX_GET_REFRESH_CACHE attribute by which a user can direct
PMIx to perform a dmodex operation, thereby obtaining a new copy of the
remote data. Anything that has been added or changed will be included.
Note that there is no current mechanism for _deleting_ information once
a process has put/commit it.

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -345,6 +345,14 @@ typedef uint32_t pmix_rank_t;
 #define PMIX_PROC_STATE_STATUS              "pmix.proc.state"       // (pmix_proc_state_t) process state
 #define PMIX_NOTIFY_LAUNCH                  "pmix.note.lnch"        // (bool) notify the requestor upon launch of the child job and return
                                                                     //        its namespace in the event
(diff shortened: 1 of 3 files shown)


jjhursey:
This should impact the discussion in Issue #1586 .

I'd like to talk through the put/commit/fence/get protocol a bit just to get it clear in my mind when dmodex is triggered and when it is not (I'll put it on the teleconf agenda for this week). In particular, I'm thinking that we need an opposing attribute to this in which the PMIx_Get is purely local - do not trigger the dmodex if the data is not present locally. But I'd like to better understand the semantic before looking into that.

rhc54:
> I'm thinking that we need an opposing attribute to this in which the PMIx_Get is purely local - do not trigger the dmodex if the data is not present locally.

Already have that attribute:
```txt
#define PMIX_IMMEDIATE                      "pmix.immediate"        // (bool) specified operation should immediately return an error from the PMIx
                                                                    //        server if requested data cannot be found - do not request it from
                                                                    //        the host RM
```

jjhursey:
I forgot about that one. Thanks!

PR #1608, opened 2020-01-28
Fix eviction time for cached events

jjhursey:
* User facing paramter is in sec, but `pmix_hotel_init` is in usec.

### DIFF
--- a/src/class/pmix_hotel.c
+++ b/src/class/pmix_hotel.c
@@ -3,6 +3,7 @@
  * Copyright (c) 2012-2016 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2012      Los Alamos National Security, LLC. All rights reserved
  * Copyright (c) 2015-2018 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2020      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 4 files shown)


jjhursey:
See analysis https://github.com/openpmix/prrte/pull/265#issuecomment-579060670 for details on the impact of this bug fix.

rhc54:
@jjhursey This looks fine, but I'm wondering - why not just change the timeout to be in seconds?? I cannot imagine every wanting to evict something from the hotel in the order of microseconds. We don't have any operation that can possibly complete in such time scales.

jjhursey:
I agree. That was my alternative approach. Let me post a PR with that fix then we can close this one if you want. Since PRRTE has the hotel data structure we will probably want the fix to be ported over there too (I can do that once we fix it here).

rhc54:
Kewl. I'll handle the PRRTE side as I've finally tracked down the problem over there. If a pending dmodex operation is in the hotel and gets evicted, we check it back in if its checkout time isn't complete (note you cannot actually set an eviction time for each tenant as it is set for the entire hotel, so you have to track the actual time yourself). This is fine for the daemon that is waiting for a particular key before sending the payload, but...

For the daemon that issued the request and is waiting for a response, checking the request back into the hotel changes its room number. Thus, when the remote daemon finally does respond, the room number it returns is stale - and we cannot match the returned payload to a request, leaving the client to hang.

I need to add (or modify) the hotel class to allow me to "recheck" an evicted occupant back into the old room. Might as well change the timeout logic at the same time.

jjhursey:
Thanks.
PR #1609 replaces this PR.

jjhursey:
PR #1609 merged - closing this variant.

PR #1609, opened 2020-01-28
Adjust hotel timeout to be in whole seconds

jjhursey:


jjhursey:
* Per discussion [here](https://github.com/openpmix/openpmix/pull/1608#issuecomment-579074072) we do not anticipate the need for usec level specification of hotel timeouts, especially for events. So just make it take whole seconds.
 * This PR replaces PR #1608
 * Original problem this solves is described in https://github.com/openpmix/prrte/pull/265#issuecomment-579060670

PR #1611, opened 2020-01-29
Properly support dmodex requests

rhc54:
A dmodex operation is triggered whenever the key being requested is not
found in the local cache. Include the specified key in the request
passed from the client to the local server. Have the server check for
that key in its own storage - if it isn't present, then pass the request
to the host thru the dmodex interface with the key being identified by
including the PMIX_REQUIRED_KEY attribute in the info array.

The rest of this fix is in the PRRTE PR that cannot be filed until this
one has been committed due to the new attribute definition.

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -61,7 +61,7 @@
 
 #include "pmix_client_ops.h"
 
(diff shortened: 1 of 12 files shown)


rhc54:
bot:ibm:xvercheck:retest

Issue #1605, opened 2020-01-24
PMIX ERROR: ERROR in file dstore_segment.c at line 110 on s390x


junghans:
Thank you for taking the time to submit an issue!

## Background information

https://github.com/hpc/libcircle/issues/39 and https://github.com/open-mpi/ompi/issues/7335

### What version of the PMIx Reference Library are you using? (e.g., v1.0, v2.1, git master @ hash, etc.)

3.1.4

### Describe how PMIx was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

dnf, pmix-3.1.4-2.fc32.s390x

### Please describe the system on which you are running

* Operating system/version: Fedora Rawhide
* Computer hardware:
* Network type:

-----------------------------

## Details of the problem

```
Running suite(s): check_queue
[d8f6aae043a2464bb6c5907f844aeb8c:19415] PMIX ERROR: ERROR in file dstore_segment.c at line 110
[d8f6aae043a2464bb6c5907f844aeb8c:19415] PMIX ERROR: OUT-OF-RESOURCE in file gds_ds21_lock_pthread.c at line 176
[d8f6aae043a2464bb6c5907f844aeb8c:19415] *** Process received signal ***
[d8f6aae043a2464bb6c5907f844aeb8c:19415] Signal: Segmentation fault (11)
[d8f6aae043a2464bb6c5907f844aeb8c:19415] Signal code: Address not mapped (1)
[d8f6aae043a2464bb6c5907f844aeb8c:19415] Failing at address: (nil)
[d8f6aae043a2464bb6c5907f844aeb8c:19415] [ 0] [0x4001bce850]
[d8f6aae043a2464bb6c5907f844aeb8c:19415] [ 1] /usr/lib64/pmix/mca_gds_ds21.so(+0x1ef0)[0x4001396ef0]
[d8f6aae043a2464bb6c5907f844aeb8c:19415] [ 2] /usr/lib64/pmix/mca_gds_ds21.so(+0x246e)[0x400139746e]
[d8f6aae043a2464bb6c5907f844aeb8c:19415] [ 3] /lib64/libmca_common_dstore.so.1(pmix_common_dstor_add_nspace+0x29a)[0x40013aaeda]
[d8f6aae043a2464bb6c5907f844aeb8c:19415] [ 4] /lib64/libpmix.so.2(+0x4e98e)[0x40011ba98e]
[...]

PR #1613, opened 2020-01-30
dstore: Fix cache size calculation

artpol84:
The bug was discovered in #1605. The previous version was returning an
unsigned representation of `-1` which resulted in 0-size mmap argument.
(see #1605 for the details)

Ref: #1618 (master PR)


gpaulsen:
@artpol84 @jjhursey I'm trying to look internally to see if we have access to an `s390x` node.
I'll report back once I've located one.

artpol84:
This one is good to go now

PR #1620, opened 2020-02-05
Re-address the collective tracker problem

rhc54:
There are several problems needing to be addressed. First, there exists a very small race condition window between registration of nspaces and/or clients and receipt of a collective operation request from a local client. We were trying to do too much in the register_client area, resulting in some potential area to miscount the number of local participants. Sadly, we weren't looking for this race condition at all in the register_nspace code. It could be that this race shouldn't happen, but I decided to protect against it just to be safe.

Second, the code in the "new_tracker" area wasn't correct. Cleanly separating out the various possible scenarios helps clarify the logic and avoid some of the problems.

I added quite a few comments to the code to help people understand the flow.

There remain some issues about detecting local vs non-local procs when
setting up the tracker. I've resolved the ones surfaced by multi-node
testing using PRRTE with both simple PMIx clients and the "double-get"
test case. This is a subset of those changes as the broader ones failed
to pass CI - hopefully, this minimum subset does and then I'll try again
with the remainder.

For this subset, we use the fact that we are being told both the number
of local procs and the total number of procs in a namespace upon call to
"register_nspace". If we are in a collective and have not yet received
the "register_nspace" call, then we defer processing of the collective
until we do.

(cherry picked from commit ce74505cc5912b417492f02daf1a9c3d4f4ced7f)
(cherry picked from commit 7fbae1bc078cca18c9ae92745958b7ced2a08242)

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -1018,10 +1018,12 @@ static inline void* pmix_calloc(size_t n, size_t m)
     }while(0)
 
 /* define a convenience macro for loading nspaces */
-#define PMIX_LOAD_NSPACE(a, b)                          \
-    do {                                                \
-        memset((a), 0, PMIX_MAX_NSLEN+1);               \
-        pmix_strncpy((char*)(a), (b), PMIX_MAX_NSLEN);  \
+#define PMIX_LOAD_NSPACE(a, b)                              \
+    do {                                                    \
+        memset((a), 0, PMIX_MAX_NSLEN+1);                   \
+        if (NULL != (b)) {                                  \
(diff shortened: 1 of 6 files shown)


jjhursey:
This should fix Issue #1595 for the v3.1 series, right? @dsolt Do you want to take it for a test drive before we merge?

This PR cherry-picked PR #1603 and PR #1606 from `master`.

dsolt:
ok, I have been testing with the version on master for a week or more now and this change addresses the issue we were seeing.   i ran a quick test on 3.1 with this branch and it also looks good.

PR #1625, opened 2020-02-07
Correct library version triplet

### LINKED ISSUE #1624: bugfix release 3.1.5rc1 bumps major soname from 2 to 3
I note the major soname bump with 3.1.5rc1 from libpmix.so.2 to libpmix.so.3 . You're signalling a significant abi change with that bump -- was that the intent for this bugfix release? It's more work downstream, as an announcement has to be made so that all pmix client packages can be rebuilt.

rhc54:
Fixes #1624 

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -77,7 +77,7 @@ date="Feb 14, 2020"
 # Version numbers are described in the Libtool current:revision:age
 # format.
 
-libpmix_so_version=5:0:2
+libpmix_so_version=4:25:2
 libpmi_so_version=1:1:0
 libpmi2_so_version=1:0:0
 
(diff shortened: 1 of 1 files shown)


jjhursey:
I think what we had before was technically correct:
```
# - If any interfaces have been added, removed, or changed since
#   the last update, increment current, and set revision to 0.
```
We modified the signature of a public interface to fit the specification (`PMIx_Allocation_request` in PR #1437). Before it had an incorrect signature, and now that is fixed. So this is a modification to a public interface and technically would require a change to the "current", which is what we had before.

However, it was not functional in its prior form so therefore the impact should be minimal/none. So that is the justification for changing only the revision number in this release.

Is that a correct understanding of this PR?

rhc54:
It is certainly a gray area. Technically, we should indeed leave it as the prior triplet since a public API did change. However, I think the rationale for going with this proposed triplet also holds water and is the right thing to do.

Issue #1633, opened 2020-02-14
soname of libmca_common_dstore


amckinstry:
I request that the soname of libmca_common_dstore be bumped from so.1 to so.3 to match pmix3

# Background information

This primarily regards Debian/Ubuntu distributions but may happen elsewhere.
We split packages to enable smooth upgrades over versions and applications to continue to use "old" shared libraries.
For pmix, we ship the _library_ packages libpmi1-pmix, libpmi2-pmix, libpmix2, and the 
_development_ packages elibpmi-pmix-dev  libpmix-dev. Its libpmix2 for version 3.1.4 because it has soname so.2. 3.1.5 bumps this to so.3 and so the new package is libpmix3 and libpmix-dev (ignoring pmi packages for the moment).

The basic idea is that the -dev package contains headers, static libs , etc. and libpmix.so files, while the shared library package contains 
libpmix.so.2 -> libpmix.so.2.2.24 symlink
libpmix.so.2.2.24 file

Similarly for libpmix3. 
*libpmix2 and libpmix3 can be co-installed by design*.
After an upgrade (especially a distro upgrade), libpmix2 may no longer exist in Debian, but old copies can remain on a users system if needed by applications. 

Now the problem: pmix breaks this by shipping:
$(LIBDIR)/pmix/*.so submodules
libmca_common_dstore.*

These are currently shipped in the libpmix2 package. Doing so within a libpmix3 package leads to conflicts; moving them into their own package is possible, but undesirable (may be revisited).
Now the subdirectory can be solved by renaming it 'pmix3' (straightforward and tested), but libmca_* remains. While technically it has not been changed, it simplifies matters a great deal if its soversion is bumped and remains identical to libpmix.

### What version of the PMIx Reference Library are you using? 

pmix 3.1.5 rc 1

### Describe how PMIx was installed 

Built from source (I'm the Debian maintainer).

### Please describe the system on which you are running

* Operating system/version:
* Computer hardware:
* Network type:

-----------------------------

## Details of the problem

Please describe, in detail, the problem that you are having, including the behavior you expect to see, the actual behavior that you are seeing, steps to reproduce the problem, etc.  It is most helpful if you can attach a small program that a developer can use to reproduce your problem.

**Note**: If you include verbatim output (or a code block), please use a [GitHub Markdown](https://help.github.com/articles/creating-and-highlighting-code-blocks/) code block like below:
```shell
shell$ prun -np 2 ./hello_world
```

[...]

PR #1635, opened 2020-02-14
Roll to rc4

rhc54:

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,4 @@
-Copyright (c) 2015-2019 Intel, Inc.  All rights reserved.
+Copyright (c) 2015-2020 Intel, Inc.  All rights reserved.
 Copyright (c) 2017-2020 IBM Corporation.  All rights reserved.
 $COPYRIGHT$
 
@@ -60,6 +60,9 @@ in accordance with an Errata update of the PMIx v3 Standard
 - PR #1620: Re-address the collective tracker problem
 - PR #1625: Fix library triplet
 - PR #1630: Fix support for PGI compiler pgcc18
+- PR #1637: Realign behavior of PMIx_Get and PMIx_Get_nb
+- PR #1640: Properly handle forwarded output when no handler registered
+
 
 
 3.1.4 -- 9 Aug 2019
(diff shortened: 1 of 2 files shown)


jjhursey:
I'm not sure this is correct, or at least I need some more information before I understand the versioning change.

In Issue #1633 the original issue was that rc1 moved `libpmix.so.2` to `libpmix.so.3` by mistake, and the request was to then bump the `.so` numbering on the other `.so`'s that we ship. That I think I get.

However, since we fixed the mistake and `libpmix.so.2` is the correct `c` value then I don't think we want to change the other `.so`'s for this generation of the library (v3.x series). Otherwise, we signal a major change in those `.so`'s (similar to Issue #1624). Maybe we just change the revision number for the v3.x series.

In the PMIx 4.x series, we can update the `c` levels on all of the `.so`'s since they are different than the v3.x versions.

rhc54:
IF you follow the discussion on #1633 plus some over on the OMPI side over time, you'll see that the absolute value of the ```c``` level is irrelevant - what the linkers focus on is the ```c-a``` result as that sets the SOVERSION number. In this case, we neglected to shift that difference to track the ```c-a``` of the main libpmix library. This causes a problem for packagers as it breaks the SOVERSION alignment of the two (common and main) libraries.

When we go to v4, we need to bump the ```c``` value such that ```c-a``` increments, thus signaling a change in the overall SOVERSION number.

rhc54:
I believe we have to make the move now, not later. As things stand, the common/dstore lib is associated with PMIx v2.x, not v3, which is why @amckinstry pointed out that it was causing downstream problems.

jjhursey:
Ok. I'm willing to go that way. I think we need a note on the release page as a heads up for anyone that might care.

rhc54:
Okay, I added those two bullets plus one for #1640. Once you okay that one, we should be good to do this and roll rc4.

jjhursey:
Let's hold this PR until we get a resolution on the discussion item [here](https://github.com/openpmix/openpmix/issues/1633#issuecomment-586649031)

rhc54:
Per the discussion, we will not fix the common SOVERSION value until the PMIx v4 release.

PR #1639, opened 2020-02-14
Remove a potential race condition

rhc54:
Do not set a timeout on a dmodex request that gets passed up to the host
as they might also have one, resulting in a race condition that can
cause segfault.


rhc54:
@dsolt I think this should fix the segfault you saw.

@jjhursey This doesn't apply to the v3.1 branch as we didn't update the dmodex code there. No need to hold that release for this fix.

PR #1651, opened 2020-02-27
Update ignores

rhc54:


rhc54:
@jjhursey Same here - the PRRTE timeout on your monster "hostname" test is simply too short and we are timing out just before we complete it.

rhc54:
This just updates ignores, so the tests are irrelevant.

PR #1661, opened 2020-03-10
Roll to v3.1.6

rhc54:
Just in case we need to release it

Update client get to avoid unnecessarily retrieving data from local server - tracks v4 and ensures we don't fail cross-version tests between the v3.1 branch and master.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -17,7 +17,7 @@
 
 major=3
 minor=1
-release=5
+release=6
 
 # greek is used for alpha or beta release tags.  If it is non-empty,
 # it will be appended to the version number.  It does not have to be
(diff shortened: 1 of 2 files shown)


jjhursey:
It looks like 2 other commits got pushed here that were not intended.

rhc54:
No, they are intended - problem is that moving the version from 3.1.5 to 3.1.6 will break cross-version as a special case for 3.1.5 was required in master. I could special-case 3.1.6 as well, and maybe that is the correct answer - but in the meantime, I tried to update with the required mods and that didn't work, so I set it aside as "low priority"

jjhursey:
Per teleconf today:
 * A fix was placed on master for a performance issue where a 'get' operation on node/app data would go through the "hash" GDS instead of "dstore". The result is more messages than necessary between the local processes and the local daemon to resolve the 'get' operation.
 * Since this is a performance optimization (albeit a common one) and not a correctness enhancement it can be viewed as lower priority.
 * If/when we move to 3.1.6 we need to assess this and make a decision on the path forward.
 * Currently, the cross-version checks are failing because more work is needed to resolve those issues with master.
 * For now, we will leave this PR open to track the conversation.

rhc54:
Dropping this as we will use v3.2.0 instead

PR #1679, opened 2020-03-19
Check hostname aliases to ensure xfer of job-level info

rhc54:
Remote nodes may know themselves by a name different than the one known
to the launcher. Provide an attribute for communicating the known
aliases and ensure that the local node info can be identified.

### DIFF
--- a/src/mca/gds/hash/gds_hash.c
+++ b/src/mca/gds/hash/gds_hash.c
@@ -165,6 +165,7 @@ typedef struct {
     pmix_list_item_t super;
     uint32_t nodeid;
     char *hostname;
+    char **aliases;
     pmix_list_t info;
 } pmix_nodeinfo_t;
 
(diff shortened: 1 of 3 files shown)


jjhursey:
bot:ibm:prrte:retest

src/mca/gds/hash/gds_hash.c
```diff
@@ -303,6 +311,89 @@ static pmix_job_t* get_tracker(const pmix_nspace_t nspace, bool create)
...
+    size_t len1, len2;
+
+    /* only check to the shortest name */
+    len1 = strlen(h1);
+    len2 = strlen(h2);
+    if (len2 < len1) {
+        len1 = len2;
+    }
+
+    if (0 == strncmp(h1, h2, len1)) {
```
jjhursey: I don't think this is correct. We have clusters where the nodes are numbered like:
```
node1
node10
node100
```
With this logic these three different nodes would all match each other.

rhc54: Shame on you!! You folks should know better than to do something like that 😦 - I believe that was taught in sys admin 101. 😄 

I suppose we could look at the name for '.' and only check to that point, though it still leaves open the question of names that start the same but differ by domain. Perhaps the only real solution is to pass down an attribute telling us what to do with domain names - i.e., to include them in tests or strip?

PR #1682, opened 2020-03-23
Ensure notification upon connect and fix PMIx_Init return code

rhc54:
Ensure we check cached notifications upon tool or client connection.
Properly return UNREACH error status from PMIx_Init when no server is
detected so the caller can decide if they want to proceed or terminate.
Silence a Coverity warning in gds/hash.


rhc54:
bot:ibm:prrte:retest

rhc54:
I can't replicate this problem no matter how many times I try - the only way to get that error is for the DVM to have died, but there is no report that this happened or why, and I cannot get the DVM to die when manually running the test locally.

PR #1685, opened 2020-03-24
Export pmix_notify_check_range so the ptl/tcp component can use it

jjhursey:


jjhursey:
I'm trying to pinpoint the CI failure first noticed here:
 * https://github.com/openpmix/openpmix/pull/1682#issuecomment-602771923

This is a PR to help me narrow it down.

jjhursey:
I'm able to reproduce inside the CI environment with 10 nodes. It seems to have to do with the second job. When the second job starts the prte exits. I'm digging.

jjhursey:
Ok this is what it is tripping on:
```
[mpiczar@3c5c6bff26ef (docker) hello_world] prun --map-by ppr:5:node hostname
prte: symbol lookup error: /workspace/exports/pmix/lib/pmix/mca_ptl_tcp.so: undefined symbol: pmix_notify_check_range
[3c5c6bff26ef:00515] PMIx_Spawn failed (-157): OPERATION-SUCCEEDED
[1]+  Exit 127                prte --daemonize --debug -mca state_base_verbose 10 --hostfile /workspace/hostfile.txt
```

With the `--daemonize` option it was dropping `prte`'s stderr. I think the function declaration just needs moved to be more visible. Let me try that.

jjhursey:
Ok let's try this patch (worked offline).

In PR #1682 the symbol `pmix_notify_check_range` was not exported so it is possible that it would not resolve as a symbol when the `_check_cached_events` is activated such that it would call that function.

This was not seen in the CI output because we used the `--daemonize` option to `prte` which hid the stderr message.

jjhursey:
bot:ibm:prrte:retest

PR #1687, opened 2020-03-25
Play it safe - export all functions/vars

rhc54:

### DIFF
--- a/src/tool/pmix_tool.c
+++ b/src/tool/pmix_tool.c
@@ -67,11 +67,11 @@
 #include "src/include/pmix_globals.h"
 #include "src/common/pmix_attributes.h"
 #include "src/common/pmix_iof.h"
+#include "src/client/pmix_client_ops.h"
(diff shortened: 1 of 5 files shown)


src/tool/pmix_tool.c
```diff
@@ -870,7 +870,7 @@ PMIX_EXPORT int PMIx_tool_init(pmix_proc_t *proc,
     return rc;
 }
 
-pmix_status_t pmix_tool_init_info(void)
+PMIX_EXPORT pmix_status_t pmix_tool_init_info(void)
```
jjhursey: I thought we only put the `PMIX_EXPORT` on the function declaration, not the definition?

jjhursey: I guess it follows the pattern from `PMIx_tool_init` above. I'm fine with it, just didn't know.

rhc54: We haven't been consistent, I fear - but I have no strong feelings either way

PR #1706, opened 2020-04-02
Silence unnecessary error msg in configure

rhc54:


rhc54:
Something broke v2.2 with itself - will have to explore that separately.

PR #1714, opened 2020-04-08
Added test_v2 directory for new version of test suite (#2)

cpshereda:
This is the start of the new PMIx test suite per @artpol84 's guidance. 

This initial version only includes pmix_test.c and all its dependencies, and nothing else, in a new subdirectory of test called test_v2. 

For now it is built using a custom Makefile and is not touched by configure, so it will not be part of the regular build process, nor enabled, for now. The idea is that for a period of time it will be run independent of/in parallel to the existing test suite.

### DIFF
--- a/test/test_v2/Makefile
+++ b/test/test_v2/Makefile
@@ -1,0 +1,35 @@
+INSTALLDIR = /home/cps/pmix-install
+INCLUDES = -I. -I../../test -I../../include -I../.. -I../../src/include
+vpath %.h .:../../test:../../include:../..:../../src/include
+LDIR = $(INSTALLDIR)/lib
+
+CC = gcc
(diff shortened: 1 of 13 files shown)


artpol84:
@rhc54 @jjhursey 

These changes are not touching any working pieces of the PMIx and introduce a new directory where the new test system will evolve. From my perspective it is good to go.

BTW I'm still not able to merge into PMIx master. I don't have this issue with OMPI anymore. @rhc54 can you look again into that? Maybe we can use this PR to debug my access rights.

rhc54:
Not quite sure what to try here - you are a member of the "Developers" team, and that team has "write" access to the repo. So everything looks okay from the GitHub perspective. Are you seeing the "merge" button grayed out?

rhc54:
I think I found it - see if the button is green for you now

artpol84:
Thanks!

Issue #320, opened 2017-02-26
dstore: keep track of the proc data stored in the dstore.


artpol84:
Analysis of the direct-modex performance of the dstore (https://github.com/pmix/master/issues/207, in particular https://github.com/pmix/master/issues/207#issuecomment-282433672) shows that there is some inefficiency there.

Looking through the server-side get logic I've noticed that the following race condition is possible:
* client1 searches dstore and doesn't found info about the procX.
* client1 sends request to the server to get the data.
* client2 searches dstore and doesn't found info about the procX.
* server retrieves procX data and stores them in the dstore and responds to client1 cleaning up this request history.
* client2 requests procX info from the server.
* server will find the procX blob in it's hash table and will publish it second time which will introduce multiple undesired effects, like key overwriting and extra dstore locking.

What needs to be done:
1. I think we don't need to store procX data in the server hash if we use dstore.
2. We need to implement dstore callback to query presence of the procX data in the dstore.

rhc54:
I believe this may now be out-of-date and somewhat overlaps #586 - can you please confirm and close if true?

artpol84:
I am sure that nothing has been done to fix that so far, so this is still relevant.
I think that #586 describes the functionality issue (that we are going to address ASAP), while this issue describes a performance problem.
Currently server will always publish the data in the dstore if he got the message from the client. But sometimes because of the concurrent dstore access server will receive multiple requests and thus publish same data twice doing useless work and introducing extra locking contention.

I would like to keep this Issue and we will address this along with #586 but with less urgency as its not a functionality bug.

artpol84:
@karasevb lets see what's happening in #586 and keep this issue in mind. Probably we will be able to solve both of this at once.

jjhursey:
@artpol84 Is this ticket ok to close? Given the #586 was resolved and the dstore enhancements that went into master?

artpol84:
I believe this is still relevant. Though not a high prio. Can discuss on the telecon next week (I'm not available this week)

rhc54:
No longer relevant

PR #1715, opened 2020-04-10
pmix_server_get: avoid creating a local tracker for existing keys

karasevb:
This commit avoids creating a local tracker for existing keys.
A local client may send a get request concurrently with a commit
request from another client, but the server may have processed the
commit request earlyer than the get request. In this case, we create
a local tracker for possibly existing keys that are added with the
completed commit request. Thus, the get request will be pended in
tracker and will be deffered. This scenario is possible when the
non-fence commit-get scheme is used and when the peer GDS
component is `dstore`.

### DIFF
--- a/src/server/pmix_server_get.c
+++ b/src/server/pmix_server_get.c
@@ -410,6 +410,25 @@ pmix_status_t pmix_server_get(pmix_buffer_t *buf,
         cb.ninfo = cd->ninfo;
         cb.key = key;
         PMIX_GDS_FETCH_KV(rc, pmix_globals.mypeer, &cb);
+        /* A local client may send a get request concurrently with
+         * a commit request from another client, but the server may
+         * have processed the commit request earlyer than the get
+         * request. In this case, we create a local tracker for
+         * possibly existing keys that are added with the completed
+         * commit request. Thus, the get request will be pended in
+         * tracker and will be deffered. This scenario is possible
+         * when the non-fence commit-get scheme is used and when
+         * the peer GDS component is `dstore`.
+         * Checking the peer storage for local keys to avoid creating
+         * a local tracker for existing keys. */
+        if ((PMIX_SUCCESS != rc) && local) {
+            PMIX_GDS_FETCH_KV(rc, cd->peer, &cb);
+            if (PMIX_SUCCESS == rc) {
+                cbfunc(rc, NULL, 0, cbdata, NULL, NULL);
+                PMIX_DESTRUCT(&cb);
+                return rc;
+            }
+        }
         PMIX_DESTRUCT(&cb);  // does not release info or key
         if (PMIX_SUCCESS != rc) {
             /* if the target proc is local, then we just need to wait */


karasevb:
Here is the client's code which reproduces the described scenario, but not 100% reproducible. For 100% repro required a server commit patch added bellow:
<details>
  <summary>client.c code</summary>

```C
#include <stdio.h>
#include <pmix.h>

#define GET_TIMEOUT 10

{
    int rc;
    pmix_proc_t myproc, proc;
    pmix_value_t kv;
    const char *key = "pml.ucx.4.1";
    pmix_value_t *val = &kv;
    int i, nprocs = 0;

    if (PMIX_SUCCESS != (rc = PMIx_Init(&myproc, NULL, 0))) {
        fprintf(stderr, "PMIx_Init failed: %s", PMIx_Error_string(rc));
        return rc;
    }

    PMIX_LOAD_PROCID(&proc, myproc.nspace, PMIX_RANK_WILDCARD);
    if (PMIX_SUCCESS != (rc = PMIx_Get(&proc, PMIX_JOB_SIZE, NULL, 0, &val))) {
        pmix_output(0, "rank %d: PMIx_Get universe size failed: %s",
                    myproc.rank, PMIx_Error_string(rc));
        return rc;
    }

    nprocs = val->data.integer;
    if (nprocs != 2) {
        pmix_output(0, "Error: proc size is %d, expected 2");
        goto exit;
    }

    kv.type = PMIX_UINT64;
    kv.data.uint64 = 1234;
    if (myproc.rank == 1) {
        if (PMIX_SUCCESS != (rc = PMIx_Put(PMIX_LOCAL, key, &kv))) {
            pmix_output(0, "Client ns %s rank %d: PMIx_Put internal failed: %s",
                        myproc.nspace, myproc.rank, PMIx_Error_string(rc));
            return rc;
        }
    }

    rc = PMIx_Fence(NULL, 0, NULL, 0);
    if (PMIX_SUCCESS != rc) {
        pmix_output(0, "rank %d: PMIx_Fence failed: %s", myproc.rank, PMIx_Error_string(rc));
    }

    if (myproc.rank == 1) {
        if (PMIX_SUCCESS != (rc = PMIx_Commit())) {
            pmix_output(0, "Client ns %s rank %d: PMIx_Commit failed: %s",
                        myproc.nspace, myproc.rank, PMIx_Error_string(rc));
            return rc;
        }
        pmix_output(0, "rank %d: commit", myproc.rank, key);
    } else {
        pmix_info_t info;
        int to = GET_TIMEOUT;

        usleep(700);
        PMIX_INFO_LOAD(&info, PMIX_TIMEOUT, &to, PMIX_INT);

        PMIX_LOAD_PROCID(&proc, myproc.nspace, 1);
        rc = PMIx_Get(&proc, key, &info, 1, &val);
        pmix_output(0, "rank %d: PMIx_Get status: %s, %lu", myproc.rank,
                    PMIx_Error_string(rc), val->data.uint64);
        if (PMIX_SUCCESS != rc) {
            pmix_output(0, "rank %d: PMIx_Get failed: %s", myproc.rank,
                        PMIx_Error_string(rc));
            goto exit;
        }
    }

    rc = PMIx_Fence(NULL, 0, NULL, 0);
    if (PMIX_SUCCESS != rc) {
        pmix_output(0, "rank %d: PMIx_Fence failed: %s",
                    myproc.rank, PMIx_Error_string(rc));
    }

exit:
    if (PMIX_SUCCESS != (rc = PMIx_Finalize(NULL, 0))) {
        fprintf(stderr, "Client ns %s rank %d:PMIx_Finalize failed: %s\n",
                myproc.nspace, myproc.rank, PMIx_Error_string(rc));
    } else {
        fprintf(stderr, "Client ns %s rank %d:PMIx_Finalize successfully completed\n",
                myproc.nspace, myproc.rank);
    }

    return 0;
}

```
</details>

<details>
  <summary>pmix_server_commit patch</summary>

```patch
diff --git a/src/server/pmix_server_ops.c b/src/server/pmix_server_ops.c
index 9328b8c7..f753ab07 100644
--- a/src/server/pmix_server_ops.c
+++ b/src/server/pmix_server_ops.c
@@ -169,6 +169,8 @@ pmix_status_t pmix_server_commit(pmix_peer_t *peer, pmix_buffer_t *buf)
     size_t sz;
     pmix_cb_t cb;
 
+    sleep(1);
+
     /* shorthand */
     info = peer->info;
     nptr = peer->nptr;

```
</details>


Hangs for 10 seconds for a timeout defined as `GET_TIMEOUT`, despite the requested key exists:
```bash
./pmix_test -n 2 -e client
```

rhc54:
Took me a while to wrap my head around this, so let me try to explain it back to you and see if I finally understood it correctly. I will use client A as the proc that commits the key, and client B as the proc that is attempting to retrieve it.

* client A commits the key, which is sent to the server
* client B issues a `PMIx_Get` call for that key. Because client B is using a dstore component, it first checks its shmem to see if the key exists. As the server has not yet processed the commit from client A, the key is not found. Therefore, client B sends a "dmodex" request to the server for the specified key
* the server serializes the two requests (the "commit" from A and the "get" from B). It processes the "commit" and (because the scope is "local") places the value in the shmem storage for that nspace. It then processes the "get", looking solely at its own "hash" store because this is a "dmodex" request. The key isn't found.
* because the key wasn't found, the server creates a local tracker and sets the timeout. However, because A is a local proc, it doesn't ask its host for the data - it assumes that it just hasn't been committed yet and waits until the timeout expires, then returning a "not found" error

Is that an accurate description? Your solution, therefore, is for the server to first check its own "hash" store, but if the key isn't found and A is a local proc, to then check the "dstore" for that nspace to see if the key was already stored there. Correct?

If I have that right, then sure - this makes sense.

karasevb:
@rhc54 Yes, you are correct.

karasevb:
I'll verify that it is relevant for the branches.

artpol84:
I think that #320 is related to this PR in the sense that we can solve that problem in a similar way.
@karasevb please have a look.

rhc54:
> I'll verify that it is relevant for the branches.

@karasevb I wouldn't bother checking the release branches. The v3 releases will immediately return "not found" instead of hanging as the dmodex changes were not brought back to them (too much change). Current thinking is that if we decide to update v3, we will simply branch master into a new v3.2, remove the v4-related things from it, and release as v3.2.

PR #1732, opened 2020-04-19
Enable search for available servers

rhc54:
Tools/debuggers need a mechanism by which they can discover the servers
available to the caller on the local node. Provide a query attribute
plus a couple of corresponding attributes for returning the information.
Extend the information placed in the rendezvous to include useful things
like the pid of the server, the uid/gid it is running under, and the
date/time it started.


rhc54:
@dirk-schubert-arm As we discussed, this provides a method for discovering the available servers:
```text
$ ./pquery PMIX_QUERY_AVAIL_SERVERS
PMIX_SERVER_INFO_ARRAY: 
	PMIX_SERVER_NSPACE:  prte-Ralphs-iMac-251042
	PMIX_SERVER_RANK:  0
	pmix.version:  4.0.0a1
	PMIX_SERVER_PIDINFO:  51042
	pmix.euid:  1000
	pmix.egid:  20
	PMIX_SERVER_START_TIME:  Sun Apr 19 14:50:44 2020
$
```

rhc54:
@jdelsign FYI - might be useful feature. Returns an array of servers that the calling uid/gid has permissions to access. Server has posted connection info, so the caller has at least the potential to connect with the returned servers (ultimately, the server's host [RM daemon] has the option to reject the connection).

PR #1740, opened 2020-04-23
Fixes cleanup the dmdx satisfied request list

karasevb:
Satisfied requests have already invoked `cbfunc` and must be destructed. Otherwise they keep alive and subject to timer event (if the timer has been enabled), which causes them to try to invoke `cbfunc` again, that leads to segfault.

### DIFF
--- a/src/server/pmix_server_get.c
+++ b/src/server/pmix_server_get.c
@@ -913,6 +913,10 @@ pmix_status_t pmix_pending_resolve(pmix_namespace_t *nptr, pmix_rank_t rank,
   cleanup:
     /* remove all requests to this rank and cleanup the corresponding structure */
     pmix_list_remove_item(&pmix_server_globals.local_reqs, &ptr->super);
+    while (NULL != (req = (pmix_dmdx_request_t*)
+                    pmix_list_remove_first(&ptr->loc_reqs))) {
+        PMIX_RELEASE(req);
+    }
     PMIX_RELEASE(ptr);
 
     return PMIX_SUCCESS;


karasevb:
bot:ibm:retest

src/server/pmix_server_get.c
```diff
@@ -913,6 +913,10 @@ pmix_status_t pmix_pending_resolve(pmix_namespace_t *nptr, pmix_rank_t rank,
   cleanup:
     /* remove all requests to this rank and cleanup the corresponding structure */
     pmix_list_remove_item(&pmix_server_globals.local_reqs, &ptr->super);
+    while (NULL != (req = (pmix_dmdx_request_t*)
+                    pmix_list_remove_first(&ptr->loc_reqs))) {
+        PMIX_RELEASE(req);
+    }
```
rhc54, line 919: Probably easier to just use the macro:

```suggestion
    PMIX_LIST_DESTRUCT(&ptr->loc_reqs);
```

karasevb: @rhc54 
```C
PMIX_LIST_DESTRUCT(&ptr->loc_reqs);
PMIX_RELEASE(ptr);
```
Thanks. I have already checked this way, but this leads to the destructors collision:
`PMIX_RELEASE(ptr)` invokes `PMIX_LIST_DESTRUCT` second time.

rhc54: If that is true, then just release the `ptr` object and don't destruct the list separately. There is no value in doing so, is there?

rhc54: Thinking about it and looking again at your patch, I suspect the underlying problem here is that the `ptr` object has a refcount greater than 1 at this point in the code. Thus, `PMIX_RELEASE(ptr)` doesn't actually run the destructor, and therefore the `local_reqs` list isn't being destructed and we see the timeout issue.

If that is what you see, then the correct solution is to find out why the refcount isn't correct and fix that issue. Otherwise, we will just leak memory.

karasevb: The sense of this PR is destruct the list separately. Once we add an item to the list it increments the ref counter for `ptr`: https://github.com/openpmix/openpmix/blob/5f898014556fbd9dcf5b7fa3f46b751ece585e64/src/server/pmix_server_get.c#L611-L614
So removing list separately is decrements the ptr refcounter [here](https://github.com/openpmix/openpmix/blob/5f898014556fbd9dcf5b7fa3f46b751ece585e64/src/server/pmix_server_ops.c#L4683), and then `PMIX_RELEASE(ptr)` invokes the destructor correctly.
I checked this code path at runtime in `gdb` - it hasn't leaks in destructors.

rhc54: Hmmm...okay, I see. Disturbing, though, as this seems like an open door for lots of developer mistakes going forward. I'll have to think about a cleaner mechanism here as doing a special case like this every time we want to destruct this object seems error prone.

However, we can tackle that separately.

PR #1748, opened 2020-04-30
Remove symbol hiding

rhc54:
Probably better just to have applications use the "embedded" lib now
that OMPI is no longer really embedding it


rhc54:
bot:ibm:xvercheck:retest

PR #1749, opened 2020-04-30
Remove PMIX_NUMA_RANK attribute

rhc54:
NUMA domains are no longer quite as crystal clear as they used to be, so
really cannot define a single "numa rank" for a proc


rhc54:
Included in PR #1758

PR #1759, opened 2020-05-08
Allow passing of PMIX_PREFIX to PMIx_tool_init

rhc54:
Tools need to be able to pass an install prefix for the PMIx they want
to use, but cannot always pass it in the environment - e.g., if they are
going to fork/exec a PMIx-based application/launcher that uses a
different PMIx library. Provide a code path that lets them pass the
PMIX_PREFIX attribute down into the installdirs framework where it can
set the PMIX_INSTALL_PREFIX field.


rhc54:
@jdelsign I believe this addresses what we discussed.

jdelsign:
@rhc54 Thanks much for the feature. It looks like the pull request is blocked, so it's not in master yet. I'll test it out once it gets merged.

PR #1765, opened 2020-05-13
ds21: fixed crash of dstor locks destructor

karasevb:
A check was added to avoid locks finalization in the error case
when the dstore segment was not created and its pointer is NULL.


jjhursey:
bot:ibm:retest

PR #1767, opened 2020-05-14
Update Lustre config logic plus add a new component prototype

rhc54:
Update lustre config logic
Tracks OMPI equivalent code

Silence Coverity warning

Add a new slingshot prototype component
To be completed - ignored for now.


rhc54:
Hmm..same Travis problem again. Everything passes, but it doesn't update the status. Will investigate later.

PR #1771, opened 2020-05-18
configure.ac: use "pax" tar format

mblockso:
The default "ustar" tar format (POSIX.1-1988) imposes a restriction that the
UID/GID is at most 2097151. Unfortunately, the configure step will *not fail*
if this condition is not met - from the GNU Automake documentation:

> "configure knows several ways to construct these formats. It will not abort if it cannot find a tool up to the task (so that the package can still be built), but 'make dist' will fail."

The "ustar" tar format has the UID/GID restriction, but it also has a maximum
8GB file size and a 256 character maximum filename length.

The "pax" tar format (POSIX.1-2001) supports "unlimited" UID/GID value and also
supports unlimited filename lengths and an unlimited file size.

This commit will switch automake to use the "pax" tar format to generate
distribution tar files.

Signed-off-by: Michael Blocksome <michael.blocksome@intel.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -94,7 +94,7 @@ pmix_show_title "Configuring PMIx"
 AC_CANONICAL_TARGET
 
 # Init automake
-AM_INIT_AUTOMAKE([foreign dist-bzip2 subdir-objects no-define 1.13.4])
+AM_INIT_AUTOMAKE([foreign dist-bzip2 subdir-objects no-define 1.13.4 tar-pax])
 
 # SILENT_RULES is new in AM 1.11, but we require 1.13.4 or higher via
 # autogen.  Limited testing shows that calling SILENT_RULES directly


mblockso:
FYI .. this same problem was recently fixed in the libfabric project -> https://github.com/ofiwg/libfabric/pull/5852

rhc54:
@jsquyres @bwbarrett Are you aware of any issues with making this change?

jsquyres:
As I commented in https://github.com/ofiwg/libfabric/pull/5852#pullrequestreview-398530883, Libfabric has a fairly narrow base for where it cares about making / distributing its tarballs.

Open MPI has a wider distribution base than Libfabric, and this change may be problematic.  Here's what the Automake docs issue the following warning about tar-pax:

> As of 2018, this format is supported by the native tar command only on GNU, FreeBSD, OpenBSD system; it is not supported by the native tar command on NetBSD, AIX, HP-UX, Solaris. There are moves to change the pax format in an upward-compatible way, so this option may refer to a more recent version in the future.

In particular, Open MPI does support NetBSD and Solaris.  It would be problematic if we produced tarballs that could not be read on those platforms.

rhc54:
Yeah, that was my concern. Given that our tarballs are less than 1Mbyte, it isn't clear to me that we really have anything to gain by doing this change.

@mblockso Is there something motivating this change? Or are you proposing it to conform with what libfabric did?

mblockso:
Sorry .. I probably wasn't clear about the root problem, which is that the "ustar" (default?) tar format limits the UID to less than 2097151.  If your UID is larger than 2097151 then the `make dist` rule will fail rather strangely. Also, the `configure` step will *not* fail because of this UID problem - which is frustrating.

So, for me, it is all about this UID issue because I can't build the tar files. If there is another tar format that resolves the UID issue then I'm all for that. Or maybe there is a way to switch the format at "autogen" time so that NetBSD/Solaris use the old format and everything else uses the new format? I dunno.

mblockso:
... or is there a way we can force tar to "ignore" the uid of the account that is creating the tar file? Can we assign the uid of the bundled files contained in the tar file?

jsquyres:
FWIW, it would be straightforward to add a UID check into `dist` rule so that it at least fails in a straightforward way.  You can't put this check in `configure` because the user that runs `configure` may not be the same user that runs `make dist`.  I realize that this is not really what you're asking for, but I can at least offer that tidbit.

mblockso:
Ah ha! Check this out:

```
# TAR_OPTIONS="--owner=0 --group=0 --numeric-owner" make dist
```
...
```
# tar -tvf pmix-4.0.0a1.tar.bz2 | head
drwxr-xr-x 0/0               0 2020-05-19 11:56 pmix-4.0.0a1/
drwxr-xr-x 0/0               0 2020-05-19 12:37 pmix-4.0.0a1/maint/
-rw-r--r-- 0/0             296 2020-05-18 10:25 pmix-4.0.0a1/maint/pmix.pc.in
-rw-r--r-- 0/0            3851 2020-05-18 10:25 pmix-4.0.0a1/INSTALL
-rw-r--r-- 0/0            3277 2020-05-18 10:25 pmix-4.0.0a1/VERSION
drwxr-xr-x 0/0               0 2020-05-19 12:37 pmix-4.0.0a1/test/
-rw-r--r-- 0/0           22115 2020-05-18 10:25 pmix-4.0.0a1/test/test_common.c
-rw-r--r-- 0/0            6796 2020-05-18 10:25 pmix-4.0.0a1/test/cli_stages.c
-rw-r--r-- 0/0           11303 2020-05-18 10:25 pmix-4.0.0a1/test/pmi_client.c
-rwxr-xr-x 0/0            3051 2020-05-18 10:25 pmix-4.0.0a1/test/run_tests02.pl.in
```

Now what is needed is a way to pass these extra arguments to tar automagically. Not sure how to do that .. we would need to change the `am__tar` variable in the generated Makefile?

mblockso:
This `TAR_OPTIONS` workaround seems to work fine for openpmix, but it fails for prrte. This is because prrte uses the "ustar" tar format in configure.ac for AM_INIT_AUTOMAKE, but openpmix does not specify a tar format.

As a result, the "tar with ustar" configure test for prrte fails and the `am__tar` variable in the Makefile gets set to `false`! Then it doesn't matter if I set `TAR_OPTIONS` or not because the tar command is never invoked - only the command `false` is invoked, which obviously causes the make to fail.

@rhc54 @jsquyres .. What should I do here? .. open an issue with prrte and close this openpmix PR?

jsquyres:
@mblockso I'm confused.  Are you saying that `make dist` fails for you with prte because `am__tar` is set to `false`?  It isn't for me; i.e., `make dist` works fine for me in openpmix and prte.

Is the ustar test failing for you in prte, and therefore it sets `am__tar` to `false`?
And if that's true, since openpmix doesn't specify a format, does that mean it's defaulting to pax for you (since, assumedly, `make dist` isn't failing for you due to `am__tar` being set to `false`)?

mblockso:
> @mblockso I'm confused. Are you saying that `make dist` fails for you with prte because `am__tar` is set to `false`? It isn't for me; i.e., `make dist` works fine for me in openpmix and prte.

Correct. For me, during **prrte** configure, some tar test "fails" because:
1. The prrte automake specifies the "tar-ustar" tar format (added 12+ years ago https://github.com/openpmix/prrte/commit/a348da4eee7db969b5ab487c09e2efc4c77a3872)
1. my UID is > 2097151

... and `configure` doesn't actually fail, it just silently sets this `am__tar` variable to `false` so you only encounter this error if you try and run `make dist` later. Very annoying.

> Is the ustar test failing for you in prte, and therefore it sets `am__tar` to `false`?

Yes. If I remove the `tar-ustar` option from prrte automake options, then it behaves just like openpmix and I am able to get past the UID problem by setting that `TAR_OPTIONS` environment variable before I run `make dist`

> And if that's true, since openpmix doesn't specify a format, does that mean it's defaulting to pax for you (since, assumedly, `make dist` isn't failing for you due to `am__tar` being set to `false`)?

I'm not sure what format is being used in openpmix configure, but it allows my UID and sets `am__tar` to something like `${TAR_tar}`. 

I don't think it is using `tar-pax` though because I have to use this `TAR_OPTIONS` environment variable before I `make dist` to specify a different UID/GID for the tar contents because my account UID/GID is too large for whatever tar format is being used.

Here's a summary ...

### current behavior: with `tar-ustar`
```
# configure
...

# head -n357 config.log | tail -n8
configure:5923: checking whether UID '11463922' is supported by ustar format
configure:5929: result: no
configure:5933: checking whether GID '2222' is supported by ustar format
configure:5936: result: yes
configure:5944: checking how to create a ustar tar archive
configure:5998: tardir=conftest.dir && eval false >conftest.tar
configure:6001: $? = 1
configure:6026: result: none

# grep am__tar config.log 
am__tar='false'

# grep "am__tar =" Makefile 
am__tar = false

# make dist
...
make[2]: Leaving directory '/home/mblockso/development/project/build/prrte'
tardir=prte-gitclone && false | BZIP2=${BZIP2--9} bzip2 -c >prte-gitclone.tar.bz2
tardir=prte-gitclone && false | eval GZIP= gzip --best -c >prte-gitclone.tar.gz
make[1]: Leaving directory '/home/mblockso/development/project/build/prrte'
if test -d "prte-gitclone"; then find "prte-gitclone" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -rf "prte-gitclone" || { sleep 5 && rm -rf "prte-gitclone"; }; else :; fi
```

### *without* `tar-ustar`
```
# git diff configure.ac
diff --git a/configure.ac b/configure.ac
index 33d9e18817..0b09203ca1 100644
--- a/configure.ac
+++ b/configure.ac
@@ -107,7 +107,7 @@ PRTE_VAR_SCOPE_POP
 # Init automake
 #
 
-AM_INIT_AUTOMAKE([foreign dist-bzip2 subdir-objects no-define 1.12.2 tar-ustar])
+AM_INIT_AUTOMAKE([foreign dist-bzip2 subdir-objects no-define 1.12.2])
 
 # SILENT_RULES is new in AM 1.11, but we require 1.11 or higher via
 # autogen.  Limited testing shows that calling SILENT_RULES directly

# configure
...

# grep -n am__tar config.log
40276:am__tar='$${TAR-tar} chof - "$$tardir"'

# grep -n "am__tar =" Makefile 
673:am__tar = $${TAR-tar} chof - "$$tardir"

# make dist
...
make[2]: Leaving directory '/home/mblockso/development/project/build/prrte'
tardir=prte-gitclone && ${TAR-tar} chof - "$tardir" | BZIP2=${BZIP2--9} bzip2 -c >prte-gitclone.tar.bz2
tar: value 11463922 out of uid_t range 0..2097151
tar: Exiting with failure status due to previous errors
tardir=prte-gitclone && ${TAR-tar} chof - "$tardir" | eval GZIP= gzip --best -c >prte-gitclone.tar.gz
tar: value 11463922 out of uid_t range 0..2097151
tar: Exiting with failure status due to previous errors
make[1]: Leaving directory '/home/mblockso/development/project/build/prrte'
if test -d "prte-gitclone"; then find "prte-gitclone" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -rf "prte-gitclone" || { sleep 5 && rm -rf "prte-gitclone"; }; else :; fi
```
### without `tar-ustar`, but *with* `TAR_OPTIONS`
```
# TAR_OPTIONS="--owner=0 --group=0 --numeric-owner" make dist
...
make[2]: Leaving directory '/home/mblockso/development/project/build/prrte'
tardir=prte-gitclone && ${TAR-tar} chof - "$tardir" | BZIP2=${BZIP2--9} bzip2 -c >prte-gitclone.tar.bz2
tardir=prte-gitclone && ${TAR-tar} chof - "$tardir" | eval GZIP= gzip --best -c >prte-gitclone.tar.gz
make[1]: Leaving directory '/home/mblockso/development/project/build/prrte'
if test -d "prte-gitclone"; then find "prte-gitclone" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -rf "prte-gitclone" || { sleep 5 && rm -rf "prte-gitclone"; }; else :; fi
```

rhc54:
So if I am following all this (and I don't claim I am), all we really need to do is remove the "tar-ustar" from the PRRTE automake options. Is this correct?

@jsquyres Given that PMIx doesn't have such an option and is working fine in OMPI tarballs, can you see any reason why we shouldn't remove it from PRRTE?

mblockso:
> So if I am following all this (and I don't claim I am), all we really need to do is remove the "tar-ustar" from the PRRTE automake options. Is this correct?

Yes .. I think that is correct.

rhc54:
So maybe the thing to do is close this PR and open one against PRRTE that just removes the "tar-ustar" option. We can let @jsquyres tell us over there if that is a problem.

Would you like to do that, or do you want me to do so?

mblockso:
I'll open the PRRTE PR .. thanks for the help Ralph!

mblockso:
Closing this PR .. discussion moved to https://github.com/openpmix/prrte/pull/579

PR #1772, opened 2020-05-18
Point to the correct file that wasn't found

rhc54:
When doing a dlopen, we apparently strip the extension from the actual
filename to identify the component it relates to, and then add accepted
extensions back to the filename when checking for file existence prior
to loading. Unfortunately, some 3rd party packages (e.g., debuggers)
take our libraries, do a little magic, and then output a copy of the
results into our lib directory - only with a new suffix. So we might get
something like "ourcomponent.so.dbg". We then strip the ".dbg" and add
".so" to it again - and complain that "ourcomponent.so is not found".

Fix the error message so it reports the name of the file it actually
couldn't find, not the name of the component, so it doesn't totally
confuse people.

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -816,15 +816,19 @@ AC_DEFUN([PMIX_SETUP_CORE],[
                                    MCA-variable-setting mechansism).  This MCA variable
                                    controls whether warnings are displayed when an MCA
                                    component fails to load at run time due to an error.
-                                   (default: enabled, meaning that
+                                   (default: enabled in --enable-debug builds, meaning that
                                    mca_base_component_show_load_errors is enabled
-                                   by default])])
+                                   by default when configured with --enable-debug])])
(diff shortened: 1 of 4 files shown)


jjhursey:
Interesting. Is this something Open MPI needs as well? Or are the debuggers only concerned about the PRRTE set of symbols?

rhc54:
I think it will hit all the libs as the debuggers do "strip" everything. Also, it would be good if OPAL output the correct file name that wasn't found - it cost me hours trying to figure out why it wasn't finding a file that clearly existed, only to finally realize it was not finding "component.so.so" but reporting it didn't find "component.so". Infuriating, as you can imagine. 🤕

jjhursey:
Yeah, this looks like one that would be difficult to pin down. Let's bring it up on the OMPI call to make them aware. FYI: @jsquyres @bwbarrett @gpaulsen

PR #1773, opened 2020-05-18
Don't require pandoc on developer builds

rhc54:
Developers don't typically refer to man pages, and requiring pandoc on
every developer's machine is proving rather controversial. So if we
detect we are building from within a git clone, only require pandoc and
build the manpages if the developer requests it. Otherwise, build the
manpages if there is an adequate pandoc available, but don't if not.


rhc54:
Odd - v2.2 fails to itself. Obviously nothing to do with this change, but probably merits separate investigation.

PR #1779, opened 2020-05-19
Add a LSF libevent.so check on Libevent check

jjhursey:


jjhursey:
This PR brings the warning message from the PRs below into OpenPMIx
 - https://github.com/open-mpi/ompi/pull/7734
 - https://github.com/open-mpi/ompi/pull/7749

It'll be rare that the user hits this warning. They would need both LSF in the default search path, and have a botched libevent install. But this could save those on LSF system a bit of digging if they are on such a system.

Issue #1782, opened 2020-05-21
python-bindings build error with master


naughtont3:
Python-binding error during `make install` with latest master of openpmix

### What version of the PMIx Reference Library are you using? (e.g., v1.0, v2.1, git master @ hash, etc.)

master at 31f8408

### Describe how PMIx was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

git clone

### Please describe the system on which you are running

* Operating system/version: Linux
* Computer hardware: x86-64
* Network type:

-----------------------------

## Details of the problem

Things configure and compile fine, but the `make install` that runs the cython has an error.

```shell

...<snip>...

make[3]: Entering directory `/home/3t4/projects/ompi-x/scratch/HEALTHY-ROOK/source/openpmix-git-br-master/bindings/python'
/usr/bin/python3 ../../bindings/python/construct.py --src="../../include"
/usr/bin/python3 ../../bindings/python/setup.py build_ext --include-dirs="../../include" --library-dirs="/home/3t4/projects/ompi-x/scratch/HEALTHY-ROOK/source/openpmix-git-br-master/_install/lib" --user
Compiling pmix.pyx because it changed.
[1/1] Cythonizing pmix.pyx
warning: pmix.pxi:875:12: Unreachable code

Error compiling Cython file:
------------------------------------------------------------
...
            j = 0
            while NULL != regattrs[n].description[j]:
                PyMem_Free(regattrs[n].description[j])
                j += 1
            PyMem_Free(regattrs[n].description)
        if regattrs[n].info != NULL:
                     ^
------------------------------------------------------------

pmix.pxi:1595:22: Object of type 'pmix_regattr_t' has no attribute 'info'

Error compiling Cython file:
------------------------------------------------------------
...
            j = 0
            while NULL != regattrs[n].description[j]:
                PyMem_Free(regattrs[n].description[j])
                j += 1
            PyMem_Free(regattrs[n].description)
        if regattrs[n].info != NULL:
                           ^
------------------------------------------------------------

pmix.pxi:1595:28: Invalid types for '!=' (Python object, void *)

Error compiling Cython file:
------------------------------------------------------------
...
            while NULL != regattrs[n].description[j]:
                PyMem_Free(regattrs[n].description[j])
                j += 1
            PyMem_Free(regattrs[n].description)
        if regattrs[n].info != NULL:
            pmix_free_info(regattrs[n].info, regattrs[n].ninfo)
                                     ^
[...]

PR #1783, opened 2020-05-22
Fix Python binding compilation

rhc54:
Update the PMIx_Register_attributes function signature. Note that there
are still some warnings coming out from different areas of the bindings
not related to this patch that need investigation.

Fixes https://github.com/openpmix/openpmix/issues/1782


naughtont3:
thanks @rhc54

Issue #1762, opened 2020-05-12
PMIX ERROR: BAD-PARAM in file base/pnet_base_fns.c at line 113


Honggang-LI:
Thank you for taking the time to submit an issue!

### What version of the PMIx Reference Library are you using? (e.g., v1.0, v2.1, git master @ hash, etc.)

good: pmix-2.1.1-1.el8.x86_64
bad:   pmix-2.2.3-1.el8.x86_64

### Describe how PMIx was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

pmix-2.1.1-1.el8.x86_64 was built from upstream released tarball pmix-2.1.1.tar.bz2 .

+ ./configure --build=x86_64-redhat-linux-gnu --host=x86_64-redhat-linux-gnu --program-prefix= --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib64 --libexecdir=/usr/libexec --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man --infodir=/usr/share/info --prefix=/usr --sysconfdir=/etc/pmix --disable-static --disable-silent-rules --enable-shared --enable-pmi-backward-compatibility --with-munge

------
pmix-2.2.3-1.el8.x86_64 was built from upstream released tarball pmix-2.2.3.tar.bz2 .
+ ./configure --build=x86_64-redhat-linux-gnu --host=x86_64-redhat-linux-gnu --program-prefix= --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib64 --libexecdir=/usr/libexec --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man --infodir=/usr/share/info --prefix=/usr --sysconfdir=/etc/pmix --disable-static --disable-silent-rules --enable-pmix-binaries --with-tests-examples --enable-shared --enable-pmi-backward-compatibility --with-munge

### Please describe the system on which you are running
RHEL-8

* Operating system/version:
* Computer hardware:
* Network type:

-----------------------------

## Details of the problem

```shell
[...]

PR #1788, opened 2020-05-28
Silence unnecessary error logs and prep for release

rhc54:
Fixes https://github.com/openpmix/openpmix/issues/1762


rhc54:
Hmm...something appears to have gotten into the pmix-tests that has caused "resolve peers" to fail for all v3.x and above. Has no impact on the v2.2 series or its ability to support the other versions. Will be investigated separately.

PR #1805, opened 2020-06-22
Ensure the PMIx server knows the local topology

rhc54:
Seeing increasing demand for the server to provide info based on the
topology. Need to advise RM's to pass their local topology pointer down
to the server so it doesn't do its own discovery.


rhc54:
@jaidayal Just need to tell PRRTE to pass the topology down. However, for your initial prototype work, it won't matter if the PMIx server does its own "topology load". So you should be good to go from here.

Issue #1768, opened 2020-05-15
Q: Type for `PMIX_EVENT_CUSTOM_RANGE`


jjhursey:
In the [pmix_common.h header](https://github.com/openpmix/openpmix/blob/master/include/pmix_common.h.in#L395) the `PMIX_EVENT_CUSTOM_RANGE` says that it takes the type:
 > (pmix_data_array_t*) array of pmix_proc_t defining range of event notification

But in the code I see it taking either a `PMIX_DATA_ARRAY` or a single `PMIX_PROC`. 
I noticed the code that it accepts either an array or a single proc [here](https://github.com/openpmix/openpmix/blob/master/src/event/pmix_event_notification.c#L1354-L1367).

So it looks like we just need to update the header (and standard) with a sentence like:
 > (pmix_data_array_t*) array of pmix_proc_t or a single (pmix_proc_t*) defining range of event notification

Does that sound right to folks?

jjhursey:
This impacts both `PMIx_Register_event_handler` and `PMIx_Notify_event`. The latter accepts the single `pmix_proc_t`, but the former returns `PMIX_ERR_EVENT_REGISTRATION` if you try to do so.

rhc54:
Should be either an array or single proc - both are acceptable.

jjhursey:
Thanks!

PR #1808, opened 2020-06-23
Fix registration of custom range

rhc54:
Should support both an array of procs AND a single proc pointer

Fixes https://github.com/openpmix/openpmix/issues/1768


rhc54:
bot:ibm:prrte:retest

PR #1816, opened 2020-06-26
Fix VPATH builds with pcc

jjhursey:
* The `generic_wrapper.1` needs a target on some systems otherwise
   it can be skiped and cause the build to fail.
   - Note that this was not necessary when building on my Mac, but was necessary when building with a two-level vpath on linux.

### DIFF
--- a/src/tools/wrapper/Makefile.am
+++ b/src/tools/wrapper/Makefile.am
@@ -31,6 +31,7 @@ MD_FILES_NOINST = \
 
 man_pages_from_md = $(MD_FILES:.1.md=.1)
 man_pages_from_md_noinst = $(MD_FILES_NOINST:.1.md=.1)
+man_pages_from_md_all = $(man_pages_from_md) $(man_pages_from_md_noinst)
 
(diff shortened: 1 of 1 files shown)


jjhursey:
I was not able to reproduce the failure around `pcc` in the `make install` even using the Docker container. The one thing I did notice is that it might not be running autogen.pl before configure which is strange, and might be the cause of issue.
 * https://github.com/openpmix/openpmix/runs/668063972

So let's give this a try to see if that's the issue.

rhc54:
@jjhursey FWIW: I just committed a bunch of changes down there. There was a significant amount of cruft from the pandoc change that was causing problems. It seemed to solve the CI problem, so this may no longer be an issue.

jjhursey:
I think I isolated it to a vpath build. I'm seeing if I can reproduce that locally to make it easier to dig into

jjhursey:
Ah it looks like @rhc54 beat me too it in https://github.com/openpmix/openpmix/pull/1815

PR #1818, opened 2020-06-26
Added infrastructure to test_v2 suite to enable basic results-checking

cpshereda:
Also added infrastructure to pass values as encoded argument from server(s) to clients for validation purposes.

'test_params' and 'validation_params' structs will be merged into a single struct in a later commit.


artpol84:
Looks good.
I'll create a counterpart issue for this test in pmix-test repo

artpol84:
@cpshereda One thing that I'd personally like to see is the separation between infrastructural changes and the introduction of a new test.
It should be easy to do - let me know if you want to sync on how to do it quickly.
But logically those are 2 separate items and I prefer to have them as 2 separate commits.

cpshereda:
@artpol84 This PR changed to infrastructure-only, per your request.

artpol84:
I’ve actually asked to split the pr on 2 commits.
But this way works as well.

cpshereda:
I'll do it that way next time.

PR #1821, opened 2020-07-06
examples: initialize var to avoid SEGV

naughtont3:


naughtont3:
Avoids a SEGV when running `./examples/server`

PR #1822, opened 2020-07-07
Delete dead code

cniethammer:
The static availsys array is initialized with doubles where the underlying
struct type uses uint64_t resulting in compiler warnings.
Removing all related code using this availsys array as it is only used in
a single path that has no side effect to any other part in the code base.

Signed-off-by: Christoph Niethammer <niethammer@hlrs.de>

### DIFF
--- a/src/mca/pstrg/lustre/pstrg_lustre.c
+++ b/src/mca/pstrg/lustre/pstrg_lustre.c
@@ -59,28 +59,6 @@ pmix_pstrg_base_module_t pmix_pstrg_lustre_module = {
     .query = query
 };
 
-typedef struct {
-    char *name;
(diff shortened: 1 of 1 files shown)


jjhursey:
@rhc54 this looks like some debugging code that was maybe leftover that we can safely remove. The PR looks good to me, but you are closer to this component so I wanted your review before we bring this in just in case it is helpful to keep this code in there.

rhc54:
I've been leaving it there as an example for @shanedsnyder - it can be removed as soon as he confirms he no longer needs it. Shane?

shanedsnyder:
Fine with me to remove this. I'm updating this code path anyways, but am not really relying on these data structures anymore.

PR #1823, opened 2020-07-08
Fix backward compatibility issue when running a 3.2 and later server …

dsolt:
…with a 3.1 and earlier client.

Signed-off-by: David Solt <dsolt@us.ibm.com>


dsolt:
Will need to some eyes on this, but I there is definitely some issue with put/get between ranks on the same node as evidenced by this test case with a 3.2 server and a 3.1 client (or master server and 3.1 client):

jjhursey:
@dsolt Ralph adjusted some backward compatibility in PR #1838 in this function. I don't know if it helps, but you might want to try the new v3.2 branch and see if it fixes the issue you were seeing.

jjhursey:
Ping. What the status of this PR?

rhc54:
FWIW: this test passes on current master

rhc54:
I verified this also works with v3.2 server and v3.1 client, so I believe this is complete

PR #1825, opened 2020-07-29
Add option to bypass local-only fence optimization

jjhursey:
* If all processes participating in the fence are local to the node
   then there is an optimization that allows the PMIx Server side
   library to complete the call internally without the upcall into
   the RM. For RMs that need to monitor the fence calls this poses a
   problem as they do not see the upcall.
   - This really only became an issue when a RM is trying to support
     a protocol like MPIR where they are taking advantage of a
     global fence in, in this case, MPI_Init to hold the processes
     in that fence operation by preventing it from completing until
     the debugger is attached and ready.
   - If this optimization is enabled and the fence participants are
     all local then the RM does not get the opportunity to hold that
     fence since they never receive the upcall.
 * This commit adds the `PMIX_MCA_pmix_server_fence_localonly_opt`
   MCA parameter that can be used to optionally disable this optimization.
   - By default the optimization is enabled so that there is no default
     behavior change.
   - For RMs that need it they can disable this optimization by
     setting this MCA parater to `0` (false) before calling the server
     initialization function.

(cherry picked from commit 8e58229a3c80e40ebd18e061d356e3ccc4addef7)

### DIFF
--- a/src/runtime/pmix_params.c
+++ b/src/runtime/pmix_params.c
@@ -22,6 +22,7 @@
  * Copyright (c) 2015-2018 Mellanox Technologies, Inc.
  *                         All rights reserved.
  * Copyright (c) 2016-2019 Intel, Inc.  All rights reserved.
+ * Copyright (c) 2020      IBM Corporation.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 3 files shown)


jjhursey:
Pr this just in case we do a 3.1.6

PR #1831, opened 2020-08-07
Created basic PMIx_Get() test case for rank positioning info

cpshereda:
Also added ability to validate values returned by basic PMIx_get calls.

This PR is the implementation for issue 64 in openpmix/pmix-tests: https://github.com/openpmix/pmix-tests/issues/64.

### DIFF
--- a/test/test_v2/test_common.c
+++ b/test/test_v2/test_common.c
@@ -45,9 +45,10 @@ void parse_cmd_client(int argc, char **argv, test_params *params, validation_par
     ;
 }
 
-void parse_cmd(int argc, char **argv, test_params *params)
+void parse_cmd(int argc, char **argv, test_params *params, validation_params *v_params)
 {
     int i;
+    uint32_t job_size;
 
     /* set output to stderr by default */
     file = stdout;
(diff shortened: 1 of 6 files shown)


test/test_v2/test_common.c
```diff
@@ -381,12 +392,174 @@ void pmixt_post_finalize(pmix_proc_t *this_proc, test_params *params, validation
...
 }
 
+/* This function is used to validate values returned from calls to Get against sidechannel validation data */
 void pmixt_validate_predefined(pmix_proc_t *myproc, const pmix_key_t key, pmix_value_t *value, validation_params *val_params)
 {
     pmix_status_t rc = PMIX_ERROR;
     size_t *size;
-    // To be developed.
-    ;
+
```
jjhursey, line 400: I would add a check to make sure the `value->type` is correct for the `key` - to catch if the server is returning a value with the wrong type (when can happen if an RM is new and accidentally puts in the wrong type).

test/test_v2/test_common.c
```diff
@@ -381,12 +392,174 @@ void pmixt_post_finalize(pmix_proc_t *this_proc, test_params *params, validation
...
+                } // other possibilities will be added here later
+                else {
+                    TEST_VERBOSE(("Check input for case PMIX_STRING: key: %s stringdata: %s",
+                        key, stringdata));
+                    exit(1);
+                }
+                free(stringdata);
+                break;
+            default:
+                TEST_VERBOSE(("No test logic for type: %d, key: %s", value->type, key));
```
jjhursey, line 557: You might consider making the `default` case an error case. This would make it easier to see if one of the cases missed.

test/test_v2/test_common.c
cpshereda: Ok great, thanks Josh.

test/test_v2/test_common.c
cpshereda: Yes, sounds good.

test/test_v2/test_common.c
cpshereda: My solution in the updated commit is to pass the type in as another parameter from the test, and then verify it matches the value->type, rather than hard code the additional checks for each type/key pair. Sound reasonable?

test/test_v2/test_common.c
cpshereda: Should be fixed.

PR #1832, opened 2020-08-07
Sync v3.2 branch with master

jjhursey:
* Based off of `master:HEAD` at https://github.com/openpmix/openpmix/commit/5ac61d028160d0958ca8d924ddca61117d9a57bd
 * Removed v4 specific items
   - Remove python bindings
   - Remove pattrs, pquery, and wrapper (pcc) tools
   - Remove frameworks: pfexec, pmdl, prm, pstrg
   - Remove components: pnet/simptest, pnet/sshot, pnet/test,
   - Sync examples back to the v3.1 levels
   - Removed v4-only interfaces and attributes
   - Move some interface declarations (common support functions) from `pmix.h`
     to `pmix_common.h` which is where they are located in the v3.1 branch.
   - A couple interfaces changed their signature in v4 so convert them
     back to the v3 versions.
 * New frameworks picked up from v4
   - pcompress (compress and zlib components)
   - preg (Expose `PMIX_REGEX` internal to PMIx only, not to users)


jjhursey:
Basic run testing passed. Posting the draft PR to see what CI turns up while I do further testing.
 - I have not tested cross-version yet. I'm assuming that here is more work there to do.

jjhursey:
I think this is ready for review. A few notes that might be worth discussion:
 * I bumped the `libpmix.so` 'revision' to '30' (v3.1.5 is '25')
   - That gives us a bit of room to release a couple of 3.1 series before they collide.
 * I did not change the prefixing behavior of the strings generated from the `PMIx_generate_regex` and `PMIx_generate_ppn` interfaces
   - My thinking is that this is an internal encoding so it shouldn't impact servers. It is visible if they inspect the strings though.
 * I did not include the `pcc` wrapper compiler in this release (since it is not in v3.1), but could if folks think it would be helpful.
 * For the descriptions of some attributes, I kept the description from 'master' since I know there were a few corrections.
 * Since I kept the `pcompress` framework from `master` I commented out the `PMIX_VALUE_COMPRESSED_STRING_UNPACK` macro in `pmix_common.h.in` (it conflicts with the definition in `pcompress`). I don't see it in the v3 standard, and I don't think anyone uses it.
 * Otherwise my basic testing came back fine along with CI testing.

rhc54:
I fixed this by removing that .m4 file from the v3.2 branch - will be in the v3.2.2 release.

PR #1833, opened 2020-08-08
build: Support hwloc and libevent cobuild

bwbarrett:
Add ability for PMIx to find hwloc and libevent packages which are
built at the same time as PMIx, such as third party dependencies
in an MPI package.  This is a slight change on the existing
embedded PMIx support, in that it does not assume that PMIx itself
is being built embedded in another library, only that some of its
dependencies may be built at the same time as PMIx.

### DIFF
--- a/config/pmix_setup_hwloc.m4
+++ b/config/pmix_setup_hwloc.m4
@@ -3,6 +3,8 @@
 # Copyright (c) 2009-2015 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2013      Los Alamos National Security, LLC.  All rights reserved.
 # Copyright (c) 2013-2020 Intel, Inc.  All rights reserved.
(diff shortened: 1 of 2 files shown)


bwbarrett:
> @bwbarrett I have no issue with these changes. However, I would like to eliminate embedded mode once you have made the corresponding changes to OMPI as nobody else is embedding PMIx, and I would like to avoid anyone doing so in the future.

That makes sense.  I left both modes to make it easier to decouple updates to PMIx and OMPI.  Once the changes to use the cobuild code are in OMPI, we can update PMIx as well.

bwbarrett:
I can't see an obvious failure with IBM CI.  Any hints?

rhc54:
Yeah, just committed a fix that should resolve the problem - the timeout on the tests was just skirting the edge of failure.

rhc54:
bot:ibm:xvercheck:retest

PR #1846, opened 2020-08-22
Added support for ppc64le build on travis-ci.

sanjaymsh:
Signed-off-by: Sanjay-Mishra<snjkmr32@gmail.com>


sanjay-cpu:
I think, then we should go with this approach only(i.e make distcheck with enabled flag).

PR #1866, opened 2020-09-28
Complete support for compute_distance API

rhc54:
Add required device-related data type support. Add function pointers to
the ploc framework to get the cpuset and compute distances to devices.
Devise a distance computation method based on device depth, ensuring
that devices on remote packages show as further away.


rhc54:
Travis passed and is hung trying to update status again. PRRTE appears to be a chicken/egg issue

Issue #1803, opened 2020-06-16
PMIX_IMMEDIATE not working?


shanedsnyder:
Thank you for taking the time to submit an issue!

### What version of the PMIx Reference Library are you using? (e.g., v1.0, v2.1, git master @ hash, etc.)

master

### Describe how PMIx was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)

Spack

### Please describe the system on which you are running

Oak Ridge Summit

-----------------------------

## Details of the problem

I had been relying on the `PMIX_IMMEDIATE` attribute to `PMIx_Get()` to return an immediate error if some specific key is not in the local kv store. This had been working in the past, but now appears to result in the same blocking behavior you'd expect if you passed no attributes at all. I know this was working in the past, but this code path is admittedly not tested that frequently, so I can't pinpoint a specific version that was working easily.

BTW, this is all tested using PRRTE master branch --- let me know if the issue is a better fit for that repo.

rhc54:
Are you sure you are using the right attribute?

- `PMIX_OPTIONAL` will return right away if the key is not currently available to the client.
- `PMIX_IMMEDIATE` goes up to the local server to request it - it should then return if the key isn't there. It should not request the info from the local host.

Neither one should block, however, so that is likely a bug.

rhc54:
@shanedsnyder I tried but was unable to replicate this problem. See my test here: https://github.com/openpmix/prrte/pull/612

Can you provide some additional info on your use-case - e.g., the particular key you were requesting?

shanedsnyder:
Sorry for the delay, this kind of fell through the cracks.

As for whether I'm mistakenly using the wrong attribute, that's possible. The intent of this code path is to check whether the given process has already put a particular key in the database. If the key isn't there, then we use some logic to create the corresponding value then add the key to the database. So basically just want to ensure that we aren't blocking for a key that maybe just hasn't been put yet, if that makes sense. Do I want PMIX_OPTIONAL then? I was thinking that PMIX_IMMEDIATE would essentially just return with an error if the key hadn't already been put in the local database, which I think is along the lines of what I'm describing.

For reference, here is the code that I'm using:

```
    snprintf(key, 512, "ssg-%s-%d-id", proc.nspace, proc.rank);
    PMIX_INFO_CREATE(info, 1);
    flag = true;
    PMIX_INFO_LOAD(info, PMIX_IMMEDIATE, &flag, PMIX_BOOL);
[...]

PR #1877, opened 2020-10-01
Fix "immediate" flag behavior at the server

rhc54:
Eliminate confusion over return status from the "defer" function between
when the "immediate" flag was given versus creation of a new tracker.

Fixes https://github.com/openpmix/openpmix/issues/1803


rhc54:
@jjhursey Probably needs to be backported to v3.2

PR #1883, opened 2020-10-08
Fixed uninitialized data bug in pmi_base64_encode_block

cpshereda:
This code doesn't appear to be utilized under normal build conditions in the master branch, but possibly the function is used elsewhere in other branches? 

I discussed it briefly with Artem and Howard, maybe we can talk about it on the weekly call to see if it is still used somewhere.

### DIFF
--- a/contrib/perf_tools/pmi2_utils.c
+++ b/contrib/perf_tools/pmi2_utils.c
@@ -58,9 +58,14 @@ static inline unsigned char pmi_base64_decsym (unsigned char value) {
 
 static inline void pmi_base64_encode_block (const unsigned char in[3], char out[4], int len) {
     out[0] = pmi_base64_encsym (in[0] >> 2);
-    out[1] = pmi_base64_encsym (((in[0] & 0x03) << 4) | ((in[1] & 0xf0) >> 4));
+
+    /* len is the length of in[] - we need to make sure we don't reference uninitialized data, hence the conditionals */
+    out[1] = 1 < len ? pmi_base64_encsym(((in[0] & 0x03) << 4) | ((in[1] & 0xf0) >> 4)) :
+                       pmi_base64_encsym((in[0] & 0x03) << 4);
     /* Cray PMI doesn't allow = in PMI attributes so pad with spaces */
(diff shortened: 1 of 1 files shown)


rhc54:
The PMI-1 and PMI-2 codes have been removed from PMIx - you'll find them over in their own repository: https://github.com/openpmix/pmi-shim. We no longer support the backward compatibility interfaces.

artpol84:
@rhc54 the fix is targeting base64 implementation that is in the PMIx codebase.
Should we we fix it in PMIx anyway? In case this functionality will be used elsewhere?

rhc54:
I'm not sure what code you are referring to - we removed the PMI-1 and PMI-2 codes, so there is nothing in PMIx that "encodes" things. I see some lingering code in the `test_v2` directory that appears to target the older PMI versions, but that should be moved to the new repo as those APIs no longer exist in PMIx.

artpol84:
Ahh, I didn't realize @cpshereda is patching contrib/perf_tools/pmi2_utils.c, so it's a contrib stuff, not the main codebase.

However, this is still an issue for main codebases:
https://github.com/open-mpi/ompi/blob/v4.0.x/opal/mca/pmix/base/pmix_base_fns.c#L666

artpol84:
btw, `test_v2` is using base64 for things unrelated to PMI1/2.
So we'll keep it there

artpol84:
This one is good to go as it fixes the perf tool

rhc54:
>However, this is still an issue for main codebases:

Only for OMPI - not for PMIx. I don't know if they will want to patch it or not. I'd suggest asking their release managers.

> btw, test_v2 is using base64 for things unrelated to PMI1/2.

For what purpose?? We don't encode/decode anything in PMIx

cpshereda:
> For what purpose?? We don't encode/decode anything in PMIx

We have a validation side channel in the test suite code that passes validation information as a command line argument. We use base64 encoding/decoding with this side channel.

It's because of this use in the test suite that I discovered the bug, we had a memory problem when I scaled it up. Since I got the original base64 routines from the stuff in contrib, I submitted the PR for that original code location. The test suite version is also fixed now but I haven't opened a PR only for that.

PR #1884, opened 2020-10-09
Checkout new shmem

rhc54:


artpol84:
@rhc54 are you going to provide the details?
I don’t remember discussing this before and it looks intrusive.
I recall we agreed that dstore will remain on par with the new shmem component.

rhc54:
We actually have been discussing the need to revamp the shared memory system for going on two years now, but it has never risen high enough on anyone's priority list to get attention. With v4, however, it has reached the point where something had to be done - the code to work around the existing dstore implementations is running into the couple of hundreds of lines now, and it is getting rather unmanageable.

Assuming this prototype proves out, it'll greatly simplify both the PMIx_Get code path and the shmem implementation, plus allowing us to finally provide pointers to the data in memory instead of copying the values into each proc. We'll lose the current dstore components, but will maintain backward compatibility via the "hash" component - so we'll meet our obligation as every proc has to have the "hash" component anyway.

I plan to write-up the details on the implementation once the prototype tests out. It's based on the HWLOC shmem implementation - Brice may have some writeup of it around the HWLOC site. We'll discuss it in the devel meetings over the next week or two - I don't expect to complete the work before then.

artpol84:
Thank you. I’ll provide some of my concerns below. You can either answer inline or we can discuss on the call.

First, I agree that workarounds are the bad thing. So +1 to get rid of them.

My first concern is that we are introducing a whole new thing and removing all the existing stuff without any overlapping releases where both types are available.

I have some thoughts on how to extend dstore to support extended job information (application, node etc) without much of complexity increase in the existing code base. Can we consider keeping dstore if we can fix that? I.e let’s not delete the backend code but rather have it disabled until/if we will extend it. If we will decide not to - then we’ll remove it. 

The other concern is the basis of this new method. IIRC you assume that you can guess a free space in the VM space. I recall that you said Brice did the investigation, but I’d like to have some documented proof that this will work on a broad set of  systems, I.e. having address randomization enabled.

Questions based on our experience with dstore:
* how will it grow when new data arrives?
* how the possible race condition between server publishing direct modex data and clients accessing unrelated keys will be handled?

rhc54:
> Thank you. I’ll provide some of my concerns below. You can either answer inline or we can discuss on the call.
> 
> First, I agree that workarounds are the bad thing. So +1 to get rid of them.
> 
> My first concern is that we are introducing a whole new thing and removing all the existing stuff without any overlapping releases where both types are available.
> 
It is rather difficult to retain the current dstore with the new code else we would be unable to remove all the workarounds. Only way would be to move all the workaround code into the dstore components, which introduces other issues beyond the simple problem of trying to move that code - e.g., we would have to come up with a method for the component to indicate that it cannot support the request and tell us to use the hash method without introducing that complication into the main code (or else the workarounds just reappear in another form).

> I have some thoughts on how to extend dstore to support extended job information (application, node etc) without much of complexity increase in the existing code base. Can we consider keeping dstore if we can fix that? I.e let’s not delete the backend code but rather have it disabled until/if we will extend it. If we will decide not to - then we’ll remove it.
> 
I have no problem if someone wants to invest more time in the dstore area, but is that really practical? In addition to implementing the array support, you'd also have to change your storage format to allow us to return a pointer to data. At the end, you'll have to modify a significant amount of the code and the result will still not be backward-compatible - I'm not sure there is much advantage to it.

One can always carry the current dstore components in a branch, or recover them at any time - they are retained in the github repository's history. I deleted them for code cleanliness as we wouldn't include inoperable code in a release, and it just seemed simpler to do so as we continue to prep for v4.

> The other concern is the basis of this new method. IIRC you assume that you can guess a free space in the VM space. I recall that you said Brice did the investigation, but I’d like to have some documented proof that this will work on a broad set of systems, I.e. having address randomization enabled.
> 
Probably the best documentation is that we have been distributing and using it in OMPI for better than two years, with no reports of any problems. If you do the math, it's pretty easy to see why. An executable binary occupies a few megabytes of the 16 exabyte virtual memory address space, so there are gigantic holes available. The only real risk is that someone does a "malloc" prior to calling `PMIx_Init` and the operating system unfortunately assigns them an address that falls within the virtual address space we are using. It's a very low probability, but not zero. However, in that case, we simply fall back to the `hash` component, so there is no interruption of service to the user.

Granted, there is some impact in the memory footprint in the fall back case, but given the improbable nature of that scenario, it seems a reasonable answer. Remember, HPC programmers are encouraged to call "init" early in their programs to ensure that any dynamic memory assignments remain local to their process - which makes it even more unlikely to see the fall back scenario.

> Questions based on our experience with dstore:
> 
>     * how will it grow when new data arrives?
> 
It doesn't - we simply allocate a new "hole" and pass that address to the client. The key to the new method is that we re-use the hash code - the only difference between the `shmem` and `hash` components is that the first one builds its hash table in the mmap'd region, while the second one builds it in the general OS-assigned malloc areas. So when we pass a new "hole" address, all the `shmem` component does is add another "head" of the hash tables to its search path.

The side benefit is that both components share the common hash table code - there literally will be a single code base that does the storage and fetch. Much less code to maintain, and allows for a degree of backward-compatibility if/when we need to extend the objects (e.g., re-designing what goes into which hash table, or adding finer separation of where data gets stored) as we aren't explicitly counting memory locations.

Again, if the client cannot add the new "hole" due to some conflict, then we fall back to the `hash` component and transfer the data down.

>     * how the possible race condition between server publishing direct modex data and clients accessing unrelated keys will be handled?

Direct modex data is sent to the client's `hash` component - no point in stuffing it into shared memory. Frankly, it isn't clear that putting full modex data into shared memory is worth doing. For one, libfabric has its own shmem implementation for sharing the address vector, so there is little point in having two of them. I know you folks are committed to UCX, but I imagine you'll have the same support at some point soon if you don't already have it - competition between the two libraries will force it, I'm sure.

In addition, the modex seems destined for a fairly short future. Slingshot definitely won't use it, the libfabric folks are in final stages of adding support for it in general (Amazon contributing the code), etc. So I'm not yet convinced of the value in expending significant effort in modex shared memory support.

Still, the current plan is to drop full modex data into shared memory, passing a new address to the client, falling back to passing the data in the hash tables if the client cannot map the address.

HTH - we can talk more at the devel meeting.

PR #1892, opened 2020-10-26
Complete slingshot scaffold

rhc54:
Cleanup checks for curl and jansson support. Ensure they properly get
included in the common/sse component and cleanup a few errors there.
Extend the pcompress framework to compress blocks of data instead of
just strings. Allow the pnet/sshot component to build when curl and
jansson support is available and the user specifically requests it to be
built by adding --with-slingshot. Extend the regex support to include
the "raw" option. Add PMIX_COMPRESSED_BYTE_OBJECT data type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -174,6 +174,7 @@ test/sshot/daemon
 test/sshot/generate
 test/sshot/server
 test/sshot/mytopo.json
+test/sshot/testcoord
 
 # coverity
 cov-int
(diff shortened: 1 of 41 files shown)


rhc54:
This scales relatively well, but could still use some scrubbing. For 10k nodes with 8 NICs/node and 8ppn, it takes about 2 sec to completely load the data on a backend compute node. Not bad and still much faster than doing a modex, but it would be nice to cut that down if possible. I'll leave further optimization to others.

rhc54:
Travis is hung, and Cross-version is failing on 3.2->3.2 indicating a problem on the 3.2 branch.

PR #1894, opened 2020-10-27
Locally cache VNI and traffic class

rhc54:
Allow host to access those values


rhc54:
Looks like something has changed on Travis that blocks it for some of the setups in one of our Travis checks - hangs in queuing the job.

PR #1911, opened 2020-11-01
construct_dictionary: fix VPATH usage

jsquyres:
Use argv.argv[0] to find the top-level OpenPMIx source directory (no
need to chdir anywhere).  Also clearly distinguish between
reading/writing to the source tree vs. the build tree.

Also, properly handle the --dryrun case (i.e., don't try to
os.replace() in the --dryrun case).

Should fix https://github.com/open-mpi/ompi/pull/8145


contrib/construct_dictionary.py
```diff
@@ -220,79 +221,98 @@ def main():
...
-        print("DICTIONARY COULD NOT BE CONSTRUCTED")
-        os.chdir(save)
-        return 1
-
-    # return to original directory
-    os.chdir(save)
+    # Write to the source tree
+    if outpath:
+        try:
+            target = os.path.join(src_include_dir, "dictionary.h")
```
jsquyres, line 311: @rhc54 This seems wrong, for two reasons:

1. We should not be writing to the source tree.
2. Writing to the build tree after the build seems like a contradiction.  This script is invoked during `make install`, after PMIx is built... but then we go back and re-write `src/include/dictionary.h`.  Shouldn't `dictionary.h` be written before the build?  And if `dictionary.h` is written at build time, shouldn't it be in the build tree?  And if it should be in the build tree, should it not be in the source tree?

rhc54 (approved):
One thing at a time 😄

PR #1912, opened 2020-11-02
Move construction of dictionary to build process

rhc54:
Construct the attribute dictionary during the "make" procedure, but only
when being built from Git clone - i.e., add the result to the tarball so
that typical users don't require Python to build PMIx.


rhc54:
Travis hang again

PR #1916, opened 2020-11-03
construct_dictionary: remove Python compat stuff

### LINKED ISSUE #1913: compile: syntax error in src/common/pmix_attributes.c
## Background information

Compiler error in src/common/pmix_attributes.c

github master @ b5c99f17a1d4c7e78cf68ca570fe888a2fe2e922

git clone

* Operating system/version: Fedora 32
* Computer hardware: x86
* Network type: ethernet

-----------------------------

## Details of the problem

```
../../pmix-git/src/common/pmix_attributes.c:28:1: error: expected ‘}’ before ‘static’
   28 | static bool initialized = false;
      | ^~~~~~
In file included from ../../pmix-git/src/common/pmix_attributes.c:26:
/home/arcurtis/src/pmix/pmix-git/src/include/dictionary.h:946:5: note: to match this ‘{’
  946 |     {.name = "PMIX_FABRIC_DEVICE_STATE", .string = "pmix.fabdev.state"
      |     ^
```

jsquyres:
The signal handler somehow was inhibiting some of the output to
dictionary.h when run with Python 3.8.x.  I don't fully understand why
-- even when no SIGINT occurred/was caught, dictionary.h would be
truncated somewhere in the middle.  Weird.

That being said, the signal handler also didn't serve much purpose in
this script, so just remove it.  In my tests, this made dictionary.h
always be fully emitted properly.

Fixes #1913.  Thanks to @tonycurtis for bringing this to our attention.


jsquyres:
@tonycurtis master/HEAD currently does not invoke `construct_dictionary.py` -- can you check out this PR branch and try invoking `contrib/construct_dictionary.py` manually again and see if it emits the entire `src/include/dictionary.h`?

I confess to not grokking why the signal handler in `construct_dictionary.py` would inhibit the full output of `dictionary.h`, but when I removed the signal handler, it seemed to consistently work properly for me.

tonycurtis:
> On Nov 3, 2020, at 1:08 PM, Jeff Squyres <notifications@github.com> wrote:
> 
> 
> @tonycurtis <https://github.com/tonycurtis> master/HEAD currently does not invoke construct_dictionary.py -- can you check out this PR branch and try invoking contrib/construct_dictionary.py manually again and see if it emits the entire src/include/dictionary.h?
> 
> I confess to not grokking why the signal handler in construct_dictionary.py would inhibit the full output of dictionary.h, but when I removed the signal handler, it seemed to consistently work properly for me.
> 

Python2 still good, but python3 now stops at:

  1862	     .description = (char *[]){"***** DEPRECATED *****", NULL}},
  1863
  1864	    {.name = "PMIX_ALLOC_NETWORK_QOS", .string = "pmix.alloc.netqos", .type = PMIX_STRING,
  1865	     .description = (char *[]){"***** DEPRECATED *****", NULL}},
  1866
  1867	    {.name = "PMIX_ALLOC_NETWORK_TYPE", .string = "pmix.alloc.nettype", .type = PMIX_STRING,
  1868	     .description = (char *[]){“

Tony

jsquyres:
@tonycurtis Thanks for your patience.  I looked deeper at the Python, and I have a better theory for what was going wrong (I think we were moving the file before it was closed/flushed).

Could you pull this branch and try running manually again?

tonycurtis:
that pr seems to be working (compiles).

but if I rm src/include/dictionary.h it doesn't get rebuilt during a make...

jsquyres:
> but if I rm src/include/dictionary.h it doesn't get rebuilt during a make...

Yes, we need to put that back in the `Makefile` (it was taken out while we were figuring this stuff out).

Thanks for your patience!  Let me amend this PR to put the "invoke `construct_dictionary.py` automatically" stuff back in...

tonycurtis:
OK!

contrib/construct_dictionary.py
```diff
@@ -9,20 +9,16 @@
 # by tools to interpret user input
 #
 
-from __future__ import print_function
```
rhc54, line 12: I'm pretty sure the config will now allow us to use Python 2.x, which means we need that "future" import, yes?

rhc54:
I just pushed a commit to restore the "future" import, remove the dictionary.h file from the repo, and remove the "python3" directive at the top of construct_dictionary.py

rhc54:
Added an error message.

@jsquyres I'm pretty sure we are hitting the "bad character" problem again when trying to read the input file. As far as I can tell, it is a bug in Python that is the problem. FWIW: the current code works fine for me.

You weren't seeing the error before because you had left "dictionary.h" in the repo - and so it wasn't running the construct_dictionary.py script in CI.

contrib/construct_dictionary.py
jsquyres: Nope -- we don't need this for `print()`.  I've tested the script with Python 2.6.6.

PR #1918, opened 2020-11-04
Eliminate the stale usock component

rhc54:
This breaks backward compatibility with the v1.x series, but we really
don't want anyone still using it anyway.


jjhursey (approved):
Per teleconf today, it should be fine to remove this for v4.
We just need to make sure that the release has a note about this backward compatibility break with v1 series - just in case we need to reference it in the future.

PR #1919, opened 2020-11-05
Plug a memory leak and optimize a bit

rhc54:
Eliminate an unnecessary malloc operation that caused a minor memory
leak every time we went thru PMIx_Get/fastpath

Thanks to @ekeever1 for finding it!

Refs https://github.com/open-mpi/ompi/issues/8166


jjhursey (approved):
This looks good to me. Thanks

PR #1922, opened 2020-11-06
Plug a memory leak and optimize a bit

jjhursey:
Eliminate an unnecessary malloc operation that caused a minor memory
leak every time we went thru PMIx_Get/fastpath

Thanks to @ekeever1 for finding it!


jjhursey:
`master` version of PR #1919

rhc54 (approved):
Thanks!

PR #1939, opened 2020-11-30
Fix dependency issue with hwloc

zhngaj:
Take hwloc as a dependency ONLY when hwloc is
NOT in embeded mode.

Signed-off-by: Jie Zhang <zhngaj@amazon.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -308,7 +308,7 @@ AC_MSG_RESULT([$LIBS])
 # Dependencies that themselves have a pkg-config file available.
 #
 PC_REQUIRES=""
-AS_IF([test "$pmix_hwloc_support_will_build" = "yes"],
+AS_IF([test "$pmix_hwloc_source" != "cobuild"],
       [PC_REQUIRES="$PC_REQUIRES hwloc"])
 AS_IF([test $pmix_libevent_support -eq 1],
       [PC_REQUIRES="$PC_REQUIRES libevent"])
--- a/configure.ac
+++ b/configure.ac
@@ -308,9 +308,9 @@ AC_MSG_RESULT([$LIBS])
 # Dependencies that themselves have a pkg-config file available.
 #
 PC_REQUIRES=""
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -308,7 +308,7 @@ AC_MSG_RESULT([$LIBS])
 # Dependencies that themselves have a pkg-config file available.
 #
 PC_REQUIRES=""
-AS_IF([test "$pmix_hwloc_support_will_build" = "yes"],
+AS_IF([test "$pmix_hwloc_source" != "cobuild"],
       [PC_REQUIRES="$PC_REQUIRES hwloc"])
 AS_IF([test $pmix_libevent_support -eq 1],
```
rhc54, line 313: Wouldn't we have the same issue with the internal libevent? IIRC, the integration is pretty close to an exact duplicate.

Or does the libevent m4 code already take that into account when setting this variable? PMIx _requires_ libevent, so it isn't like pmix_libevent_support can be anything other than 1. I'm okay with leaving this question for separate resolution - just wondering why we don't have the same issue here.

configure.ac
```diff
@@ -308,7 +308,7 @@ AC_MSG_RESULT([$LIBS])
 # Dependencies that themselves have a pkg-config file available.
 #
 PC_REQUIRES=""
-AS_IF([test "$pmix_hwloc_support_will_build" = "yes"],
+AS_IF([test "$pmix_hwloc_source" != "cobuild"],
```
rhc54, line 311: Probably want to keep the "will_build" check along with the "cobuild" one - we shouldn't mandate an hwloc dependency if we are not building support for it in PMIx

rhc54 (commented):
Couple of questions..

rhc54:
@zhngaj I pushed a commit that addresses my questions, so this is now good to go! Thanks!

zhngaj:
Sorry, rhc54@, did not check the thread in time. Thanks for your update.

PR #1864, opened 2020-09-23
SPEC add RPM *-devel packages

plesn:
This proposition adds *-devel RPMs containing headers and shared objects symbolic links when `build_all_in_one_rpm == 0`

With the libpmi split there is therefore 4 package in this situation: `pmix, pmix-devel, pmix-libpmi, pmix-libpmi-devel`

pmix-devel provides:
```
./opt/pmix/3.1.6/include
./opt/pmix/3.1.6/include/pmix.h
./opt/pmix/3.1.6/include/pmix_common.h
./opt/pmix/3.1.6/include/pmix_extend.h
./opt/pmix/3.1.6/include/pmix_rename.h
./opt/pmix/3.1.6/include/pmix_server.h
./opt/pmix/3.1.6/include/pmix_tool.h
./opt/pmix/3.1.6/include/pmix_version.h
./opt/pmix/3.1.6/lib/libmca_common_dstore.la
./opt/pmix/3.1.6/lib/libmca_common_dstore.so
./opt/pmix/3.1.6/lib/libpmix.la
./opt/pmix/3.1.6/lib/libpmix.so

```
In the same way
pmix provides only libpmix.so.* and libmca_common_dstore.so.* and no headers
pmix-libpmi provide libpmi{,2}.so.*
pmix-libpmi-devel provide libpmi{,2}.so pmi{,2}.h

This proposition is on v3.1, same can be done on master but for some reason on master `make_tarball` fails.

PR #1941, opened 2020-12-01
.spec: add *-devel packages

plesn:
Cherry pick of PR #1864 from v3.1 on v3.2.
Creates *-devel RPMs when `build_all_in_one_rpm == 0`. This is usefull for us as we move to v3.2.


rhc54 (approved):
I verified this is in master branch as well, so we are good to go. Thanks!

PR #1945, opened 2020-12-13
Update to utilize new tools functions

rhc54:
Provide a simplified method for indirect launch by tools, pushing much
of the interconnect handshake down into the library.


rhc54:
PRRTE is being updated to follow this patch.

PR #1953, opened 2020-12-18
Work on fixing Python bindings

rhc54:

### DIFF
--- a/bindings/python/pmix.pxi
+++ b/bindings/python/pmix.pxi
@@ -222,8 +222,10 @@ cdef int pmix_load_argv(char **keys, argv:list):
     for a in argv:
         pya = a
         if isinstance(a, str):
+            print("LOADARGV ENCODING")
             pya = a.encode('ascii')
         keys[n] = strdup(pya)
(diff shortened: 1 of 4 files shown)


bindings/python/pmix.pxi
```diff
@@ -222,8 +222,10 @@ cdef int pmix_load_argv(char **keys, argv:list):
     for a in argv:
         pya = a
         if isinstance(a, str):
+            print("LOADARGV ENCODING")
             pya = a.encode('ascii')
         keys[n] = strdup(pya)
+        print("LOADARGV ", n, pya)
```
jjhursey, line 228: It looks like a bit of debugging snuck in here.

rhc54: Yeah, Danielle left a lot of debugging statements in this code. I'm not worrying about it for now and even will be adding a few more to keep chasing down the problems - will remove all of them once we fully debug the bindings.

jjhursey: Sounds good.

PR #1943, opened 2020-12-03
Test to get job info for other ranks + custom rank placement

cpshereda:
This PR is the implementation of https://github.com/openpmix/pmix-tests/issues/66.

It includes a new client test to get job information for all ranks, plus the ability to place ranks on servers (nodes) in a user-specified configuration.

Also included is a commit with bug fixes to prior tests.

PR #1960, opened 2020-12-21
Bug fixes to Makefile, pmix_test.c, and all client tests.

cpshereda:
Enables tests to build correctly.

### DIFF
--- a/test/test_v2/Makefile
+++ b/test/test_v2/Makefile
@@ -1,6 +1,7 @@
-INSTALLDIR = $(HOME)/pmix-install
-INCLUDES = -I. -I../../test -I../../include -I../.. -I../../src/include
-vpath %.h .:../../test:../../include:../..:../../src/include
+#Fix below to be your local openpmix installation
+INSTALLDIR = /opt/cps/pmix-install
+INCLUDES = -I. -I$(INSTALLDIR)/include -I../../include -I../.. -I../../src/include
+vpath %.h .:$(INSTALLDIR)/include:../../include:../..:../../src/include
 LDIR = $(INSTALLDIR)/lib
 
 CC = gcc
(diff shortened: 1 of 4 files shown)


cpshereda:
This PR splits out the bugfix that's present in 1943 so that we can open an issue.

artpol84 (approved):

test/test_v2/Makefile
```diff
@@ -1,6 +1,7 @@
-INSTALLDIR = $(HOME)/pmix-install
-INCLUDES = -I. -I../../test -I../../include -I../.. -I../../src/include
-vpath %.h .:../../test:../../include:../..:../../src/include
+#Fix below to be your local openpmix installation
+INSTALLDIR = /opt/cps/pmix-install
```
jjhursey, line 2: @cpshereda Do you really mean to hardcode this to `/opt/cps/pmix-install`? This seems incorrect to me.

Could you change it to
```
INSTALLDIR ?= $(HOME)/pmix-install
```

Then you can redefine it via an envar:
```
shell$ INSTALLDIR=/my/path/pmix-install make
```

cpshereda: You're right... it is fixed in the way you describe in the subsequent commit (PR #1943) but we separated this one out so that I could quickly open an issue that required a bugfix to this suite, and I didn't backport the change.

Anyway, I'm fine with force-pushing a changed version that has the same Makefile as PR #1943, so I'll do that and you can decide what next.

cpshereda: @jjhursey The fix is there. I'm going to open another PR for this.

cpshereda: @jjhursey Not sure how to handle this. Right now I have a force-pushed version of the same branch with a modified commit. We can do something with that, or I can add a commit, but will that then make merging PR #1943 more difficult? 

The reason for the rush on this was to enable me to open an issue that required whoever looks at it (Ralph?) to build the test suite.

jjhursey: We can just fix in PR #1943 - I have a comment [on that PR](https://github.com/openpmix/openpmix/pull/1943#discussion_r547321971) with a minor change request.

Thanks!

PR #1967, opened 2020-12-23
Update v3.2 with event fixes

rhc54:


q-p:
I think this improved matters but didn't entirely fix them: With this PMIx version (the internal one in Open-MPI 4.0.6rc2) I can still produce hangs when multiple independent MPI executables are running in parallel on the same machine, one of which pretty much only does `MPI_Init` and `Finalize` (i.e. it does next to no work in-between them and doesn't use any other MPI methods).

It always seems to be this lone "quick" binary which — occasionally — ends up hanging on shutdown

```
Thread 1 (Thread 0x7f483efe7a40 (LWP 20198)):
#0  0x00007f4839ca0017 in pthread_join () from /lib64/libpthread.so.0
#1  0x00007f4838f08890 in opal_thread_join (t=0x226ca68, thr_return=0x0) at threads/thread.c:74
#2  0x00007f4838f07f3c in stop_progress_engine (trk=0x226c990) at runtime/opal_progress_threads.c:120
#3  0x00007f4838f08448 in opal_progress_thread_finalize (name=0x7f48391003b0 "OPAL-wide async progress thread") at runtime/opal_progress_threads.c:228
#4  0x00007f4839602229 in rte_finalize () at ess_singleton_module.c:511
#5  0x00007f48395ad609 in orte_finalize () at runtime/orte_finalize.c:77
#6  0x00007f483ddc8166 in ompi_mpi_finalize () at runtime/ompi_mpi_finalize.c:478
#7  0x00007f483ddfed38 in PMPI_Finalize () at pfinalize.c:54
#8  0x00007f483e9988e0 in main (argc=<optimized out>, argv=<optimized out>) at ../infrastructure/test/TestMain.cpp:73
#9  0x00007f48398eb555 in __libc_start_main () from /lib64/libc.so.6
#10 0x00000000004169ee in _start ()
```

and the progress thread in
```
Thread 9 (Thread 0x7f4827fff700 (LWP 20275)):
#0  pmix_notify_check_range (rng=0x7f4839376c08 <pmix_globals+776>, proc=0x2378eec) at event/pmix_event_notification.c:1237
#1  0x00007f48390447c9 in progress_local_event_hdlr (status=0, results=0x0, nresults=0, cbfunc=0x0, thiscbdata=0x0, notification_cbdata=0x2378e30) at event/pmix_event_notification.c:441
#2  0x00007f4838f93d0b in process_event (sd=-1, args=4, cbdata=0x2379bb0) at pmix3x.c:274
#3  0x00007f4838f72100 in event_process_active_single_queue (base=0x226cfe0, activeq=0x226bff0) at event.c:1370
#4  0x00007f4838f72374 in event_process_active (base=0x226cfe0) at event.c:1440
#5  0x00007f4838f729c7 in opal_libevent2022_event_base_loop (base=0x226cfe0, flags=1) at event.c:1644
#6  0x00007f4838f07eee in progress_engine (obj=0x226ca68) at runtime/opal_progress_threads.c:105
#7  0x00007f4839c9eea5 in start_thread () from /lib64/libpthread.so.0
#8  0x00007f48399c79fd in clone () from /lib64/libc.so.6
```

When I then manually kill the binary in question it (or one of the other MPI-using binaries that managed to run and completely successfully) seems to have output (it is being captured for unit tests)

```
      [some_host.dlr.de:21587] pmix_mca_base_component_repository_open: unable to open mca_pnet_tcp: perhaps a missing symbol, or compiled for a different version of OpenPMIx (ignored)
      [some_host.dlr.de:21587] pmix_mca_base_component_repository_open: unable to open mca_pnet_test: perhaps a missing symbol, or compiled for a different version of OpenPMIx (ignored)
```
at some point probably leading up to the hang. But it might be from one of the "succeeding" binaries, not the hanging one.

When I run all my (unit test) binaries one by one, I don't see a hang; it seems related to multiple (independent) mpi processes starting up / finalizing in an overlapping manner.

PR #1971, opened 2020-12-24
Complete revision of Python bindings

### LINKED ISSUE #1965: PMIx_Get of PMIX_NODEID, PMIX_HOSTNAME fails when using hash instead of dstore

up-to-date git master, installed from clone

Problem appears to be hw-independent; however, at present, test suite is run on single-node Linux without an RM.

-----------------------------

## Details of the problem

Build the test_v2 test suite from the current master (in test/test_v2):
 -To do so, you will have to first modify the Makefile in that directory to point to your local installation of pmix (this is handled better in a subsequent commit, sorry).

Assuming bash, from test_v2 directory:
```shell
shell$ make
shell$ export PMIX_MCA_gds_base_verbose=100
shell$ ./pmix_test -n 1 -v -e test_get_basic
```
(lots of output, uses dstore, should succeed)
Then,
```shell
shell$ export PMIX_MCA_gds=hash
shell$ ./pmix_test -n 1 -v -e test_get_basic
...
==17714== [1608661904.085563] ERROR [test_get_basic.c:61:main]: Client ns smoky_nspace-0 rank 0: PMIx call failed: NOT-FOUND
...
```
(fails at the Get of PMIX_NODEID)

rhc54:
Use the queue class to threadshift all server-level responses instead of
doing a timer delay

-----------------------

Let hash search for node/app info on specified proc
If the process has been specified in a PMIx_Get and its node or app
information is requested, then search the proc-level data to get it.

Fixes #1965

-----------------------

Ensure Python progress thread doesn't return early

-----------------------

### DIFF
--- a/bindings/python/pmix.pxi
+++ b/bindings/python/pmix.pxi
@@ -92,49 +92,9 @@ cdef void iofhdlr_cache(capsule, ret):
     cdef pmix_pyshift_t *shifter
     shifter = <pmix_pyshift_t*>PyCapsule_GetPointer(capsule, "iofhdlr_cache")
     pyiofhandler(shifter[0].idx, shifter[0].channel, &shifter[0].source,
-            shifter[0].payload, shifter[0].info, shifter[0].ndata)
+                 shifter[0].payload, shifter[0].info, shifter[0].ndata)
     if 0 < shifter[0].ndata:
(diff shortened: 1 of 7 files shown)


rhc54:
bot:ibm:xvertool:retest

PR #1973, opened 2020-12-26
Update the client get per v4.0 Standard

rhc54:
If the process rank is specified and node/app info is requested, then if
they didn't provide the hostname or appnum, assume they are looking for
info on the node/app to which that proc belongs


rhc54:
bot:mellanox:retest

PR #1981, opened 2020-12-27
Revert the changes to client_get

rhc54:
Defer to v4.1 release

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -351,9 +351,6 @@ PMIX_EXPORT pmix_status_t PMIx_Get_nb(const pmix_proc_t *proc, const pmix_key_t 
                         PMIX_VALUE_GET_NUMBER(rc, &info[n].value, appnum, uint32_t);
                         if (PMIX_SUCCESS != rc) {
                             PMIX_ERROR_LOG(rc);
-                            if (copy) {
-                                PMIX_INFO_FREE(iptr, nfo);
-                            }
                             return rc;
                         }
                     }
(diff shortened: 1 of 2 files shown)


rhc54:
Not sure I understand - do you mean "why didn't we revert these changes in the newer branches"? It has been a long time, but I believe the reason is that we wanted those changes in the newer releases. I suspect it was reverted here because there are corresponding changes in the newer branches that are too impactful to bring back to v4.0.

Is there some motivation for the question?

awlauria:
sorry - I meant to say these changes were reverted in the v4.0 branch, which is fine.

But the commit message made it seem like it was intended to go into v4.1 and beyond, and it doesn't appear these commits (not the reverts) actually did. I was wondering if there's a reason for that.

I'm asking because I'm finding backporting fixes to be riddled with conflicts from master, and I noticed a series of old commits that haven't found their way to release branches.

rhc54:
I see. Yes, backporting from master to the v4.x series is challenging. I'm afraid there is no easy solution. The primary reason is that the v4.x series continues to include the `dstore` subsystem which, due to its limitations, requires that the `gds` store/fetch code be left untouched in those series. For example, the logic in `pmix_client_get.c` (the concern of this PR) is vastly different from that in the master branch - for a very good reason.

So you cannot necessarily bring changes that touch anything in the store/fetch arena back to the v4.x branches. Unfortunately, that also means being careful when touching the other API code chains as they often interact with the `gds` subsystem - after all, PMIx is pretty focused on providing data, so you'd expect the APIs to interact with the data store.

Configure logic, updating utilities or the `ptl` subsystem - these are all easy to do. Anything like what we've been working on recently (especially the changes to the fetch logic over the last few days) is going to have to be very carefully scrutinized and is likely to not make the transition. If you are hitting conflicts, it should be considered a "red flag" that you might be touching areas of concern.

awlauria:
Makes sense - thanks

PR #1983, opened 2020-12-28
Step 1: remove dstore

rhc54:
Step 1: remove dstore

Take a stepwise approach to a new shmem implementation by first removing
the current dstore. This is done in one standalone commit to allow for
simple restoration if needed.

----------------------

Only test once for non-existent key to speedup the test

----------------------


rhc54:
Will resolve cross-version issues in Step 2 where we address the client_get code

PR #1985, opened 2020-12-29
Step 2: update client_get and hash component

rhc54:
Update client_get function and the gds/hash component to support v4. Fix
a few miscellaneous issues.

Add malloc option to support getting memory from the OS as well as from
an alternate memory pool. We will use this later for the new shmem
component.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -76,13 +76,10 @@ date="Mar 10, 2018"
 # format.
 
 libpmix_so_version=0:0:0
-libpmi_so_version=0:0:0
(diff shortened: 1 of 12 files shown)


rhc54:
This PR now fixes the cross-version:Client test (master, of course, continues to fail)

rhc54:
This appears to fix most of the problems. Still to resolve are the cases where v2.2 or v3.1 are the server. Let's commit this for now so things become a little clearer.

PR #1988, opened 2020-12-30
Add support for realloc to memory allocator

rhc54:


rhc54:
Progress! Done with "master" as server - still have a couple of problems when early versions are the server

PR #1991, opened 2021-01-01
Fix server/get for gds=hash

rhc54:
The client/get function does not pass the requested key to the server in pre-v3.2 versions. This means that the server in those versions doesn't know what key it needs to respond with. Currently, it only provided job-level info for procs in a different namespace, and provided no mechanism for retrieving proc-posted data from the other namespace.

It's a trivial fix that is required to provide forward-compatibility - i.e., so that clients of v3.2 and above can get proc-posted data from different namespaces when only the gds/hash is enabled. This is an edge case, but it becomes the default (and only) case when working with clients from v4.1 and above, so let's fix it in each series.

(cherry picked from commit d8792e13991868ab3f7d8f5db54593d1fd59a9c6)


rhc54:
xversion of master will be fixed shortly

PR #1992, opened 2021-01-01
Remove incorrect special case

rhc54:


rhc54:
master fails xversion without this PR

PR #1999, opened 2021-01-05
finalize: set hostname in global to NULL

hppritcha:
otherwise applications that try to init/finalize pmix multiple times segfault
in second call to pmix_rte_finalize.


hppritcha:
see https://gist.github.com/hppritcha/c78ccdfb48db573bad29aa8e4ce9e2eb for an example test case.

rhc54 (approved):

PR #2002, opened 2021-01-07
Add the PMIx Standard version level to the version string

jjhursey:
* PMIx Standard version can be adjusted in the VERSION file
 * `OpenPMIx 4.1.0a1 (PMIx Standard: 4.0)`

### DIFF
--- a/.github/actions/mlnx/entrypoint.sh
+++ b/.github/actions/mlnx/entrypoint.sh
@@ -45,7 +45,7 @@ build_dir=${work_dir}/build
 rpm_dir=${work_dir}/rpms
 tarball_dir=${work_dir}/tarball
 
-pmix_ver=`cat VERSION | grep "major=" | cut -d "=" -f 2``cat VERSION | grep "minor=" | cut -d "=" -f 2`
+pmix_ver=`cat VERSION | grep "^major=" | cut -d "=" -f 2``cat VERSION | grep "^minor=" | cut -d "=" -f 2`
 
 make_opt="-j$(nproc)"
 
--- a/VERSION
+++ b/VERSION
@@ -25,12 +25,16 @@ release=0
 
 greek=a1
 
+# PMIx Standard Compliance Level
+std_major=4
+std_minor=0
+
 # If repo_rev is empty, then the repository version number will be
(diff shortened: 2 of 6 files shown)


jjhursey:
The Mellanox CI seems to be failing all of the tests with an error like:
```
++ '[' '4
41
0' -ge 31 ']'
/entrypoint.sh: line 94: [: 4
41
0: integer expression expected
```
Without being able to see the output file it is parsing I'm not sure what might be going on here. It might be related to the version string - it's hard to tell.

Can @amaslenn or someone from Mellanox (@artpol84 do you know the right person to flag here?) take a look?

jjhursey:
Per teleconf - let's see if we can add the git hash to the output on git builds - not tarball.

jjhursey:
I added the git revision to the output string for git checked out builds. The string is generated from `git describe` inside of the `config/pmix_get_version.sh` script.

 * For git repos it looks like:
   - `OpenPMIx 4.1.0a1, repo rev: v1.1.3-2710-gc3f68e00 (PMIx Standard: 4.0)`
 * For non git repos (e.g., tarballs) it looks like:
   - `OpenPMIx 4.1.0a1 (PMIx Standard: 4.0)`

jjhursey:
I poked around at the Mellanox script a bit, and think I know what the problem is. It's the evaluation of this:
 * https://github.com/openpmix/openpmix/blob/master/.github/actions/mlnx/entrypoint.sh#L48

On master without this PR it evaluates to `41` with this PR it evaluates to `4 41 0` because we defined `std_major` and `std_minor` which are picked up by this expression. If we change the expression per below then we are fine:
```diff
- pmix_ver=`cat VERSION | grep "major=" | cut -d "=" -f 2``cat VERSION | grep "minor=" | cut -d "=" -f 2`
+ pmix_ver=`cat VERSION | grep "^major=" | cut -d "=" -f 2``cat VERSION | grep "^minor=" | cut -d "=" -f 2`
```

I just pushed that change to this PR which should resolve the issue.

abouteiller (approved):

rhc54 (approved):

PR #2004, opened 2021-01-08
Fix Python binding build for VPATH

### LINKED ISSUE #2003: python bindings seem confused for 4.0.0 build
This is on Debian Unstable Linux, with pmix 4.0.0

For reference, we do two builds - a static build to get .a files, and then a shared build - shared is the default used by applications, but the .a libs are there for users to build optimised.

This means we do builds in subdirectories, `debian/shared-build `and   `debian/static-build`.
This works but when the `--enable-python-bindings` flag is used, it fails, as it presumes that 
`$TOPDIR/include` == `$BUILDDIR/include".

This is fixed in current Debian builds with a hack that copies all the contents of the include files to each other, between the config and build stages.

It would be good to get a fix for this included in any 4.0.1 release.

jjhursey:
* We need to include both the autogenerated include directory
   (in the VPATH) and the static include directory (in src/include)
   when building the python bindings.
 * Fixes #2003

### DIFF
--- a/README
+++ b/README
@@ -12,7 +12,7 @@ Copyright (c) 2006-2020 Cisco Systems, Inc.  All rights reserved.
 Copyright (c) 2006-2011 Mellanox Technologies. All rights reserved.
 Copyright (c) 2006-2012 Oracle and/or its affiliates.  All rights reserved.
 Copyright (c) 2007      Myricom, Inc.  All rights reserved.
-Copyright (c) 2008      IBM Corporation.  All rights reserved.
+Copyright (c) 2008-2021 IBM Corporation.  All rights reserved.
 Copyright (c) 2010      Oak Ridge National Labs.  All rights reserved.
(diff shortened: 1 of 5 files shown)


rhc54 (commented):
Only thing I'm unsure of is the README change. You should only need to install either the rpm's or the pip3 - you shouldn't need to install both of them. I've never had to do both. Is there some reason to imply that one must do both of these?

jjhursey:
That's what I had to do to get it to work on a RHEL 7.6 box. The RPM wasn't enough, I needed to do the pip3 as well. I figured I would document it somewhere.

rhc54 (approved):
Okay - it shouldn't do any harm to install both, so no problem with including it.

rhc54:
@jjhursey Oh, don't forget to cherry-pick this over to v4.0.1 - and thanks for doing this!

PR #2007, opened 2021-01-09
Disable IPv6 by default

rhc54:
We are getting errors from the "bind" function on ARM machines due to IPv6 interfaces. We don't normally see those, and typically we disable them by default - but PMIx didn't set that default for some reason. Correct that here.

If someone cares about IPv6 support, they will need/want to further debug this behavior

(cherry picked from commit 7427abfa4f8c704de5514490abfdf2f7eb823701)


rhc54:
bot:ibm:xvertool:retest

PR #2019, opened 2021-01-13
Restore the pnet/opa component

rhc54:
Refs https://github.com/open-mpi/ompi/pull/8359

### DIFF
--- a/src/mca/pnet/opa/Makefile.am
+++ b/src/mca/pnet/opa/Makefile.am
@@ -1,0 +1,60 @@
+# -*- makefile -*-
+#
+# Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
+#                         University Research and Technology
+#                         Corporation.  All rights reserved.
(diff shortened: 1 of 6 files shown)


hppritcha:
@rhc54 looks interesting.   What is this opa management software that the configure.m4 is looking for?  Is that a released software package?  I don't see a /usr/include/opamgt on our OPA HFI1 clusters.

rhc54:
At one time, I had started to implement the remaining PMIx fabric feature set for OPA - providing the fabric topology, traffic reports, etc. This requires access to the fabric manager, which is what the `opamgt` package contains. It is distributed as a standalone rpms (`opa-libopamgt` and `opa-libopamgt-devel`).

Note that I didn't restore that portion of the plugin as it wasn't complete - I only restored the part that provides the authentication key.

rhc54:
Oh, I should have said - note that the configure.m4 doesn't require `opamgt`. If it isn't present, it just ignores it (and would have turned off that portion of the plugin)

PR #2034, opened 2021-01-25
Added multiserver (> 2) capability to test suite

cpshereda:
Last set of changes required to permit new test suite to run with more than 2 servers, plus a few other updates.


test/test_v2/test_init_fin.c
```diff
@@ -21,6 +21,11 @@ int main(int argc, char *argv[]) {
     test_params params;
     validation_params v_params;
 
+/*
```
artpol84, line 24: Lets make it an option

artpol84 (approved):

PR #2041, opened 2021-01-29
Various HWLOC support changes plus one OFI config fix

rhc54:
Provide an envar version of the HWLOC shmem rndvz info
Allow lower-level libs to attach to the HWLOC shmem region without requiring them to link against PMIx.

-----------------------

Fix bugs in OFI configure and HWLOC component
Protect against --with-ofi being "yes". Add missing function call to hwloc component to get shared memory size

-----------------------

### DIFF
--- a/config/pmix_check_ofi.m4
+++ b/config/pmix_check_ofi.m4
@@ -112,7 +112,7 @@ AC_DEFUN([_PMIX_CHECK_OFI],[
            AS_IF([test ! -z "$with_ofi_libdir" && \
                          test "$with_ofi_libdir" != "yes"],
                  [pmix_ofi_libdir=$with_ofi_libdir],
-                 [if test ! -z "$with_ofi"; then
+                 [if test ! -z "$with_ofi" && test "$with_ofi" != "yes"; then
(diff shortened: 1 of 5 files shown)


jjhursey:
We should create a little code snippet that shows how a library would use the new envars - something that just initializes hwloc properly. That would make it easier for those impacted libraries to make the necessary code adjustments. The nice thing about using envars here is that it is easy to test against for RMs that might not provide them (yet).

The standard would be a good place once this is brought over there. In the mean-time maybe we can create a FAQ item (either on the wiki or website) with the code snippet that we can point folks to.

jjhursey:
This was filed as https://github.com/pmix/pmix-standard/pull/333 but later withdrawn. Should we remove the code from OpenPMIx or are folks still experimenting with it?

rhc54:
> This was filed as [pmix/pmix-standard#333](https://github.com/pmix/pmix-standard/pull/333) but later withdrawn. Should we remove the code from OpenPMIx or are folks still experimenting with it?

Done in https://github.com/openpmix/openpmix/pull/2165

PR #2042, opened 2021-01-29
Fix various PGI warnings.

awlauria:


awlauria:
PMIx side of https://github.com/open-mpi/ompi/issues/7686

rhc54 (approved):
Wow - I can pretty much guarantee we won't avoid generating more warnings in the future if PGI is going to be that pedantic!

Appreciate your patch - just wish there was a way to make it a one-off, but I don't see how.

jjhursey (approved):
Thanks for cleaning this up!

rhc54:
Travis is overwhelmed.

rhc54:
@awlauria Unfortunately, this blew up on OMPI CI under clang with lots of errors like this:
```text
/home/ubuntu/workspace/open-mpi.build.compilers/Compiler/clang38/src/3rd-party/openpmix/src/atomics/sys/atomic_stdc.h:123:1: error: address argument to atomic operation must be a pointer to _Atomic type ('int32_t *' (aka 'int *') invalid)
PMIX_ATOMIC_STDC_DEFINE_FETCH_OP(or, 32, int32_t, |)
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/ubuntu/workspace/open-mpi.build.compilers/Compiler/clang38/src/3rd-party/openpmix/src/atomics/sys/atomic_stdc.h:112:16: note: expanded from macro 'PMIX_ATOMIC_STDC_DEFINE_FETCH_OP'
        return atomic_fetch_ ## op ## _explicit ((type *) addr, value, memory_order_relaxed) operator value; \
               ^                                 ~~~~~~~~~~~~~
<scratch space>:105:1: note: expanded from here
atomic_fetch_or_explicit
^
/usr/lib/llvm-3.8/bin/../lib/clang/3.8.0/include/stdatomic.h:153:34: note: expanded from macro 'atomic_fetch_or_explicit'
#define atomic_fetch_or_explicit __c11_atomic_fetch_or
```
Any ideas what the heck is wrong?

rhc54:
Sadly, I encountered the same errors when trying to compile on the Mac - not a surprise as that is also clang. I had no choice but to revert this commit. However, I was able to reapply all the changes _except_ those made to the atomics as those didn't have a problem (https://github.com/openpmix/openpmix/pull/2048)

So you'll just need to revisit the atomics, I'm afraid.

awlauria:
@rhc54 I'll take a look.

awlauria:
@rhc54 it looks like clang has their own builtin atomics functions that take a volatile void *, whereas the gcc and pgi compilers don't take a volatile.

I opened up https://github.com/openpmix/openpmix/pull/2049 in an attempt to fix it.

rhc54:
Nice! Of course clang would have to be difficult - we always seem to hit peculiarities there.

Thanks for doing this!

PR #2043, opened 2021-01-29
Fix configury so that PMIx is compiled with -O3 by default.

awlauria:


awlauria:
Just browsing the code, it looks like this is also in the v3 series. Should it get cherry-picked to v3.1/.2?

rhc54:
> Should it get cherry-picked to v3.1/.2?

Definitely 3.2 - wouldn't hurt to do 3.1 as well, if you don't mind and have time. Thx!

rhc54:
Don't worry about v4.0 - we have a bunch of stuff that needs to go over there, and I'll just add this to the list.

awlauria:
Done. Thanks!

awlauria:
I actually tracked down the real culprit, since something didn't seem right, seeing as how PRRTe has the same configury logic and doesn't reproduce for me. Looks like there is an old developer-override line here:

```
#################### Early development override ####################
if test "$WANT_DEBUG" = "0" && test -z "$enable_debug" && test "$PMIX_DEVEL" = "1"; then
    WANT_DEBUG=1
    echo "--> developer override: enable debugging code by default"
fi
#################### Early development override ####################
```
it's only for builds brought down in a `git clone`. I can force push to this PR if you still want it removed, but not sure if there's still value with that.

awlauria:
If not, feel free to close these PR's.

rhc54:
Let me ponder that a bit. Not sure of the right answer.

awlauria:
I updated the PR if you want to take it. If not, no worries. Thanks

rhc54:
My hesitation is that the OMPI community has always refrained from dictating a particular level of optimization. The right level depends strongly upon the application and compiler, and it is nearly impossible to pick something that will work well in all cases. So we always said that it was up to the user to specify an optimization level, if desired.

This seems to violate that understanding, and I'm not sure that's a good thing. Need to discuss that with folks a bit more.

As for removing the default debug build for git checkouts - I don't recall if that is still the norm in OMPI or not. I'll have to check. I have no strong feelings about it.

awlauria:
Building ompi and prrte from scratch give me the -O3, in my experience. That's why PMIx caught my eye, and caused me to go down the hole on this. I checked PRRTE, and this logic isn't in the configury that I can see.

This shouldn't be happening for release builds, since the override looks for a .git directory to see if it's a developer build or not. So tarballs and releases _should_ always be built with -O3.

rhc54:
Hmm...digging a little deeper, and checking the history in my email repos, I think I see what happened. There was a bit of a tussle over performance comparisons against MPICH - they were taking OMPI "out-of-the-box" using "git clone" vs a highly optimized version of their lib...and we were getting trashed. So back in 2008, the default was changed to be optimized.

We also apparently set it to "-O3", but _only_ for gcc. All other compilers default to "-O" as a more conservative approach as we otherwise ran into all those problems I mentioned.

PMIx isn't a performant lib, but I have no problem with conforming to OMPI and PRRTE.

PR #2045, opened 2021-01-29
v3.2: Fix configury so PMIx is compiled with -O3 by default.

awlauria:
(cherry picked from commit fcb79a6)

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -1018,15 +1018,11 @@ else
     AC_MSG_RESULT([no])
     WANT_DEBUG=0
 fi
-#################### Early development override ####################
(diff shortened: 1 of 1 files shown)


rhc54:
Why isn't this the same patch as https://github.com/openpmix/openpmix/pull/2043? The cherry-pick refers to some commit on your fork, not to something on this repo - I suspect things have changed since you originally filed this and it needs to be refreshed to match #2043?

awlauria:
Apologies - forgot to update the cherry-picks. They should be good now.

rhc54 (approved):
No worries - appreciate it!

PR #2046, opened 2021-01-29
Correctly handle precedence for first/last overall events

rhc54:
If someone specifies a first or last overall event and then registers an event relative to them, we need to correctly handle that both during registration and notification.


rhc54:
Travis successfully completed, but once again failed to update the status.

PR #2051, opened 2021-02-02
Fix clang compliler regression.

### LINKED ISSUE #2050: Atomics broken under Clang
Tip of master when brought over to OMPI:

/home/ubuntu/workspace/open-mpi.build.compilers/Compiler/clang38/src/3rd-party/openpmix/src/atomics/sys/atomic_stdc.h:250:5: error: member reference base type 'volatile void' is not a structure or union
    atomic_flag_clear ((volatile void *) lock);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/lib/llvm-3.8/bin/../lib/clang/3.8.0/include/stdatomic.h:181:63: note: expanded from macro 'atomic_flag_clear'
#define atomic_flag_clear(object) __c11_atomic_store(&(object)->_Value, 0, __ATOMIC_SEQ_CST)

awlauria:
Make special exception for Clang built-in C11 atomics.

Fixes #2050.

Regression caused by #2049.

### DIFF
--- a/src/atomics/sys/atomic_stdc.h
+++ b/src/atomics/sys/atomic_stdc.h
@@ -228,13 +228,21 @@ typedef atomic_flag pmix_atomic_lock_t;
 static inline void pmix_atomic_lock_init (pmix_atomic_lock_t *lock, bool value)
 {
     (void)value;
+#ifdef PMIX_HAVE_CLANG_BUILTIN_ATOMIC_C11_FUNC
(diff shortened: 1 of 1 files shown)


awlauria:
Verified it builds correctly with pgcc v 20.11-0 and clang 9.0.1

PR #2053, opened 2021-02-03
Fix check for IBM xl compilers for v13.1 and later.

awlauria:
By default newer xlc compilers only define __ibmxl__ now.

https://www.ibm.com/support/knowledgecenter/en/SSXVZZ_13.1.6/com.ibm.xlcpp1316.lelinux.doc/compiler_ref/xlmacros.html


rhc54:
Unrelated error

PR #2059, opened 2021-02-05
configury: fix _PMIX_CHECK_PACKAGE_LIB() macro

ggouaillardet:
use the correct name for the cached variable
(so it can be properly un-cached)

This is a backport from open-mpi/ompi@7da50433f799c98de82e2963d2fb5ea5d21663fb


rhc54:
huh - looks like something is out-of-sync with PRRTE. Nothing to do with this PR, though - I'll investigate separately.

PR #2062, opened 2021-02-09
Remove PMIX_BUFFER datatype and extend macro definition

rhc54:

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -1289,7 +1289,7 @@ typedef uint16_t pmix_data_type_t;
 #define PMIX_APP                        23
 #define PMIX_INFO                       24
 #define PMIX_PDATA                      25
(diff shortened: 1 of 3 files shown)


jjhursey:
I'm looking at this a bit more closely and don't understand why it is deprecating `PMIX_BUFFER`. It is still used in the openpmix codebase in some of the client and server code (though all of those instances might be protected with `V1` peer checks). Is there a better replacement for it?

rhc54:
PMIX_BUFFER isn't a publicly exposed data type - it is internal to the library. So the datatype doesn't belong in the public-facing header files as there is no corresponding definition of the structure.

We do continue to use the buffer datatype internally. Externally, the `pmix_data_buffer_t` is the corresponding public-facing definition, which is why that replaced `PMIX_BUFFER` in the `pmix_value_t` and other places.

PR #2074, opened 2021-02-16
Validate CFLAGS individually.

awlauria:
- Came across this when trying to debug other xlc compiler
  issues. The old configury logic would only set certain flags
  if the compiler vendor was labeled as 'gnu'. Ie - any non-gcc
  build would have far fewer CFLAGS compared to a gcc build.

  This change cropped up due to this commit:
  a2b2712d368c258782cbcf083742822ffe0dbc35
  which changed the way vendors are labeled. Before pretty much
  all vendors were labeled as gnu,
  Now, however, the vendor may be labeled as 'gnu'/'ibm'/
  'portland group'/ect. This change will validate and set CFLAGS
  for any compiler instead of assuming they are only valid for gcc.

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -17,6 +17,8 @@
 #                         reserved.
 # Copyright (c) 2016      Intel, Inc. All rights reserved.
 # Copyright (c) 2021      Nanook Consulting.  All rights reserved.
+# Copyright (c) 2021      IBM Corporation.  All rights reserved.
+#
 # $COPYRIGHT$
 #
 # Additional copyrights may follow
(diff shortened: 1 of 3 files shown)


awlauria:
Came up when trying to figure out this ompi issue:
https://github.com/open-mpi/ompi/pull/8458
if it looks good I'll propagate to ompi/prrte as well.

rhc54:
It looks okay to me, but I'm no expert on some of these things.

rhc54:
@awlauria Please let me know when you think this is ready and I'll commit - bringing it over to PRRTE makes sense too.

awlauria:
@rhc54 thanks. Should be done shortly. Just checking a few things.

awlauria:
@rhc54 pushed a minor change. I think this is good to go

awlauria:
bot:mellanox:retest

awlauria:
bot:mlnx:retest

awlauria:
retest

awlauria:
bot:retest

awlauria:
I can't seem to trigger a mlnx re-test.

rhc54:
bot:mellanox:retest

rhc54:
it can take quite awhile to circle around their queue

rhc54:
let's just assume that Mellanox has gone belly-up again

Issue #2079, opened 2021-03-02
Python bindings fail on install


jjhursey:
I'm running with `master` and discovered the following build issue with the Python bindings.

```
./configure --prefix=/home/pmixer/tmp-build-CB5NR/install-pmix \
    --with-libevent=/home/pmixer/local/libevent --with-hwloc=/home/pmixer/local/hwloc \
    --enable-python-bindings
make -j 20
make -j 20 install
```

The `make install` will fail with this message:
```
make[2]: Entering directory `/home/pmixer/tmp-build-CB5NR/openpmix/bindings/python'
make[3]: Entering directory `/home/pmixer/tmp-build-CB5NR/openpmix/bindings/python'
/usr/bin/python3 ../../bindings/python/construct.py --src="../../include" --include-dir="../../include"
make[3]: Nothing to be done for `install-data-am'.
BAD DIMENSION 2
make[3]: *** [install-exec-local] Error 1
make[3]: Leaving directory `/home/pmixer/tmp-build-CB5NR/openpmix/bindings/python'
make[2]: *** [install-am] Error 2
make[2]: Leaving directory `/home/pmixer/tmp-build-CB5NR/openpmix/bindings/python'
make[1]: *** [install-recursive] Error 1
make[1]: Leaving directory `/home/pmixer/tmp-build-CB5NR/openpmix/bindings'
make: *** [install-recursive] Error 1
```

rhc54:
I found it - noodling over the fix

jjhursey:
Still failing
 * https://github.com/openpmix/openpmix/pull/2080#issuecomment-789133950

jjhursey:
Some notes:
 * Building with `Cython: yes (0.19)` failed
 * Building with `Cython: yes (0.29.21)` passed for Ralph, but failed with `(0.29.22)` for me.

Failure is:
```
/usr/bin/python3 ../../bindings/python/construct.py --src="../../include" --include-dir="../../include"
make[3]: Nothing to be done for `install-data-am'.
Traceback (most recent call last):
  File "../../bindings/python/construct.py", line 530, in <module>
    main()
  File "../../bindings/python/construct.py", line 516, in main
    harvest_constants(options, "pmix.h", constants, definitions)
  File "../../bindings/python/construct.py", line 30, in harvest_constants
    lines = inputfile.readlines()
  File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
[...]

PR #2080, opened 2021-03-02
Fix Python binding build

rhc54:
Replace array of char in pmix_proc_stats_t with a single
character as that is all that is actually required.

Fixes https://github.com/openpmix/openpmix/issues/2079

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -2073,7 +2073,7 @@ typedef struct pmix_proc_stats {
     pid_t pid;
     char *cmd;
     /* process stats */
-    char state[2];
+    char state;
     struct timeval time;
     float percent_cpu;
     int32_t priority;
--- a/src/mca/bfrops/base/bfrop_base_copy.c
+++ b/src/mca/bfrops/base/bfrop_base_copy.c
@@ -1275,7 +1275,7 @@ pmix_status_t pmix_bfrops_base_copy_pstats(pmix_proc_stats_t **dest,
     if (NULL != src->cmd) {
         p->cmd = strdup(src->cmd);
     }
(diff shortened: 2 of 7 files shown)


jjhursey:
Huh. I just checked out master and 'make install' has this error:
```
/usr/bin/python3 ../../bindings/python/construct.py --src="../../include" --include-dir="../../include"
make[3]: Nothing to be done for `install-data-am'.
  File "../../bindings/python/construct.py", line 530, in <module>
    main()
  File "../../bindings/python/construct.py", line 516, in main
    harvest_constants(options, "pmix.h", constants, definitions)
  File "../../bindings/python/construct.py", line 30, in harvest_constants
    lines = inputfile.readlines()
  File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 217: ordinal not in range(128)
```

rhc54:
What version of Python are you trying to build against?

rhc54:
Never mind - it saids 3.6, which is odd is that is what I use and I am not seeing any errors.

rhc54:
FWIW: here are my versions
```text
Cython: yes (0.29.21)
Python: yes (3.6.12)
```
Trying with fresh clone

rhc54:
No joy - builds clean. Afraid I'll have to pass this debug over to you as I can't make it fail 🤷‍♂️

jjhursey:
My Cython seems to be a bit older
```
Bindings
-----------------------
Cython: yes (0.19)
Python: yes (3.6.8)
```

PR #2082, opened 2021-03-03
Start bringing v4.0.x up to v4.0.1rc

rhc54:
Begin pulling cherry-picks across from master. This begins _after_ the dstore was removed from master, so we'll have to see how far we can get before things break down. We also need to decide how much of master to pull across. This PR brings over a couple of new macros, which seems innocent enough. However, it stops short of adding the PSTATS support as I'm not sure we want to go there in the v4.0 series.


rhc54:
I'm not sure if we are caught in an inbetween state here, or if there is something actually wrong. I'm loath to go in and fix things as that could quite likely cause conflicts with upcoming cherry-picks.

So I think what I'll do is just keep this PR and continue adding more CPs to it. Hopefully, things will magically work at some point. Otherwise, we can do spot fixes once we get to where we want to stop.

jjhursey:
For the `Special Builds` it's tripping on the Python fix which has yet to be cherry-picked. So I assume that will resolve itself. For PRRTE, I agree with waiting to spot fix once all the changes are brought over. I suspect it will resolve itself. No need to spend time on that right now.

jjhursey:
Test out timestamp with 'special builds'
bot:ibm:builds:retest

rhc54:
Replaced by #2100

PR #2085, opened 2021-03-03
Added basic fence test and timed wildcard fence test with no data exchange

cpshereda:
Also added functions to support this basic fence.


test/test_v2/test_clientapp_funcs.c
```diff
@@ -223,3 +223,29 @@ void pmixt_validate_predefined(pmix_proc_t *myproc, const pmix_key_t key,
...
+        }
+	}
+    gettimeofday(&local_end, NULL);
+    usecs = (local_end.tv_sec * 1000000) + local_end.tv_usec
+        - ((local_start.tv_sec * 1000000) + local_start.tv_usec);
+    avg_fence = ((double)usecs)/1E8;
+
+    TEST_VERBOSE(("Simple PMIx_Fence, avg time: %lf", avg_fence));
+	return avg_fence;
+}
```
jjhursey, line 251: The mix of tabs and spaces here bugs me. Can you clean that up? In general, we prefer spaces not tabs.

cpshereda: @jjhursey  Sure and good catch, I don't really know how that happened. I'm using VS Code and it should be defaulting to spaces. I'll fix it and look into why it arose.

test/test_v2/test_common.c
```diff
@@ -184,6 +184,8 @@ void parse_cmd(int argc, char **argv, test_params *params, validation_params *v_
             fprintf(stderr, "\t-t <>    set timeout\n");
             fprintf(stderr, "\t-o out   redirect clients logs to file out.<rank>\n");
             fprintf(stderr, "\t-d str   assign ranks to servers, for example: 0:0,1:1:2,3,4\n");
+            fprintf(stderr, "\t-m num   fence time multiplier (similar to an error bar)\n");
```
artpol84, line 187: This is misleading.
these options are specific to the fence test only.
We need to allow tests to pass the callback here that can parse test-specific items and append corresponding options to the --help output

artpol84: Also need an option to enable fence timing.
I suggest having it disabled by default as it requires a quiet system for stable results.

artpol84: so the launch cmd should be something like:
```
./pmix_test -s 4 -n 16 -d '0:0,1,3,5,7,9;1:2,4,6,8;2:10,12;3:11,13,14,15' -- ./$test <test args>
```
The pmix_test should take whatever comes after the "-" symbol (or whatever separator that is convenient) and treat it as an exec command. And it needs to append to it the parameters option provided by the pmix_test itself,

artpol84 (changes requested):

hppritcha:
I'd like to see this merged in soon irrespective of things like the help notes.  we can patch that up later.

cpshereda:
> I'd like to see this merged in soon irrespective of things like the help notes. we can patch that up later.

We discussed it today, I should be able to implement the help changes needed this week and push a revised commit.

test/test_v2/test_common.c
cpshereda: should work according to your specification now.

cpshereda:
All changes requested should now be implemented @hppritcha @artpol84 . I did leave the additional basic fence test case in the second commit (test_fence_basic). It includes some functionality that's not tested in the wildcard case.

Artem, will you review and merge?

Howard, I will update the CI script as well.

hppritcha (approved):

hppritcha:
this looks good to me merging.

PR #2097, opened 2021-03-12
Fix shadowing

awlauria:


src/mca/ploc/hwloc/ploc_hwloc.c
```diff
@@ -683,22 +689,8 @@ pmix_status_t setup_topology(pmix_info_t *info, size_t ninfo)
...
-                    if (end) {
-                       *end = '\0';
-                    }
-                    pmix_output(0, "%s", line);
-                }
-                fclose(file);
-            }
+            return ploc_hwloc_print_maps();
         }
-        return PMIX_SUCCESS;
```
jjhursey, line 701: Don't we still want to return `PMIX_SUCCESS` here in the case where verbosity is not set?

awlauria: Yes, this is right. Thanks!

Fixed it.

rhc54 (approved):
LGTM!

PR #2103, opened 2021-03-17
Fix leaks

awlauria:
There's more to come, only 3 tracked down as of now.

### DIFF
--- a/src/include/pmix_globals.c
+++ b/src/include/pmix_globals.c
@@ -154,6 +154,9 @@ static void ncdcon(pmix_nspace_caddy_t *p)
 static void ncddes(pmix_nspace_caddy_t *p)
 {
     if (NULL != p->ns) {
+        if(NULL != p->ns->jobbkt) {
+            PMIX_RELEASE(p->ns->jobbkt);
+        }
         PMIX_RELEASE(p->ns);
     }
 }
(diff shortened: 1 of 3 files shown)


src/server/pmix_server_ops.c
```diff
@@ -4771,6 +4771,14 @@ static void tdes(pmix_server_trkr_t *t)
     if (NULL != t->info) {
         PMIX_INFO_FREE(t->info, t->ninfo);
     }
+    pmix_nspace_caddy_t *nm, *nm_next;
```
rhc54, line 4774: Just wondering - would it make more sense to simply replace `PMIX_DESTRUCT(&t->nslist)` with `PMIX_LIST_DESTRUCT(&t->nslist)` here, and then add `PMIX_RELEASE(p->jobbkt)` to the `pmix_nspace_caddy_t` destructor on line 154 of src/include/pmix_globals.c? Seems to me like we always want to have these things removed/destructed.

awlauria: I like it, done.

rhc54:
The Mellanox failure is odd as that change passed before it was committed - it has nothing to do with this one. I'll try to replicate and see what is going on.

src/include/pmix_globals.c
```diff
@@ -154,6 +154,7 @@ static void ncdcon(pmix_nspace_caddy_t *p)
 static void ncddes(pmix_nspace_caddy_t *p)
 {
     if (NULL != p->ns) {
+        PMIX_RELEASE(p->ns->jobbkt);
```
rhc54, line 157: I'm not sure, but you may need to check for `p->ns->jobbkt` being `NULL` prior to trying to release it - probably safer to do so.

rhc54 (approved):

PR #2131, opened 2021-03-25
build: Allow autogen.pl to be run from a tarball

bwbarrett:
Add two files missing from dist tarballs which are needed for
autogen.pl to succeed when run in a dist tarball.


rhc54:
bot:ibm:prrte:retest

PR #2135, opened 2021-03-25
Enable static MCA components by default and remove unused MCA configure code

bwbarrett:
Default to a slightly nicer default on large systems of building MCA components into the base library, rather than having hundreds of stat/open calls during initialization.

Also, remove the unused direct all interface logic in pmix_mca.


NEWS
```diff
@@ -56,7 +58,8 @@ and macros, changes to the library include:
  - Detect/avoid conflict with LSF version of "libevent"
  - Auto-detect and forward envars from various identified programming models
    (e.g., OpenMPI, OpenSHMEM) and fabrics
-
+ - Change the default component build behavior to prefer building components
+   as partof libpmix.so instead of individual DSOs.
```
bwbarrett, line 62: sigh; will fix the silly typo.

PR #2137, opened 2021-03-27
Set the default for enable-mca-dso

rhc54:
Some components should always be built as dso's to allow
for situations where the libs aren't present on the backend.
Default to supporting that situation, allowing a user to
override it if they want everything static.


rhc54:
See discussion https://github.com/openpmix/prrte/issues/871

PR #2138, opened 2021-03-28
Tool finalize crash due to ref count issue

jjhursey:
* Found while debugging PRRTE's indirect.c
   - PMIx_tool_init followed by a PMIx_Spawn
   - PMIx_tool_set_server switches to the IL
   - PMIx_tool_finalize will crash when destructing the
     `pmix_globals.iof_requests` array because of a 0 ref count
     on the `p->requestor` pointer.
 * Determined that this is because when switching servers
   we do not PMIX_RETAIN the peer pointer that becomes `myserver`
   so it gets destructed in pmix_tool.c when it cleans up the
   `pmix_server_globals.clients` list before calling `pmix_rte_finalize`.
   - By retaining the object, it is then cleaned up properly in finalize.

### DIFF
--- a/src/tool/pmix_tool.c
+++ b/src/tool/pmix_tool.c
@@ -7,7 +7,7 @@
  *                         All rights reserved.
  * Copyright (c) 2016      Mellanox Technologies, Inc.
  *                         All rights reserved.
- * Copyright (c) 2016      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2016-2021 IBM Corporation.  All rights reserved.
  * Copyright (c) 2021      Nanook Consulting.  All rights reserved.
  * $COPYRIGHT$
  *
@@ -1953,6 +1953,7 @@ static void retry_set(int sd, short args, void *cbdata)
 
     /* switch the active server - we are in an event, so
      * it is safe to do so */
+    PMIX_RETAIN(peer);
     pmix_client_globals.myserver = peer;
     pmix_globals.connected = true;
 


jjhursey:
I noticed that in the `indirect.c` example from PRRTE the `PMIx_tool_finalize` was [commented out](https://github.com/openpmix/prrte/blob/40ce30cbfd98cc55d184090208808c30082093cb/examples/debugger/indirect.c#L481). By adding it back I was able to generate a core file during finalize which showed the following.

```
(gdb) bt
#0  0x00003fffa280fc30 in raise () from /lib64/libc.so.6
#1  0x00003fffa2811f8c in abort () from /lib64/libc.so.6
#2  0x00003fffa2805754 in __assert_fail_base () from /lib64/libc.so.6
#3  0x00003fffa2805844 in __assert_fail () from /lib64/libc.so.6
#4  0x00003fffa2dc1294 in iofreqdes (p=0x3fff9c0057f0) at pmix_globals.c:264
#5  0x00003fffa2c3c714 in pmix_obj_run_destructors (object=0x3fff9c0057f0)
    at /smpi_dev/jjhursey/dev/pmix/pmix-master/src/class/pmix_object.h:553
#6  0x00003fffa2c3d31c in pmix_rte_finalize () at runtime/pmix_finalize.c:138
#7  0x00003fffa2c515e4 in PMIx_tool_finalize () at tool/pmix_tool.c:1579
#8  0x0000000010003eb8 in main (argc=6, argv=0x3fffda8e1158) at indirect.c:485
(gdb) up
#1  0x00003fffa2811f8c in abort () from /lib64/libc.so.6
(gdb) 
#2  0x00003fffa2805754 in __assert_fail_base () from /lib64/libc.so.6
(gdb) 
#3  0x00003fffa2805844 in __assert_fail () from /lib64/libc.so.6
(gdb) 
#4  0x00003fffa2dc1294 in iofreqdes (p=0x3fff9c0057f0) at pmix_globals.c:264
264	        PMIX_RELEASE(p->requestor);
```

Digging around a bit I found it releasing the IL peer early even though we made it 'myserver' in `PMIx_tool_set_server`.

I won't claim this fixes all of the references around this server transition, but it does seem to fix this one.

rhc54 (approved):

PR #2143, opened 2021-03-30
ptl: prevent free of uninited suri variable

acolinisi:
This fixes an assert failure in the memory allocator at `free(suri)`
when running:

	pterm

There is a path from the start of that function to the free() that
leaves suri uninitialized: with none of the connection identifiers set,
through the clause

	else if (1 < pmix_list_get_size(&connections))

Signed-off-by: Alexei Colin <acolin@isi.edu>


rhc54 (approved):
Thx!

PR #2149, opened 2021-03-31
Some mods/cleanup of debugger definitions and handling

rhc54:
The host/debugger needs a way to communicate which ranks are to
be debugged - eg., which are to be stopped on exec. We had been
assuming that it would be all of them as that is the typical
case, but that isn't a restriction. Modify the debugger attrs
to pass a rank (or an array of ranks) instead of just "bool".
Allow the app to indicate where they are stopped, should they
choose to do so. Replace `WAIT_FOR_NOTIFY` with `READY_FOR_DEBUG`
as release by notification is probably the least-used use-case.

### DIFF
--- a/examples/client.c
+++ b/examples/client.c
@@ -173,7 +173,7 @@ int main(int argc, char **argv)
      * directive is provided so that something like an MPI implementation
      * can do some initial setup in MPI_Init prior to pausing for the
      * debugger */
-    if (PMIX_SUCCESS == (rc = PMIx_Get(&proc, PMIX_DEBUG_WAIT_FOR_NOTIFY, NULL, 0, &val))) {
+    if (PMIX_SUCCESS == (rc = PMIx_Get(&proc, PMIX_DEBUG_STOP_IN_APP, NULL, 0, &val))) {
         /* register for debugger release */
         DEBUG_CONSTRUCT_LOCK(&mylock);
         PMIX_INFO_CREATE(info, 1);
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -684,9 +684,16 @@ typedef uint32_t pmix_rank_t;
 
 
 /* debugger attributes */
-#define PMIX_DEBUG_STOP_ON_EXEC             "pmix.dbg.exec"         // (bool) job is being spawned under debugger - instruct it to pause on start
-#define PMIX_DEBUG_STOP_IN_INIT             "pmix.dbg.init"         // (bool) instruct job to stop during PMIx init
-#define PMIX_DEBUG_WAIT_FOR_NOTIFY          "pmix.dbg.notify"       // (bool) block at desired point until receiving debugger release notification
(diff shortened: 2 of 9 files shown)


rhc54:
The PRRTE tests will be fixed by the corresponding commit over there.

jjhursey:
PMIx Standard changes in PR https://github.com/openpmix/openpmix/pull/2149/
 * Goal: Allow debugger a way to communicate which ranks are to be debugged (could be a subset)
 * `PMIX_DEBUG_STOP_ON_EXEC` type changed from `bool` to `pmix_rank_t/pmix_data_array_t`
   - `PMIX_RANK_WILDCARD` can be used to indicate all processes in the namespace
 * `PMIX_DEBUG_STOP_IN_INIT` type changed from `bool` to `pmix_rank_t/pmix_data_array_t`
   - `PMIX_RANK_WILDCARD` can be used to indicate all processes in the namespace
 * `PMIX_DEBUG_STOP_IN_APP` added with type `pmix_rank_t/pmix_data_array_t`
   - `PMIX_RANK_WILDCARD` can be used to indicate all processes in the namespace
 * `PMIX_BREAKPOINT (char*)` string ID of the breakpoint where the process(es) is/are waiting
   - For example, `"pmix-init"` set when `PMIX_DEBUG_STOP_IN_INIT`
   - (TODO) Missing event for the `PMIX_DEBUG_STOP_ON_EXEC` case
 * Deprecated `PMIX_DEBUG_WAITING_FOR_NOTIFY`
   - Replaced by `PMIX_READY_FOR_DEBUG`
 * Add `PMIX_READY_FOR_DEBUG` event indicating that all of the requested processes have stopped at the location required (e.g., `PMIX_DEBUG_STOP_IN_INIT`)
   - Accompanied by `PMIX_BREAKPOINT` info key indicating where the proc is waiting as a custom string
 - (TODO) The `client.c` example does not issue a `PMIx_Notify_event(PMIX_READY_FOR_DEBUG` when `PMIX_DEBUG_STOP_IN_APP` is detected

jjhursey:
Discussed these interface modifications on the PMIx Standard ASC call April 8, 2021:
 * Concern about changing the type of definition of `PMIX_DEBUG_STOP_ON_EXEC` and `PMIX_DEBUG_STOP_IN_INIT` for those tools that already targeted/released with the v4.0 versions.
 * Suggest:
   * Keeping the old definitions of `PMIX_DEBUG_STOP_ON_EXEC` and `PMIX_DEBUG_STOP_IN_INIT` as `bool`. Possibly start the deprecation process on them.
   * Create new `PMIX_DEBUG_STOP_ON_EXEC_LIST` and `PMIX_DEBUG_STOP_IN_INIT_LIST` versions with the `pmix_rank_t/pmix_data_array_t`.
   * All other modifications seem ok and will be proposed via either provisional or deprecated processes.
 * I have volunteered to make the change in OpenPMIx before we release, and propose the changes to the PMIx Standard in the 2Q 2021 meeting.

rhc54:
You might first wonder "has anyone actually implemented against those?" - because the answer is "NO".

PR #2151, opened 2021-04-01
Remove debug

rhc54:


rhc54:
@dwootton-ny @jjhursey We have a problem in the debugger CI. Your CI is looking at the debugger source codes in the "master" branch instead of in the PR. So if we do something that requires a change to the debugger examples, we cannot pass the CI. Could you please correct the CI script? I suspect this needs to be done in Jenkins?

drwootton:
@rhc54 I don't know anything about Jenkins and probably don't have access. I backed out the additional tests in the CI script until I can resolve the points in this pull request

PR #2153, opened 2021-04-01
atomics: Only support C11 and GCC builtin atomics

bwbarrett:
Remove the inline assembly and hand-crafted assembly version
of the atomics used in PMIx, always relying on either the
GCC-style builtin or C11 atomic operations.  All compilers
PMIx supports include support for one of the two, even if
at a slightly less than optimal performance, and this allows
for a big code removal.

Also remove the high resolution timer code, as it was unused
in PMIx.


src/atomics/sys/atomic_stdc.h
```diff
@@ -73,11 +75,11 @@ static inline void pmix_atomic_wmb(void)
 
 static inline void pmix_atomic_rmb(void)
 {
-#    if PMIX_ASSEMBLY_ARCH == PMIX_X86_64
+#    if defined(PMIX_ATOMIC_X86_64)
     /* work around a bug in older gcc versions (observed in gcc 6.x)
      * where acquire seems to get treated as a no-op instead of being
      * equivalent to __asm__ __volatile__("": : :"memory") on x86_64 */
-    pmix_atomic_mb();
+    __asm__ __volatile__("" : : : "memory");
```
bwbarrett, line 82: This might actually be a problem.  Curious to see what IBM tests think of this.  Also not sure how to fix it if they get angry.

rhc54:
@drwootton Can you please update the signatures? I've tried, but I can't seem to get them to match.

PR #2155, opened 2021-04-07
Fix some minor errors

rhc54:
Ensure we pass the desired scope on a PMIx_Get call.
Don't reinitialize nspace object on multiple calls to
PMIx_server_register_nspace
Allow the client to go up to the server when looking
for a newly assigned context ID


jjhursey:
Should this be cherry-picked to v4.0?

PR #2156, opened 2021-04-07
Fix some minor errors

rhc54:
Ensure we pass the desired scope on a PMIx_Get call.
Don't reinitialize nspace object on multiple calls to
PMIx_server_register_nspace
Allow the client to go up to the server when looking
for a newly assigned context ID

(cherry picked from commit 705ece510a2551bcca53ae2ef3a7721789daac78)


rhc54:
@jjhursey I'm not sure if this will really work with PRRTE's updated PMIx_Connect support. The server portion is fine, but the change to pmix_client_get.c is risky since all that dstore-circumvention logic is in there. Did my best but someone probably needs to check.

PR #2171, opened 2021-04-14
Attempt to fix the curl support

rhc54:
Replicate what OMPI does for mca/common by adding some
magic sauce. Add the mca/common/sse subdir to the top-level
makefile.am. Fix a typo in the curl m4


rhc54:
@bwbarrett When you get time, could you please take a look at this? I've got it to the point where the config.log shows that the mca/common/sse library should be built. However, I cannot find any of the symbols (either the ones from the sse code or curl symbols) in the resulting libpmix. I'm not sure if I'm doing something wrong or those symbols are just "hidden" somehow. Regardless, with this change `prte` no longer errors out with a missing curl symbol - not sure what that means.

rhc54:
bot:ibm:retest

PR #2177, opened 2021-04-23
Cleanup the v4.0.1 references and update to v4.1

rhc54:


rhc54:
master branch needs to be updated before it can pass cross-version tests

PR #2186, opened 2021-05-01
Silence some gcc11 warnings

rhc54:
It's a subtle point, but gcc11's warnings of potential
string overflows are technically valid. If someone were
to pass a string longer than PMIX_MAX_KEYLEN to one of
the four functions that took a pmix_key_t argument, then
that argument would overflow and impact the surrounding
memory. So go back to the v3.0 function prototypes as
they are technically safer.


rhc54:
bot:ibm:xver:retest

rhc54:
bot:ibm:xvertool:retest

rhc54:
bot:ibm:xvercheck:retest

PR #2191, opened 2021-05-06
Remove unnecessary shift in make_dist_tarball script

mike0042:
Signed-off-by: Michael Karo <mike0042@gmail.com>


rhc54:
Odd - I don't recall re-enabling Travis. We stopped using them a fair time ago. I'll look into disabling it again.

PR #2214, opened 2021-06-23
Transfer stdout/err formatting to PMIx

rhc54:
Transfer the PRRTE output formatting code to the PMIx library
so that tools can also provide a similar capability. This
includes the ability to output XML format, timestamps,
and tagged output as well as the ability to specify a
file format and/or directory where the output is to be
written.

Please see the OpenPMIx wiki page for explanation of
this feature and how to use it.


rhc54:
PRRTE failure expected - requires corresponding change in that repo.

rhc54:
LANL CI frozen trying to fetch repo

PR #2217, opened 2021-06-24
Initialize val before get in case get isn't successful

MattBBaker:
I was trying out the Python interface to PMIx and noticed this bug. It looks like if the `PMIx_Get()` code path encounters an error, than `val` never gets set and Python crashes with a 'use before assignment' error when trying to return `val`. This PR add an assignment to `val` so that it is always initialized before the return


rhc54:
Kewl - thx! Could you please add the "Signed-off-by: Name \<email\>" to the commit?

MattBBaker:
@rhc54 Updated the commit message

PR #2220, opened 2021-06-28
Some cleanups of the event notification and keepalive support

rhc54:
Ensure that the event notification chain stops when it should by
tracking the interim status returned by each handler. When the keepalive
pipe goes away, generate an event with a known named affected proc
so that the host can catch it.


rhc54:
@hppritcha @cpshereda Not sure what happened to your v2 test Makefile - it looks ok to me, but fails:
```
make: *** No rule to make target 'test_fence_basic.o', needed by 'test_fence_basic'.  Stop.
```

Issue #2142, opened 2021-03-29
prte fails to start due to missing "curl" symbols if PMIx configure --with-curl


rhc54:
The `src/mca/common/sse` component depends upon `libcurl` and only builds if `--with-curl` is given and a suitable `libcurl` found at the indicated (or standard) location. Once it is built, however, `prte` fails to run with an error:
```text
$ prte
dyld: Symbol not found: _curl_global_cleanup
  Referenced from: /Users/rhc/pmix/build/clean/lib/libpmix.0.dylib
  Expected in: flat namespace
 in /Users/rhc/pmix/build/clean/lib/libpmix.0.dylib
Abort trap: 6 (core dumped)
$
```

rhc54:
This still persists on both master and v4.0.1

rhc54:
@bwbarrett You can replicate this as stated above - just build PMIx `--with-curl` and have `libcurl` available. I did try a fix for it in https://github.com/openpmix/openpmix/pull/2171 but without success. Appreciate the assist!

bwbarrett:
Closing as the issue is fixed in master and common/sse is not in the v4.0 or v4.1 release branches.

PR #2223, opened 2021-06-30
common/sse: Ensure library links against libcurl

bwbarrett:
Unify on a single naming convention for the library produced for
the common/sse component, with libmca_common_sse.la.  Unlike
traditional components that are actual DSOs, common components
aren't dynamically loaded; they're dynamically linked by other
loaded components and are stand-alone shared libraries, so the
naming convention that is necessary for other components is
not necessary for common components.

This solves a bug with the common/sse component not being able
to find libcurl symbols when built as a static component.  The
*_LIBADD rule was only for the dynamic build mode and was missing
for the static build mode.

Refs #2142

### DIFF
--- a/src/mca/common/sse/Makefile.am
+++ b/src/mca/common/sse/Makefile.am
@@ -25,19 +25,19 @@ sources = \
 
 lib_LTLIBRARIES =
 noinst_LTLIBRARIES =
-comp_inst = libmca_common_sse.la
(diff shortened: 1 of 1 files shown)


rhc54 (approved):
Thanks Brian!

rhc54:
We are having known problems with the LANL CI, so no issue here.

PR #2224, opened 2021-06-30
common/dstore: Fix inconsistent Makefile.am

bwbarrett:
The Makefile.am for common/dstore was inconsistent in the
flags set for the static and dynamic build cases.  While
this was not causing any problems today, it could lead
to the problems seen with common/sse on the master branch.

This change is similar to 936bf79 on master, in that it
unifies on a single naming convention for the library produced
by common components.

At the same time, remove some unneeded flags, like the LTDLINC
CPPFLAGS, which are not used in this component.

### DIFF
--- a/src/mca/common/dstore/Makefile.am
+++ b/src/mca/common/dstore/Makefile.am
@@ -1,7 +1,9 @@
 #
 # Copyright (c) 2018      Mellanox Technologies.  All rights reserved.
-#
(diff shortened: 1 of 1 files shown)


bwbarrett:
This is not a cherry pick, as the common-dstore component is not on master.

rhc54 (approved):

PR #2227, opened 2021-07-01
Revert change to PMIx_Abort

rhc54:
Do not call "exit" from within the function - just return an "unreachable" error if the PMIx server is no longer available. Retain the changes to the "simptest" program.

### DIFF
--- a/src/client/pmix_client.c
+++ b/src/client/pmix_client.c
@@ -1061,8 +1061,7 @@ PMIX_EXPORT pmix_status_t PMIx_Finalize(const pmix_info_t info[], size_t ninfo)
     return PMIX_SUCCESS;
 }
 
-PMIX_EXPORT pmix_status_t PMIx_Abort(int status, const char msg[],
-                                     pmix_proc_t procs[], size_t nprocs)
+PMIX_EXPORT pmix_status_t PMIx_Abort(int flag, const char msg[], pmix_proc_t procs[], size_t nprocs)
 {
     pmix_buffer_t *bfr;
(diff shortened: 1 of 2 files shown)


rhc54:
@bwbarrett Here is the reversion we discussed. Let me know if you need/want me to make the corresponding change on the OMPI side.

rhc54:
LANL CI is still borked, so we are ignoring it for now.

rhc54:
@bwbarrett @abouteiller Based on today's discussion, I dug into the OMPI code a bit more and was surprised to find that `ompi_rte_abort` already calls `_exit` as it should. I then tried to follow the path upward to see where/how that gets called, and found it eventually leads to the predefined OMPI errhandlers.

I can therefore only surmise that for some reason those predefined errhandlers are not getting called when the MPI layer gets the "lost connection" event notification. I then dug further and found that:

* OMPI does _not_ register for a default event handler - this is a major mistake. You need to capture and check default events in case something goes wrong
* OMPI does register an event handler for the `PMIX_ERR_PROC_ABORTED` event in ompi_mpi_init.c - but that event is never being generated anywhere that I can find in OMPI, PMIx, or PRRTE. I'm not sure why anyone thought that event would be coming?
* OMPI also registers an event handler for `PMIX_DEBUGGER_RELEASE` in ompi_rte.c for use by debuggers. No help here.

I have a branch where I fix the problem by adding the `PMIX_ERR_LOST_CONNECTION` code to the event handler registration in ompi_mpi_init.c (will post it tomorrow). You may want to do a general review of event registration.

PR #2228, opened 2021-07-02
Some cleanups for client finalize and IOF output

rhc54:
Don't close the send/recv socket for a client when it
finalizes - let the socket close upon client termination
and do the cleanup at that time. Be a little better
on setting the IOF output default for a server. Write
IOF output directly to stdout/stderr instead of pusing
into events to avoid potential loss of output upon
termination.


rhc54:
Needs corresponding change in PRRTE

PR #2233, opened 2021-07-07
Improved handling of C bytes to Python objects

MattBBaker:
When using the Python bindings I noticed some strange behavior with namespaces and `PMIX_BYTE_OBJECT` values. I made some bug fixes to the Python bindings that makes my use case work, and it also seems to fix some of the tests. The `PMIX_BYTE_OBJECT` use case is fixed by leveraging Cython's type casting ability to convert C `char[]`'s into Python `bytes` classes and the namespace bug is fixed by similarly casting `char[]` into `bytes` and then decoding those into a Python `string`.

### DIFF
--- a/bindings/python/pmix.pxi
+++ b/bindings/python/pmix.pxi
@@ -1274,10 +1274,7 @@ cdef dict pmix_unload_value(const pmix_value_t *value):
         pyns = str(value[0].data.proc[0].nspace)
         return {'value':{'nspace':pyns, 'rank':value[0].data.proc[0].rank}, 'val_type':PMIX_PROC}
     elif PMIX_BYTE_OBJECT == value[0].type:
-        mybytes = <char*> PyMem_Malloc(value[0].data.bo.size)
-        if not mybytes:
-            return PMIX_ERR_NOMEM
-        memcpy(mybytes, value[0].data.bo.bytes, value[0].data.bo.size)
+        mybytes = <bytes>value[0].data.bo.bytes[:value[0].data.bo.size]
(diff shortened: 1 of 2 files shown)


bindings/python/pmix.pxi
```diff
@@ -1274,10 +1274,7 @@ cdef dict pmix_unload_value(const pmix_value_t *value):
         pyns = str(value[0].data.proc[0].nspace)
         return {'value':{'nspace':pyns, 'rank':value[0].data.proc[0].rank}, 'val_type':PMIX_PROC}
     elif PMIX_BYTE_OBJECT == value[0].type:
-        mybytes = <char*> PyMem_Malloc(value[0].data.bo.size)
-        if not mybytes:
-            return PMIX_ERR_NOMEM
-        memcpy(mybytes, value[0].data.bo.bytes, value[0].data.bo.size)
+        mybytes = <bytes>value[0].data.bo.bytes[:value[0].data.bo.size]
```
rhc54, line 1277: Is there an issue here with persistence? The bytes in the `value` struct are not guaranteed to remain upon return from a function call - will this automatically "copy" them?

MattBBaker: The Cython documentation says that this creates a new bytes object with a copy of the C string, https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#type-casting

rhc54: Perfect! Thx.

rhc54 (approved):

PR #2234, opened 2021-07-09
Allow servers to use PMIx_Spawn

rhc54:
Enable PMIx servers to call "spawn" so that the PMIx
server library can process directives - e.g., indicating
that IOF is to be sent to a particular target.


rhc54:
bot:ibm:prrte:retest

PR #2244, opened 2021-07-19
Add prelim support for MPICH

rhc54:
Harvest some envars, set a few on fork

### DIFF
--- a/src/mca/pmdl/mpich/pmdl_mpich_component.c
+++ b/src/mca/pmdl/mpich/pmdl_mpich_component.c
@@ -1,0 +1,96 @@
+/* -*- Mode: C; c-basic-offset:4 ; indent-tabs-mode:nil -*- */
+/*
+ * Copyright (c) 2004-2008 The Trustees of Indiana University and Indiana
+ *                         University Research and Technology
+ *                         Corporation.  All rights reserved.
+ * Copyright (c) 2004-2005 The University of Tennessee and The University
+ *                         of Tennessee Research Foundation.  All rights
+ *                         reserved.
+ * Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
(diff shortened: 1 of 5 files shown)


rhc54:
@raffenet @zhenggb72 @wesbland

I have made a first pass at adding MPICH support into PMIx - just harvesting/forwarding envars and setting some for each process. I'm not sure entirely what you folks support. I based this first cut on what I saw Hydra doing from your latest 4.0 alpha tarball.

Can you please take a look and tell me what I got wrong and/or what might be missing? I'm working next on Hydra-like support in PRRTE and should have something for you to review there soon. These will be too late for the next release (which is going to happen in the next week), but we can include them in a subsequent release in about a month.

raffenet:
Just acknowledging that I have seen this and will dig into it when I have some time. Thanks!

rhc54:
bot:ibm:prrte:retest

src/mca/pmdl/mpich/pmdl_mpich_component.c
```diff
@@ -0,0 +1,96 @@
...
+    },
+    .include = NULL,
+    .exclude = NULL
+};
+
+static pmix_status_t component_register(void)
+{
+    pmix_mca_base_component_t *component = &mca_pmdl_mpich_component.super.base;
+
+    mca_pmdl_mpich_component.incparms = "MPIR_*";
```
zhenggb72, line 68: I think it is safe to use "MPIR_CVAR" for all MPICH env vars.

rhc54: Good point. Question: Hydra has options about forwarding envars. It wasn't quite clear to me if those options apply to these envars, or were aimed at general envars other than the MPICH-specific ones. Any insights would be appreciated.

PR #2252, opened 2021-08-03
Misc Python fixes

MattBBaker:
There are threeish different bugs being fixed here. One is string decoding, similar to several previous fixes.

One of the biggest changes is changing some instances of `PyMem_Malloc/PyMem_Free` with the system `malloc/free`. The reason for this is that in some code paths, such as called in `PMIx_Spawn()` will lead to PMIx internally calling the `realloc()` function. When the libc `realloc()` function sees a pointer from `PyMem_Malloc()` it crashes the program. So this change will use the system malloc when allocating data structures that will be given to PMIx rather than being given to Python. 

The last change is declare the `pyeventhandler()` cb function with the `with gil` clause. Without it the Python interpreter won't set up the Python thread state and won't be able to access the Python datatypes inside of the callback since it is called in a C context.


rhc54 (approved):
Thanks! I'll roll to v4.1.1 when done with CI and committed.

PR #2264, opened 2021-09-02
Avoid blocking in the stdin read handler

rhc54:
There is a very tiny window of time where the host might be calling
down into the PMIx server library (which does a threadshift) and
the read handler attempts to call up into the host from within the
handler (which is in a PMIx progress thread event). If we hit that
window, the host may be blocked waiting for the PMIx progress thread
to make progress while the PMIx progress thread is waiting for the
host's progress thread to do something...and we threadlock.

So don't block in the read handler. In general, we don't want to
block anywhere in an event, so we probably need to review the code
to ensure this isn't happening elsewhere.


rhc54:
bot:ibm:prrte:retest

PR #2275, opened 2021-09-11
Add a Standard extension value to the compliance version

rhc54:
Since OpenPMIx is moving past the current official Standard,
begin tracking how far beyond it has gone by adding a simple
"std_extension" value to the current "std_major/minor"
values.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -26,8 +26,23 @@ release=0
 greek=a1
 
 # PMIx Standard Compliance Level
+# The major and minor numbers indicate the version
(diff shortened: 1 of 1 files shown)


jjhursey:
@rhc54 I was going to roll up a release of the PMIx v4.1 standard in the next couple of weeks. Probably an RC this week, then release shortly thereafter. Any reason to hold it from OpenPMix's perspective?

rhc54:
No, that sounds good!

PR #2283, opened 2021-09-13
man: Fix path check to see if pandoc is required

bwbarrett:

### DIFF
--- a/config/pmix_setup_man_pages.m4
+++ b/config/pmix_setup_man_pages.m4
@@ -79,7 +79,7 @@ AC_DEFUN([_PMIX_SETUP_PANDOC],[
                   AC_MSG_RESULT([no - adequate pandoc installation not found])
                   PANDOC=
                   PMIX_ENABLE_MAN_PAGES=0],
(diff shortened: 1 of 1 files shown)


rhc54 (approved):
Sigh - thx!

rhc54:
Hey @bwbarrett - call me confused, but when I look at the OMPI version of this, I see:
```
          [AS_IF([test ! -f "$srcdir/ompi/mpi/man/man5/MPI_T.5"],
```
So why is `srcdir` correct there, but not here? Is there something more fundamental wrong here?

bwbarrett:
> Hey @bwbarrett - call me confused, but when I look at the OMPI version of this, I see:
> 
> ```
>           [AS_IF([test ! -f "$srcdir/ompi/mpi/man/man5/MPI_T.5"],
> ```
> 
> So why is `srcdir` correct there, but not here? Is there something more fundamental wrong here?

The `$top_srcdir` part was me being overly cautious.  The important part is that both `$top_srcdir` and `$srcdir` point at the top level directory (at least, at that point in configure), so it was the missing `src/` in the path that caused the problem.

PR #2289, opened 2021-09-16
Added test_v2 to autoconf/automake processing

cpshereda:
Also some cleanup of v2 test suite code.


cpshereda:
@rhc54 Please let me know if the v2 tests build and work for you now with this update.

PR #2309, opened 2021-10-02
Enable ultra-picky compiler options

rhc54:
The picky compiler options have been turned "off" for
some time due to lots of warnings being converted to
errors. Do the cleanup necessary to re-enable those
options.

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -1060,6 +1060,9 @@ if test "$WANT_PICKY_COMPILER" = "0" && test -z "$enable_picky" && test "$PMIX_D
 fi
 #################### Early development override ####################
 
+AC_DEFINE_UNQUOTED(PMIX_PICKY_COMPILERS, $WANT_PICKY_COMPILER,
+                   [Whether or not we are using picky compiler settings])
+
 #
 # Developer debugging
 #
--- a/config/pmix_check_vendor.m4
+++ b/config/pmix_check_vendor.m4
@@ -15,6 +15,7 @@ dnl Copyright (c) 2013-2020 Intel, Inc.  All rights reserved.
 dnl Copyright (c) 2015      Cisco Systems, Inc.  All rights reserved.
 dnl Copyright (c) 2015      Research Organization for Information Science
 dnl                         and Technology (RIST). All rights reserved.
+dnl Copyright (c) 2021      Nanook Consulting.  All rights reserved.
 dnl $COPYRIGHT$
 dnl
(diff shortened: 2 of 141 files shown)


rhc54:
Hmmm...the LANL test fails here:
```text
util/cmd_line.c: In function 'pmix_cmd_line_make_opt3':
util/cmd_line.c:979:26: error: 'e.ocl_otype' may be used uninitialized in this function [-Werror=maybe-uninitialized]
     option->clo_otype = e->ocl_otype;
                         ~^~~~~~~~~~~
cc1: all warnings being treated as errors
```
The problem is that the variable `e` is passed into that routine, so I'm not sure how I can "initialize" it there. What I could do is just set it there, especially since we are eliminating that struct on another branch.

rhc54:
PGI just doesn't like picky compiler options:
```text
make[3]: Entering directory '/workspace/openpmix-src/src/include'
  CC       pmix_globals.lo
"/opt/nvidia/hpc_sdk/Linux_ppc64le/20.11/compilers/include-gcc49/_cplus_macros.
          h", line 20: error: zero used for undefined preprocessing identifier
          "__cplusplus"
  #if __cplusplus >= 201103L
      ^

"/opt/nvidia/hpc_sdk/Linux_ppc64le/20.11/compilers/include/_cplus_preinclude.h"
          , line 42: error: use of non-prototype function declarator
  extern void __builtin_pgi_va_start();
              ^

"/opt/nvidia/hpc_sdk/Linux_ppc64le/20.11/compilers/include/_cplus_preinclude.h"
          , line 43: error: use of non-prototype function declarator
  extern void* __builtin_pgi_va_arg();
               ^

"/workspace/openpmix-src/src/atomics/sys/atomic_stdc.h", line 163: error:
          ellipsis with no explicit parameters is nonstandard
  PMIX_ATOMIC_STDC_DEFINE_FETCH_OP(add, 32, int32_t, +)
  ^

4 errors detected in the compilation of "pmix_globals.c".
```
I'm trying to turn these options off when we see PGI, but I don't seem able to do so. Without seeing the configure output, I'm helpless.

@jjhursey Is there any way I can get the config.log or you could look to see why configure.ac isn't able to determine that we have PGI and avoid setting the picky options?

jjhursey:
I just saw the comment. It looks like PGI is passing now. Let me know if you still want me to pull the config.log and I can send it your way.

rhc54:
Thanks! Yeah, I figured out a way to solve it based on other things I saw in the configure code.

PR #2311, opened 2021-10-04
Enable address sanitizer

rhc54:
Fix an address issue in the pif/posix area so that
the address sanitizer passes


rhc54:
@bwbarrett Could you please take a look at this? I have a couple of concerns:

* the address sanitizer requires the ASAN library. I have tried to put a check in for it, but it doesn't work on Apple Mac - they have the ASAN support in something I can't identify. I'm therefore not sure this check is working.
* how do we "know" if the CI is using these flags or not? We don't see the configure output, so I'm not sure how we gain confidence that this "picky" compiler switch isn't going to break people.

Any thoughts?

rhc54:
bot:ibm:prrte:retest

Issue #2319, opened 2021-10-06
Compile error: ARM 21.1 compilers in util/show_help_lex.c


tonycurtis:
## Background information

Compile error using ARM 21.1 compilers

### What version of the PMIx Reference Library are you using? 

git head @ 93bbaed5cd86b2b0245ace56cfeb99df6be3c30c

### Describe how PMIx was installed 

clone

### Please describe the system on which you are running

* Operating system/version: CentOS 8.1
* Computer hardware: aarch64
* Network type: IB

-----------------------------

## Details of the problem

```
make[2]: Entering directory '/autofs/nccs-svm1_envoy_od/arcurtis/src/pmix/build/src'
  CC       util/show_help_lex.lo
util/show_help_lex.c:753:13: error: misleading indentation; statement is not part of the previous 'if' [-Werror,-Wmisleading-indentation]
            if ( ! (yy_state_buf) )
            ^
util/show_help_lex.c:751:9: note: previous statement is here
        if ( ! (yy_state_buf) )
        ^
util/show_help_lex.c:1187:3: error: misleading indentation; statement is not part of the previous 'if' [-Werror,-Wmisleading-indentation]
                return yy_is_jam ? 0 : yy_current_state;
                ^
util/show_help_lex.c:1184:2: note: previous statement is here
        if ( ! yy_is_jam )
        ^
2 errors generated.
```

rhc54:
😄 I just have to either figure out how to turn "off" the picky options for that file, or come up with a replacement for Flex. Will ponder for a bit.

PR #2321, opened 2021-10-07
Silence Flex warnings from show_help

rhc54:
Move the flex processing to its own directory and
turn off picky compilers there

Fixes https://github.com/openpmix/openpmix/issues/2319


rhc54:
The "pmix_pr_tester" passed but failed to set the status.

PR #2337, opened 2021-10-11
add a missing helpfile (util) and few fixups

naughtont3:
NOTE: Pruned a few of the help sections that were not
used, and added "PMIx" in text to clarify origin of msg.

### DIFF
--- a/src/util/os_dirpath.c
+++ b/src/util/os_dirpath.c
@@ -114,12 +114,7 @@ int pmix_os_dirpath_create(const char *path, const mode_t mode)
         }
 
         /* Now that we have the name, try to create it */
-        if (0 != mkdir(tmp, mode)) {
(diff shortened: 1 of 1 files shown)


rhc54:
Makes sense, but you still have to test the return from `mkdir` or else the "picky" compiler options will fail. You don't have to do anything intelligent in response - just need to check it.

rhc54:
Add `--enable-devel-check` to your configure line and you can see - I'm going to try and add it to the CI to help. Unfortunately, my plan to default developers to the "picky" configuration fell thru as it messed up the packagers.

naughtont3:
I rebuilt w/ the `--enable-devel-check` and the `src/util` area built w/o warning/errors.  

Note: There are some unrelated issues in the tests & example trees, but GCC seemed to be ok w/o checking retval of mkdir().

Issue #2328, opened 2021-10-09
Coverity warnings in test_v2


rhc54:
```text
________________________________________________________________________________________________________
*** CID 373434:  API usage errors  (PRINTF_ARGS)
/test/test_v2/test_server.c: 874 in server_read_cb()
868                         strncpy(fence_sig[fence_idx][p].nspace, msg_hdr.procs[p].nspace, 255);
869                         // loop over all nodes to find which node has this proc
870                         for (n = 0; n < g_num_nodes; n++) {
871                             temp_nodeid = -1;
872                             // look for a match for this fence proc within the procs on this node
873                             for (i = 0; i < nodes[n].pmix_local_size; i++) {
   CID 373434:  API usage errors  (PRINTF_ARGS)
   Argument "nodes[n].pmix_local_size" to format specifier "%d" was expected to have type "int" but has type "unsigned long".
874                                 TEST_VERBOSE(("Node: %d Local size: %d rank: %d", n, nodes[n].pmix_local_size, nodes[n].pmix_rank[i]));
875                                 if (nodes[n].pmix_rank[i] == msg_hdr.procs[p].rank) {
876                                     temp_nodeid = nodes[n].pmix_nodeid;
877                                     TEST_VERBOSE(("Node: %d found for proc: %d", temp_nodeid, nodes[n].pmix_rank[i]));
878                                     break;
879                                 }

** CID 373433:  API usage errors  (PRINTF_ARGS)


________________________________________________________________________________________________________
*** CID 373433:  API usage errors  (PRINTF_ARGS)
/test/test_v2/server_callbacks.c: 124 in fencenb_fn()
118                              size_t ninfo, char *data, size_t ndata, pmix_modex_cbfunc_t cbfunc,
119                              void *cbdata)
120     {
121         size_t n;
122     
123         for (n = 0; n < nprocs; n++){
   CID 373433:  API usage errors  (PRINTF_ARGS)
   Argument "nprocs" to format specifier "%d" was expected to have type "int" but has type "unsigned long".
124             TEST_VERBOSE(("Participating in this fence on node: %d procs[%d].nspace: %s,"
125                           " procs[%d].rank: %d, nprocs: %d", my_server_id, n, procs[n].nspace, n,
126                           procs[n].rank, nprocs));
127         }
128     
129         /* commented out, for now, to allow for the possibility of timeouts */

** CID 373432:  API usage errors  (PRINTF_ARGS)


________________________________________________________________________________________________________
[...]

PR #2338, opened 2021-10-11
Fixed format specifier problems in v2 test suite

cpshereda:
-Also changed to use pmix_strncpy() in several spots

Should address https://github.com/openpmix/openpmix/issues/2328.


rhc54:
We'll pull it in and let Coverity take another look - only way to tell, unfortunately.

PR #2370, opened 2021-10-22
Cleanup the configure logic

rhc54:
Remove "embedded" mode and the hwloc/libevent "header" options
and rely instead on the CPPFLAGS/LDFLAGS/LIBS combination to
find things. The effectively removes the OMPI-specific
customizations of the PMIx configure logic, which contained
errors that compensated for errors in the OMPI side.


rhc54:
bot:lanl:retest

PR #2375, opened 2021-10-28
Remove stale qsort checks

rhc54:
We don't need to worry about this legacy issue any more.


naughtont3:
@rhc54 was this just a single item cleanup or part of a bigger need.  I ask b/c we have old devel machine with ancient gcc that ran into this today.  We can probably add newer compiler, but wanted to know if these m4 file cleanups were part of bigger item.

rhc54:
Just a single item cleanup, though I'm surprised you encountered it as (per the comment in the code) it was restricted to a problem with stone age Solaris distributions
```
dnl There was some mentioning of broken qsort happened for Solaris that could
dnl cause qsort to return a bad pointer which could cause some badness.
dnl The problem should have been corrected with these patches from SunSolve.
dnl Solaris 10 should be free from this problem.
dnl
dnl 5.8_sparc #108827-27 or later
dnl 5.8_x86   #108828-28 or later
dnl 5.9_sparc #112874-20 or later
dnl 5.9_x86   #114432-07 or later
```

PR #2389, opened 2021-11-09
delete use of `PMIX_CHECK_BROKEN_QSORT` refs #2375

naughtont3:

### DIFF
--- a/src/include/pmix_config_bottom.h
+++ b/src/include/pmix_config_bottom.h
@@ -433,18 +433,6 @@ typedef PMIX_PTRDIFF_TYPE ptrdiff_t;
 #        define vsnprintf pmix_vsnprintf
 #    endif
 
-/*
- * Some platforms (Solaris) have a broken qsort implementation.  Work
- * around by using our own.
- */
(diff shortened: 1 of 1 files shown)


rhc54:
Ouch - thanks for catching it. Interesting that it didn't trip anyone/anywhere else 🤷‍♂️

naughtont3:
Yes, that worried me too.  The gcc thing is likely a red hearing, just part of other things going on today.  But surprised this did not cause a problem anywhere else when using the v4.1/v4.2 branches or rc5 tag.  

Also not sure how this slipped by when master has is correctly.

rhc54:
> Also not sure how this slipped by when master has is correctly.

I still see it in master...actually thought you were fixing it there first 😄

naughtont3:
Doh!  Too many branches/machines.  I guess I didn't have a fresh master on desktop.

naughtont3:
Well, that does make me feel a little better. I was very confused as to why it was only on the branches.  Turns out it was just me that was confused.  Added the master PR.

rhc54:
Thx! Much appreciated

PR #2396, opened 2021-11-12
Fix the network support components

rhc54:
When we allocate information to support a particular fabric,
we pack that info into a buffer and then transport the "blob"
to the backend for unpacking and processing. During the unpack
operation, we tried to be clever and simply assign the blob
into the various buffer fields.

Bad idea - missed a critical one and that caused the unpack
to fail. Instead, define a suitable macro for this purpose
to ensure that ALL the required buffer fields get set.

Also, OPA fabric is device type 0x208 - correct the typo.

Thanks to Howard Pritchard and Michael Heinz for identifying
the problem.


mwheinz:
I've verified that this corrects the issue.

PR #2404, opened 2021-11-23
Select all adapters with IPV4 addresses within specified subnet ranges.

drwootton:
Change processing for the --pmixmca ptl_base_if_include and --pmixmca ptl_base_if_exclude options so that all adapters that are within the specified subnet range are selected or excluded, respectively.

The changes to the split_and_resolve function create a list of unique adapter names selected for processing in the order they are seen, and that list is returned to the caller.

The use case here is: Given a system with three ethernet devices:

eth0: 192.168.1.10
eth1: 192.168.2.10
eth2: 9.10.11.10

All three interfaces can route to other nodes, but we only want to use eth0 and eth1 for TCP traffic for this job.
Currently, I need to specify either the individual devices or non-overlapping subnets: --pmixmca ptl_base_if_include 192.168.1.0/24,192.168.2.0/24 because if I specify --pmixmca ptl_base_if_include 192.168.0.0/16 only one of the two devices would be selected

With this fix, I can specify --pmixmca ptl_base_if_exclude 192.168.1.0/16 and pick up both interfaces.

Related Open MPI Issue: #5818
Signed-off-by: David Wootton <dwootton@us.ibm.com>

### DIFF
--- a/src/mca/ptl/base/ptl_base_fns.c
+++ b/src/mca/ptl/base/ptl_base_fns.c
@@ -1405,8 +1405,8 @@ static char *pmix_getline(FILE *fp)
  */
 char **pmix_ptl_base_split_and_resolve(char **orig_str, char *name)
 {
-    int i, ret, save, if_index;
-    char **argv, *str, *tmp;
+    int i, n, ret, if_index, match_count, interface_count;
+    char **argv, **interfaces, *str, *tmp;
     char if_name[PMIX_IF_NAMESIZE];
     struct sockaddr_storage argv_inaddr, if_inaddr;
     uint32_t argv_prefix;
@@ -1420,9 +1420,22 @@ char **pmix_ptl_base_split_and_resolve(char **orig_str, char *name)
     if (NULL == argv) {
         return NULL;
     }
-    for (save = i = 0; NULL != argv[i]; ++i) {
+    interface_count = 0;
+    interfaces = NULL;
+    for (i = 0; NULL != argv[i]; ++i) {
         if (isalpha(argv[i][0])) {
-            argv[save++] = argv[i];
+            /* This is an interface name. If not already in the interfaces array, add it */
+            for (n = 0; n < interface_count; n++) {
+                if (0 == strcmp(argv[i], interfaces[n])) {
+                    break;
+                }
+            }
+            if (n == interface_count) {
+                pmix_output_verbose(20,
+                                    pmix_ptl_base_framework.framework_output,
+                                    "ptl:tool: Using interface: %s ", argv[i]);
+                pmix_argv_append(&interface_count, &interfaces, argv[i]);
+            }
             continue;
         }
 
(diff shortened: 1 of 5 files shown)


drwootton:
An OpenMPI Coverity scan found a couple potential problems in source that is almost identical to this fix, so those should be resolved before merging these changes as they likely need to be replicated here.

src/mca/ptl/base/ptl_base_fns.c
```diff
@@ -1458,33 +1469,47 @@ char **pmix_ptl_base_split_and_resolve(char **orig_str, char *name)
...
-                            pmix_net_get_hostname((struct sockaddr *) &if_inaddr), if_name);
-        argv[save++] = strdup(if_name);
-        free(tmp);
     }
 
-    /* The list may have been compressed if there were invalid
-       entries, so ensure we end it with a NULL entry */
-    argv[save] = NULL;
+    /* Mark the end of the interface name array with NULL */
+    if (NULL != interfaces) {
```
rhc54, line 1508: You don't need nor want to add a NULL to the end of this array. The `pmix_argv_append` function ensures that there is a NULL entry at the end of the array.

src/mca/ptl/base/ptl_base_fns.c
```diff
@@ -1458,33 +1469,47 @@ char **pmix_ptl_base_split_and_resolve(char **orig_str, char *name)
...
+        interfaces[interface_count] = NULL;
+    }
+    pmix_argv_free(argv);
     free(*orig_str);
-    *orig_str = pmix_argv_join(argv, ',');
-    return argv;
+    *orig_str = pmix_argv_join(interfaces, ',');
```
rhc54, line 1513: I don't understand this statement - isn't the `orig_str` the list of interfaces to include/exclude? You definitely do _not_ want to free and overwrite that info as those are MCA params - this is likely to cause the param system to segfault upon clean termination.

drwootton: My understanding is that in the original code, argv had the mca parameter list specifying the interfaces to be included/excluded. 

The original code performed an in-place update of the passed argv array, either keeping the original specification if it was an interface name, and replacing it with the first matching interface if it was a subnet specification. There's some validation of the parameters, so in the original code, invalid parameters are skipped, meaning the resulting argv array may be shortened, so a null is appended following the last replacement argv to terminate the array at that point.

The original code then freed whatever we in orig_str and replacing it by building a comma delimited string of whatever is in the new, updated argv array.

My changes were intended to replicate this functionality, but now since a subnet specification can expand to one or more interface names, a new array containing the final list of interfaces to be returned is needed.

It seemed I was making this work as it did originally, but now returning multiple interface names for a subnet spec. 

I could be missing something, but my tests have been successful with this change.

drwootton: To clarify, orig_str was what had the original -mca parameter list amd the argv array was created from that by splitting at commas. But still, in the original code, orig_str was freed and replaced at the end of the function.

rhc54:
Regardless of what was done in the past, we strive to cleanup mistakes as we go. I downloaded your branch  and corrected a few things (uninitialized variable, warnings on input parameter mismatches), and cleaned up the code. I also rebased it against the upstream master.

This should now be good to go once it passes CI.

PR #2412, opened 2021-12-06
Fix library linking in static link cases

bwbarrett:
Fix logic to figure out what libraries to link with an application.  In the dynamic case, no longer include dependent libraries, since the aren't needed (this was a missing macro expansion in the wrapper compiler).  In the dynamic case, we need to thread the MCA libraries into the wrapper compiler.  Finally, switch from using pkg-config package requires and list the libraries out explicitly, due to problems with libevent and the MCA system (details in the commit message).

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -882,6 +882,34 @@ AC_DEFUN([PMIX_SETUP_CORE],[
     # executables) need perl.
     AC_PATH_PROG(PERL, perl, perl)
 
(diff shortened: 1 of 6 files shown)


rhc54 (approved):
I don't pretend to know enough to properly review what was done, but it (a) seems reasonable, and (b) the comments clearly explain what is happening. So LGTM!

Thanks!

rhc54:
bot:ibm:retest

PR #2417, opened 2021-12-10
The PMIX_HAVE_LIBEV and PMIX_HAVE_LIBEVENT flags must always be defined

rhc54:
Regardless of what was specified, if anything.


rhc54:
@bwbarrett Just FYI - it built fine on my M1 laptop, but failed on my Intel iMac. Probably something different about how I set the configure flags.

PR #2420, opened 2021-12-14
Respect the nocopy qualifier

rhc54:
If we are a tool, then IO must be forwarded to us in case
we have clients that need servicing. However, if the user
asked that we not copy IO to stdout/err (because they are
having the server put it into a file), then we need to
respect that request.


rhc54:
@naughtont3

PR #2421, opened 2021-12-15
test_v2: use static declaration for client parser

hppritcha:
declare the parse_fence_client static for safety


cpshereda (approved):
Thanks for catching this, Howard.

PR #2422, opened 2021-12-15
test_v2: add a simple modex direct test

hppritcha:
test can toggle between full data exchange/no data exchange for the PMIx_Fence

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -191,6 +191,7 @@ test/test_v2/test_get_peers
 test/test_v2/test_helloworld
 test/test_v2/test_init_fin
 test/test_v2/test_fence_partial
+test/test_v2/test_modex_basic
 
 # coverity
 cov-int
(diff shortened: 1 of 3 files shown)


hppritcha:
This test brings out an issue with the pmix_test server implementation, namely the case of using the PMIx_Fence without the data exchange feature will hang when using multiple servers (except for the special case of 2 servers and one client per server).

To run the test without issues with multiple servers run in this fashion
```
./pmix_test -v -s 2 -n 4 -- ./test_modex_basic --pmix-collect-data
```
or with other s and n values.

To observe the hang run with these options:
```
./pmix_test -v -s 2 -n 4 -- ./test_modex_basic 
```
the special case of s = 2 and n=2 passes.

hppritcha:
@cpshereda this is the test case we were discussing today.

hppritcha:
Oh I should add the test runs successfully with prte/prun.

hppritcha:
closing this PR

PR #2424, opened 2021-12-16
build: Be more flexible in library handling

bwbarrett:
Add ability to both add additional libraries into the link
dependencies for libevent, libev, and hwloc (useful if they
have unexpected libraries for static linking) and to disable
library checks for places where we get it really wrong and
users resort to setting LIBS directly.


rhc54 (approved):
Thx!

PR #2433, opened 2022-01-04
Python IOF pull fixes

MattBBaker:
This fixes several bugs in the Python bindings. One is to release the GIL around the `PMIx_IOF_pull` function and adds GIL protection around the `pyiofhandler()` handler.

The next one is more tricky and may need to be broken out if controversial. Inside of the `pmix_pyshift_t` there is a `pmix_byte_object_t *` that never gets set up in the `pyiofhandler()` in the 'handle not found' code path and causes codes to crash when trying to initialize the bad memory. The change I made changed `pmix_byte_object_t` to not be a pointer anymore, then allocate the bytes element and initialize in the 'handle not found' code path and frees it after it retries. 

This fixes my bug in the short term, though longer term I want to take a closer look at the event loop in general.

### DIFF
--- a/bindings/python/pmix.pxi
+++ b/bindings/python/pmix.pxi
@@ -58,7 +58,7 @@ class myLock(threading.Event):
 
 ctypedef struct pmix_pyshift_t:
     char *op
-    pmix_byte_object_t *payload
+    pmix_byte_object_t payload
     size_t idx
     pmix_modex_cbfunc_t modex
     pmix_status_t status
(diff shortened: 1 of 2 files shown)


rhc54:
I don't have an issue with the change to `pmix_pyshift_t` so long as it doesn't break any other code that currently uses it in the bindings. I assume you checked for that (if not, please do and submit whatever fixes are required), so I'm good with this change.

PR #2439, opened 2022-01-20
Fix race condition when processing pending get requests

jjhursey:
* Make sure to remove the matching pending request from the list before
   calling the client callback. Otherwise the client could race ahead and
   determine that there is an outstanding pending request at the bottom of
   the get operation. In which case it is putting the main thread to sleep
   forever (hang) thinking that another request will be incoming which it
   won't because it was already processed.

Signed-off-by: Joshua Hursey <jhursey@us.ibm.com>

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -7,7 +7,7 @@
  *                         All rights reserved.
  * Copyright (c) 2016-2018 Mellanox Technologies, Inc.
  *                         All rights reserved.
- * Copyright (c) 2016      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2016-2022 IBM Corporation.  All rights reserved.
  * Copyright (c) 2021      Nanook Consulting.  All rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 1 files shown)


jjhursey:
I put WIP on this for the moment while I do some pressure testing on this change.

src/client/pmix_client_get.c
```diff
@@ -490,6 +490,7 @@ static void _getnb_cbfunc(struct pmix_peer_t *pr, pmix_ptl_hdr_t *hdr, pmix_buff
                         "pmix: get_nb looking for requested key");
     PMIX_LIST_FOREACH_SAFE (cb, cb2, &pmix_client_globals.pending_requests, pmix_cb_t) {
         if (PMIX_CHECK_NSPACE(lg->p.nspace, cb->pname.nspace) && cb->pname.rank == lg->p.rank) {
+            pmix_list_remove_item(&pmix_client_globals.pending_requests, &cb->super);
```
rhc54, line 493: I'm a tad confused - it looks like all we are doing is moving the removal earlier. Is there some code path that would skip the removal? Is the problem that the cbfunc was getting called while the request is still on the list? Just wondering if we have an event problem - i.e., that something was supposed to be done in an event and we are violating that rule.

jjhursey: From what I could tell it was this ordering that was the problem:
```
            cb->cbfunc.valuefn(rc, val, cb->cbdata);
            pmix_list_remove_item(&pmix_client_globals.pending_requests, &cb->super);
```
So the callback is called and the client raced around to another PMIX_Get before the `pmix_list_remove_item` was called.
Just flipping the two lines of code allowed the test to pass - such as:
```
            pmix_list_remove_item(&pmix_client_globals.pending_requests, &cb->super);
            cb->cbfunc.valuefn(rc, val, cb->cbdata);
```
Moving it earlier in the function to where we first know that we no longer need it on the list seemed approprate.

rhc54: Okay, we can take this for now. I'll take a look as it sounds like we have an event problem - the client shouldn't be able to get down into the "get" processing until the prior one is complete.

PR #2441, opened 2022-01-22
Enable ability to bind PMIx progress thread

rhc54:
In some scenarios, it is desirable to bind progress threads
to one or more specific cores - e.g., a core designated for
utility usage and not assigned application processes. Provide
an attribute by which the user can specify the core(s) to
which the internal PMIx progress thread is to be bound, and
to indicate if the binding is mandatory (i.e., return an
error if the thread cannot be bound).

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -733,6 +733,9 @@ AC_DEFUN([PMIX_SETUP_CORE],[
     AC_PROG_GREP
     AC_PROG_EGREP
 
+    # This check must come after PMIX_CONFIG_THREADS
+    AC_CHECK_FUNCS([pthread_setaffinity_np])
+
     ##################################
     # Visibility
     ##################################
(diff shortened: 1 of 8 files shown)


rhc54:
@bwbarrett Could you please take a peek at this? I couldn't find a more portable solution, but this should purportedly work in many environments.

rhc54:
Discovered that the configure logic for detecting the
pthread_setaffinity_np function was always failing. I've
tried several variations, but either it always passes or
it always fails. I can't seem to reliably detect that the
function is available. This represents the latest attempt.
It always passes.

config/pmix.m4
```diff
@@ -683,6 +683,29 @@ AC_DEFUN([PMIX_SETUP_CORE],[
           [AC_DEFINE_UNQUOTED([HAVE_UNIX_BYTESWAP], [1],
                               [whether unix byteswap routines -- htonl, htons, nothl, ntohs -- are available])])
 
+    #
```
bwbarrett, line 686: From my local testing, the `AC_CHECK_FUNCS([pthread_setaffinity_np])` test is sufficient, with one minor change.  You need to move the check until after the call to `PMIX_CONFIG_THREADS`.  If there's a platform where `pthread_setaffinity_np()` is a compiler builtin instead of a library function (and I doubt such a place exists), then `AC_CHECK_FUNCS` would fail.  But the problem here was just that the lack of the `-pthread` or `-lpthread` (depending on platform) meant that there was no library providing `pthread_setaffinity_np` in the link command.

Moving `AC_CHECK_FUNCS([pthread_setaffinity_np])` to below the call to `PMIX_CONFIG_THREADS` resulted in a "yes" on modern Linux and a "no" on MacOS.

rhc54: Verified it works - thx!

PR #2444, opened 2022-01-28
Remove unneeded atomics code

rhc54:
PMIx is a single-threaded code base. We are either
in the user's thread, or we are in the PMIx progress
thread. Objects are not allowed to be accessed by
both threads at the same time. Therefore, we don't
need all that atomics code - we only need the rmb/wmb
functions to handle the threadshifts.


rhc54:
@bwbarrett Per our Slack-chat

src/include/atomic.h
```diff
@@ -55,7 +55,9 @@
 #ifndef PMIX_SYS_ATOMIC_H
 #define PMIX_SYS_ATOMIC_H 1
 
+#ifdef HAVE_STDATOMIC_H
```
bwbarrett, line 58: To avoid more header checks, I'd just put this include in the C11 block.  The C11 atomics test won't be true if stdatomic.h exists.  And if you're not using C11 atomics, this header's no use to you.  That would save yet another header test during configrue.

src/class/pmix_object.h
```diff
@@ -338,11 +336,11 @@ static inline pmix_object_t *pmix_obj_new_debug_tma(pmix_class_t *type, pmix_tma
         do {                                                                         \
             assert(NULL != ((pmix_object_t *) (object))->obj_class);                 \
             assert(PMIX_OBJ_MAGIC_ID == ((pmix_object_t *) (object))->obj_magic_id); \
-            pmix_obj_update((pmix_object_t *) (object), 1);                          \
+            ((pmix_object_t *) (object))->obj_reference_count++;                     \
```
bwbarrett, line 339: We'd probably want to do a bit of performance analysis, but the other option, which is where I was leaning, is to just have a pthread_lock in the base object and protect the ref counting that way.  That way, if you do have thread contention somewhere accidentally, you don't have silent data corruption.

bwbarrett (approved):

jjhursey:
bot:ibm:prrte:retest

src/include/atomic.h
```diff
@@ -55,14 +55,47 @@
...
+    __asm__ __volatile__("" : : : "memory");
+#    else
+    atomic_thread_fence(memory_order_acquire);
+#    endif
+}
+
 #elif PMIX_ATOMIC_GCC_BUILTIN
-#    include "atomic_gcc_builtin.h"
+
+static inline void pmix_atomic_rmb(void)
```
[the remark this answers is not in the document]
rhc54: I honestly don't care about the ordering - I just used what had already been there for ages. I can reorder them.

src/class/pmix_object.h
rhc54: Done - we aren't a performant library, so the cost of the lock is somewhat irrelevant.

src/include/atomic.h
rhc54: Done

src/include/atomic.h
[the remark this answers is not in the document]
rhc54: Done

bwbarrett (approved):

rhc54:
Sigh - looks like the tests violate the thread safety rules 🤷‍♂️  Doubt I'll have time anytime soon to fix those - they aren't easily understood. I might try and see what has to be replaced just to make those work. It isn't the PMIx code base that's the problem - it is the tests themselves.

bwbarrett:
I can take a look, but it won't be until after the one-sided fixes are in OMPI (another week or two).

rhc54:
@bwbarrett I found the problem and fixed it - missing mutex init in the object constructor

PR #2453, opened 2022-02-04
Handle string literals in environ to unblock hwloc versions

bwbarrett:
Address https://github.com/openpmix/openpmix/issues/2450 by being more robust about environ handling when merging environments for application launch.

### DIFF
--- a/.github/actions/mlnx/Dockerfile
+++ b/.github/actions/mlnx/Dockerfile
@@ -1,7 +1,6 @@
 FROM centos:7.6.1810
 
 RUN \
-    yum install -y epel-release; \
     yum install -y perl perl-Data-Dumper \
         automake libtool flex make bzip2 git which rpm-build libevent-devel pandoc hwloc hwloc-devel
 
--- a/.gitignore
+++ b/.gitignore
@@ -119,6 +119,7 @@ test/pmi_client
 test/pmix_client
 test/pmix_test
 test/pmix_client2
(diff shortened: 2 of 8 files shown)


jjhursey:
bot:ibm:prrte:retest

rhc54 (approved):
LGTM! I haven't searched to see where else we touch the environment, so we may hit the "trip wire" again when broader experience is gained. At least we have tools we can use to fix that if/when it happens.

bwbarrett:
@rhc54 I looked through every use of environ and this was the only place where we were directly mucking with environment strings (vs doing searches or just calling pmix_setenv(..., &environ).  So we should be ok.

rhc54:
I can't tell if the PRRTE CI ran again or not - did anybody copy the timestamp on it before retest?

rhc54:
Current test timestamp is `Fri Feb  4 16:27:13 EST 2022`

bot:ibm:prrte:retest

rhc54:
Yeah, this looks like a true problem. My best guess is that something in the new `pmix_environ_merge_inplace` function is segfaulting when we try to locally "spawn".

rhc54:
@bwbarrett Just to be clear - there is no problem in your new code. The only issue was a typo where you checked the return status from `pmix_environ_merge_inplace` - you had `==` instead of `!=`. So no need to adjust/add to the tests you wrote.

PR #2454, opened 2022-02-07
test: Fix dereference issue in environ test

bwbarrett:
Fix CID 375078, caused by ec0daf3, by checking that the return of
getenv() is not NULL before passing it to strcmp().


rhc54 (approved):
LGTM

PR #2457, opened 2022-02-15
pmix_iof.c: malloc buffer before memcpy().

awlauria:
Found this when testing mpirun/prterun --output dir=/foo/bar/
option.


rhc54 (approved):
Nice catch - thx!

PR #2458, opened 2022-02-16
examples: add a test to simulate open mpi PMIx group usage

hppritcha:
usage.

The current algorithm in Open MPI for supporting MPI 4 Sessions support
makes use of an extended CID (communicator ID) concept.  This approach
is leading to problems supporting Sessions for PMLs that can't handle
exCID information easily.

A much simpler approach to handing the exCID exchange mechanism is
demonstrated in this PR.  By making use of PMIx dmodex feature in
combination with PMIx Group functionality, it should be much
easier to support MPI Sessions in PMLs other than OB1.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,6 +84,7 @@ examples/pubi
 examples/server
 examples/tool
 examples/group
+examples/group_dmodex
 examples/asyncgroup
 examples/hello
 examples/bad_exit
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -24,7 +24,9 @@ headers = examples.h
 
 AM_CPPFLAGS = -I$(top_builddir)/src -I$(top_builddir)/src/include -I$(top_builddir)/include -I$(top_builddir)/include/pmix
 
-noinst_PROGRAMS = client client2 dmodex dynamic fault pub pubi tool debugger debuggerd alloc jctrl group asyncgroup hello nodeinfo  abi_no_init abi_with_init group_lcl_cid
+noinst_PROGRAMS = client client2 dmodex dynamic fault pub pubi \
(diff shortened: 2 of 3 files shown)


hppritcha:
@rhc54 if you could do a sanity check on this example it would be much appreciated.

rhc54:
Hmmm...it will certainly work, but only in environments that support the dmodex operation. However, I'm not sure I understand what it actually accomplishes. Every process has to call "group_construct", and so the CID is going to be returned to each of them during execution of that operation. So why do I need to perform a dmodex to retrieve it? Is the issue that the info is in the "result" array and not accessible via `PMIx_Get`? If so, you can always just store it internally with `PMIx_Store_internal` and then `PMIx_Get` it somewhere else in the code.

rhc54:
> one question that comes to mind, is there a way via PMIx that a proc can remove a key/value from a given namespace?

Hmmm...no, not currently - I imagine we could add something, but it might be tricky. Lots of race conditions if we are trying to remove it globally (as opposed to just within the calling proc's data store).

hppritcha:
okay about the removal.  

The value being put here is representing what we are now calling a local CID.  That's different on each proc (or likely will be) in the group.  The PMIx Group CiD is very useful here at discriminating local CIDs associated with different communicators created via MPI_Comm_create_from_group.   I'm using the dmodex approach because we can't call a PMIx_Fence to exchange the data as MPI_Comm_create_from_group may be called concurrently from multiple threads in a process - MPI 4 says this call is thread safe.    

Would you have time for a short webex?

rhc54:
I believe we resolved this via the qualified value being returned in the group construct data. @hppritcha Can we either clean up the conflict and commit, or drop/close this PR?

rhc54 (approved):

PR #2471, opened 2022-02-28
pmix_cmd_line.c: Stop parsing at the executable.

awlauria:
Otherwise, the parser will try to match command line
options for the application to prterun.

Refs open-mpi/ompi#10044


rhc54:
This does seem to fix it - thx!

PR #2481, opened 2022-03-02
Move the Special builds to GitHub Actions

jjhursey:
* These run in the same container as the cross-version checks
   which are already running in OpenPMIx's GitHub Actions.

### DIFF
--- a/.github/workflows/run-special.yml
+++ b/.github/workflows/run-special.yml
@@ -1,0 +1,31 @@
+name: OpenPMIx Special Builds Testing
+
+on:
+  pull_request:
+    # We don't need this to be run on all types of PR behavior
+    # See
+    #  - https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request
+    #  - https://frontside.com/blog/2020-05-26-github-actions-pull_request
+    types:
+      - opened
+      - synchronize
+      - edited
+      - reopened
+
+env:
+  # Use docker.io for Docker Hub if empty
+  REGISTRY: docker.io
+  # github.repository as <account>/<repo>
+  IMAGE_NAME: jjhursey/pmix-xver-tester
+
+jobs:
+  special-builds:
+    runs-on: ubuntu-latest
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - name: Check out the code
+        uses: actions/checkout@v1
+      # Run the container tester
+      - name: Special Builds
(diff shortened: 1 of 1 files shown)


jjhursey:
This move the Special builds to GitHub Actions.

It currently fails when running on an x86 system compiling the python bindings.
 * https://github.com/jjhursey/openpmix/runs/5398396777?check_suite_focus=true

```
./configure --prefix=/home/pmixer/tmp-build-htNom/install-static --disable-shared --enable-static --with-libevent=/home/pmixer/local/libevent --with-hwloc=/home/pmixer/local/hwloc --enable-python-bindings
```

The error is something with the interplay between cython and GCC it seems
```
gcc -pthread -shared -Wl,-z,relro -Wl,-z,now -g -Wl,-z,relro -Wl,-z,now -g build/temp.linux-x86_64-3.6/home/pmixer/tmp-build-htNom/openpmix/bindings/python/pmix.o -L/home/pmixer/tmp-build-htNom/install-static/lib -L/usr/lib64 -L/home/pmixer/.local/lib -Wl,--enable-new-dtags,-R/home/pmixer/.local/lib -lpmix -lpython3.6m -o build/lib.linux-x86_64-3.6/pmix.cpython-36m-x86_64-linux-gnu.so
/usr/bin/ld: /home/pmixer/tmp-build-htNom/install-static/lib/libpmix.a(showhelp_lex.o): relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: /home/pmixer/tmp-build-htNom/install-static/lib/libpmix.a(keyval_lex.o): relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
error: command 'gcc' failed with exit status 1
make[3]: *** [Makefile:795: install-exec-local] Error 1
```

rhc54:
Hmmm...I see something on that gcc line that speaks of "cpython", which is a definite conflict with "cython". I wonder if cpython is installed and somehow getting pulled in here? I can take a look in a bit.

Thanks for setting this up!

jjhursey:
I filed an Issue with a few more details on the issue (and how to reproduce in the container on a mac)
 * https://github.com/openpmix/openpmix/issues/2482 

It's curious that the ppc64le version of the same container doesn't seem to mind. Odd for sure.

rhc54:
Yeah, that is strange. I'll try to poke at it soon. For now, this should be okay to go as it isn't a check of the Python build itself, but rather the setup for doing such checks.

jjhursey:
Sounds good. I'll leave the IBM version running for at least a week. I won't turn it off until the GH Actions version is running clean so you can compare them in case that is helpful.

PR #2483, opened 2022-03-02
Add shared-memory segment utility functions.

samuelkgutierrez:


rhc54:
The special builds failure is due to a problem in building the Python bindings and unrelated - I need to turn off the Python bindings check until it can be resolved.

PR #2484, opened 2022-03-02
Add macros for variadic PMIX_NEW, PMIX_CONSTRUCT.

samuelkgutierrez:
For PMIX_NEW(), one argument means the caller does not want a custom memory
allocator, namely the common case.  Two arguments means caller wants a
custom (TMA) memory allocator.

Similarly, PMIX_CONSTRUCT() now takes a maximum of three arguments: two
arguments means caller does not want a custom memory allocator---the
common case. Three arguments means the caller wants a custom (TMA) memory
allocator.


rhc54:
bot:ibm:builds:retest

rhc54:
Bah - we may still be trying to run the Python test. Will continue to poke at that one.

PR #2486, opened 2022-03-03
Silence warnings in pmixt_validate_predefined().

samuelkgutierrez:
Silence "reading 512 bytes from a region of size X" warnings generated
by GCC 11.2.0. This may be a spurious warning, but once that is
generated nonetheless. The problem seems to come from a function
definition with a static length char buffer (pmix_key_t) with call-sites
passing string constants that are strictly shorter than the size of
pmix_key_t. This usage seems atypical in the codebase, so perhaps there
is a better way to silence these particular warnings.

A similar warning is generated at a call-site in
test/test_v2/server_callbacks.c at line 323 involving a call to
PMIx_server_register_nspace(), but this one is stickier, so someone else
should look into that one. Note: changing the string constant to a char*
variable does not silence this one. Pesky, pesky.

### DIFF
--- a/test/test_v2/test_clientapp_funcs.c
+++ b/test/test_v2/test_clientapp_funcs.c
@@ -256,7 +256,7 @@ void pmixt_post_finalize(pmix_proc_t *this_proc, test_params *l_params, validati
 
 /* This function is used to validate values returned from calls to PMIx_Get against
    side channel validation data */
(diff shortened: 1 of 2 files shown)


rhc54:
Yeah, we see that complaint quite a bit from the latest compilers. Not entirely sure how to resolve it in general yet. All I've managed to do was to follow the calling chain and ensure that everyone passed/recvd pmix_key_t's.

PR #2487, opened 2022-03-04
Provide a "hook" for PRRTE to init a small amount of PMIx

rhc54:
Now that PRRTE directly depends on PMIx classes and frameworks,
we need to ensure that some small portion of PMIx is setup
prior to PRRTE attempting to use it. Create a new "pmix_init_util"
for this purpose. Need to isolate that prototype in its own
include header to avoid pulling in a bunch of other stuff.

The PMIx interface framework was the immediate problem as PRRTE
tried to traverse a list prior to it being constructed.

### DIFF
--- a/src/mca/pif/base/base.h
+++ b/src/mca/pif/base/base.h
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2010      Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2016      Intel, Inc.  All rights reserved.
+ * Copyright (c) 2022      Nanook Consulting.  All rights reserved.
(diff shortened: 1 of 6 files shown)


rhc54:
@awlauria @bwbarrett This was one of two causes of the problems you encountered. I'm not completely happy with it as PRRTE will initialize the installdirs framework before PMIx can pass down any user-provided info directives. So when PRRTE uses PMIx, we lose that ability. This shouldn't be a problem when PRRTE is directly used (e.g., for the DVM or by OMPI's `mpirun`) as the PRRTE tools rely on envars for adjusting installdirs.  So I'm not convinced it is an issue - just something to watch out for going forward.

PR #2493, opened 2022-03-08
examples/dmodex.c: Fix some bugs.

awlauria:
- Fence after the loop to prevent a hang if one rank fails.
- Fix check for a local rank.
- Make sure the namespace and rank is always copied in proc.

### DIFF
--- a/examples/dmodex.c
+++ b/examples/dmodex.c
@@ -15,7 +15,7 @@
  * Copyright (c) 2011      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2013-2020 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Mellanox Technologies, Inc.  All rights reserved.
- * Copyright (c) 2019      IBM Corporation.  All rights reserved.
+ * Copyright (c) 2019-2022 IBM Corporation.  All rights reserved.
  * Copyright (c) 2021-2022 Nanook Consulting  All rights reserved.
  * $COPYRIGHT$
  *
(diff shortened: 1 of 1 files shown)


awlauria:
This test passes for me on a single node and 2 nodes.

awlauria:
bot:retest

rhc54:
> This test passes for me on a single node and 2 nodes.

Well...yes, though you have eliminated a significant element of the test - namely, the entire "wait until data is published" portion of the code. Thus, you won't see any issues with the hotel eviction/recheck code path, etc.

In other words, this "fixed" the bug by simply ignoring the faulty code path 😄

awlauria:
I see what you mean regarding the first fence  - I thought it was required to distribute the data before the Get() calls. Will remove it.

The second fence (after the loop) should be after the `done:` label, or else ranks on a failure will go straight to PMIx_Finalize(), while those who succeeded will be stuck in the fence. I think this one should stay as is.

examples/dmodex.c
```diff
@@ -227,16 +228,16 @@ int main(int argc, char **argv)
         }
     }
 
+done:
+    /* finalize us */
+
     /* call fence so everyone waits before leaving */
-    proc.rank = PMIX_RANK_WILDCARD;
-    if (PMIX_SUCCESS != (rc = PMIx_Fence(&proc, 1, NULL, 0))) {
+    if (PMIX_SUCCESS != (rc = PMIx_Fence(NULL, 0, NULL, 0))) {
```
rhc54, line 235: Just an FYI (so no need to change anything): `PMIx_Fence` with a `NULL` proc argument defaults to "all procs in my nspace".

PR #2495, opened 2022-03-09
Modify pmix_show_help_init() to take an addition helpdir.

awlauria:
This way prte help files can be added to the search path.


rhc54:
I think this is fine. I've been thinking of ways to eliminate the PRRTE `show_help` code (replace with the PMIx version). The only hangup is the sending of the messages to the DVM controller and that proc's aggregation code. We might be able to work around that by integrating the recv and aggregation into the `prted/pmix/pmix_server_gen.c` and using the `PMIx_Log` function as the go-between. This would also resolve the problem of getting `show_help` output to the tool.

Needs some finagling to make it all work.

PR #2497, opened 2022-03-10
Add a couple of print APIs and update pquery to use one

rhc54:
I'm tired of writing pretty-print code in every app, so
create two new APIs:

PMIx_Info_string - prints a pmix_info_t
PMIx_Value_string - prints a pmix_value_t

Update pquery to print out the results using the new
APIs. Also update ignores


rhc54:
bot:ibm:prrte:retest

PR #2500, opened 2022-03-10
pmix_hotel_recheck(): Remove this function. 

awlauria:
There's no consumers of it at this point, and likely is not needed.

See discussion in https://github.com/openpmix/prrte/pull/1265.


rhc54 (approved):
Thanks!

PR #2502, opened 2022-03-11
Add missing function and improve error message

rhc54:
We were declaring pmix_argv_split_with_empty but never instantiated
it - do so. Include the original dlopen error when reporting possible
reasons the library could not be loaded.

(cherry picked from commit 94dbda0a426154080ca77cf7326a5186ff0ba93d)

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -1824,9 +1824,11 @@ static inline void pmix_argv_free(char **argv)
 #define PMIX_ARGV_FREE(a)  pmix_argv_free(a)
 
 static inline
-char **pmix_argv_split(const char *src_string, int delimiter)
+char **pmix_argv_split_inter(const char *src_string,
+                             int delimiter,
+                             bool include_empty)
 {
-    char arg[256];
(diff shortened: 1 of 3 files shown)


rhc54:
OOOOO...PRRTE master will not build against this branch any more due to the v5.0 restriction. Need to update the CI.

jjhursey:
Is the following the correct pairing?
 * OpenPMix < v5 must use PRRTE v2.x branch
 * OpenPMIx >= v5 must use PRRTE master branch

I can take a look at the CI config to sort that out.

rhc54:
Yeah, that sounds right. Technically, you could use OpenPMIx >= v5 against an older PRRTE branch, and it might be worth checking that option out as well as against PRRTE master (perhaps as part of the "special builds"?) to ensure backward compatibility. Just can't use older versions of OpenPMIx with PRRTE master and beyond due to the code reuse between them.

Much appreciate the help!

jjhursey:
Ok. I'll work on that this week and let you know via Slack when it's ready.

PR #2518, opened 2022-03-15
Prohibit Python bindings with non-shared lib builds

rhc54:
Cython takes issue with purely static builds, so error
out if the user requests the Python bindings but has
specified --disable-shared.

Also fix a problem with the "canonicalize_path" function
in configure - new versions of MacOS (i.e., starting with
Mojave) do _not_ have a /usr/bin/python, but do typically
have a "python" in the path. However, this python is often
v3, so make the syntax "futureproof" to handle the situation

(cherry picked from commit 4c0cd800ca9db76159b72099a039333ac61ac9b5)

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -1267,6 +1267,14 @@ if test "$enable_python_bindings" != "yes"; then
     WANT_PYTHON_BINDINGS=0
 else
     AC_MSG_RESULT([yes])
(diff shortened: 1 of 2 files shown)


rhc54:
huh - another one got the wrong pmix version

bot:ibm:prrte:retest

rhc54:
Ah, yes - cannot build PRRTE against this version. Duh.

PR #2520, opened 2022-03-16
Fix a crash with --enable-debug

awlauria:
- Revert "Leak: Always free ns->jobbkt in pmix_nspace_caddy_t destructor."

This reverts commit https://github.com/openpmix/openpmix/commit/aeb4145563eb3ed48a5445009213b806df2ccbfa.

Not sure why I did this, ns->jobbkt is free'd in the ns
destructor, called two lines down. If it isn't getting released,
that is a leak somewhere else where its reference count is incremented.
This was the wrong fix.

FWIW - that leak elsewhere may have since been fixed, causing me
to see this bug now.

This corrects a crash with --enable-debug where remote daemons
will crash with an assert on jobbkt's ref count.

- Initialize size for getsockopt().

This parameter is technically an in and out
for getsockopt() calls. Reported as uninitialized
by valgrind.

### DIFF
--- a/src/include/pmix_globals.c
+++ b/src/include/pmix_globals.c
@@ -184,9 +184,6 @@ static void ncdcon(pmix_nspace_caddy_t *p)
 static void ncddes(pmix_nspace_caddy_t *p)
 {
     if (NULL != p->ns) {
-        if (NULL != p->ns->jobbkt) {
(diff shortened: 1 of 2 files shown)


rhc54 (approved):
Using that `pmix_socklen_t` that way seems kinda weird, but that actually is what the function signature is looking for, so this is fine.

PR #2525, opened 2022-03-17
gds:shmem: Checkpoint mostly server-side work.

samuelkgutierrez:
Checkpoint mostly server-side work storing info objects into a
shared-memory backing store.

### DIFF
--- a/src/class/pmix_object.h
+++ b/src/class/pmix_object.h
@@ -396,7 +396,6 @@ static inline pmix_object_t *pmix_obj_new_debug_tma(pmix_class_t *type, pmix_tma
     ((type *)pmix_obj_new_debug_tma(PMIX_CLASS(type), tma, __FILE__, __LINE__))
 
 #else
-// TODO(skg) Implement, test without debug.
 #define PMIX_NEW_NO_TMA(type)   ((type *)pmix_obj_new_tma(PMIX_CLASS(type), NULL))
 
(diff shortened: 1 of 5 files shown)


samuelkgutierrez:
Looks like the `dmodex` test is failing. I'm not sure why.

rhc54:
@awlauria I'm not sure if we still have a problem in the test - I don't see anything immediately wrong in the output, but it just stopped. If I read it correctly, it only ran for 13 seconds? Could you please investigate?

rhc54:
@samuelkgutierrez Nothing to do with this change - we may still need some debugging of the test itself.

awlauria:
@rhc54 the test isn't timing out, it's hitting more PMIx_Get() failures.

```

15:44:45 Client ns prterun-94d8d5204348-144@1 rank 0: PMIx_Get prterun-94d8d5204348-144@1-0-remote failed: NOT-FOUND
15:44:45 prterun-94d8d5204348-144@1:21 Remote value for prterun-94d8d5204348-144@1:0 successfully retrieved
```
I wonder if the timeout in the test just needs to be increased?

awlauria:
The output of that test is a little...verbose...I opened up https://github.com/openpmix/pmix-tests/pull/123 to clean that up.

Should make it easier to see what went wrong.

awlauria:
bot:ibm:retest

samuelkgutierrez:
Thank you @awlauria and @rhc54.

PR #2526, opened 2022-03-17
Ci test

jjhursey:


jjhursey:
bot:ibm:prrte:retest

jjhursey:
Ok so the [build worked fine](https://gist.github.com/fdd59a4034e9d16b23981f6af09b1c85) this time. For the OpenPMIx `v4.2` branch it chose the PRRTE `v2.1` branch instead of master.
```
17:28:23 Using PRRTE Branch: v2.1 (target OpenPMIx branch v4.2)
```
It failed on the `dmodex` test case which I think is expected at this point.

Let's retest for good measure
bot:ibm:prrte:retest

jjhursey:
Actually it looks like a `dmodex` build error:
```
==========================
Building PMIx dmodex
==========================
dmodex.c:29:10: fatal error: src/include/pmix_config.h: No such file or directory
 #include "src/include/pmix_config.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

@awlauria I'm assuming that this doesn't work on the v2.1 branch. Can we improve the build to skip that test if the necessary file is not present?

rhc54:
I don't see any reason for the dmodex.c test to rely on anything other than the public PMIx headers, so this is probably just a leftover from the original version.

rhc54:
Should be fixed by https://github.com/openpmix/pmix-tests/pull/124

rhc54:
bot:ibm:prrte:retest

rhc54:
Rather odd output - I'll have to check this separately. It is quite possible that the PMIx v4.2 branch will return something different as there are some bug fixes in PMIx master that haven't been backported yet.

awlauria:
I might have a look today to see what needs to come back for this to be fixed in v4.2.

jjhursey:
I am happy to leave this PR open if you all want to bounce off of it a bit. But the backend CI script is fixed with the following logic and that seems to be working
```
# if testing OpenPMIx v4*
#   use PRRTE v2.1
# else
#   use PRRTE master
```

I also made a similar change for PRRTE PRs
```
# if testing PRRTE v2*
#   use OpenPMIx v4.2
# else
#   use OpenPMIx master
```

PR #2527, opened 2022-03-17
Protect critical zone in pmix_obj_update().

awlauria:
Fix a very subtle bug where it is
possible that another thread will win the race
and update obj_reference_count before its value is returned.
This can result in asserts being hit with debug builds.

Cache the updated value before the lock is let go and return it
seems to fix intermittent crashes in various MPI routines.

### DIFF
--- a/src/class/pmix_object.h
+++ b/src/class/pmix_object.h
@@ -709,9 +709,9 @@ static inline int pmix_obj_update(pmix_object_t *object, int inc)
         perror("pthread_mutex_lock()");
         abort();
     }
-    object->obj_reference_count += inc;
+    ret = (object->obj_reference_count += inc);
     pthread_mutex_unlock(&object->obj_lock);
-    return object->obj_reference_count;
+    return ret;
 }
 
 END_C_DECLS


awlauria:
Hmm , interesting. This passed the dmodex test but failed the debugger test.

bot:ibm:retest

rhc54 (approved):
Very interesting - I wonder if we need to put a `PMIX_ACQUIRE_OBJECT` at the beginning and a `PMIX_POST_OBJECT` at the end of this routine like we do elsewhere when getting an object on a thread? It basically just ensures that a value is updated in memory prior to another thread accessing it. We've had that problem in some environments, which might explain why this particular problem is hard to nail down.

rhc54:
FWIW: we sometimes see the debugger test fail due to timing issues between when things start and try to connect. Always resolved by just rerunning the test.

awlauria:
I don't think it's necessary. pthread_mutex_lock()/unlock should guarantee that other threads see the updated value with a built-in memory barrier.

rhc54:
Okay, sounds good.

PR #2528, opened 2022-03-17
Add a little extra safety

rhc54:
Use the PMIX_ACQUIRE_OBJECT and PMIX_POST_OBJECT to
ensure that any updates to the object reference count
get properly stored before some other thread can
access them.

### DIFF
--- a/src/class/pmix_object.h
+++ b/src/class/pmix_object.h
@@ -133,6 +133,8 @@
 #include <stdio.h>
 #include <errno.h>
 
+#include "src/include/pmix_atomic.h"
+
 BEGIN_C_DECLS
 
 #if PMIX_ENABLE_DEBUG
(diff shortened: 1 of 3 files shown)


rhc54:
@awlauria I'm not sure these are fully required, but it might be a good safety measure given what you were seeing. There might be a tiny performance hit, but we never juggle ref counts in a critical path (and such a thing doesn't really exist in PMIx or PRRTE). Anyway, see what you think - no strong opinion either way here.

awlauria:
I don't think it's necessary, as `pthrad_mutex_*()`, as I understand it, have hardware memory barriers built-in to them. I've never seen any memory inconsistency when using them, and haven't seen the need for additional barriers.

If I were to take a position I'd say leave them out for now.

PR #2530, opened 2022-03-18
Improve handling of dependencies with static libraries

bwbarrett:
Series of patches that improves PMIx's dependency management.  We now add module requires fields for the dependencies where we found a pkg-config module, we track the static library dependencies, and we're much more consistent in what flags we add to the wrapper and pkg-config module file depending on how PMIx is configured.


bwbarrett:
With this patch series, on Ubuntu 20.04, I can compile pmix with `--enable-static --disable-shared --with-curl --with-jansson` with all the dependencies that are in stock Ubuntu and have PRRTE master link against it without having to specify any extra libraries.  I made progress in static linking an executable, but need to build a static libc and ran out of time.  But all the library tracking looks right.

jsquyres:
Love it!

I didn't review this closely, but I did read all the commit messages and skim the contents.  I think that this is an awesome set of steps forward.

rhc54:
I'd echo what Jeff said - awesome! I have no hope of understanding all that was done, but am thankful someone smarter than me was willing to work thru all that complexity 😄

PR #2532, opened 2022-03-18
v4.2: Protect critical zone in pmix_obj_update().

awlauria:
Fix a very subtle bug where it is
possible that another thread will win the race
and update obj_reference_count before its value is returned.
This can result in asserts being hit with debug builds.

Cache the updated value before the lock is let go and return it
seems to fix intermittent crashes in various MPI routines.

(cherry picked from commit 56a46caa737da90d934a902769315bf3e949d3fd)

### DIFF
--- a/src/class/pmix_object.h
+++ b/src/class/pmix_object.h
@@ -540,9 +540,9 @@ static inline int pmix_obj_update(pmix_object_t *object, int inc)
         perror("pthread_mutex_lock()");
         abort();
     }
-    object->obj_reference_count += inc;
+    ret = (object->obj_reference_count += inc);
     pthread_mutex_unlock(&object->obj_lock);
-    return object->obj_reference_count;
+    return ret;
 }
 
 END_C_DECLS


rhc54:
I'm not sure of what to do with these one-off transfers. The problem is that we can easily lose track of what went over and what hasn't yet been moved, but should be. It also gets things out-of-sync - though that might not impact this particular PR, it can lead to conflicts with other PRs that happened to touch a common code path.

Also still wrestling with how we commit to master. Some people do tiny atomistic commits that span multiple weeks - if not squashed or rebased, these little commits pepper the history and make it very hard to find them all when it is time to update the release. It's one major cause of all those "conflicts" you mentioned elsewhere.

So I'm scratching my head a bit and will try to figure something out. Not trying to discourage anyone - just explaining why this might get delayed while thinking is underway.

awlauria:
@rhc54 ideally patches that should be brought back to release branches are done so immediately after going into master, or shortly after. Unfortunately that isn't always the case. 

In OMPI I've generally brought things back to the release branches in a similar manner. it makes the PR's easier to digest and review. I have not seen an impact in it making other things more difficult to back port or screwing up the history.

awlauria:
Not saying either way is wrong - both work.

I can try to group them more together in my next pr. The three I've posted are small enough and isolated enough to not impact anything else I think

rhc54:
Not a problem - I'm not sure there is an ideal solution. What I've been trying to do is once a week go back and scan all the commits to master, transferring them in sequence to the 4.2 release branch. Then I cherry-pick those fixes (since they will inevitably include conflict resolutions) back to v4.1, again maintaining the sequence.

🤷‍♂️ Far from perfect. I think I'm going to set the merge button to "rebase and merge" to fix the problem of commits from a given PR peppering the history so it is easier to find things that need to be ported. The problem isn't quite as big as it used to be as that organization isn't providing contributions any more, but it does still occasionally happen.

PR #2535, opened 2022-03-18
v4.2: Update the configure logic to track master

rhc54:


rhc54:
The `dmodex` test is very unhappy with the v4.2/v2.1 combination. Separate issue

PR #2536, opened 2022-03-18
v4.2: Fix delayed local get requests

rhc54:


rhc54:
@awlauria This change actually caused a daemon to segfault in the `dmodex` test - not exactly sure why. So far as I can tell, the PRRTE v2.1 branch has the necessary fixes in it, but something about the combination is unhappy.

awlauria:
@rhc54 I think this can be closed.

rhc54:
Agreed

PR #2543, opened 2022-03-23
Dmodex v4.2

awlauria:


src/client/pmix_client_get.c
```diff
@@ -594,12 +594,7 @@ static void _getnb_cbfunc(struct pmix_peer_t *pr,
     PMIX_LIST_FOREACH_SAFE(cb, cb2, &pmix_client_globals.pending_requests, pmix_cb_t) {
         if (PMIX_CHECK_NSPACE(proc.nspace, cb->pname.nspace) && cb->pname.rank == proc.rank) {
             pmix_list_remove_item(&pmix_client_globals.pending_requests, &cb->super);
-            if (PMIX_SUCCESS != ret) {
```
rhc54, line 597: As stated in the "master" branch version, removing this code block is incorrect.

rhc54: As per the master branch PR, just remove commit https://github.com/openpmix/openpmix/pull/2543/commits/acca23a317fb8e2f61eb675b93fe6c80bee3fb01 and this should be fine.

awlauria: thanks, done.

src/client/pmix_client_get.c
```diff
@@ -626,6 +626,28 @@ static void _getnb_cbfunc(struct pmix_peer_t *pr,
...
+            else {
+               /* if we are both using the "hash" component, then the server's peer
+                * will simply be pointing at the same hash tables as my peer - no
+                * no point in checking there again */
+                if (0 != strcmp(pmix_globals.mypeer->nptr->compat.gds->name, pmix_client_globals.myserver->nptr->compat.gds->name)) {
+                    pmix_output_verbose(2, pmix_client_globals.get_output,
+                                        "pmix: get_nb searching for key %s for proc %s, - %s",
+                                        cb->key, PMIX_NAME_PRINT(&proc), pmix_client_globals.myserver->nptr->compat.gds->name);
+                    PMIX_GDS_FETCH_KV(rc, pmix_client_globals.myserver, cb);
+                }
```
rhc54, line 638: Argh - I should have caught this earlier. The "if success" clause below needs to be inside of the above "if" clause. Otherwise, you are checking `rc` again for no reason.

rhc54: Needs to go back to master branch as well - sorry for missing it earlier 😞

awlauria: I think it is correct in master - this is a bad cherry-pick. thanks, will fix.

rhc54: thx!

src/client/pmix_client_get.c
```diff
@@ -615,7 +615,18 @@ static void _getnb_cbfunc(struct pmix_peer_t *pr,
...
+            }
+            else {
```
rhc54, line 629: errrr...I don't think that is correct. The problem now is that if the second fetch succeeds, you won't extract the value from the `cb` struct so it can be passed down. I believe you could make this work if you replace this "else" with `if (PMIX_SUCCESS == rc)`, though you might need to check it to be sure.

rhc54: Yeah, looks good - thx!

src/client/pmix_client_get.c
```diff
@@ -615,7 +615,18 @@ static void _getnb_cbfunc(struct pmix_peer_t *pr,
                                 "pmix: get_nb searching for key %s for proc %s",
                                 cb->key, PMIX_NAME_PRINT(&proc));
             PMIX_GDS_FETCH_KV(rc, pmix_globals.mypeer, cb);
-            if (PMIX_SUCCESS == rc) {
+            if (PMIX_SUCCESS != rc) {
```
rhc54, line 618: Just curious - in the master branch, we dealt with the case where `PMIX_OPERATION_SUCCEEDED` was returned, but here we don't. Is that an oversight, or does this release branch never return that status?

awlauria: That I don't know, I was wondering the same thing, but have not looked into it.

awlauria:
bot:ibm:gnu:retest

awlauria:
bot:ibm:pgi:retest

src/client/pmix_client_get.c
rhc54: Probably safer to add the check - worst case it does nothing (as the hash never returns that code). However, if we ever backport something to the hash module that _does_ return "succeeded", it will save us chasing our tails trying to figure out what broke.

awlauria:
Sure, I make a small change to account for that.

rhc54 (approved):

PR #2544, opened 2022-03-24
pmix_cmd_line.c: For MPI tools, disable help output for arguments.

awlauria:
This was causing options such as '--hostfile h/-h' to fail, when they
should just work, IE 'h' and '-h' are valid hostfile names.

Refs: https://github.com/open-mpi/ompi/issues/10156

### DIFF
--- a/src/util/pmix_cmd_line.c
+++ b/src/util/pmix_cmd_line.c
@@ -98,6 +98,8 @@ int pmix_cmd_line_parse(char **pargv, char *shorts,
     char *ptr, *str, **argv;
     pmix_cmd_line_store_fn_t mystore;
 
+    bool is_mpi_tool = (NULL != getenv("OMPI_VERSION"));
+
     /* the getopt_long parser reorders the input argv array, so
      * we have to protect it here - remove all leading/trailing
      * quotes to ensure we are looking at simple options/values */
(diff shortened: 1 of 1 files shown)


jsquyres:
I don't really grok the layering of software here; I thought the solution would end up in Open MPI's schitzo component rather than needing to touch the core.  

To be clear: I'm really only advocating for Open MPI's `mpirun` behavior -- it's unfortunate that this also has to touch PMIx core.  Should there be a different / in-band way for schitzo components to indicate that they want this behavior (vs. relying on an out-of-band mechanism, such as the `OMPI_VERSION` env variable)?

jsquyres:
Yo @rhc54 -- @awlauria and I were chatting in Slack.  What if the OMPI schitzo component effectively copies `pmix_cmd_line.c` into itself (appropriately symbol-prefixed/etc.) and then modifies it for however OMPI wants to handle parsing?

Then OMPI's schitzo command line parsing decisions won't affect anyone else.

rhc54:
This is the better solution as the problem impacts all personalities: https://github.com/openpmix/openpmix/pull/2545

What you choose to do for the more general question (e.g., revert back to old parser) is up to you.

jsquyres:
Works for me.  We were just trying to avoid impacting the other personalities.

PR #2545, opened 2022-03-24
Remove the "<option> help" support

rhc54:
It has been correctly pointed out that a number of the cmd line
options could accept "help" as an input value as opposed to
meaning a request for "help". Until we can figure out a way to
separate those instances out from legitimate requests for help,
let's remove that method.

We therefore only accept two methods for requesting help:

- a directive to the cmd itself: e.g., "prte --help"

- an option to the help cmd line option: e.g., "--help map-by"

Hopefully, we will someday come up with a solution - perhaps
need to take a closer look at how "hydra" currently supports it.

Closes https://github.com/openpmix/openpmix/pull/2544

### DIFF
--- a/src/util/help-cli.txt
+++ b/src/util/help-cli.txt
@@ -12,7 +12,7 @@
 # This is the US/English general help file for PRTE CLI options.
 #
 [help]
-Help can be requested in three ways:
+Help can be requested in two ways:
 
 - a directive to the cmd. For example, the command
     "%s --help" or "%s -h"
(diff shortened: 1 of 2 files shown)


awlauria:
Shouldn't it also be removed on lines 194 and 311?

rhc54:
> Shouldn't it also be removed on lines 194 and 311 in pmix_cmd_line.c?

No and yes. No, you don't want to remove that earlier code block as that is what supports the e`--help <option>` request. There is no ambiguity there.

Yes, we do want to remove the latter code block as it parses, for example, `-h help` to be `--help help` while we want it to parse to `--host help` to specify a host with name "help". I've also fixed the problem where someone gives a short option and includes an argument when that option doesn't accept one.

awlauria (approved):

PR #2550, opened 2022-03-27
Support the common "-np" option

rhc54:
Seems to be pretty commonly used, so support it
as a special case.


rhc54:
bot:ibm:prrte:retest

PR #2551, opened 2022-04-05
Add PMIX_QUERY_ABI_VERSION key for PMIx_Query_info

jjhursey:
* `PMIx_Query_info` can now query the ABI level via the `PMIX_QUERY_ABI_VERSION` key
 * The `PMIX_QUERY_ABI_VERSION` is only resolvable locally to the caller and should not be sent to the server. So add a mechanism to, when this situation is identified, strip out this key from the query sent to the server. When the server response comes back then add the locally resolved information before passing it back to the client.
 * Add the PMIx Standard ABI version level to the `VERSION` file (similar to the PMIx Standard level)
   * Since the PMIx Standard ABI is not yet finalized set it to `0.0` for now. I'm assuming that we will change that to `5.0` once the ABI is ratified.
 * Ref
   * PMIx Standard PR: https://github.com/pmix/pmix-standard/pull/397
   * PMIx Standard ABI Discussion: https://github.com/pmix/pmix-standard/issues/365

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,6 +91,8 @@ examples/daemon_error_notify
 examples/launcher
 examples/log
 examples/showkeys
+examples/abi_no_init
+examples/abi_with_init
 
 include/pmix_version.h
 include/pmix_rename.h
--- a/VERSION
+++ b/VERSION
@@ -4,6 +4,7 @@
 # Copyright (c) 2013      Mellanox Technologies, Inc.
 #                         All rights reserved.
 # Copyright (c) 2014-2016 Intel, Inc. All rights reserved.
+# Copyright (c) 2022      IBM Corporation. All rights reserved.
 
 # This is the VERSION file for PMIx, describing the precise
 # version of PMIx in this distribution.  The various components of
@@ -43,6 +44,17 @@ greek=a1
 std_major=4
 std_minor=1
 std_extension=2
+
+# PMIx Standard ABI Compliance Level
+# The major and minor numbers indicate the version
+# of the official PMIx Standard ABI that is supported
+# by this release. Generally, the PMIx Standard ABI
+# should match the PMIx Standard Compliance Level
+# major/minor levels. However, they are listed separately
+# for maximal flexibility.
+# Since no PMIx Standard ABI exists at the moment, set to "0.0"
+std_abi_major=0
+std_abi_minor=0
 
 # If repo_rev is empty, then the repository version number will be
 # obtained during "make dist" via the "git describe --tags --always"
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -12,7 +12,7 @@ dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
 dnl Copyright (c) 2006-2021 Cisco Systems, Inc.  All rights reserved
 dnl Copyright (c) 2007      Sun Microsystems, Inc.  All rights reserved.
-dnl Copyright (c) 2009-2021 IBM Corporation.  All rights reserved.
+dnl Copyright (c) 2009-2022 IBM Corporation.  All rights reserved.
 dnl Copyright (c) 2009      Los Alamos National Security, LLC.  All rights
 dnl                         reserved.
 dnl Copyright (c) 2009-2011 Oak Ridge National Labs.  All rights reserved.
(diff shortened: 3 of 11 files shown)


rhc54:
Hmmm...that turned out much larger than I expected. I grok what you are doing with the local resolve, but I don't understand why the things that are returned via the GDS lookup aren't included in the "locally resolved" data. It seems like we should count them in the new tracking structures and not ask the server for those values - yes?

It also isn't clear to me what I should be returning for the "ABI version". If I have APIs and/or structures that extend the official v5.0 standard, then do I report "5.0+"? If I don't implement one or more APIs, or don't support some struct, do I report "5.0-"? If a subsequent release of my implementation adds yet another API, do I add another "+" to the ABI version? If I follow everything else but modify one structure, then do I report "NOTHING" - or is it "5.0ish"?

Just trying to understand what this version string really does for anyone. It seems to me that you still have to check for the presence of an API and/or structure, and there still is no guarantee that it will match the reported version. So far as I can see, if you are trying to interface to a library using `dlsym` semantics, then you are always at risk unless you try to mark every single API and struct - and then I still have to understand what to return if that entity is modified from the standard, and the tool has to figure out how to interpret what I said (or else maybe it's just "abort").

Note that none of the OpenPMIx release branches or the master branch exactly conforms to any PMIx Standard version - we have APIs and structs in all of them that extend the relevant standard. So none of them could report a pure ABI version number.

rhc54:
Just to be clear: I like what this PR does with respect to handling the locally resolved values. My question was really towards extending that slightly to handle the case where the values are retrieved via the GDS and therefore don't require help from the server. I honestly hadn't considered the case where someone would ask for the ABI version along with a set of other keys - thanks for taking that on!

My questions about what ABI version to return from OpenPMIx were truly intended as questions. I'm honestly not sure how to convey the actual ABI of the library relative to that of the Standard - or at least, how to do so in a way that has some meaning and value to a user. My suggestions aren't facetious, but were meant to try and pose some possibilities. I'm not in love with any of them, but I couldn't come up with anything better. Maybe you folks can?

jjhursey:
Yeah, it turned out a bit bigger than expected because the query API doesn't have good support for partially available local information

 I was curious about the GDS thing too. The code I added just before the `PMIX_GDS_FETCH_KV` for the ABI key worked if I asked just for the ABI key. If I asked for something else then it asked the server for all of the information and dropped anything that we had locally. The alternative would be to strip the locally found information and store it on the caddy to be filled in when it comes back from the server. Similar to what we do in a later step. We could probably do something at the `PMIX_GDS_FETCH_KV` call point too, but I didn't want to take that on with this change. I was afraid the patch would continue to grow and I was trying to keep it fairly focused. Something to maybe iterate on on top of these changes.

As for the ABI reporting that is definitely something we need to discuss (if you can join the IAWG meeting on Monday we can chat about it there). My thinking is that the PMIx Standard ABI represents a stable set of interfaces and whatnot with an associated version, say `5.0`.
 * **Case Equal**: An implementation that provides that stable set is compliant and can report `5.0`.
 * **Case Equal + Some Additions**: An implementation that adds to that set is still compliant (so if you add a new function foo) since anyone that relies on the set defined in the PMIx Standard ABI will still work (they just won't use the new stuff).
 * **Case Modified**: If an implementation modifies something in the ABI then it is no longer compliant (meaning they changed, for example, the signature of `PMIx_Init`) and cannot report `5.0`.
 * **Case Subset**: An implementation that provides only a subset of the ABI is not compliant and cannot report `5.0`. For API's that are missing then a `dlsym` should tell you if the symbol isn't there. So there is still a way to probe a bit to see what might be there, but you are not guaranteed that the signature is the same. We could also consider adding a `<5.0` notation for implementations that provide a subset of "5.0". I don't know how common this case will be.
 * Note that it is still fine if an implementation wants to return not supported for various APIs. They just need to define the symbols to be ABI compliant thus signaling a link/dlsym compatability.

So it's fine if OpenPMIx contains other interfaces. I think OpenPMix currently report the PMIx Standard level as something like 9.8.7 where the last digit signifies the delta from the "9.8" standard. I don't think the ABI needs that third digit.

rhc54:
> I was curious about the GDS thing too. The code I added just before the `PMIX_GDS_FETCH_KV` for the ABI key worked if I asked just for the ABI key. If I asked for something else then it asked the server for all of the information and dropped anything that we had locally. The alternative would be to strip the locally found information and store it on the caddy to be filled in when it comes back from the server. Similar to what we do in a later step. We could probably do something at the `PMIX_GDS_FETCH_KV` call point too, but I didn't want to take that on with this change. I was afraid the patch would continue to grow and I was trying to keep it fairly focused. Something to maybe iterate on on top of these changes.

Yeah, IIRC it kinda happened by default. Originally, I had the code always go to the server - idea being that you were querying about some server-level info. Then Artem wanted to use the interface to effectively do a `PMIx_Get` on multiple keys using a single call - I think he thought it would be faster that way. So we added the GDS fetch code, but I did it thinking that you would either be asking for system-level info (in which case you had to go up) or we were doing Artem's scenario (in which case you would not go up).

Doing the general case makes sense but would take more effort. I think your infrastructure should make it relatively easy to do, but I agree it can be done separately.

rhc54:
> As for the ABI reporting that is definitely something we need to discuss (if you can join the IAWG meeting on Monday we can chat about it there).

I can do that - added it to my calendar. You thinking makes sense to me.

rhc54:
We can figure out the right thing to return - regardless, this infrastructure looks like a good step forward to support this query, and for the backend implementation of the query function in general. Thanks!

PR #2560, opened 2022-04-13
Checkpoint gds/shmem work.

samuelkgutierrez:
Please note that this commit reimplements---internally within the gds
shmem component---some core PMIx functionality. The reason for this
approach is to first prototype the infrastructure required to implement
this component, with the ultimate goal of eventually finding a
reasonable way to reintegrate the duplicated infrastructure. This
approach minimizes sweeping changes to core infrastructure used by other
components, while also supporting forward progress in gds/shmem development.
Don't panic :).

### DIFF
--- a/src/class/pmix_object.h
+++ b/src/class/pmix_object.h
@@ -211,12 +211,29 @@ typedef void (*pmix_destruct_t)(pmix_object_t *);
 
 /* types **************************************************************/
 
-/* memory allocator for objects */
+/** Memory allocator for objects */
 typedef struct pmix_tma {
+    /** Pointer to the TMA's malloc() function. */
     void *(*malloc)(struct pmix_tma *, size_t);
+    /** Pointer to the TMA's calloc() function. */
+    void *(*calloc)(struct pmix_tma *, size_t, size_t);
+    /** Pointer to the TMA's realloc() function. */
     void *(*realloc)(struct pmix_tma *, void *, size_t);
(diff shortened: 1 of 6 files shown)


rhc54:
@samuelkgutierrez Looks like there is something not quite right in the class header:
```text
/github/workspace/src/class/pmix_object.h: In function 'pmix_tma_strdup':
/github/workspace/src/class/pmix_object.h:260:36: error: macro "strdup" passed 2 arguments, but takes just 1
         return tma->strdup(tma, src);
                                    ^
/github/workspace/src/class/pmix_object.h:260:9: warning: return from incompatible pointer type [enabled by default]
         return tma->strdup(tma, src);
```

samuelkgutierrez:
> @samuelkgutierrez Looks like there is something not quite right in the class header:
> 
> ```
> /github/workspace/src/class/pmix_object.h: In function 'pmix_tma_strdup':
> /github/workspace/src/class/pmix_object.h:260:36: error: macro "strdup" passed 2 arguments, but takes just 1
>          return tma->strdup(tma, src);
>                                     ^
> /github/workspace/src/class/pmix_object.h:260:9: warning: return from incompatible pointer type [enabled by default]
>          return tma->strdup(tma, src);
> ```

Hi @rhc54. Thank you. I don't think this compiler likes that the TMA has a function pointer named `strdup`. I'm fixing this now and will send a fix shortly.

PR #2566, opened 2022-04-16
Make pmix_common.h stand alone

rhc54:
When moving the now-deprecated macros to pmix_deprecated.h, we
had those macros invoke their replacement APIs. However, those
APIs are defined in pmix.h (where they belong). This causes users
who included only "pmix_common.h" in their files to break if they
invoked one of the now-deprecated macros.

Resolve this by including the function prototypes in pmix_deprecated.h.
Since we want them to be seen as true APIs, also include them
(in a protected form) in pmix.h, with comments documenting their
use.


rhc54:
bot:lanl:retest

PR #2583, opened 2022-05-14
Complete support for group-level info

rhc54:
Allow processes to contribute information during calls to
PMIx_Group_construct that are shared with all construct
participants. The values are returned to the participants
as "qualified" values, each value qualified by the assigned
group context ID.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,7 +103,8 @@ src/include/pmix_config.h
 src/include/pmix_config.h.in
 src/include/pmix_config.h.in~
 src/include/pmix_frameworks.h
-src/include/dictionary.h
+src/include/pmix_dictionary.c
+src/include/pmix_dictionary.h
 
 src/mca/pinstalldirs/config/pinstall_dirs.h
 src/mca/pdl/base/static-components.h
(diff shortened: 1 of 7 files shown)


rhc54:
@hppritcha Sorry for the long delay. This completes your requested support of group-specific information that is shared among "group construct" participants and returned as values qualified by the assigned group context ID. Note that I was able to implement this solely within PMIx itself - no changes are required to the RTE or host side.

It would be difficult, but not impossible, to backport this to PMIx v4.x. Biggest problem is the reliance on the "qualified value" support, which the "dstore" in those earlier versions cannot support. I would prefer not to try and tackle that backport and instead leave this for a v5 release. Please let me know if that is a major problem for you.

And again - thanks for your patience!

rhc54:
@hppritcha Please see `examples/group_lcl_cid.c` for an example of how to use this new feature. Note (as that example shows) that the qualified values provided in the group info persist after the group has been destructed - values are removed during the usual post-termination cleanup.

hppritcha:
great!  I'll try this out.

PR #2587, opened 2022-05-18
configure.ac: update directory space check

jsquyres:
PMIX_CANONICALIZE_PATH broke on MacOS when `/usr/bin/python`
disappeared (in favor of `/usr/bin/python3`) in MacOS Monterrey.  Make
things simpler by not canonicalizing the path to begin with.

Specifically: just look at `pwd`, $srcdir, and $prefix.  Even if
$srcdir and/or $prefix are relative to the build dir, the union of all
the checks will ensure that none of the target paths will contain
spaces.

Back-ported from
https://github.com/open-mpi/ompi/commit/6992c642437af1135ccb6f159683853fe63c58cc

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -10,7 +10,7 @@
 #                         University of Stuttgart.  All rights reserved.
 # Copyright (c) 2004-2005 The Regents of the University of California.
 #                         All rights reserved.
-# Copyright (c) 2006-2020 Cisco Systems, Inc.  All rights reserved
+# Copyright (c) 2006-2022 Cisco Systems, Inc.  All rights reserved
 # Copyright (c) 2006-2008 Sun Microsystems, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


rhc54:
Looks like @bwbarrett approved these over on OMPI - since he's the author of the removed code, I'll take his approval over there as indicative that this does what he wanted.

rhc54:
Thank _you_ for the assist! I never saw the problem because MacPorts puts a `python` in my default path.

Definitely should go to the v4.x branches. We do still have people using the v3 series (due to some Slurm configure issues), so probably should do v3.2 if you don't mind.

jsquyres:
> Definitely should go to the v4.x branches. We do still have people using the v3 series (due to some Slurm configure issues), so probably should do v3.2 if you don't mind.

Made PRs #2588, #2589, #2590 for the v4.x branches.  Looks like this code does not exist on the v3.2 branch, so no PR was needed there.

PR #2591, opened 2022-05-18
pmix_fd: cap the max FD to try to close

jsquyres:
On some OS's (e.g., macOS), the value returned by
sysconf(_SC_OPEN_MAX) can be set by the user via "ulimit -n X", where
X can be -1 (unlimited) or a positive integer. On macOS in particular,
if the user does not set this value, it's unclear how the default
value is chosen.  Some users have reported seeing arbitrarily large
default values (in the billions), resulting in a very large loop over
close() that can take minutes/hours to complete, leading the user to
think that the app has hung.  To avoid this, ensure that we cap the
max FD that we'll try to close.  This is not a perfect scheme, and
there's uncertainty on how the macOS default value works, so we
provide the pmix_maxfd MCA var to allow the user to set the max FD
value if needed.

This commit is inspired by https://github.com/open-mpi/ompi/pull/10360
and https://github.com/open-mpi/ompi/issues/10358.

Thanks to Scott Sayres for raising the issue.

### DIFF
--- a/src/runtime/pmix_params.c
+++ b/src/runtime/pmix_params.c
@@ -12,7 +12,7 @@
  *                         All rights reserved.
  * Copyright (c) 2006      Los Alamos National Security, LLC.  All rights
  *                         reserved.
- * Copyright (c) 2008-2015 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2008-2022 Cisco Systems, Inc.  All rights reserved.
  * Copyright (c) 2009      Oak Ridge National Labs.  All rights reserved.
  * Copyright (c) 2010-2014 Los Alamos National Security, LLC.
  *                         All rights reserved.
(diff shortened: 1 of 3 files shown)


rhc54:
Ick, that would be bad! Thanks for the fix! Ordinarily, we would need to port this back to PRRTE as well, but I'm working on updating the PMIx and PRRTE release branches to at least pickup the PMIx utilities so we don't have to keep multi-fixing things. For now, I think this is good enough.

jsquyres:
BTW, should I back-port this to the v4.x and v3.2 branches?

rhc54:
Yeah, my bad - sorry for the goose chase. The local fork/exec feature entered in the v4.x series, and so there was no need to close fd's prior to that point.

I have PMIx v4.2 now supporting all the PRRTE branches, so I'll commit that shortly. Includes this change. I'll start looking at the PMIx v4.1/0 series next.

PR #2619, opened 2022-06-12
Checkpoint gds/shmem work.

samuelkgutierrez:
For a single client it appears that the sharing of hash tables and lists
works. pmix_gds_shmem_fetch() requires work to accommodate multiple
clients.

Other notable changes include:
* Revert changes introduced in 'TMA: Integrate more custom memory allocator
  infrastructure'. The base object system doesn't appear to require the
  changes introduced in that commit.

### DIFF
--- a/src/class/pmix_object.c
+++ b/src/class/pmix_object.c
@@ -66,13 +66,13 @@ static const int increment = 10;
 /*
  * Local functions
  */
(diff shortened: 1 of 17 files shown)


rhc54:
Sounds like good progress - hooray!

PR #2634, opened 2022-07-02
Update show-help system

rhc54:
Allow authors to include a link to another file/topic
in their descriptions, with "show_help" picking up
the referenced info and inserting it into the output
at that point. Syntax is:

```
#include#filename#topic
```

Remove dependency on Flex for this subsystem.


rhc54:
@jsquyres You might have interest in this feature over there.

PR #2648, opened 2022-07-23
Minor fixes to the build process

jsquyres:
Ran across some minor build issues with the OpenPMIx master when building Open MPI's main branch on an old RHEL 6 machine (with Python 2).

See the commit messages in the 2 individual commits for details.


rhc54:
Thanks - much appreciated!

PR #2652, opened 2022-07-25
v4.1: Python bindings fixes

awlauria:
@rhc54 just bringing this over at Jeff's request. I noticed you already have a PR opened for v4.2 that includes these. If you have plans to bring this in a large sync feel free to close - or if you have no intention on developing this branch further.

### DIFF
--- a/contrib/construct_dictionary.py
+++ b/contrib/construct_dictionary.py
@@ -1,6 +1,6 @@
 #
 # Copyright (c) 2020      Intel, Inc.  All rights reserved.
-# Copyright (c) 2020      Cisco Systems, Inc.  All rights reserved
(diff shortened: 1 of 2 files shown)


rhc54:
Yeah, I'll pick it up with the bigger PR - right now, that one is failing in the debugger stuff. I need to check what PRRTE branch that CI runs against. If it is 2.1, then that would explain the problem.

awlauria:
Sounds good, thanks.

 Just to make sure, this PR is for v4.1  not v4.2. Is this going in a bigger PR for v4.1 (or not applicable anymore?)

rhc54:
Ah crud - I didn't see it was v4.1. Let's hold off as I need to reset that branch to shed too many of the v4.2 changes. We want to restrict it to just bug fixes. I'll sweep this up when I do that update.

PR #2655, opened 2022-07-28
Forward stdin to apps directly fork/exec'd by PMIx

rhc54:
We allow tools to directly fork/exec intermediate
launchers via the PMIx library, so we need to support
forwarding of any stdin for those tools to the spawned
children.

### DIFF
--- a/src/client/pmix_client_spawn.c
+++ b/src/client/pmix_client_spawn.c
@@ -261,6 +261,11 @@ PMIX_EXPORT pmix_status_t PMIx_Spawn_nb(const pmix_info_t job_info[], size_t nin
     /* if we are a server, then process this ourselves */
     if (PMIX_PEER_IS_SERVER(pmix_globals.mypeer) &&
         !PMIX_PEER_IS_LAUNCHER(pmix_globals.mypeer)) {
+
+        if (NULL == pmix_host_server.spawn) {
+            return PMIX_ERR_NOT_SUPPORTED;
+        }
+
         cd = PMIX_NEW(pmix_setup_caddy_t);
         if (NULL == cd) {
             return PMIX_ERR_NOMEM;
--- a/src/common/pmix_iof.c
+++ b/src/common/pmix_iof.c
@@ -1585,6 +1585,10 @@ void pmix_iof_write_handler(int sd, short args, void *cbdata)
         if (0 == output->numbytes) {
             /* don't reactivate the event */
             PMIX_RELEASE(output);
+            if (2 < wev->fd) {  // close the channel
+                close(wev->fd);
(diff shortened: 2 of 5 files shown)


rhc54:
bot:ibm:gnu:retest

rhc54:
bot:ibm:gnu:retest

rhc54:
@jjhursey Can I get some help on this? This might be a chance to try and chase down a real problem. The GNU test keeps failing when running "hostname" across 5 nodes using `ppr:5:node`. It always fails, yet the failures do not repeat in detail - one of the nodes always fails to generate any output, but which node changes run-to-run. So with 5 nodes, we always get four nodes of output.

I cannot reproduce this behavior, though I'll keep trying to find some way to do so. Can you perhaps dig into it a bit and see where the problem might lie? I'd suggest adding `--display map` to ensure things actually got mapped as a starting point, and adding `--leave-session-attached` in case a daemon is trying to report a problem.

Let me know if you don't have time and I'll try to hack up the CI scripts to generate debug output. Just don't want to do that as in-person is always better, and I dislike banging those scripts around and then reverting everything.

jjhursey:
I'll take a look today.

jjhursey:
bot:ibm:retest

jjhursey:
I was able to reproduce consistently with:
```
prterun --map-by ppr:5:node --hostfile /opt/mpi/etc/hostfile hostname < /dev/null 
```
It produces a segv. I'm having trouble getting the core file for some reason.

Without the `< /dev/null ` it runs fine.

jjhursey:
```
(gdb) bt
#0  0x00007fffb60fbf88 in pmix_iof_read_local_handler () from /opt/ci/support/exports/pmix/lib/libpmix.so.0
#1  0x00007fffb5f39e24 in event_process_active_single_queue (base=base@entry=0x1000ceb3110, activeq=0x1000ceb35e0, 
    max_to_process=max_to_process@entry=2147483647, endtime=endtime@entry=0x0) at event.c:1691
#2  0x00007fffb5f3a9b0 in event_process_active (base=0x1000ceb3110) at event.c:1783
#3  event_base_loop (base=0x1000ceb3110, flags=<optimized out>) at event.c:2006
#4  0x00007fffb60cbc0c in progress_engine () from /opt/ci/support/exports/pmix/lib/libpmix.so.0
#5  0x00007fffb5ab9718 in start_thread () from /lib64/libpthread.so.0
#6  0x00007fffb59cab58 in clone () from /lib64/libc.so.6
```

Not a debug build, but at least points to the function. Let me know if you need more information.

rhc54:
Thanks Josh - the `< /dev/null` was the key to reproducing it.

PR #2657, opened 2022-07-29
Protect "create" macros from zero entries

### LINKED ISSUE #2656: PMIX_INFO_CREATE(m, 0) not supported
master branch

-----------------------------

## Details of the problem

The macro `PMIX_INFO_CREATE` does either not support `n=0` that is not forbidden by the standard at least or it relies on `calloc(0)` to return `NULL` which is not necessarily the case according to the manpage:

> The  calloc() function allocates memory for an array of nmemb elements of size bytes each and returns a pointer to the allocated memory.  The memory is set to zero.  If nmemb or size is 0, then calloc() returns either NULL, or a unique pointer value that can later be successfully passed to free(). 

```
#define PMIX_INFO_CREATE(m, n)                                  \
    do {                                                        \
        pmix_info_t *_i;                                        \
        (m) = (pmix_info_t*)pmix_calloc((n), sizeof(pmix_info_t));   \
	    if (NULL != (m)) {                                      \
            _i = (pmix_info_t*)(m);                             \
            _i[(n)-1].flags = PMIX_INFO_ARRAY_END;              \
        }                                                       \
    } while (0)
```

I would have proposed to change it to 
```
        (m) = (n) ? (pmix_info_t*)pmix_calloc((n), sizeof(pmix_info_t)) : NULL;   \
```
but I an not certain enough about all implications to directly open a pull request in this case.

In my opinion, supporting `n=0` makes sense, since it enables using the macro like that and so ease things:
```
size_t ninfo = 0;
pmix_info_t *info;

if (cond1) ninfo++;
if (cond2) ninfo++;

PMIX_INFO_CREATE(info, ninfo);

size_t i = 0;
if (cond1) PMIX_INFO_LOAD(&info[i], ...);
if (cond2) PMIX_INFO_LOAD(&info[i], ...);

PMIx_Whatever(...,info, ninfo, ...);

PMIX_INFO_FREE(info, ninfo);
```

rhc54:
Passing zero for the number of objects to create
isn't explicitly forbidden, so protect the
macros from that use-case.

Fixes #2656 

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -2188,16 +2188,24 @@ typedef struct pmix_coord {
     .dims = 0                       \
 }
 
-#define PMIX_COORD_CREATE(m, d, n)                                      \
-    do {                                                                \
-        pmix_coord_t *_m;                                               \
-        _m = (pmix_coord_t*)pmix_calloc((d), sizeof(pmix_coord_t));     \
-        if (NULL != _m) {                                               \
-            _m->view = PMIX_COORD_VIEW_UNDEF;                           \
-            _m->dims = (n);                                             \
-            _m->coord = (uint32_t*)pmix_calloc((n), sizeof(uint32_t));  \
-            (m) = _m;                                                   \
-        }                                                               \
+#define PMIX_COORD_CREATE(m, d, n)                                              \
+    do {                                                                        \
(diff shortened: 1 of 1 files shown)


rhc54:
bot:ibm:gnu:retest

PR #2660, opened 2022-07-30
Fix retrieval of node/app/session-realm info

rhc54:
Per the Standard, one can retrieve information about a process
from the node/app/session realms in two ways:

(a) by passing the proc ID along with a key that belongs to
    one of those realms. The PMIx library is responsible
    for identifying such keys and retrieving the specified
    info from the corresponding default realm. If the user
    wishes the key to be retrieved from a non-default realm,
    then they must include the attribute specifying that realm.

(b) by passing an invalid proc rank (e.g., wildcard or undef)
    and including attributes that specify the realm from which
    the given key is to be retrieved along with whatever
    further identification is required - e.g., PMIX_NODE_INFO
    combined with the hostname of the node whose information
    is being fetched.

This commit fixes both methods and provides an example (nodeinfo)
in the examples directory for using this feature.

(cherry picked from commit 77fc96c5a045060810d23ba8080c62fbc074aefe)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,6 +94,7 @@ examples/showkeys
 examples/abi_no_init
 examples/abi_with_init
 examples/group_lcl_cid
+examples/nodeinfo
 
 include/pmix_version.h
 include/pmix_rename.h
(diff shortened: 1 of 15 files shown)


rhc54:
Failure of xversion with master is known and expected as we are trying to update master. The GNU failure is unrelated and persistent - will be investigated separately.

PR #2667, opened 2022-08-06
Add helper macro, sort proc arrays, and fix multi-node group info

rhc54:
[Add helper macro](https://github.com/openpmix/openpmix/commit/288418b8dd2d84c47bfefd7d62b45f8caefab562)

...for checking GDS component of a peer

[Sort proc arrays to remove order sensitivity and fix multi-node grp info](https://github.com/openpmix/openpmix/commit/2496ae8a43eb433c0c11767674ab4700aff2e43d)

Remove the ordering sensitivity when doing collective operations
across arrays of procs as this is causing problems for users.
Indeed, backends are sorting the arrays given to them precisely
for this reason - so let's do it as a service to avoid future
complications.

Group info needs to be stored as "global" scope so it can be
properly retrieved.

### DIFF
--- a/examples/group_lcl_cid.c
+++ b/examples/group_lcl_cid.c
@@ -188,10 +188,10 @@ int main(int argc, char **argv)
     if (NULL != results) {
         cid = 0;
         PMIX_VALUE_GET_NUMBER(rc, &results[0].value, cid, size_t);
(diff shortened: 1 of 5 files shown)


rhc54:
bot:ibm:gnu:retest

Issue #2676, opened 2022-08-12
configure can't handle the case that $PREFIX/lib64 is a symlink to $PREFIX/lib


lyu:
### What version of the PMIx Reference Library are you using?
`4.2.0rc1`

### Describe how PMIx was installed 
From release tarball

### Please describe the system on which you are running

* Operating system/version: Rocky Linux 8.4 aarch64

-----------------------------

## Details of the problem
As the title says, when `$PREFIX/lib64` is a symlink to `$PREFIX/lib`, configure confuses itself
```
checking for libevent library (event_core) in /lustre/home/user/.local... configure: error: Found library event_core in both /lustre/home/user/.local/lib and
/lustre/home/user/.local/lib64.  This has confused configure.  Please add --with-libevent-libdir=PATH to configure to help disambiguate.
```
This also happens for prrte 3.0.0rc1
Some linux distros also do this symlink (e.g. Arch). Maybe pmix/prrte can check the actual `libdir` field of the `.pc` files?

rhc54:
Yick - that symlink doesn't sound like a good idea! It breaks the convention that `lib` is where you'll find 32-bit libraries, so this is going to cause all kinds of problems for library developers.

I'll have to ponder that one. For now, you should be able to do as the error message indicates and tell configure the actual path you want it to use.

lyu:
Well, if I understand it correctly, Arch is the one who is trying to adhere to https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard#Directory_structure, but I get your point. The suggested configure options do work.

rhc54:
Interesting - I hadn't seen that one. Thanks for passing it along! I can see the sense in what they did, although I must admit that I haven't seen it in the wild (or at least we didn't realize it!). The symlink must be something new, though, as we've been building fine until now.

I'll look into the configure logic - we should be able to handle the symlink better than we do, and I'll check into your suggestion as well.

rhc54:
@lyu Could you please try https://github.com/openpmix/openpmix/pull/2677 and let me know if that fixes the symlink issue?

lyu:
@rhc54 Unfortunately it doesn't, configure errors out with the same error message.
I have libevent, hwloc and zlib installed under `$HOME/.local`, do you need other info that I can provide?

rhc54:
I've updated that patch - could you please give it another try? I had fixed one spot but missed that there was another. I have them both fixed now.

lyu:
[...]

PR #2677, opened 2022-08-13
Do not error out if lib is symlinked to lib64

rhc54:
Distros are now creating a symlink between /lib and /lib64, and
between /usr/lib and /usr/lib64, in accordance with recent
agreements on file system conventions. Thus, if we detect that
the package file exists in both places, check that the two
locations are not symlinks before declaring ambiguity and
exiting with an error.

Fixes https://github.com/openpmix/openpmix/issues/2676


rhc54:
Confirmed to solve the problem both by the original reporter and an offlist packager.

PR #2689, opened 2022-08-17
Need to replace the entire proc in fence with group member

rhc54:
Replacing only the nspace leaves the group rank instead of
the proc's nspace rank, so we need to replace the entire
proc entry with the grp member's ID.

(cherry picked from commit 5ea332842a2b5ec7c6b8675c119e0a27697fed5c)


rhc54:
bot:ibm:retest

PR #2694, opened 2022-08-19
Output IO as singleton, support background commands

rhc54:
When running as a singleton, setup to handle any IO from spawned
children. Do not identify an ampersand as a cmd line "tail" as
this falsely implies it is an application.


rhc54:
bot:ibm:retest

PR #2698, opened 2022-08-21
Fix the buildrpm script

rhc54:
Current code never finds home directory RPMBUILD, so
simplify it so it works.


rhc54:
bot:ibm:retest

PR #2717, opened 2022-08-25
Release GIL before registering event handler

MattBBaker:
Newfound bug can result in deadlocks in the Python bindings when calling `PMIx_Register_event_handler()`. Add change to drop GIL before registering a new event handler.


rhc54:
This is just a testing sync issue - I'll take care of it in a bit.

rhc54:
bot:ibm:retest

Issue #2711, opened 2022-08-24
Potential SEGFAULT in gds/hash after checking session info


krempel-pt:
Version: 4.2.0

I hit another problem, this time leading to a segfault in pmix:

Somehow I generated a mismatch of session id running into the following path in `pmix_gds_hash_check_session()`

https://github.com/openpmix/openpmix/blob/81c14d7df04bbc3a2048f936038b61d07e066353/src/mca/gds/hash/gds_utils.c#L169-L175

This has been called by `pmix_gds_hash_process_session_array()`:

https://github.com/openpmix/openpmix/blob/81c14d7df04bbc3a2048f936038b61d07e066353/src/mca/gds/hash/process_arrays.c#L598-L600

`sptr` then is not checked to be not NULL but later is dereferenced:

https://github.com/openpmix/openpmix/blob/81c14d7df04bbc3a2048f936038b61d07e066353/src/mca/gds/hash/process_arrays.c#L625

For completeness here is the stack trace:
```gdb
Program terminated with signal SIGSEGV, Segmentation fault.                                                                                                                                                                                                                                                                                           
#0  0x00007fccc9a11910 in _pmix_list_append (list=0xb0, item=0x7fccc0004380, FILE_NAME=0x7fccc9ac2230 "../../../../../src/mca/gds/hash/process_arrays.c", LINENO=625) at PMIx/pmix-4.2.0/src/class/pmix_list.h:564                                                                                                    
564         item->pmix_list_prev = sentinel->pmix_list_prev;                                                                                                                                                                                                                                                                                          
[Current thread is 1 (Thread 0x7fccc89ff640 (LWP 26971))]                                                                                                                                                                                                                                                                                             
(gdb) bt       
#0  0x00007fccc9a11910 in _pmix_list_append (list=0xb0, item=0x7fccc0004380, FILE_NAME=0x7fccc9ac2230 "../../../../../src/mca/gds/hash/process_arrays.c", LINENO=625) at PMIx/pmix-4.2.0/src/class/pmix_list.h:564
#1  0x00007fccc9a18a85 in pmix_gds_hash_process_session_array (val=0x1235630, trk=0x7fccc0007c70) at ../../../../../src/mca/gds/hash/process_arrays.c:625
#2  0x00007fccc9a05daf in hash_cache_job_info (ns=0x7fccc0001ee0, info=0x1235200, ninfo=7) at ../../../../../src/mca/gds/hash/gds_hash.c:198
[...]

PR #2719, opened 2022-08-26
Make the session info array support conform to the standard

rhc54:
It was noted that the session ID doesn't have to be the first
item in the array, and that one can indeed pass multiple
session info arrays in "register nspace" - only one can have
a non-default (UINT32_MAX) ID as that assigns it to the job.
Also noted that all jobs should have access to not only the
info from a non-default ID assigned to that job, but also to
all info assigned to the default ID as that represents
info available to all.

Thanks to @krempel-pt for pointing these out and providing
a reproducer.

Fixes https://github.com/openpmix/openpmix/issues/2711


rhc54:
bot:ibm:retest

PR #2731, opened 2022-08-31
Do not set the buffer type in construct

rhc54:
Need to let the first pack command set the buffer type
so it gets set to that of the remote side. Otherwise,
you get an error if the other side is different than that
of the sender.

(cherry picked from commit 0c94393c332f624ae533144b622795acef604d67)


rhc54:
Looks like I got it!

jjhursey:
Thanks. My testing must have missed that one.

rhc54:
You had to hit a very specific scenario, which is why it took a little head-scratching to understand. However, the error message was spot on, as usual for Python, so really all I had to do was RTFM to fix it.

PR #2738, opened 2022-09-06
Convert the MCA parameter "pmix_mca_base_component_show_load_errors"

rhc54:
to be a flexible mechanism to specify when (and when not) to emit
warnings about errors when trying to load DSO components.

Port of https://github.com/open-mpi/ompi/pull/10763

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -10,7 +10,7 @@ dnl Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
 dnl                         University of Stuttgart.  All rights reserved.
 dnl Copyright (c) 2004-2005 The Regents of the University of California.
 dnl                         All rights reserved.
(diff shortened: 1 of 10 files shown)


jsquyres:
@rhc54 Be aware that https://github.com/open-mpi/ompi/pull/10763 is still a proposal; I don't know if that's what Open MPI is ultimately going to go with or not.

rhc54:
I know - however, it looks reasonable to me and we are going to need something like it for the same reasons you cited. We already have complaints about the library requirement from Cray users about PRRTE. We don't actually link against libraries here (at least, so far) in the GPU, RM, and fabric frameworks - but we can see the day coming.

If we diverge on this it is probably okay. I suspect we will track what you do for the most part so that your users don't find it all too confusing (i.e., you have to do something for OMPI and then something else for PMIx and again for PRRTE since that is your mpirun).

PR #2740, opened 2022-09-06
Revert deletion of common Makefile.am required by autogen.pl

cniethammer:
The missing common Makefile.am resulted in the following error from autogen.pl:
> Missing src/mca/common/Makefile.am at ./autogen.pl line 96.

### DIFF
--- a/src/mca/common/Makefile.am
+++ b/src/mca/common/Makefile.am
@@ -1,0 +1,25 @@
+#
+# Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
+#                         University Research and Technology
+#                         Corporation.  All rights reserved.
+# Copyright (c) 2004-2005 The University of Tennessee and The University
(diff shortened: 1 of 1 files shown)


rhc54:
I don't think we want to have a "dead" file in the source. The right answer really is to modify `autogen.pl` so it doesn't _require_ the mca/common directory to exist. Would you mind changing this to address that approach? Or if you prefer, I'll take care of it.

Note that we aren't seeing any problems with `autogen.pl`, so I'm wondering if the issue here is that you need to clean your cloned directory tree to remove a stale fragment? The directory should not exist today.

cniethammer:
OK, got it.
By default I always clean my checkouts with `git clean -df`. However, this did not succeed in removing the `common/` directory here as there is a left over entry in the .gitignore file for `src/mca/common/sse/sse_lex.c`.

I'll update the PR to clean the `.gitignore` in a moment.

PR #2741, opened 2022-09-06
Clean up left over .gitignore entry after mca/common/* removal

cniethammer:
Clean up left over .gitignore entry after mca/common/* removal in 4f8a4bdc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,7 +111,6 @@ src/include/pmix_dictionary.h
 src/mca/pinstalldirs/config/pinstall_dirs.h
 src/mca/pdl/base/static-components.h
 src/mca/pinstalldirs/base/static-components.h
-src/mca/common/sse/sse_lex.c
 
 src/tools/pattrs/pattrs
 src/tools/pevent/pevent


rhc54:
This is really odd. I just did a fresh clone of the repo, and I don't see any src/mca/common directory in it. Can you checkout a fresh clone and verify you see this problem?

rhc54:
Looking at your change, I'm even more confused. `pmix_dictionary.h` is a generated file and is not part of the git repo. You therefore cannot edit/change it via PR.

cniethammer:
> This is really odd. I just did a fresh clone of the repo, and I don't see any src/mca/common directory in it. Can you checkout a fresh clone and verify you see this problem?
A fresh clone does work for me as well and does not include a common directory. So that is fine.
Problem was as it is a submodule in open MPI that I had to update...
As said it was just a problem with the .gitignore file for me at the end, that kept the directory to be cleaned.

cniethammer:
> Looking at your change, I'm even more confused. `pmix_dictionary.h` is a generated file and is not part of the git repo. You therefore cannot edit/change it via PR.

I am confused ... I did not include this file in the PR looking at https://github.com/openpmix/openpmix/pull/2741/files

rhc54:
I see what this does - not sure why it would have anything to do with clean operation, but I have no problem with updating the global ignore file.

PR #2759, opened 2022-09-15
Update Github actions to use submodules

jjhursey:
Using https://github.com/marketplace/actions/checkout-submodules

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
+      with:
+        submodules: recursive
     - name: Build
       uses: ./.github/actions/mlnx
       with:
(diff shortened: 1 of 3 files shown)


rhc54:
Yeah, this is probably the right approach. My changes to `pmix-tests` didn't accomplish anything, so I'll revert them.

jjhursey:
You can probably keep them in pmix-tests since it's going to be the 'right thing to do' if a user is looking at the code.

rhc54:
I'll leave them in for now so long as they "do no harm" - can revert later if needed

rhc54 (approved):

rhc54:
so why are the IBM ones failing "during PREP"?

jjhursey:
bot:ibm:retest

jjhursey:
Looks like a corner case. I just pushed a fix. Should be good now.

PR #2762, opened 2022-09-20
llvm/oneapi: fixes to bring pmix up to iso c99

hppritcha:
compliance at least as far as the LLVM on which the Intel OneAPI 2022 release is based (14).  LLVM is getting pickier about what it lets through in terms of ISO C99 compliance when using -Werror, etc.

Related to https://reviews.llvm.org/D122983

Without this patch and without disabling developer flags (which is very difficult when building within spack without adding yet another variant) one gets these sorts of errors at compile time.

pmix_net.c:168:22: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]

threads/thread.c:109:27: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes] int pmix_tsd_keys_destruct()
                          ^
                           void

base/bfrop_base_frame.c:68:13: error: variable 'count' set but not used [-Werror,-Wunused-but-set-variable]
        int count = 0;

Signed-off-by: Howard Pritchard <hppritcha@gmail.com>

### DIFF
--- a/src/mca/bfrops/base/bfrop_base_frame.c
+++ b/src/mca/bfrops/base/bfrop_base_frame.c
@@ -63,10 +63,10 @@ int pmix_bfrops_base_output = 0;
 
 static int pmix_bfrop_register(pmix_mca_base_register_flag_t flags)
 {
+    PMIX_HIDE_UNUSED_PARAMS(flags);
+
     if (PMIX_MCA_BASE_REGISTER_DEFAULT == flags) {
         /* do something to silence warning */
-        int count = 0;
-        ++count;
     }
     pmix_bfrops_globals.initial_size = PMIX_BFROP_DEFAULT_INITIAL_SIZE;
     pmix_mca_base_var_register("pmix", "bfrops", "base", "initial_size", "Initial size of a buffer",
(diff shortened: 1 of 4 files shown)


rhc54:
Yes, FWIW we are trying to "scrub" all those things. Sadly, different LLVM compilers (either version or vendor) yield very different results, so it is a game of "whackamole" to ferret them all out. Thanks for contributing the oneapi ones!

rhc54:
@hppritcha Looks like something is stuck on the LANL CI - can you take a look?

src/mca/bfrops/base/bfrop_base_frame.c
```diff
@@ -65,8 +65,6 @@ static int pmix_bfrop_register(pmix_mca_base_register_flag_t flags)
 {
     if (PMIX_MCA_BASE_REGISTER_DEFAULT == flags) {
         /* do something to silence warning */
```
rhc54, line 67: You'll need to change this or we'll get errors from other environments. Just change that little code block to:
```c
PMIX_HIDE_UNUSED_PARAMS(flag);
```

hppritcha:
yes looks like the jenkins update borked up remote agents.

hppritcha:
i disabled the project but doesn't look like i can cancel the pending ci run.

rhc54:
Got it. We need that one change I flagged, and that should restart the CI anyway, yes?

hppritcha:
updated with the hide parameter macro.

PR #2763, opened 2022-09-21
Add action to check commits on pull requests

Joe-Downs:
The action checks each commit in a pull request to see if

* the committer / author do not match bad patterns ("root@", "localhost", etc.)
* a "Signed-off-by:" line exists
* a "cherry picked from" message appears, if required by config inputs (not required by default)
  * if the message exists, ensure the referenced commit exists

If all commit checks pass, the action successfully completes and the action check for the PR passes. If any commit check fails, the action check fails as well. Additionally, a comment is posted to the pull request, describing which commits failed which test and suggesting the user fix them.

Signed-off-by: Joe Downs <joe@dwns.dev>

-------
### Comment Example:
image

-------

@rhc54 Finally got the ompi commit checker moved to [its own repo](https://github.com/open-mpi/pr-git-commit-checker) as [you requested](https://github.com/open-mpi/ompi/pull/10789). It should be ready to go right now, however, I'm not sure if you or anyone else wants to make some changes. For the most part, these don't affect the operation of the action, but maintainers may want to change for personal taste / project requirements.

1. Filename (`pr-target.yaml`)
2. Action name (`PR Checks CI`)
3. Tracking reference (`@v1.0.0`)

Any of these can be changed if needed. I just gave them names that I thought make sense. I think probably the biggest point of discussion is no. 3, and perhaps @jsquyres has some thoughts on it.

I set it up to track the [v1.0.0 release](https://github.com/open-mpi/pr-git-commit-checker/releases/tag/v1.0.0). Which is of course the most stable, but means the file needs to be updated if we ever add more to the action. To save that (admittedly minor) trouble, I think it would be best to instead track a `v1` branch. That way, the YAML file doesn't have to be updated to make use of any changes. Honestly, we may never add anything extra, but maybe there could be some QoL updates in the future, or other small additions.

Lastly, do you also want the labeler for this repo and PRRTE?  If so, I can easily add that to this YAML. (It is in [its own repo](https://github.com/open-mpi/pr-labeler) as well). However, it might need a minor tweak, due to this repo's difference in label names.

### DIFF
--- a/.github/workflows/pr-target.yaml
+++ b/.github/workflows/pr-target.yaml
@@ -1,0 +1,28 @@
+name: PR Checks CI
+
+# We're using pull_request_target here instead of just pull_request so that the
+# action runs in the context of the base of the pull request, rather than in the
+# context of the merge commit. For more detail about the differences, see:
+# https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
+on:
+    pull_request_target:
+        # We don't need this to be run on all types of PR behavior
+        # See https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request
+        types:
+          - opened
+          - synchronize
+          - edited
+
+permissions: {} # none
+
+jobs:
+  check:
+    permissions:
+      pull-requests: write
+    name: Check Commits
+    runs-on: ubuntu-latest
+    steps:
+      - name: Pull Request Commit Checker
+        uses: open-mpi/pr-git-commit-checker@v1.0.0
+        with:
+          token: "${{ secrets.GITHUB_TOKEN}}"


rhc54:
The labeler would also be nice to have - much appreciate the assist here! Would it make sense to put these yaml files in the same repo? Just thinking of logistical ease.

Joe-Downs:
I'm not sure I understand what you mean by "same repo." Together in a separate repo?

rhc54:
Sorry - it is irrelevant to this PR. When looking at the open-mpi project area, it is apparent that you are storing each "checker" setup in its own repository. I was only asking about bundling them together in one repo, though I later wondered if that would mean taking "all or nothing" when adopting them. Probably not worth the consideration.

Thanks again for the assist!

jsquyres:
@rhc54 I think I understand what you're asking.  We just recently put the different github actions in separate repos so that we can export them as Github Marketplace Actions.  They're 3 different things, so we expose them as 3 different Github Actions.  As Marketplace Actions, we can basically just reference that Github Marketplace Action in each of OMPI / PMIx / PRRTE repos without needing to actually keep the Python code synchronized between all the repos (e.g., in this PR as `uses: open-mpi/pr-git-commit-checker@v1.0.0`).

Github Actions don't work via Git Submodules, either (i.e., we can't have a Git submodule in `.github/workflows` that GitHub will automatically treat as Github Actions).  So Marketplace Actions -- and therefore separate repos -- seemed like the Right Way to avoid all the code duplication.

Make sense?

jsquyres:
@rhc54 Also, now that this PR is merged, once you see it run successfully a few times, you can remove the "Commit email checker" and "Signed-off-by checker" github web hooks on this repo:

Screen Shot 2022-09-21 at 9 49 18 AM

rhc54:
Got it - and thanks for the explanation!

Joe-Downs:
So @rhc54, I can add the labeler as well, but there's a slight hiccup. In the OMPI repo, the `Target:` labels match the branch names exactly. (e.g., `Target: v5.0.x` goes with branch `v5.0.x`) All the labeler does is concatenate `Target: ` with the branch name and add that as a label.

If we add it here, either the labels need to be changed, or I can make the labeler a bit more fuzzy. I think changing the labels would be simpler, but I'm not sure if you're okay with that.

rhc54:
Changing the labels is easier and fine with me. I'll do them in the morning.

PR #2769, opened 2022-09-22
build: Update to use OAC_C_COMPILER_VENDOR

bwbarrett:
Update to use OAC version of COMPILER_VENDOR rather than the internal version.  The OAC version only exports the cache variable, rather than a specified variable name, and will report Clang as clang and not gnu, so update the code to deal with those two changes.


rhc54 (approved):
Looks good to me - thx!

rhc54:
We have never seen this error before - perhaps a new compiler or something?
```text
  CC       pmix_globals.lo
"pmix_globals.c", line 733: error: ellipsis with no explicit parameters is nonstandard
      va_start(ap, x);
```
I cannot find anything wrong with that statement. Any ideas? I'll rerun things just in case it is a false flag.

rhc54:
bot:ibm:retest

rhc54:
Found one reference to that error message indicating it is due to the particulars of how the compiler actually implements the variadic function - apparently, in at least one case it is triggered by adding `-strict-ansi` to the compile cmd.

Seriously, could one of you IBM folks check to see if the PGI compiler was somehow changed recently (like in the last hour or two)? The referenced code has been there for a very long time without once being flagged as problematic. Frankly, from what I can find, there is no solution other than fiddling with compile cmd flags to turn it off.

rhc54:
Duh - this PR changes the compiler vendor identification. So I suspect it is the cause of this new error, probably either by changing the ID of vendor or perhaps resulting in a change to the compiler cmd line flags.

bwbarrett:
Yeah, I'm sure it has something to do with this commit.  But as long as you can't get to the full output of configure, there's no way to debug it.  I don't have easy access to the PGI compiler, so IBM's going to have to help here.

bwbarrett:
The IBM issue was real and sorting out the "pgi" vs. "portland group" seems to have fixed it.  Rebased on the HEAD of master and repushed this morning so that we can get this committed.

PR #2770, opened 2022-09-23
Update v4.2 with fresh commits from master

awlauria:


rhc54:
Thanks again - totally forgot to do this.

PR #2773, opened 2022-09-27
Plug a memory leak

darbyShaw:
Ensure we release the data array which is allocated in pmix_bfrops_base_copy_darray.

Signed-off-by: Heena Sirwani <heena.sirwani@itwm.fraunhofer.de>


rhc54:
Good catch! This highlighted a few other places that need fixing, but I can do that now that I've been focused on it.

PR #2780, opened 2022-10-05
Fix compiler error from unused function parameters

cniethammer:
Newer gcc versions warn about unused-parameter - which results in compilation errors after 93bbaed5cd86b2b0245ace56cfeb99df6be3c30c.
This fixes this by applying __pmix_attribute_unused__ to the reported code part and parameters.


rhc54 (changes requested):
I appreciate it, but that isn't how we handle this. Please use the `PMIX_HIDE_UNUSED_PARAMS` macro

cniethammer:
Thanks for pointing out. This looks much nicer. I updated the PR accordingly.

rhc54 (approved):

PR #2798, opened 2022-10-26
Add const qualifier to pset_name

krempel-pt:
Since pset_name in PMIx_server_define_process_set() is not to be modified by the function, it should be declared as const, especially to avoid compiler warnings about discarded qualifiers.

Signed-off-by: Stephan Krempel <krempel@par-tec.com>


krempel-pt:
I know, that currently the standard does not have the `const` qualifier there, I therefor opened a discussion to change that: https://github.com/pmix/pmix-standard/issues/426

rhc54 (approved):
No issues - I think this is fine.

rhc54:
Ah, I see the CI problem. Let me push a change into your branch to address it.

PR #2800, opened 2022-11-02
Intel oneapi (and probably llvm): patch to allow

hppritcha:
pmix tests to compile using icx, icpx, etc.

Without this patch intel oneapi 2022 series fails to compile openpmix if tests are built.

### DIFF
--- a/test/test_common.h
+++ b/test/test_common.h
@@ -286,7 +286,6 @@ typedef struct {
                 }                                                                                 \
             }                                                                                     \
         } else {                                                                                  \
-            int _count;                                                                           \
             _cbdata.in_progress = 1;                                                              \
             PMIX_VALUE_CREATE(_val, 1);                                                           \
             _cbdata.kv = _val;                                                                    \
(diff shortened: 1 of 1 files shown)


awlauria (approved):

test/test_common.h
```diff
@@ -286,7 +286,8 @@ typedef struct {
                 }                                                                                 \
             }                                                                                     \
         } else {                                                                                  \
-            int _count;                                                                           \
+            int _count = 0;                                                                       \
+            PMIX_HIDE_UNUSED_PARAMS(_count);                                                      \
```
rhc54, line 290: So...oneapi is claiming that a variable _must_ be used in _all_ "if" clauses or else it is "unused"???? That strikes me like a compiler bug - it's an absurd requirement. Or am I misreading this?

bwbarrett: That seems like a weird one to raise a warning about, but without seeing the actual warning, it's hard to tell why it is upset.

However, it seems like a much better change would be to just remove the declaration on line 289 and move it to line 299.  Why declare the variable in a larger scope than needed, especially if you're doing work to mask its use?

hppritcha: i like @bwbarrett 's idea. i change it that way and hopefully intel (actually clang) will stop complaining.

hppritcha: i think i'll just remove the variable.  looks like it was probably being used for some kind of debugging.

bwbarrett (approved):

rhc54 (approved):

PR #2805, opened 2022-11-10
Update Python iof_push() function to take null byte object

MattBBaker:
If `PMIX_IOF_PUSH_STDIN` is set in the info object to a call to `PMIx_IOF_Push()` than the byte object parameter can be NULL. The Python bindings will try and do an allocation anyways, and in some cases this can result in a real pointer instead of a NULL, which will cause problems. This updates the iof_push() function to allow for this behavior.
Also use `malloc()` instead of `PyMem_Malloc()`.
Signed-off-by: Matthew Baker <matthew@voltrondata.com>

### DIFF
--- a/bindings/python/pmix.pyx
+++ b/bindings/python/pmix.pyx
@@ -3380,21 +3380,24 @@ cdef class PMIxTool(PMIxServer):
         ndirs       = 0
 
         # convert data to pmix_byte_object_t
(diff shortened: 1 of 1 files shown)


rhc54:
@MattBBaker I can move these to the v4.2 branch - is there an urgent need to release a v4.2.3? Or is next available release adequate?

rhc54:
bot:ibm:retest

MattBBaker:
Next available release is good, no rush for a new release right now

PR #2815, opened 2022-11-21
Modify the pmix_output system

rhc54:
Replace the PMIX_OUTPUT_VERBOSE macro with direct
calls to pmix_output_verbose so it always compiles, even in non-debug configurations. Inline the "check_verbosity" function to minimize the performance impact of calling that function - requires changing the "info" array into a properly-prefixed global array.

Retain the PMIX_OUTPUT_VERBOSE definition so that
legacy users (e.g., PRRTE v3.0.0) are not impacted, but direct that it always be compiled "in".


rhc54:
bot:ibm:retest

PR #2832, opened 2022-11-30
Allow gds function fallback to hash.

samuelkgutierrez:
For a strict subset of gds module function pointers, allow for the use of NULL to indicate that another module should be used instead, namely hash.

The impacted functions here are: store, assemb_kvs_req, and accept_kvs_resp.

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -9,6 +9,7 @@
  *                         All rights reserved.
  * Copyright (c) 2016-2022 IBM Corporation.  All rights reserved.
  * Copyright (c) 2021-2022 Nanook Consulting.  All rights reserved.
+ * Copyright (c) 2022      Triad National Security, LLC. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -48,7 +49,7 @@
 
 #include "src/class/pmix_list.h"
 #include "src/mca/bfrops/bfrops.h"
-#include "src/mca/gds/gds.h"
+#include "src/mca/gds/base/base.h"
 #include "src/mca/pcompress/base/base.h"
 #include "src/mca/ptl/base/base.h"
 #include "src/threads/pmix_threads.h"
--- a/src/common/pmix_query.c
+++ b/src/common/pmix_query.c
@@ -27,6 +27,7 @@
 #include "src/mca/bfrops/bfrops.h"
 #include "src/mca/pstrg/pstrg.h"
 #include "src/mca/ptl/base/base.h"
+#include "src/mca/gds/base/base.h"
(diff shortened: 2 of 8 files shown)


rhc54:
Errr...I don't think we want to do that - we should be cycling over all active components looking for someone to respond. If the indicated component cannot handle it, then we just need to default to the "hash" component. Something like this:
```c
#define PMIX_GDS_ASSEMB_KVS_REQ(s, p, r, k, b, c)                   \
    do {                                                            \
        pmix_gds_base_module_t *_g = (p)->nptr->compat.gds;         \
        (s) = PMIX_SUCCESS;                                         \
        if (NULL == _g->assemb_kvs_req) {                           \
            if (0 == strcmp(_g->name, "hash") {                     \
                (s) = PMIX_ERR_NOT_SUPPORTED;                       \
            } else {                                                \
                _g = pmix_globals.mypeer->nptr->compat.gds;         \
            }                                                       \
        }                                                           \
        if (NULL != _g->assemb_kvs_req) {                           \
            pmix_output_verbose(1, pmix_gds_base_output,            \
                                "[%s:%d] GDS ASSEMBLE REQ WITH %s", \
                                __FILE__, __LINE__, _g->name);      \
            (s) = _g->assemb_kvs_req(r, k, b, (void *) c);          \
        }                                                           \
    } while (0)
```

samuelkgutierrez:
> Errr...I don't think we want to do that - we shouldn't be cycling over all active components looking for someone to respond. If the indicated component cannot handle it, then we just need to default to the "hash" component. Something like this:
> 
> ```c
> #define PMIX_GDS_ASSEMB_KVS_REQ(s, p, r, k, b, c)                   \
>     do {                                                            \
>         pmix_gds_base_module_t *_g = (p)->nptr->compat.gds;         \
>         (s) = PMIX_SUCCESS;                                         \
>         if (NULL == _g->assemb_kvs_req) {                           \
>             if (0 == strcmp(_g->name, "hash") {                     \
>                 (s) = PMIX_ERR_NOT_SUPPORTED;                       \
>             } else {                                                \
>                 _g = pmix_globals.mypeer->nptr->compat.gds;         \
>             }                                                       \
>         }                                                           \
>         if (NULL != _g->assemb_kvs_req) {                           \
>             pmix_output_verbose(1, pmix_gds_base_output,            \
>                                 "[%s:%d] GDS ASSEMBLE REQ WITH %s", \
>                                 __FILE__, __LINE__, _g->name);      \
>             (s) = _g->assemb_kvs_req(r, k, b, (void *) c);          \
>         }                                                           \
>     } while (0)
> ```

Thank you, Ralph. Okay, I'll give that method a shot.

PR #2844, opened 2022-12-06
Add PMIX_CPUSET_DESTRUCT macro

krempel-pt:
Add macro for symmetry reasons

Signed-off-by: Stephan Krempel <krempel@par-tec.com>

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -2286,6 +2286,11 @@ typedef uint8_t pmix_link_state_t;
 
 
 /****   PMIX CPUSET    ****/
+/* Bring function definition across for symmetry macro. We have to do this as
+ * there are people who only included pmix_common.h if they were using macros
+ * but not APIs */
+PMIX_EXPORT void PMIx_Cpuset_destruct(pmix_cpuset_t *cpuset);
+
 typedef struct{
     char *source;
     void *bitmap;
@@ -2300,7 +2305,8 @@ typedef struct{
 #define PMIX_CPUSET_CONSTRUCT(m) \
     memset((m), 0, sizeof(pmix_cpuset_t))
 
-#define PMIX_CPUSET_DESTRUCT(m)
+#define PMIX_CPUSET_DESTRUCT(m) \
+    PMIx_Cpuset_destruct(m);
 
 #define PMIX_CPUSET_CREATE(m, n)    \
     do {                                                                    \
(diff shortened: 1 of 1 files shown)


rhc54:
If we are going to add that macro, then we should have it do something - it cannot just be empty as that is misleading. There is a function for this purpose:
```c
PMIX_EXPORT void PMIx_Cpuset_destruct(pmix_cpuset_t *cpuset);
```
If you look in `deprecated.h`, you'll see how to add the macro correctly. You'll need to move the function definition over there as was done with the other macro forms. It's a little weird as the macro would be coming into the code as already "deprecated", but that's the only way to provide the symmetry you are seeking.

krempel-pt:
Ok, then here is a new suggestion, not make it deprecated from the beginning.

But perhaps I missed something, how is the intended way to construct and destruct `pmix_cpuset_t`?

To use `PMIX_CPUSET_CONSTRUCT` and `PMIx_Cpuset_destruct()` does not feel intuitive to me at all.

icompres:
Hi Stephan,

Ralph can explain this better, but the move from macros to functions was
decided, as far as I understand, as a path towards the binary interface
changes that are incoming in the PMIx standard.  The macros are kept around
for the time being, to minimize impact to existing users.

Isaías A. Comprés Ureña

rhc54:
Yeah, we probably should instead add a `PMIx_Cpuset_construct` function. Somewhat of a shame as all it does is zero out the memory, but we can make it inline so implementations don't have to provide it. The macro versions can then be deprecated like we did for the other macros.

krempel-pt:
> Yeah, we probably should instead add a `PMIx_Cpuset_construct` function. Somewhat of a shame as all it does is zero out the memory, but we can make it inline so implementations don't have to provide it. The macro versions can then be deprecated like we did for the other macros.

Sounds like a good plan for the future, but I just realized, that the PMIx standard version 4.1 explicitly mentions macros `PMIX_CPUSET_CONSTRUCT(m)` and `PMIX_CPUSET_DESTRUCT(m)`, so i think they should be included both in OpenPMIx 4.x versions for standard compiance, and not be deprecated for the moment.

In current draft for the next standard version those macros are in, too, but this could be subject to change.

rhc54:
Okay, the decision was to move all of the construct/create/destruct/free macros to function calls for ABI purposes. We will retain the macro definitions over here for backward compatibility. This will be fairly extensive as it covers all the data types, so I'll take the action item to do it - will happen slowly (probably while watching football and/or basketball games).

jjhursey:
FYI: I filed a ticket on the PMIx Standard to fix this up on that side
 * https://github.com/pmix/pmix-standard/issues/428

krempel-pt:
> I have completed converting all macros into functions - the macros are still present for backward compatibility. Symmetry is also provided, so I'm closing this for now.

Many thanks.

PR #2846, opened 2022-12-06
Checkpoint gds/shmem modex data retrieval.

samuelkgutierrez:

### DIFF
--- a/src/client/pmix_client_fence.c
+++ b/src/client/pmix_client_fence.c
@@ -201,7 +201,6 @@ static pmix_status_t unpack_return(pmix_buffer_t *data)
     
     /* provide an opportunity to store any data (or at least how to access
      * any data) that was included in the fence */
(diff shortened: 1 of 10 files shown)


src/client/pmix_client_fence.c
```diff
@@ -200,7 +201,12 @@ static pmix_status_t unpack_return(pmix_buffer_t *data)
     
     /* provide an opportunity to store any data (or at least how to access
      * any data) that was included in the fence */
-    PMIX_GDS_RECV_MODEX_COMPLETE(data);
+    // TODO(skg) Is this peer correct? It seems to work.
```
rhc54, line 204: yes, that is the correct peer - sorry for the error!

samuelkgutierrez: Thank you. Just making sure. I removed the comment.

PR #2848, opened 2022-12-07
Change function signature of mark_modex_complete().

samuelkgutierrez:
mark_modex_complete() now returns pmix_status_t.


rhc54 (approved):
Good thought!

PR #2850, opened 2022-12-08
Improve shared-memory segment attach in gds/shmem.

samuelkgutierrez:
* Generalize segment attach code.
* Implement client-side attach for modex data.


rhc54:
bot:ibm:retest

Thu Dec 8 00:37:57 UTC 2022

PR #2852, opened 2022-12-08
gds/shmem: Fix crash in server_mark_modex_complete().

samuelkgutierrez:
* Make sure that a modex exchange has occurred before server_mark_modex_complete() was called.
* Incorporate minor cleanup unrelated to fix.

### DIFF
--- a/src/mca/gds/shmem/gds_shmem.c
+++ b/src/mca/gds/shmem/gds_shmem.c
@@ -19,41 +19,15 @@
 #include "gds_shmem_utils.h"
 #include "gds_shmem_store.h"
 #include "gds_shmem_fetch.h"
-// TODO(skg) This will eventually go away.
-#include "pmix_hash2.h"
-
-#include "src/client/pmix_client_ops.h"
-#include "src/server/pmix_server_ops.h"
 
 #include "src/util/pmix_argv.h"
 #include "src/util/pmix_error.h"
 #include "src/util/pmix_output.h"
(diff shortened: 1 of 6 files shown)


rhc54:
> * Make sure that a call to server_store_modex() occurred before server_mark_modex_complete() was called.

Just curious - with the change I made, I can see a way that could technically happen, but I don't see how it happens in practice. What test case caused it to trigger?

samuelkgutierrez:
Hi, Ralph.

If you comment out the code referenced [here](https://github.com/openpmix/openpmix/pull/2852/files#diff-35580c6334e25b873bde80a9f458bfc499ff7d52570c0213e0671ea216d706f3R1280-R1283), `run_tests00.pl` will reproduce the issue I'm seeing. Don't forget (because I always do) to enable the shmem component by changing 

```c
#define PMIX_GDS_SHMEM_DISABLE 1
```
to 
```c
#define PMIX_GDS_SHMEM_DISABLE 0
```
in `src/mca/gds/shmem/gds_shmem.h`.

Before your change `nslist` in `server_mark_modex_complete()` was empty. Now, it doesn't seem like I can use that property as a work bypass. I'm sure I'm missing something, so let's talk offline.

samuelkgutierrez:
I updated pull request message because it was misleading.

rhc54:
We'll take this for now as it includes things beyond just that 3-line fix - we can come back to address that later.

PR #2862, opened 2022-12-12
Address to-do items in gds/shmem.

samuelkgutierrez:
* Expose an MCA parameter that influences the ultimate sizes of the shared-memory segments used to back gds/shmem data.
* Emit shared-memory usage statistics during job termination when appropriate.
* Incorporate other minor modifications that happened to be captured during this work.

### DIFF
--- a/src/mca/gds/shmem/gds_shmem.c
+++ b/src/mca/gds/shmem/gds_shmem.c
@@ -101,10 +101,10 @@ strtost(
 }
 
 /**
- * 8-byte address alignment.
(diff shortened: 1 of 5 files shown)


rhc54:
Just a suggestion - you might add an MCA param to control emitting the usage stats.

samuelkgutierrez:
> Just a suggestion - you might add an MCA param to control emitting the usage stats.

Makes sense. Consider this taken care of because the stats are only emitted when the component verbosity is cranked up.

PR #2889, opened 2022-12-23
Add initial support for PMIx_Get operations on groups.

darbyShaw:
The following commit adds support for PMIx_Get operations on groups for realm (node, process) attributes that depend on the process rank. The support is only for PMIx reserved keys and for groups constructed with embedded barriers.

### DIFF
--- a/src/client/pmix_client.c
+++ b/src/client/pmix_client.c
@@ -211,6 +211,7 @@ pmix_client_globals_t pmix_client_globals = {
     .singleton = false,
     .pending_requests = PMIX_LIST_STATIC_INIT,
     .peers = PMIX_POINTER_ARRAY_STATIC_INIT,
+    .groups = PMIX_LIST_STATIC_INIT,
     .get_output = -1,
     .get_verbose = 0,
     .connect_output = -1,
@@ -621,6 +622,7 @@ PMIX_EXPORT pmix_status_t PMIx_Init(pmix_proc_t *proc, pmix_info_t info[], size_
                          PMIX_FWD_STDERR_CHANNEL, pmix_iof_write_handler);
 
     /* setup the globals */
+    PMIX_CONSTRUCT(&pmix_client_globals.groups, pmix_list_t);
     PMIX_CONSTRUCT(&pmix_client_globals.pending_requests, pmix_list_t);
     PMIX_CONSTRUCT(&pmix_client_globals.peers, pmix_pointer_array_t);
     pmix_pointer_array_init(&pmix_client_globals.peers, 1, INT_MAX, 1);
@@ -1088,6 +1090,7 @@ PMIX_EXPORT pmix_status_t PMIx_Finalize(const pmix_info_t info[], size_t ninfo)
         PMIX_LIST_DESTRUCT(&pmix_server_globals.iof);
         PMIX_LIST_DESTRUCT(&pmix_server_globals.iof_residuals);
     }
+    PMIX_DESTRUCT(&pmix_client_globals.groups);
 
     if (0 <= pmix_client_globals.myserver->sd) {
         CLOSE_THE_SOCKET(pmix_client_globals.myserver->sd);
--- a/src/tool/pmix_tool.c
+++ b/src/tool/pmix_tool.c
@@ -645,6 +645,7 @@ PMIX_EXPORT int PMIx_tool_init(pmix_proc_t *proc, pmix_info_t info[], size_t nin
     pmix_globals.iof_flags.local_output = outputio;
 
     /* setup the globals */
+    PMIX_CONSTRUCT(&pmix_client_globals.groups, pmix_list_t);
     PMIX_CONSTRUCT(&pmix_client_globals.pending_requests, pmix_list_t);
     PMIX_CONSTRUCT(&pmix_client_globals.peers, pmix_pointer_array_t);
     pmix_pointer_array_init(&pmix_client_globals.peers, 1, INT_MAX, 1);
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -81,6 +81,10 @@ static pmix_status_t process_request(const pmix_proc_t *proc, const char key[],
     pmix_status_t rc;
     pmix_value_t *ival;
     size_t n;
+    pmix_group_t *grp;
+    pmix_cb_t cb2;
+    uint32_t running_size = 0, jsize = 0;
+    pmix_kval_t *kv;
 
     /* if the proc is NULL, then the caller is assuming
      * that the key is universally unique within the caller's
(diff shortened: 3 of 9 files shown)


rhc54:
I'm a little confused by the embedded barrier requirement. The requesting proc must have participated in a group construct operation or else it wouldn't know about the group. This means that the local PMIx server must know about the group as well. We require that the PMIx server translate the groupID,rank to its corresponding nspace,rank prior to doing anything as the data in storage is only done by nspace,rank.

So if the server translates the provided ID, then there is no problem with the host being able to handle any request. Looking at this change, I'm wondering if the issue is the missing translation at the beginning of the server "get" code? You might look at the server "fence" op for an example of how we handled it.

darbyShaw:
> I'm a little confused by the embedded barrier requirement. The requesting proc must have participated in a group construct operation or else it wouldn't know about the group. This means that the local PMIx server must know about the group as well. We require that the PMIx server translate the groupID,rank to its corresponding nspace,rank prior to doing anything as the data in storage is only done by nspace,rank.
> 
Embedded barrier is assumed because at the moment, I force immediate or localonly get operations and do not send a dmodex request if the the namespace is not known. I thin  that the namespace has to be known if if the group construction was done with an embedded barrier because I think only then (unless I am wrong) the endpoint info of the participating namespaces becomes local to the participating servers.
 
> So if the server translates the provided ID, then there is no problem with the host being able to handle any request. Looking at this change, I'm wondering if the issue is the missing translation at the beginning of the server "get" code? You might look at the server "fence" op for an example of how we handled it.

There is no problem with the host being able to handle the request. The problem is being able to translate get operations on groups used as namespaces in pmix_proc_t structures from clients. I understand the fence op and the get op has been changed in a similar fashion. The group information is translated to actual nspace, rank back to the client requesting the information via PMIx_Get.

rhc54:
I guess my point was that if the groupID,rank has been changed to the corresponding nspace,rank, then you can just leave the rest of the code the same and remove the comment about requiring an embedded barrier. The host will only see the nspace,rank ID, and that is all it needs for a dmodex operation. The data will come back tagged with nspace,rank, but that's okay as it has to be stored that way anyway.

The request "caddy" likewise should only contain nspace,rank so that the "satisfy_request" function can properly match the returned data (which will all be in nspace,rank) with the requesting proc.

All servers involved in the group construct are required to have a complete set of data (all job-level data plus any data posted by the participating procs) at the completion of the construct operation - whether or not it contains a barrier. So I don't think you need any restrictions here - just convert the IDs upon receiving them and everything else should just be the same.

darbyShaw:
OK. I will remove the embedded barrier assumption and check the dmodex flow. The satisfy_request needs not only the translated namespace and rank but also the group name and rank in the request since the client on receiving the reply in its getnb callback function checks its datastore with the group information and it has no idea that the received translated proc in the reply corresponds to the group query it made. So the server sends the byte object of the translated proc information with the group proc that was present in the request made by the client. So now the datastore stores the proc information both by the translated proc and the queried group proc. I think this can be made better but for later?

rhc54:
> The satisfy_request needs not only the translated namespace and rank but also the group name and rank in the request since the client on receiving the reply in its getnb callback function checks its datastore with the group information and it has no idea that the received translated proc in the reply corresponds to the group query it made. 

Sounds like the client needs to translate the IDs as well, then - we cannot have the system storing the data twice as the footprint will become too large.

darbyShaw:
I also think that storing translations at every client of the group on the same node would lead to duplication if done in memory. Either the hash component instead has to instead store pointers for group information or may be a separate translation query could be introduced to the server that does not return the proc byte object but only the translated proc.

rhc54:
Okay, I'm confused - the client already knows the nspace,rank of all participating members of the group. It has to know them in order to construct the group. One thing we've debated about is the requirement that each participant list the membership in the same order in the construct call - we finally decided to relax that requirement and make the call order insensitive. Thus, the membership returned by the server from the call becomes the definition for group rank.

This also makes things consistent with the async group formation (invite/join) as there is no ordering of procs one can use until the group definition is returned upon completion of the op.

So if we wrap all this together, it may be that the correct place to do the translation is in the client prior to passing any calls up to the server or accessing any local/shmem data storage. The client should use the returned membership array to do the translation. Yes, that means storing the membership in each client - but that has to happen anyway or else the app cannot use it. Small footprint when compared to the job-level data, especially at scale.

Make sense?

darbyShaw:
I am not so sure if the construct call can be made order insensitive. Then all procs in the group have different views of what rank they are referring to when they want to use a specific rank for a specific proc and  may be that's why all procs should agree on a unique mapping. The standard document also mentions that the server is responsible for translations but of course that can change over time.

darbyShaw:
Ah, I see that the membership returned by the server becomes the definition. That should solve the problem then. But does that mean that the client library developers like MPI or GPI have to maintain the group membership and not the pmix client library?

rhc54:
> The standard document also mentions that the server is responsible for translations but of course that can change over time.

Yeah, we should take that out of the standard - it's up to the implementation to decide where to do the translation. The only point we were trying to make is that the neither the app nor the host have to do it - the PMIx library will do it somewhere between the app and the host environment.

> But does that mean that the client library developers like MPI or GPI have to maintain the group membership and not the pmix client library?

No, they shouldn't need to do so. Think of it this way. Each client app has a different notion of the rank of a process in the group. So they will call functions like `PMIx_Get` using a groupID and the rank of the target proc based on their _local_ view of the membership. The PMIx client library then translates that ID into the global nspace,rank of the actual proc.

So the fact that one app rank will refer to `nspace,15` as `group,2` and another app rank might refer to it as `group,10` is really irrelevant - each app's PMIx client will translate it to `nspace,15` before it goes anywhere and to access any internal information.

That said, we _do_ return the final group membership upon completion of the construct operation, so apps and these other libraries could translate based on that ordering (since that is what PMIx does) should they choose to do so. If you think about it, it's the only way to know the final group membership and ordering when using the async group construct. So even though someone might currently use the blocking construct, it might be good to keep things consistent in case they later decide to switch to the async construct.

Make sense? I can help with the changeover as it is somewhat intrusive, if you like - or I'm happy to advise and let you handle it.

darbyShaw:
Things are a bit clearer now. Thank you very much for the insights. I would like to volunteer to do the changeover if you don't mind with your advice of course.

rhc54:
Please do! I truly appreciate your contributions - always well done 😄

darbyShaw:
I now made the following changes:

- Removed the group translation logic from the server as discussed above.
- Since it would be very convenient if all the clients in the group had the same group view and the same group-level rank view for the participating clients, I sort the group members in the group before adding it to the pmix_client_globals.groups list at each client. This helps in maintaining the same group view across all clients and facilitates analogous usage of group names and namespaces (for eg., every process in the namespace has the same view of its own rank and others' ranks).  We can then expose the group-level ranks to applications as process ranks when we add processes dynamically to jobs.
- At the client side for get operations, when a group name and group rank is used in the proc_t structure, I translate the proc to the actual nspace, rank by finding out job sizes of the nspaces that use wildcard ranks in the group membership. So suppose I have two nspaces in the group with wildcard ranks: nspace1, wildcard and nspace2, wildcard and the job sizes of the two namespaces are 5, is I perform a group get op using group, rank 5, it gets translated to nspace2, rank 0 by subtracting the job size of nspace 1 from the group-level rank.
- I also move the pmix_group_t definition to pmix_globals from pmix_server_ops since the struct is used by both client and server.

Please let me know if the changes make sense. However there is no rush and I wish you good holidays.

src/client/pmix_client.c
```diff
@@ -1065,6 +1067,7 @@ PMIX_EXPORT pmix_status_t PMIx_Finalize(const pmix_info_t info[], size_t ninfo)
         PMIX_LIST_DESTRUCT(&pmix_server_globals.iof);
         PMIX_LIST_DESTRUCT(&pmix_server_globals.iof_residuals);
     }
+    PMIX_DESTRUCT(&pmix_client_globals.groups);
```
rhc54, line 1070: This needs to be `PMIX_LIST_DESTRUCT`

src/tool/pmix_tool.c
```diff
@@ -644,6 +644,7 @@ PMIX_EXPORT int PMIx_tool_init(pmix_proc_t *proc, pmix_info_t info[], size_t nin
     pmix_globals.iof_flags.local_output = outputio;
 
     /* setup the globals */
+    PMIX_CONSTRUCT(&pmix_client_globals.groups, pmix_list_t);
```
rhc54, line 647: Need to add the corresponding `PMIX_LIST_DESTRUCT` to the tool's "finalize" routine

rhc54 (commented):
Overall, this looks really good! I do have one question, however. I have on my "to-do" list to handle the case where every process involved in a group construct doesn't necessarily know all the procs that will be included in the group when complete. There are use-cases where, for example, a proc might know of additional members that need to be included. I'm adding an attribute for that purpose.

What this means is that the host environment is responsible for assembling the final group membership, which is then returned to the client (and the user) upon completion of "construct". It is this array that defines the group ranks - not what the user provides to us at time of calling "construct". It's fine for us to sort that array prior to returning it to the user. Note that there is a sorting routine in the util area of the code that can be used for that purpose.

I might be missing it, but is that the array (the one returned upon completion of the "construct" operation) the one you are using here? Or are you using the one provided by the caller? If the latter, then we should change that to use the returned array. I honestly don't remember what the current code does.

We also need to highlight that difference in the Standard, but that's a separate issue.

darbyShaw:
Actually, I was quite inclined towards using the array returned by the server in the PMIX_GROUP_MEMBERSHIP attribute, if that's similar to what you are saying. But I saw that neither the PMIx server nor the host server (PRRTE) return such information and wasn't sure if the collective group construction operation was allowed to proceed if the client's requested view of membership  is not respected. 

I think the async group construction op using group invite allows clients to have the inviting proc control the group membership. I am quite interested in this operation and I see that it is expected that every invited client receives details of group membership using the attribute PMIX_GROUP_MEMBERSHIP in the event handler's callback for the PMIX_GROUP_CONSTRUCT_COMPLETE event. But this information is given directly to the application. Would it be possible to intercept this information at the PMIx client side?

I would be happy to change the implementation based on your insights. Thank you.

rhc54:
Yeah, I need to do some work on the PRRTE side of things to correctly handle the construction operation. When I initially did it, we didn't have these issues of procs not knowing all the members of the group prior to construct, nor was the ordering problem identified back then. Since those came up, I've been somewhat tied up with other things.

Let me finish my current little project (should be done later today) and then I'll address the PRRTE end of things.

> But this information is given directly to the application. Would it be possible to intercept this information at the PMIx client side?

The event actually flows thru the client library on its way to the application's event callback function, so easy enough to intercept. IIRC, we may already do that for some items. The complete flow is:
```text
host --> PMIx server library --> PMIx client library --> app callback function
```
The server library picks out information from events before passing them along - no reason why the client library can't do the same.

darbyShaw:
The translation at the client requires the job info of all the namespaces participating in group in the GDS. The job info is not exchanged during group construct for example for spawned namespaces. Should all the job info of all namespaces of local participants in the GROUP_ENDPT_DATA be added at group construct?

rhc54:
> On Jan 4, 2023, at 1:16 AM, Darby Shaw ***@***.***> wrote:
> 
> 
> The translation at the client requires the job info of all the namespaces participating in group in the GDS. The job info is not exchanged during group construct for example for spawned namespaces. Should all the job info of all namespaces of local participants in the GROUP_ENDPT_DATA be added at group construct?
> 

Yes, the job info should be getting passed. Honestly thought it was already happening - will have to take a look in a bit.

rhc54:
I see you have updated this - where do we stand on it? Is it ready to go?

I think I still owe you some work in PRRTE - yes? Need to return the final group membership when responding to "construct", and I believe I need to include all posted info and job info for the participants. Does that sound correct?

I believe I have a branch somewhere that is close to being ready...can check.

darbyShaw:
Yes, I am yet to add processing of PMIX_GROUP_MEMBERSHIP info at the pmix server side which is to be passed to the clients and used in their callbacks to store a view of the group. I would need the info from the PRRTE side.

Yes, and also the posted and job info needs to be passed to the participants.

rhc54:
I'll go ahead and commit this to avoid creating lots of conflicts with my changes.

PR #2890, opened 2022-12-23
Add const qualifier to nspace strings

krempel-pt:
Just adds some missing const qualifiers to avoid meaningless compiler warnings when using `PMIx_Proc_load` with const namespace string.


rhc54 (approved):
Thanks for catching this!

PR #2892, opened 2022-12-23
configure: abort in 32-bit environments

Joe-Downs:
OpenPMIX does not support 32-bit environments, so make configure fail early.

Signed-off-by: Joe Downs <joe@dwns.dev>


rhc54:
Much appreciated, but I'd like to hold off on committing this until we see the conclusion from https://github.com/open-mpi/ompi/issues/11248

rhc54:
Confirmed that OMPI is going to similarly remove 32-bit support.

PR #2894, opened 2022-12-24
python: updates for Python bindings

jsquyres:
- Separate build vs. install: the Python bindings will now be built during "make all", and installed during "make install" (vs. being built and installed during "make install").
  - Fix regression introduced in 9cd0aa4f where the Python bindings were not actually installed in $prefix.
- Update from Python distutils to setuptools.  We still use setup.py instead of the newer setup.cfg or pyproject.toml, but that can be upgraded later.
  - Suppress "setup.py" deprecated warnings during the installation.
- Fix keywords listing in the Python package
- Setup dependencies properly so that the Python bindings will properly re-build if any of its dependent header files change.
- Use a workaround to make Cython work properly in VPATH builds.
- Update "make uninstall" for the Python bindings to be a bit more robust.
- Make "make clean" remove the generated constants files.
- Removed some stale kruft from setup.py; updated to use setuptools instead of disttools.
- Add bindings/python/requirements.txt

Signed-off-by: Jeff Squyres <jeff@squyres.com>

### DIFF
--- a/bindings/python/Makefile.am
+++ b/bindings/python/Makefile.am
@@ -23,36 +23,123 @@
 # $HEADER$
 #
 
-# we do NOT want picky compilers down here
+# We do NOT want picky compilers down here
 CFLAGS = $(PMIX_CFLAGS_BEFORE_PICKY)
-PYTHONPATH=$PYTHONPATH:$(DESTDIR)$(prefix)/lib/python$(PYTHON_VERSION)/site-packages
 
-helpers = setup.py pmix.pyx pmix.pxi construct.py
+helpers = \
+        setup.py \
+        construct.py
 
-EXTRA_DIST = $(helpers)
+src_files = \
+        pmix.pxi \
+        pmix.pyx
+
+EXTRA_DIST = $(helpers) $(src_files)
 
 if WANT_PYTHON_BINDINGS
 
-install-exec-local: $(helpers)
+harvested_headers = \
+        $(top_srcdir)/include/pmix.h \
+        $(top_builddir)/include/pmix_common.h \
+        $(top_srcdir)/include/pmix_deprecated.h \
+        $(top_srcdir)/include/pmix_server.h \
+        $(top_srcdir)/include/pmix_tool.h
+
+constants_files = \
+        pmix_constants.pxd \
+        pmix_constants.pxi
+
+# Create pmix_constants.pxd and pmix_constants.pxi by harvesting a
+# bunch of PMIx header files These will be generated in the PMIx build
+# dir.
+$(constants_files): construct.py $(harvested_headers)
 	$(PYTHON) $(top_srcdir)/bindings/python/construct.py --src="$(top_builddir)/include" --include-dir="$(top_srcdir)/include"
-	PMIX_BINDINGS_TOP_SRCDIR=$(PMIX_TOP_SRCDIR) PMIX_BINDINGS_TOP_BUILDDIR=$(PMIX_TOP_BUILDDIR) \
-		$(PYTHON) $(top_srcdir)/bindings/python/setup.py build_ext --library-dirs="$(DESTDIR)$(libdir)" --user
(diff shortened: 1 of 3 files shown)


jsquyres:
@rhc54 Try this out and see if it works for you.  Do you have any CI tests that build/install/test the Python bindings, perchance?

rhc54:
Mellanox CI is failing during srpm build:
```shell
make[3]: *** No rule to make target `pmix.pyc', needed by `distdir'.  Stop.
```
The "special builds" CI tests building the bindings and it also is failing:
```shell
make[2]: Entering directory '/home/pmixer/tmp-build-hjY2K/openpmix/bindings/python'
/usr/bin/python3 ../../bindings/python/construct.py --src="../../include" --include-dir="../../include"
/usr/bin/python3 ../../bindings/python/construct.py --src="../../include" --include-dir="../../include"
PMIX_BINDINGS_TOP_SRCDIR=/home/pmixer/tmp-build-hjY2K/openpmix PMIX_BINDINGS_TOP_BUILDDIR=/home/pmixer/tmp-build-hjY2K/openpmix \
	/usr/bin/python3 ../../bindings/python/setup.py build_ext --library-dirs="/home/pmixer/tmp-build-hjY2K/openpmix/src/.libs" --user
  File "../../bindings/python/setup.py", line 134, in <module>
    sys.exit(main())
  File "../../bindings/python/setup.py", line 128, in main
    package_setup(package_name, package_vers)
  File "../../bindings/python/setup.py", line 85, in package_setup
    category=setuptools.SetuptoolsDeprecationWarning)
AttributeError: module 'setuptools' has no attribute 'SetuptoolsDeprecationWarning'
make[2]: *** [Makefile:852: pmix.c] Error 1
make[2]: Leaving directory '/home/pmixer/tmp-build-hjY2K/openpmix/bindings/python'
make[1]: *** [Makefile:687: all-recursive] Error 1
```

rhc54:
The cross-version thing is an anomaly - we get it now and then. The other is real:
```shell
running install
Checking .pth file support in /home/pmixer/tmp-build-gcwZN/install-pmix/lib64/python3.6/site-packages/
/usr/bin/python3 -E -c pass
TEST FAILED: /home/pmixer/tmp-build-gcwZN/install-pmix/lib64/python3.6/site-packages/ does NOT support .pth files
error: bad install directory or PYTHONPATH

You are attempting to install a package to a directory that is not
on PYTHONPATH and which Python does not read ".pth" files from.  The
installation directory you specified (via --install-dir, --prefix, or
the distutils default setting) was:

    /home/pmixer/tmp-build-gcwZN/install-pmix/lib64/python3.6/site-packages/

and your PYTHONPATH environment variable currently contains:

    '$PYTHONPATH:/home/pmixer/tmp-build-gcwZN/install-pmix/lib/python3.6/site-packages'

Here are some of your options for correcting the problem:

* You can choose a different installation directory, i.e., one that is
  on PYTHONPATH or supports .pth files

* You can add the installation directory to the PYTHONPATH environment
  variable.  (It must then also be on PYTHONPATH whenever you run
  Python and want to use the package(s) you are installing.)

* You can set up the installation directory to support ".pth" files by
  using one of the approaches described here:

  https://setuptools.readthedocs.io/en/latest/easy_install.html#custom-installation-locations


Please make the appropriate changes for your system and try again.
```
This was the "warning" I had fixed before.

jsquyres:
Yeah, I'm looking at the special builds fail.

rhc54:
FWIW: you have to add the install path to the `PYTHONPATH` to make it go away.

jsquyres:
> FWIW: you have to add the install path to the `PYTHONPATH` to make it go away.

I thought I did -- I see it in the output.  Not sure it didn't propagate properly (it does on my machine).  Checking...

rhc54 (approved):
Wow - that turned out to be _way_ more work than I expected!!! Thanks so much for doing it.

PR #2895, opened 2022-12-25
v4.2.x: python: updates for Python bindings

jsquyres:
- Separate build vs. install: the Python bindings will now be built during "make all", and installed during "make install" (vs. being built and installed during "make install").
  - Fix regression introduced in 9cd0aa4f where the Python bindings were not actually installed in $prefix.
- Update from Python distutils to setuptools.  We still use setup.py instead of the newer setup.cfg or pyproject.toml, but that can be upgraded later.
  - Suppress "setup.py" deprecated warnings during the installation.
- Fix keywords listing in the Python package
- Setup dependencies properly so that the Python bindings will properly re-build if any of its dependent header files change.
- Use a workaround to make Cython work properly in VPATH builds.
- Update "make uninstall" for the Python bindings to be a bit more robust.
- Make "make clean" remove the generated constants files.
- Removed some stale kruft from setup.py; updated to use setuptools instead of disttools.
- Add bindings/python/requirements.txt

Signed-off-by: Jeff Squyres <jeff@squyres.com>
(cherry picked from commit 214d5dffd1d2798baa96f5ce0c3085051903659a)

### DIFF
--- a/bindings/python/Makefile.am
+++ b/bindings/python/Makefile.am
@@ -23,36 +23,123 @@
 # $HEADER$
 #
 
-# we do NOT want picky compilers down here
+# We do NOT want picky compilers down here
 CFLAGS = $(PMIX_CFLAGS_BEFORE_PICKY)
-PYTHONPATH=$PYTHONPATH:$(DESTDIR)$(prefix)/lib/python$(PYTHON_VERSION)/site-packages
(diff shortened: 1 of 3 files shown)


rhc54:
bot:ibm:retest

PR #2897, opened 2023-01-04
Enable size estimates for modex-like operations

rhc54:
We collect data and pass it up to the host for distribution during fence operations that support traditional modex plus group and connect ops. Ensure we include an estimate fo the resulting storage size. Also clean up a few memory leaks.


rhc54:
bot:ibm:retest

Issue #1424, opened 2019-08-29
[Q]: pmix_base_async_modex - HOWTO


angainor:
Hi, Ralph,
I am a little confused about the async modex. In OpenMPI doc I read

> Setting the `pmix_base_async_modex` [...] causes the system to lookup the endpoint information for a peer only at first message. Thus, instead of collecting endpoint information for all processes, only the endpoint information for those processes this peer communicates with will be retrieved. 

Looking at the code, it seems that in some of the code paths in `ompi_mpi_init.c` no fence is executed
```
    /* If we have a non-blocking fence:
     * if we are doing an async modex, but we are collecting all
     * data, then execute the non-blocking modex in the background.
     * All calls to modex_recv will be cached until the background
     * modex completes. If collect_all_data is false, then we skip
     * the fence completely and retrieve data on-demand from the
     * source node.
     *
     * If we do not have a non-blocking fence, then we must always
     * execute the blocking fence as the system does not support
     * later data retrieval. */
    if (NULL != opal_pmix.fence_nb) {
        if (opal_pmix_base_async_modex && opal_pmix_collect_all_data) {
            /* execute the fence_nb in the background to collect
             * the data */
        } else if (!opal_pmix_base_async_modex) {
            /* we want to do the modex */
        }
        /* otherwise, we don't want to do the modex, so fall thru */
    } else if (!opal_pmix_base_async_modex || opal_pmix_collect_all_data) {
       [...]
    }
```
So if we want async modex and don't collect all data, then there is no fence. Do I understand correctly that in such case I do not need a fence at all, and a call to `PMIx_Get` will get the value for me from the source? I'm confused, because from your comment here #1404 it seems that the data isn't circulated without a fence.

Then again, I have a program, in which I set a string key, and then I `PMIx_Get` the value in any of the ranks without running the fence, so it seems to work. 

Could you explain how that works?

Thanks!

rhc54:
When you do a fence and request that data be circulated, then we essentially perform an allgather operation. At the end, every daemon winds up with a complete copy of all the data posted by the individual procs. When a proc then requests data from some proc on a different node, the request is locally satisfied from the daemon's copy of the info.

[...]

PR #2900, opened 2023-01-06
Fix return of job-info for get operation with wildcard rank.

darbyShaw:
When the rank is a wildcard, a NULL key is required in order to fetch all job-info from the GDS for the given namespace.

Signed-off-by: Heena Sirwani <heena.sirwani@itwm.fraunhofer.de>

### DIFF
--- a/src/server/pmix_server_get.c
+++ b/src/server/pmix_server_get.c
@@ -294,7 +294,7 @@ pmix_status_t pmix_server_get(pmix_buffer_t *buf, pmix_modex_cbfunc_t cbfunc, vo
         PMIX_CONSTRUCT(&pbkt, pmix_buffer_t);
         /* they are asking for the job-level info for (or at least a
          * reserved key from) this nspace */
-        rc = get_job_data(nptr->nspace, cd, key, &pbkt);
+        rc = get_job_data(nptr->nspace, cd, NULL, &pbkt);
         if (PMIX_SUCCESS != rc) {
             PMIX_DESTRUCT(&pbkt);
             return rc;


rhc54:
Hmmm....I may be missing something here. This change means that the server will return the entire job info for the nspace if the rank is wildcard, regardless of the key? I believe the intent was as you state in your commit message - "When the rank is a wildcard, a NULL key is required in order to fetch all job-info from the GDS for the given namespace." This allows the client to retrieve a single piece of information if that is what is required.

Just trying to understand the motivation behind the change.

darbyShaw:
I think that the get request reaches the server only if the nspace is not known to the client or the request is for data which is not yet put or present in the GDS yet. In the first case, the client has a part of  code in the get_data function that replaces the rank by wildcard so that all job info can be cached even if the query was for only one key of one rank. In the second case, I think that the request is for a non PMIx reserved key and may be it should be associated with a rank or its scope should be global. If it is associated with a rank then the above code change will not affect the request. If the key scope is global then the get_job_data function in the code path above only retrieves internally scoped data so I assume that a wildcard query for a non PMIx reserved key has no semantic or is not defined yet.

rhc54:
Let's set aside the session/node/app-level data keys as they obey their own rules and focus solely on the job/proc-level values. Let's also limit the discussion solely to the master branch as the prior release branches have a lot of ugly logic to workaround limitations in the dstore GDS components that are being eliminated going forward.

Looking at just that code path, it's clear that the comments in the code and the code itself conflict in this regard. The comments indicate that the server is expected to return _all_ data associated with the job (for a wildcard rank), while the code only returns the specific key that was requested. I remember this conflict well, unfortunately, so let me try to capture it here.

We used to operate on the belief that the host would provide all job-related info prior to process startup, and thus all reserved job/proc-level keys were fully known at that time. Thus, failure to find a reserved key meant that it wasn't available. We also thought that apps would only `PMIx_Put` non-reserved keys for sharing during a `PMIx_Fence` operation. We held those keys at the server until a client asked for one of them - and then we returned all of the keys for that target proc under the assumption that the client would likely be asking for more of them.

Thus, there was no way for a client to ask the server for job/proc-related info. We subsequently opened that door a little by allowing the query when the nspace of the target proc (i.e., the proc whose information was being requested) differed from the requestor. Again, we returned all the info once any key was requested.

Over time, we found that different hosts provided varying amounts of the job/proc-level info at startup, and so apps would `PMIx_Put` any missing reserved keys so that the keys would be shared via the fence operation and then could be retrieved. In other words, if the host didn't provide a value for something like `PMIX_LOCAL_RANK` and I had another way of computing it for myself, then I could `PMIx_Put` my value, do a fence, and then retrieve it for all other ranks in the job. Creating non-reserved keys for all these things was painful and an unnecessary burden, and so we widened the opening to support it by allowing the client to query the server for reserved keys in addition to non-reserved keys.

The net result of all this is that the difference between "reserved" and "non-reserved" keys has been eliminated and the comments in the code need to be updated to reflect this change (I will take care of it).

This leaves one question: should the server return all of the target job/proc data when any key is requested, or just the requested key? It can be more efficient if multiple keys are going to be queried, but if shmem isn't used or if only one piece of information is actually needed it can lead to unnecessarily large footprints. So we've been trying to review it on a per-case basis.

If you can provide some background as to what led you to this change, that would be helpful in making the decision.

darbyShaw:
I was trying to work with a prototype that avoids the overhead of having PMIx groups and yet is able to obtain information about processes from other namespaces. I was interested mainly in the job sizes and process node hostnames and local ranks. I was able to query the job size from the server using a job rank as wildcard and was hoping to get all the job info in return (similar to the discussion here https://github.com/openpmix/openpmix/issues/1424). But that didn't happen and I couldn't query hostnames after that either because the client expects that is has node information available but it doesn't have it and it doesn't query the server either (probably is has something to do with how the PMIx_server_register_nspace function works). I made the change since I read the comments and I thought the behaviour in the comments  could facilitate what I want and made a change to confirm if that was what was intended. Now I know that it was not intended and I am back to using PMIx groups :)

darbyShaw:
Also, this part of the code confused me a little.
```
doget:
    cb->info = iptr;
    cb->ninfo = nfo;
    PMIX_GDS_FETCH_KV(rc, pmix_client_globals.myserver, cb);
    if (PMIX_SUCCESS == rc) {
        pmix_output_verbose(5, pmix_client_globals.get_output,
                            "pmix:client data found in server-provided data");
        cb->status = process_values(cb);
        goto done;
    }
    pmix_output_verbose(5, pmix_client_globals.get_output,
                        "pmix:client data NOT found in server-provided data");

    /* if we are both using the "hash" component, then the server's peer
     * will simply be pointing at the same hash tables as my peer - no
     * no point in checking there again */
    if (!PMIX_GDS_CHECK_COMPONENT(pmix_client_globals.myserver, "hash")) {
        /* check the data in my hash module */
        PMIX_GDS_FETCH_KV(rc, pmix_globals.mypeer, cb);
        if (PMIX_SUCCESS == rc) {
            pmix_output_verbose(5, pmix_client_globals.get_output,
                                "pmix:client data found in internal hash data");
            cb->status = process_values(cb);
            goto done;
        }
    }
    pmix_output_verbose(5, pmix_client_globals.get_output,
                        "pmix:client job-level data NOT found");
```
The comment "_if we are both using the "hash" component, then the server's peer will simply be pointing at the same hash tables as my peer - no point in checking there again_" was confusing. I understand that the client will only have the tables if the server has them but the reverse is not implied and that they are not necessarily pointers but copies if the gds for both is hash.

rhc54:
> I was trying to work with a prototype that avoids the overhead of having PMIx groups and yet is able to obtain information about processes from other namespaces. I was interested mainly in the job sizes and process node hostnames and local ranks. I was able to query the job size from the server using a job rank as wildcard and was hoping to get all the job info in return (similar to the discussion here #1424). But that didn't happen

One question that might help guide us to an answer is: how did you know the nspace of the other job? You would need to know it to do any query about it, so how did you obtain that piece of information?

Obtaining information about other random namespaces is probably better done using the "query" API. We are still feeling our way towards a good split between "get" and "query", but increasingly it seems that "get" is better used when dealing with namespaces you have some connection to (either via "connect" or "group"), while "query" is best used to get information about things you don't have any connection to (like file systems or other random jobs in the system).

If "query" isn't getting you what you need, then we might want to address that by extending support there rather than modifying "get" - every time we touch the latter, we impact a lot of people.

rhc54:
> The comment "_if we are both using the "hash" component, then the server's peer will simply be pointing at the same hash tables as my peer - no point in checking there again_" was confusing. I understand that the client will only have the tables if the server has them but the reverse is not implied and that they are not necessarily pointers but copies if the gds for both is hash.

A little code organization explanation is required here. A client process has at least two GDS components available to it. One is always the "hash" component which stores copies of data locally in the process itself. The client can cache information there that it wants to reuse internally and not expose to anyone else, plus it can keep a copy of any other proc's info if necessary. The pointer to this component is stored on the `pmix_globals.mypeer` object.

The other component the client tracks is the GDS component being used to access information provided by its server. This is stored in the `pmix_client_globals.myserver` peer. If shared memory is being used, then this component contains the shared memory hooks to access data placed there by the server. If "hash" is being used, then this component will be the same as the one being used by the client to store its own internal info.

So if "hash" is being used to handle data from the server, there is no need to run a "fetch" across it twice - if the data were available, we would have gotten it the first time. If the component being used to handle data from the server is different, then we do need to check both places as the server might have stored it where we could get it, or we may have gotten it directly and put it in our internal storage.

Hope that helps

darbyShaw:
Thank you for the detailed explanation.

PR #2906, opened 2023-01-11
Change policy on master branch versioning

rhc54:
It has been noted that PMIx maintains compatibility between its master branch and all outstanding release branches. In the past, we locked the master branch to a libtool triplet of 0:0:0, which translates to a libtool .so versioning of 0. This contrasts with the .so=2 values of the release branches and can be confusing as it implies that the master branch is not, in fact, compatible with those branches.

Change the policy to have the master branch maintain a libtool triplet equal to the highest found on any release branch.

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -106,9 +106,12 @@ date="Unreleased developer copy"
 # All changes in these version numbers are dictated by the PMIx
 # release managers (not individual developers).  Notes:
 
-# 1. Since these version numbers are associated with *releases*, the
-# version numbers maintained on the PMIx Github trunk (and developer
-# branches) is always 0:0:0 for all libraries.
+# 1. Since the PMIx master branch is maintained to always be compatible
+#    with all prior releases (i.e., we never remove an interface), the
+#    libtool 'c', 'r', and 'a' values are maintained equal to the highest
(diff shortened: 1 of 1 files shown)


VERSION
```diff
@@ -106,17 +106,20 @@ date="Unreleased developer copy"
...
 
-# 1. Since these version numbers are associated with *releases*, the
-# version numbers maintained on the PMIx Github trunk (and developer
-# branches) is always 0:0:0 for all libraries.
+# 1. Since the PMIx master branch is maintained to always be compatible
+#    with all prior releases (i.e., we never remove an interface), the
+#    libtool 'c' and 'a' values are maintained equal to the highest values
+#    found on any given release branch. This ensure that the .so value
+#    remains the same as on the release branches, thereby indicating
+#    compatibility.
```
bwbarrett, line 114: You actually need to maintain all three across the board.  The 4.2 branch also has the version as 9:0:7, so the code is right, but the comment is wrong (I believe).

bwbarrett (approved):

VERSION
rhc54: oh, yeah - I changed the code and forgot to update the comment. I'll fix that - thanks for catching it!

PR #2914, opened 2023-01-18
No need to fetch info arrays for job info

rhc54:
[No need to fetch info arrays for job info](https://github.com/openpmix/openpmix/pull/2914/commits/0b1b74a75f0eacd6767cfb092453873ec51f3835)

We no longer have GDS components that do not
understand info arrays, so we don't need to
worry about providing that info to the client's
"hash" component.


rhc54:
bot:ibm:retest

PR #2918, opened 2023-01-19
ptl/base: retry recv() when it encounter EAGAIN or EWOULDBLOCK

wzamazon:
currently, pmix_ptl_base_recv_blocking() return
PMIX_ERR_TEMP_UNAVAILABLE when recv() encounter EAGAIN or EWOULDBLOCK.

This caused client to close socket, and re-establish connection in try_connect().

The close of socket caused server to receive the SIGPIPE signal, and got aborted.

To address the issue, this patch makes pmix_ptl_base_recv_blocking() to retry recv() when it encountered EAGAIN or EWOULDBLOCK

Signed-off-by: Wei Zhang <wzam@amazon.com>


wzamazon:
@rhc54 Please take a look

wzamazon:
Not sure what the error of IBM CI is. Is there a way to see the log?

rhc54:
Looks like it had some kind of network failure that broke the initial container prep - nothing to do with the PR.

PR #2927, opened 2023-01-20
Add option to abort on component find failure

rhc54:
If a specified component (i.e., PMIX_MCA_foo=a,b,c) cannot be found or refuses to load, provide an MCA param to indicate that we should error out. Otherwise, ignore the situation (reporting it if show_load_errors is set) - the framework itself will abort if none of the components are available.

(cherry picked from commit 4012cc8dbfec92662a7be96a1b9fe249d5d76f57)


rhc54:
bot:ibm:retest

PR #2938, opened 2023-01-25
Turn off the "format" attribute on Solaris

rhc54:
According to the docs, this is the one that would
trigger complaints about NULL pointers in va_args, even though the error talks about the "sentinel"
attribute. Might as well swat them both.


rhc54:
bot:ibm:retest

PR #2940, opened 2023-01-27
gds:shmem Decouple session and job data.

samuelkgutierrez:
IBM's CI found an issue with how we were storing job and session data in shared-memory. This commit checkpoints our work toward fixing that issue. I still need to cleanup some code, but my initial tests look promising. The solution here that appears to work requires that session data be stored in distinct shared-memory segments.

### DIFF
--- a/src/mca/gds/shmem/gds_shmem.c
+++ b/src/mca/gds/shmem/gds_shmem.c
@@ -19,11 +19,6 @@
 #include "gds_shmem_store.h"
 #include "gds_shmem_fetch.h"
 
-#include "src/util/pmix_argv.h"
-#include "src/util/pmix_environ.h"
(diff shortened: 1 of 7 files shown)


rhc54:
Something to keep in mind: it is possible for a PRRTE DVM to support multiple sessions. I'm working on that use-case now.

samuelkgutierrez:
Hopefully this way will support that use case 🤞🏼

rhc54:
Wouldn't sweat it too much just yet - I will probably be creating a session-specific "shepherd DVM" any way. If it needs to use the "hash" component to get the session info from the system DVM, that isn't a big deal - might even be better for security reasons. Just nice to keep options open if it isn't too much trouble. Keep me posted so I can adapt as needed.

PR #2955, opened 2023-02-06
Fix configure for __builtin_expect support.

samuelkgutierrez:


rhc54:
Afraid I don't know enough to evaluate the impact of this - what precisely does this do? What is the impact of the change?

samuelkgutierrez:
This fixes configure so that the value of `PMIX_C_HAVE_BUILTIN_EXPECT` in src/include/pmix_config.h is properly calculated. Before this change it was always evaluating to `0`. All this does is provide access to `__builtin_expect()` via `PMIX_UNLIKELY()` and `PMIX_LIKELY()` to provide the compiler with additional branch prediction info.

rhc54:
ok, thanks

PR #2968, opened 2023-02-09
Add missing thread support to client functions

rhc54:
Need to acquire and post objects passed via the
thread handler. Some environments are more
sensitive about it.


samuelkgutierrez:
bot:ibm:retest

PR #2990, opened 2023-02-22
Avoid large stack allocations in pmix_bfrops_base_tma_setenv().

samuelkgutierrez:


rhc54:
Thanks Sam! It was the memory management over the asprintf calls that made me hesitate and call for help. Much appreciated.

samuelkgutierrez:
> Thanks Sam! It was the memory management over the asprintf calls that made me hesitate and call for help. Much appreciated.

Glad I could help, @rhc54!

PR #2997, opened 2023-02-24
Plug memory leaks.

samuelkgutierrez:
Plug memory leaks identified by memcheck on the prterun side.


rhc54:
I'm not sure I fully grok these changes, but one comment: you must not free a pointer that is passed to a function as there is no way to `NULL` it when done - the pointer still points to garbage space. This is why the macros free the base storage of a class like `pmix_coord_t` while the function does not. This was noted in the Standard when we made the change to functions versus macros. I cannot tell for sure, but your comments seem to imply you are free'ing pointers passed to functions? If so, we shouldn't do that.

rhc54:
Agreed that we would in fact free the base memory - more changes enroute.

PR #2998, opened 2023-02-24
Convert the majority of PMIX_*_FREE() macros.

samuelkgutierrez:
Convert the majority of PMIX_*_FREE() macros by moving (or adding) the frees into the underlying function calls. Also NULLify the incoming pointers after freeing the base pointers where appropriate.

More to come, but the marked FIXME locations need more thought.

### DIFF
--- a/include/pmix_deprecated.h
+++ b/include/pmix_deprecated.h
@@ -544,12 +544,11 @@ PMIX_EXPORT pmix_info_t* PMIx_Info_list_get_info(void *ptr, void *prev, void **n
 #define PMIX_VALUE_CREATE(m, n) \
     (m) = PMIx_Value_create(n)
 
-// free(m) is NOT called inside PMIx_Value_destruct(), so do it here.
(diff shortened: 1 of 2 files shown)


rhc54:
Hmmm...isn't this going to get a little confusing if some free functions actually free the base memory, but others don't? Is the difference just a question of not yet having added the free inside the function? Or is there an underlying reason why we can't be consistent? I agree that the number of exceptions appears small, but that makes in even more likely that someone will make a mistake.

Also a note to myself: need to add a hard rule in the code style doc that use of these functional forms is prohibited inside the PMIx and PRRTE code bases. Developers _must_ use the macro forms to ensure that things get set to NULL after free so we don't have stale pointers hanging around the code.

samuelkgutierrez:
You're right, good point: we should be consistent here. I pushed some changes that hopefully address the inconsistencies you pointed out. Nice catch.

rhc54:
Thx!

PR #3020, opened 2023-03-18
gds/shmem: Update backing file permissions if directed.

samuelkgutierrez:


samuelkgutierrez:
bot:ibm:retest

rhc54:
bot:ibm:retest

The IBM CI may be down for the weekend, but let's give it one more try just to see if they can run.

src/mca/gds/shmem/gds_shmem.c
```diff
@@ -365,6 +365,9 @@ job_construct(
 ) {
     PMIX_GDS_SHMEM_VVOUT_HERE();
 
+    job->uid = getuid();
```
rhc54, line 368: I think you want the effective uid - i.e., should use `geteuid`

src/mca/gds/shmem/gds_shmem.c
```diff
@@ -1978,9 +2001,31 @@ server_add_nspace(
...
+    for (size_t i = 0; i < ninfo; ++i) {
+        if (PMIX_CHECK_KEY(&info[i], PMIX_USERID)) {
+            const uid_t nuid = (uid_t)info[i].value.data.uint32;
+            PMIX_GDS_SHMEM_VOUT(
+                "%s: updating nspace=%s UID from %zd to %zd",
+                __func__, nspace, (size_t)job->uid, (size_t)nuid
+            );
+            job->uid = nuid;
+            job->chown = true;
+            // We don't care about any other keys, so bail.
```
rhc54, line 2024: You probably should check for `PMIX_GRPID` as well and update the group permissions too if it doesn't match the value returned by `getegid`. Many sites use the group ID as an accounting entity and we should allow for sharing between members of the group.

src/mca/gds/shmem/gds_shmem.h
```diff
@@ -166,6 +166,10 @@ typedef struct {
 
 typedef struct {
     pmix_list_item_t super;
+    /** User ID */
+    uid_t uid;
+    /** Change owner? */
+    bool chown;
```
rhc54, line 172: Need to add `grpid` and `bool chgrp` fields.

rhc54 (changes requested):

rhc54 (approved):

src/mca/gds/shmem/gds_shmem.c
```diff
@@ -1004,6 +1040,15 @@ shmem_segment_create_and_attach(
     }
     // Attach to the shared-memory segment.
     rc = shmem_attach(job, shmem_id, (uintptr_t)base_addr);
+    if (PMIX_UNLIKELY(PMIX_SUCCESS != rc)) {
+        PMIX_ERROR_LOG(rc);
```
rhc54, line 1044: Just curious: do we really want to errorlog this here? The errorlog message won't mean anything to a user other than "something went wrong". If the intent is to warn them and then failover to the `hash` component, we probably should provide a "show help" instead so we can give them more info on what happened, note that we will failover and the implications of it, and how to turn "off" the message for future runs.

I'm not sure where your failover occurs, so perhaps this isn't the right place?

samuelkgutierrez: Thank you, Ralph. You're right. I haven't implemented the failover yet because I want to try these changes so I don't fool myself into thinking one bit is working while another is not. I'll get the failover code in another pull request.

rhc54: Very good - thx!

samuelkgutierrez:
bot:ibm:retest

rhc54:
bot:ibm:retest

samuelkgutierrez:
bot:ibm:retest

samuelkgutierrez:
bot:ibm:retest

naterini:
Can this be included with the v5.0 branch too?

rhc54:
Yes - I have to bring several things over today. Will do so now.

PR #3024, opened 2023-03-20
Return final group membership to client

rhc54:
It is possible that the group membership may have been altered during group_construct - e.g., if one or more participants adds members that are not visible to all. Ensure we pass back the actual final membership, and use that to define the group in the client's storage

### DIFF
--- a/examples/group.c
+++ b/examples/group.c
@@ -16,7 +16,7 @@
  * Copyright (c) 2013-2020 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015      Mellanox Technologies, Inc.  All rights reserved.
  * Copyright (c) 2019      IBM Corporation.  All rights reserved.
- * Copyright (c) 2021-2022 Nanook Consulting.  All rights reserved.
+ * Copyright (c) 2021-2023 Nanook Consulting  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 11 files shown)


rhc54:
@darbyShaw This provides the group membership when combined with https://github.com/openpmix/prrte/pull/1710. I see the endpoint info being added as well, so I think the only thing still missing is the job-level info for procs in other namespaces - is that correct?

darbyShaw:
> @darbyShaw This provides the group membership when combined with [openpmix/prrte#1710](https://github.com/openpmix/prrte/pull/1710). I see the endpoint info being added as well, so I think the only thing still missing is the job-level info for procs in other namespaces - is that correct?

Yes, correct.

rhc54:
@darbyShaw It's a pretty big change, but I cleaned up all the server-level code for tracking groups so we avoid duplication with the client. I also fixed a race condition that was causing me to lose add_members directives. I assume you have some tests you can use to exercise these things (e.g., to ensure the PMIx_Get and PMIx_Fence procs correctly translate group entries)? If so, can you (a) test these changes (you need the corresponding PRRTE changes), and (b) share them so I can try them here as well?

You'll note that I extended the `examples/group.c` code to exercise "add_members" and to provide an example of how to use `PMIx_Query_info` to obtain the list of currently active namespaces and a range of info on each of them. The info includes the job size, and the hostname and local rank for each proc in the namespace. Would that help resolve your problem of wanting to get such info without having to create "groups"?

Issue #3036, opened 2023-04-06
python bindings failing spawn with non-supported


naughtont3:
## Background information

It seems the changes in March may have caused a problem with the python bindings.  We are seeing an existing test fail with spawn `NOT-SUPPORTED` after upgrade to latest openpmix.  

### What version of the PMIx Reference Library are you using? 

 - openpmix master @ 5b883068 
 - prrte master @ ef659d7ec3

### Please describe the system on which you are running

* Operating system/version:
* Computer hardware:
* Network type:

-----------------------------

## Details of the problem

Still testing to try and pinpoint but looks like the initialization is off and the `pmixservermodule` does not return the `spawn` key, so thinks it is now supported.

```shell

bash:$ prte --report-uri /tmp/dvm.uri &
[1] 28330
bash:$ DVM ready

bash:$ prun --dvm-uri file:/tmp/dvm.uri --np 2 /usr/bin/hostname
login06
login06
bash:$ 
bash:$ python3 tiny2.py 
init RC 0
spawn RC -47
ERROR -  NOT-SUPPORTED
bash:$ prun --dvm-uri file:/tmp/dvm.uri --np 2 /usr/bin/hostname
login06
login06
bash:$ pterm --dvm-uri file:/tmp/dvm.uri
TERMINATING DVM...DONE
[1]+  Done                    prte --report-uri /tmp/dvm.uri
bash:$ 
```

```
bash:$ cat tiny2.py 
import pmix, sys
tool = pmix.PMIxTool()
rc,proc = tool.init( [ {"key": pmix.PMIX_SERVER_URI, "value": "file:/tmp/dvm.uri", "val_type": pmix.PMIX_STRING}])
print("init RC", rc)
app = { 'cmd':"/usr/bin/hostname", 'maxprocs':2, 'my_id': 1}
info = []
rc,nspace = tool.spawn(info, [app])
print("spawn RC", rc)
if rc != pmix.PMIX_SUCCESS:
    print("ERROR - ",tool.error_string(rc))
    sys.exit(1)

print("DONE")
sys.exit(0)
```

naughtont3:
note, did quick check with previous commit before python/binding change in march(214d5dffd1d2798baa96f5ce0c3085051903659a) and above check was fine.

rhc54:
[...]

PR #3037, opened 2023-04-11
Ensure tools send requests to their server

rhc54:
Tools may register server module function pointers to allow them to handle certain requests. The Python bindings are an example were the pointers can be registered, even though the host tool doesn't support them. Ensure that tools only use those pointers for requests that come up to them from clients, and don't call themselves back for those operations.


rhc54:
@naughtont3 Would you please check to see if this fixes #3036 ?

PR #3051, opened 2023-04-19
Improve shared-memory reference counting.

samuelkgutierrez:


rhc54:
Question: do we need to make the same changes to the general object system?

samuelkgutierrez:
> Question: do we need to make the same changes to the general object system?

No, I don't think so. I don't understand why this code was causing problems in shared-memory segments, but it would occasionally.

PR #3056, opened 2023-04-24
Ensure tools always send their event registration to their servers

rhc54:
If the event needs to be registered and the tool is connected, then the tool needs to send the registration to its server.


rhc54:
bot:ibm:retest

PR #3061, opened 2023-05-06
Handle the case of unbound proc. Also ensure topology is available

rhc54:
Ensure the topology is available to singletons. Properly handle the cpuset case for unbound procs.


rhc54:
RTD is busted

Issue #3062, opened 2023-05-08
PMIx group are not removed when PMIx_Group_destruct'ed


ggouaillardet:
This can be seen with the `PMIx` used by Open MPI `main` branch

Refs. open-mpi/ompi#11593

in `src/client/pmix_client_group.c`, newly created PMIx groups are appended to `pmix_client_globals.groups` but they are never removed from it. Consequently, when a new group is created with the same name, the first group with the same name is picked from the list in `PMIx_Group_destruct_nb`

The `PMIx` program below can be used in order to evidence the hang.
It has to be run with 2 ranks.

```c
#include <stdio.h>
#include <pmix.h>

int main(int argc, char *argv[]) {
    pmix_proc_t myproc;
    pmix_proc_t _procs[2], *procs;

    PMIx_Init(&myproc, NULL, 0);

    int i;
    size_t nprocs;
    for (i=0; i<2; i++) {
        memcpy(_procs+i, &myproc, sizeof(pmix_proc_t));
        _procs[i].rank = i;
    }

    pmix_info_t pinfo, *results = NULL;
    size_t nresults;

    PMIX_INFO_LOAD(&pinfo, PMIX_GROUP_ASSIGN_CONTEXT_ID, NULL, PMIX_BOOL);

    /* group "tag" contains ranks {0,1} */
    PMIx_Group_construct("tag", _procs, 2, &pinfo, 1, &results, &nresults);

    PMIx_Group_destruct("tag", NULL, 0);

    if (0 == myproc.rank) {
        /* group "tag" that was previously created with ranks {0,1} has been destructed
         * and is now re-created with the local rank */

        PMIX_INFO_LOAD(&pinfo, PMIX_GROUP_ASSIGN_CONTEXT_ID, NULL, PMIX_BOOL);
        PMIx_Group_construct("tag", _procs, 1, &pinfo, 1, &results, &nresults);

        /* this destruction will hang */
        PMIx_Group_destruct("tag", NULL, 0);
    }

    pmix_proc_t proc;
    PMIX_PROC_CONSTRUCT(&proc);
    PMIX_LOAD_PROCID(&proc, myproc.nspace, PMIX_RANK_WILDCARD);
    PMIx_Fence(&proc, 1, NULL, 0);

    PMIx_Finalize(NULL, 0);

    return 0;
}

```

rhc54:
[...]

PR #3063, opened 2023-05-09
Remove the group from client tracking when destruct is done

rhc54:
Once group destruct has completed, remove the group from the client's tracking list.

Refs #3062


ggouaillardet:
Thanks @rhc54, it works for me!

PR #3071, opened 2023-05-10
Disqualify gds/shmem if C11 atomics are not available.

samuelkgutierrez:
gds/shmem's use of pmix_shmem currently requires C11 atomics for reference counting. Disqualify gds/shmem if C11 atomics are not available. Also update how we implement reference counting in anticipation of future modifications (e.g., explicit memory barriers).

### DIFF
--- a/src/mca/gds/shmem/configure.m4
+++ b/src/mca/gds/shmem/configure.m4
@@ -18,7 +18,9 @@ AC_DEFUN([MCA_pmix_gds_shmem_CONFIG], [
     dnl address space is probably too small for the 'virtual memory hole'
     dnl finding that we do here. Below assumes support for only 32- and 64-bit
     dnl architectures.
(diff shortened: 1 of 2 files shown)


rhc54:
Hmmm...I believe that PMIx requires C11 atomics for the entire library, so I'm not sure why we need to protect gds/shmem specifically?

samuelkgutierrez:
I just noticed on a machine with an XL compiler that `PMIX_ATOMIC_C11` is defined as 0 in `src/include/pmix_config.h`. Is that expected if C11 atomics are required? Maybe I'm misunderstanding this variable's meaning.

rhc54:
Don't know - check to see if OMPI requires C11 atomics (pretty sure they do). If so, then we are free to do so too and should just error out if they aren't available.

samuelkgutierrez:
Okay, I can take a look. May I have these changes merged as-is for now? I would like to continue testing. Once we get the compiler bits sorted, I can make the necessary modifications.

PR #3074, opened 2023-05-15
util/pmix_shmem.c: toggle builtin atomics routine based on configure

wenduwan:
`./configure ... --disable-builtin-atomics` is currently broken. This patch fixes that.

### DIFF
--- a/src/util/pmix_shmem.c
+++ b/src/util/pmix_shmem.c
@@ -41,14 +41,25 @@ static inline void
 inc_ref_count(
     pmix_shmem_header_t *header
 ) {
+#if PMIX_USE_GCC_BUILTIN_ATOMICS
     (void)pmix_atomic_fetch_add_32(&header->ref_count, 1);
+#else
+    pmix_atomic_rmb();
+    ++(header->ref_count);
+#endif
 }
 
 static inline bool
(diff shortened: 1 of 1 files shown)


wenduwan:
@rhc54 @samuelkgutierrez Could you please take a look? Thanks!

samuelkgutierrez:
Thank you, @wenduwan. Based on my testing last week, I don't feel confident that the old way I had implemented reference counting was correct, especially on system architectures that can aggressively reorder memory operations.

Why do you disable built-in atomics? I'm curious because this change makes me a little nervous given how touchy this part of the code has been. Maybe we can figure out another approach.

lrbison:
This is an alternative: https://github.com/openpmix/openpmix/pull/3075

Which is to use c11 atomics or GCC builtin atomics based on configure options, and implement the same `pmix_atomic_fetch_add_32` style functions using c11 atomics.

wenduwan:
@lrbison Thanks! Closing this PR in favor of https://github.com/openpmix/openpmix/pull/3075

wenduwan:
@samuelkgutierrez This was caught in our internal testing matrix, not that we want to disable builtin atomics.

PR #3075, opened 2023-05-15
Define pmix atomic ops for c11 atomics

lrbison:
Define pmix atomic ops for c11 atomics

Previously these ops were only available when using GCC atomics, but now
they can be used when C11 atomics are selected.

### DIFF
--- a/src/include/pmix_atomic.h
+++ b/src/include/pmix_atomic.h
@@ -18,7 +18,7 @@
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2018-2020 Intel, Inc.  All rights reserved.
  * Copyright (c) 2021-2022 Nanook Consulting  All rights reserved.
- * Copyright (c) 2021      Amazon.com, Inc. or its affiliates.  All Rights
+ * Copyright (c)           Amazon.com, Inc. or its affiliates.  All Rights
  *                         reserved.
  * Copyright (c) 2023      Triad National Security, LLC. All rights reserved.
  * $COPYRIGHT$
@@ -80,6 +80,22 @@ static inline void pmix_atomic_rmb(void)
 #    endif
 }
 
+#define PMIX_ATOMIC_DEFINE_OP(type, bits, operator, name)                   \
+    static inline type                                                      \
+    pmix_atomic_fetch_##name##_##bits(pmix_atomic_##type *addr, type value) \
+    {                                                                       \
+        return atomic_fetch_##name##_explicit(                              \
+            addr, value, memory_order_relaxed);                             \
+    }                                                                       \
+                                                                            \
+    static inline type                                                      \
+    pmix_atomic_##name##_fetch_##bits(pmix_atomic_##type *addr, type value) \
+    {                                                                       \
+        return atomic_fetch_##name##_explicit(                              \
+            addr, value, memory_order_relaxed) operator value;              \
+    }
+
+/* end of PMIX_USE_C11_ATOMICS */
 #elif PMIX_USE_GCC_BUILTIN_ATOMICS
 
 static inline void pmix_atomic_wmb(void)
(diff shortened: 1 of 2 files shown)


src/include/pmix_atomic.h
```diff
@@ -112,12 +127,12 @@ static inline void pmix_atomic_rmb(void)
         return __atomic_##name##_fetch(addr, value, __ATOMIC_RELAXED);         \
     }
 
+#endif
+
```
wenduwan, line 131: In theory we are missing a branch here. Maybe force fail if neither conditions were met?

samuelkgutierrez (approved):
Looks good, @lrbison! I like @wenduwan's suggestion: have the #else branch #error out, but you implemented the important bits.

lrbison:
This is already captured at configure time with ` --disable-builtin-atomics --disable-c11-atomics`
```
*** Atomics
checking if gcc supports GCC inline assembly... yes
configure: error: No usable atomics implementation found.  Cannot continue.
```

But I've added an `#error` path here.

samuelkgutierrez:
Perfect, thank you.

lrbison:
I want to make sure a sanity check passes, you might notice I implemented the `atomic_fetch_add` by manually adding the value that c11 stdatomics returns as the pre-operation value to the supplied value, likewise with sub and xor etc....

wenduwan (approved):

wenduwan:
nit typo in commit message ` Remove uneeded header from pmix_shem.c `

samuelkgutierrez:
I just pulled and tested your branch, @lrbison. Something isn't quite right with the C11 version. The reference counting is off somehow because the files created by `gds/shmem` after running `make check` are still around.

samuelkgutierrez:
The semantics of the different variants found [here](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html) may be helpful in tracking down the issue.

samuelkgutierrez:
[0001-Update-C11-atomics.patch](https://github.com/openpmix/openpmix/files/11483536/0001-Update-C11-atomics.patch)

@lrbison, if you are interested, applying this patch to the head of your pull request seems to fix the issues I'm seeing.

samuelkgutierrez:
Your approach is cleaner than the patch I provided. It looks like the only issue with your fix was the swapping of where the additional op is used.

lrbison:
> was the swapping of where the additional op is used.

Just shows I can't work under pressure, I was being interrogated by an 8 year old at the same time!

I've re-pushed, and confirmed ibm tests are passing in mpi main (except for `ialltoallv_somezeros`).

I configured using ` --without-tests-examples` but couldn't find the gds/shmem files you mentioned.  How were you testing?

samuelkgutierrez (approved):
Tested, looks good on my machine and on a PPC system using GCC.

samuelkgutierrez:
Haha. I understand! I tested by running `make check` and looking in `/tmp` for any leftover file of the form `pmix-gds-shmem.node0-smoky_nspace-*`. That's a clue that reference counting isn't working as expected.

I tested your changes and they look good to me. Are the `ialltoallv_somezeros` particular to the changes or expected failures?

lrbison:
The `ialltoallv_somezeros` caused me some pause.  It fails with:
```
./include/ofi_mem.h:437: ofi_buf_free: Assertion `ofi_atomic_dec32(&ofi_buf_region(buf)->use_cnt) >= 0' failed.
```
which is of course atomics related.

I re-tested using the alternative code path (--enable-builtin-atomics) and I find a hang instead.  It does pass in open mpi's 4.1's release, so I'm guessing it's a regression but it's unrelated to this change.

lrbison:
Re-pushed just to fix typo in commit message.

I think this is ready for merge.

PR #3079, opened 2023-05-27
Respect "--" on cmd line

rhc54:
Do not process beyond the "--" marker, taking everything beyond that point as belonging to the application


rhc54:
Not needed

PR #3084, opened 2023-06-02
Fix missing "simptest.h" file

rhc54:
The simptest.h file in the test/simple directory
was missing from the tarball. This fix involves
no change in library source code - it is strictly
a change to a Makefile in the test area.


rhc54:
PRRTE does not support this release branch, so tests involving PRRTE will automatically fail.

PR #3089, opened 2023-06-08
Fix a couple of typos

rhc54:
Destruct the list and not just the list object during client finalize. Simptool test should use PMIx_tool_finalize instead of client finalize.


awlauria:
bot:ibm:retest

PR #3100, opened 2023-06-23
update hwloc object type compatibility

bgoglin:


rhc54:
Looks like IBM CI has gone down once again. @awlauria is this going to be an ongoing problem? I'm wondering if we should just turn it off/ignore it.

rhc54:
bot:ibm:retest

PR #3111, opened 2023-07-27
Repair the pattrs utility

rhc54:
Fix some bitrot in the attribute query support. Resolve inconsistency in passed structure type. Handle the case of returning an array of values vs returning a single string value. Improve formatting of output by adding spaces between function results.


rhc54:
bot:ibm:retest

PR #3112, opened 2023-08-03
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH` to make builds reproducible.

See https://reproducible-builds.org/ for why this is good and https://reproducible-builds.org/specs/source-date-epoch/ for the definition of this variable.

Without this patch, I got variations in our `openmpi5` package:
```diff
+++ new//usr/lib64/mpi/gcc/openmpi5/share/doc/pmix/html/developers/frameworks.html      2023-07-25 00:00:00.000000000 +0000
@@ -109,7 +109,7 @@
 – most services and functionality are implemented through MCA
 components.</p>
 <p>Here is a list of all the component frameworks in PMIx as of
-this writing August-2023:</p>
+this writing September-2039:</p>
```

This patch was done while working on reproducible builds for openSUSE.

Signed-off-by: Bernhard M. Wiedemann bwiedemann suse de

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -17,7 +17,12 @@
 # -- Project information -----------------------------------------------------
 
 import datetime
-mydate = datetime.datetime.now()
+import os
+import time
+
+mydate = datetime.datetime.utcfromtimestamp(
+    int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
+)
 year = mydate.year
 month = mydate.strftime("%B")
 


rhc54:
We already look for and set `SOURCE_DATE_EPOCH` during configure. @jsquyres - do we need to simply capture that setting in the docs generator?

rhc54:
bot:ibm:retest

rhc54:
Interestingly enough, this patch now allows someone to build the PMIx docs and claim copyright into the future years by simply setting an envar - see https://github.com/open-mpi/ompi/pull/11847 which raises that issue.

Should there be logic to disallow passing a future date?

rhc54:
Guess we allow this envar to override the build date elsewhere, so might as well be consistent - though I still have no idea how you got the error you report (unless you deliberately set your system clock to some bizarre future time?).

PR #3127, opened 2023-08-07
docs: update readthedocs.org config

jsquyres:
RTD is removing the option to use pre-installed packages.  Instead, we just need to tell them where our Python requirements.txt file is located so that they'll install exactly those packages.


rhc54:
FWIW: I had already filed this in https://github.com/openpmix/openpmix/pull/3126. Looks like the two are identical. I don't care which one we take, do you?

Issue #3128, opened 2023-08-07
Abort in listen_thread when many fds open


dgloe-hpe:
### What version of the PMIx Reference Library are you using? 

4.2.2

### Describe how PMIx was installed 

From RPMs

### Please describe the system on which you are running

* Operating system/version: SLES 15 SP4
* Computer hardware: Intel Xeon
* Network type: Slingshot

-----------------------------

## Details of the problem

Our process which uses the PMIx library is getting an abort signal if we fork while we have many file descriptors open (1049 in one case I analyzed). The backtrace looks like this:

```
(gdb) bt full
#0 0x00007f618014ac6b in raise () from /lib64/libc.so.6
No symbol table info available.
#1 0x00007f618014c305 in abort () from /lib64/libc.so.6
No symbol table info available.
#2 0x00007f6180190a97 in __libc_message () from /lib64/libc.so.6
No symbol table info available.
#3 0x00007f6180228812 in __fortify_fail () from /lib64/libc.so.6
No symbol table info available.
#4 0x00007f6180226c40 in __chk_fail () from /lib64/libc.so.6
No symbol table info available.
#5 0x00007f618022871a in __fdelt_warn () from /lib64/libc.so.6
No symbol table info available.
#6 0x00007f617a01670a in listen_thread (obj=<optimized out>) at base/ptl_base_listener.c:153
__d = <optimized out>
rc = <optimized out>
max = <optimized out>
addrlen = 128
pending_connection = <optimized out>
timeout = {tv_sec = 549755813888, tv_usec = 26038368}

readfds = {fds_bits = {0 <repeats 16 times>}}
lt = <optimized out>
#7 0x00007f6180ed76ea in start_thread () from /lib64/libpthread.so.0
No symbol table info available.
#8 0x00007f618021794f in clone () from /lib64/libc.so.6
No symbol table info available.
(gdb)
```

I believe the line of code that's failing is this:
```
FD_SET(lt->socket, &readfds);
```

If lt->socket is greater than FD_SETSIZE, this could be the cause of the buffer overflow. I would suggest either checking that lt->socket is smaller than FD_SETSIZE first, or use an alternative to select that doesn't have this limitation.

rhc54:
Hmmm...that is indeed problematic. I'm surprised we haven't encountered this before since that code has been in the wild for nearly 8 years now.

[...]

PR #3129, opened 2023-08-08
Switch to using event lib for connections

rhc54:
The "select" function has an inherent limitation on the number of file descriptors that can be open at any given time. In some environments, this can cause problems - so switch to using the event library for harvesting client-server connection requests.

Note: we were directly using "select" to avoid dropped connections due to an incoming connection flood at large scale. This was copied from an "mpirun" implementation. PMIx, however, doesn't have that flood problem - at most, one might see a few hundred local clients attempting to connect at the same time. Probably not an issue, but worth remembering.

### DIFF
--- a/src/mca/ptl/base/base.h
+++ b/src/mca/ptl/base/base.h
@@ -14,7 +14,7 @@
  * Copyright (c) 2014-2020 Intel, Inc.  All rights reserved.
  * Copyright (c) 2015-2020 Research Organization for Information Science
  *                         and Technology (RIST).  All rights reserved.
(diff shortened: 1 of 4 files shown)


rhc54:
bot:ibm:retest

rhc54:
Unstable CI strikes again!

gpaulsen:
@rhc54 Are you planning to backport this to v4.2.x ?  Open MPI v5 is interested in this, and I believe the original poster sited openpmix v4.2.2 in #3128

rhc54:
Already done - sitting in the v4.2 branch awaiting eventual release as 4.2.6

PR #3131, opened 2023-08-08
Multiple commits

rhc54:
[Switch to using event lib for connections](https://github.com/openpmix/openpmix/commit/fbfced2a8417010b074a496a13f46979863ead3e)

The "select" function has an inherent limitation on the
number of file descriptors that can be open at any given
time. In some environments, this can cause problems - so
switch to using the event library for harvesting client-server
connection requests.

Note: we were directly using "select" to avoid dropped
connections due to an incoming connection flood at large
scale. This was copied from an "mpirun" implementation.
PMIx, however, doesn't have that flood problem - at most,
one might see a few hundred local clients attempting to
connect at the same time. Probably not an issue, but worth
remembering.

(cherry picked from commit https://github.com/openpmix/openpmix/commit/6e8d12dd4d8fa3d276411aa1eba6107a4f90d2d5)

[Roll to version 5.0.1](https://github.com/openpmix/openpmix/commit/42830c0e08283a80ce2d16e987c2ddc45ceb5f6d)

bot:notacherrypick

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -16,7 +16,7 @@
 
 major=5
 minor=0
-release=0
+release=1
 
 # greek is used for alpha or beta release tags.  If it is non-empty,
 # it will be appended to the version number.  It does not have to be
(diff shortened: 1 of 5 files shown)


rhc54:
bot:ibm:retest

PR #3134, opened 2023-08-16
show_help: strip leading/trailing blank lines

jsquyres:
When we read a [topic] from a help file, strip any leading or trailing blank lines from the array of strings returned to the caller.

This is helpful in the context of https://github.com/openpmix/prrte/pull/1765


rhc54 (approved):
looks reasonable to me!

PR #3135, opened 2023-08-21
Fix Sphinx RTD theme HTML table cell word wrapping

jsquyres:
See individual commit messages for details:

* s/Open MPI/PMIx/ in a few places that were missed
* Override some CSS to fix HTML table cell word wrapping, and remove the workarounds that we had to effect word wrapping.


rhc54 (approved):
You have been very busy! Much appreciated!

PR #3137, opened 2023-08-24
Cleanup and temp fix

rhc54:
[Improve error handling in setup_topology](https://github.com/openpmix/openpmix/commit/97f1a17702f8d37965d1c82a300bd94b971af885)

Ensure we go thru the progression and then do our own
discovery if there are problems.

[Temp fix - disable gds/shmem component](https://github.com/openpmix/openpmix/commit/9d817ba2e976688e16b2b8c80d623bd15e88e4c5)

Disable the gds/shmem component unless specifically
requested using the PMIX_MCA_gds envar.

### DIFF
--- a/src/hwloc/pmix_hwloc.c
+++ b/src/hwloc/pmix_hwloc.c
@@ -319,33 +319,40 @@ tryxml:
     PMIX_GDS_FETCH_KV(rc, pmix_client_globals.myserver, &cb);
     if (PMIX_SUCCESS == rc) {
         file = popstr(&cb);
-        rc = load_xml(file);
-        free(file);
+        if (NULL == file) {
+            rc = PMIX_ERR_NOT_FOUND;
+        } else {
+            rc = load_xml(file);
+            free(file);
+        }
         cb.key = NULL;
         PMIX_DESTRUCT(&cb);
-        if (PMIX_SUCCESS == rc) {
-            pmix_output_verbose(2, pmix_hwloc_output,
-                                "%s:%s v2 xml adopted", __FILE__, __func__);
-            /* record locally in case someone does a PMIx_Get to retrieve it */
-            kv.key = PMIX_TOPOLOGY2;
-            kv.value = &val;
-            val.type = PMIX_TOPO;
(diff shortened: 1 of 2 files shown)


rhc54:
@samuelkgutierrez Could you please look at commit https://github.com/openpmix/openpmix/pull/3137/commits/9d817ba2e976688e16b2b8c80d623bd15e88e4c5 and PR https://github.com/openpmix/openpmix/pull/3136?

I'm hitting a strange error. If I add two attributes to `pmix_common.h` in PR https://github.com/openpmix/openpmix/pull/3136, any attempt to retrieve information from the gds/shmem component fails - it just returns garbage. If I disable the component, then everything works fine.

I have gone thru the code to look for places where we might have hit some limit (e.g., triggered an auto-resize of a pointer array), but nothing is apparent. The gds/hash component is fine with the change. Only thing I can think of is something peculiar to the gds/shmem code.

I'd appreciate any assistance. Meantime, I am disabling gds/shmem by default (can be overridden by setting `PMIX_MCA_gds=shmem,hash` in the environment) so work can continue.

rhc54:
The xversion tests for client and `make check` were failing when running the PR against the `master` branch. My testing was solely against the xversion scenario in a Linux container. However, the only change between those two branches was the addition of the two attribute definitions. No other code change.

It all worked fine on my Mac - but gds/shmem is disabled there.

I tried removing the definitions, and it all passed. Tried restoring the definitions and removing two others - and it all passed. Increasing the number of attributes - and it failed. Very weird.

rhc54:
I've confirmed that for some odd reason, it only happens during xversion tests between the two branches. I then tried running the v5.0 branch as the server with the master branch as the client - and that worked. I tried reversing the two, and that worked as well.

So since the master branch has been updated with the commit from that PR, the whole thing doesn't make much sense. I'll remove that temp fix and let's just see how things go from here. It looks like there may have been some weirdness about the PR branch.

samuelkgutierrez:
We don't do anything special regarding attribute definitions, so strange indeed! Thank you for looking more into this, @rhc54.

PR #3160, opened 2023-09-15
Add a new Github Action

jsquyres:
Add a few build tests Github Actions:

* Ubuntu
* MacOS

Do both with non-vpath and vpath, and both of those with and without sphinx.

Also do a distcheck.


jsquyres:
@rhc54 gave a 👍 in Slack.  Merging.

PR #3164, opened 2023-09-24
asm: Remove unneeded asm setup tests

bwbarrett:
PMIx has never supported building atomics from assembly files (as opposed to inline assembly or compiler builtins), but had the configure tests to support that mode.  The configure tests cause configure to fail on MacOS with XCode 15, so remove them.

Note: PMIx 4.0 and later are not impacted by the issue, because they had a more comprehensive rewrite of the assembly infrastructure.


rhc54:
This branch cannot pass CI as it is too old - we don't update the CI here as the branch has passed out of support.

Issue #1819, opened 2020-06-28
PMIX ERROR: NO-PERMISSIONS in file ../../../../../../src/mca/common/dstore/dstore_base.c at line 234


racedowling:
Thank you for taking the time to submit an issue!

### What version of the PMIx Reference Library are you using? (e.g., v1.0, v2.1, git master @ hash, etc.)
bash$ mpiexec -V
mpiexec (OpenRTE) 3.1.3

### Describe how PMIx was installed (e.g., from a source/distribution tarball, from a git clone, from an operating system distribution package, etc.)
apt-get install mpi4py
pip3 install mpi4py

### Please describe the system on which you are running

* Operating system/version: Raspbian Buster (Debian)
* Computer hardware: Raspberry Pi 4
* Network type: Ethernet

-----------------------------

## Details of the problem

Please describe, in detail, the problem that you are having, including the behavior you expect to see, the actual behavior that you are seeing, steps to reproduce the problem, etc.  It is most helpful if you can attach a small program that a developer can use to reproduce your problem.

bash$ mpiexec /slurmfs/usr/bin/python3 calculate.py
pi is approximately 3.1418009868930938, error is 0.0002083333033007
[clust-1:10216] PMIX ERROR: NO-PERMISSIONS in file ../../../../../../src/mca/common/dstore/dstore_base.c at line 234

This happens every time mpiexec it is run.

**Note**: If you include verbatim output (or a code block), please use a [GitHub Markdown](https://help.github.com/articles/creating-and-highlighting-code-blocks/) code block like below:
```shell
shell$ prun -np 2 ./hello_world
```

jjhursey:
Sorry for the delay. We believe that this issue was fixed in [PMIx v2.2.4](https://github.com/openpmix/openpmix/releases/tag/v2.2.4). Open MPI 3.1.3 ships with PMIx 2.1.4.

You can try to update your Open MPI version through the package manager. Alternatively, you could build the latest Open MPI from source - the latest is [Open MPI v4.0.4](https://www.open-mpi.org/software/ompi/v4.0/)

rhc54:
Closed as this looks to have been fixed.

Twilight-Shuxin:
Hi, I am experiencing this same error here
```
PMIX ERROR: NO-PERMISSIONS in file ../../../../../../../../../../opal/mca/pmix/pmix3x/pmix/src/mca/common/dstore/dstore_base.c at line 237
[...]

PR #3169, opened 2023-10-03
Make checking min versions consistent

rhc54:
We have two dependent libraries we care about, and minimum version requirements on both of them. Let's connect them to the min values in the VERSION file and "standardize" the check configure code to make it easier to understand and maintain.

Also, since we require a min HWLOC version of 1.11, there is no longer a need to check for at least 1.8 so we know we have hwloc_topology_dup.

### DIFF
--- a/config/pmix_setup_hwloc.m4
+++ b/config/pmix_setup_hwloc.m4
@@ -1,4 +1,4 @@
-# -*- autoconf -*-
+# -*- shell-script -*-
 #
 # Copyright (c) 2009-2020 Cisco Systems, Inc.  All rights reserved
 # Copyright (c) 2013      Los Alamos National Security, LLC.  All rights reserved.
@@ -16,7 +16,7 @@
 # MCA_hwloc_CONFIG([action-if-found], [action-if-not-found])
 # --------------------------------------------------------------------
 AC_DEFUN([PMIX_SETUP_HWLOC],[
-    PMIX_VAR_SCOPE_PUSH([pmix_hwloc_dir pmix_hwloc_libdir pmix_check_hwloc_save_CPPFLAGS])
+    PMIX_VAR_SCOPE_PUSH([pmix_hwloc_dir pmix_hwloc_libdir])
 
     AC_ARG_WITH([hwloc],
                 [AS_HELP_STRING([--with-hwloc=DIR],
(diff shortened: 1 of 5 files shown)


rhc54:
@jsquyres I think there is a problem with the Dockerfile used in these xversion tests. They say they are using HWLOC 1.11.13, but we are generating an error indicating that it is below 1.11.0. I've tested it locally and all works fine with that HWLOC version, so I'm wondering if the Dockerfile is picking up some older version.

Note that we used to accept all the way down to 1.5.0, so moving to 1.11.0 is a significant change. However, it seems reasonable to me - note that PRRTE and OMPI both have that as min version in their VERSION files.

Can you take a peek and see if you can spot a problem? Otherwise, I'm going to consider those Dockerfiles suspect as I do see some odd messages out of them.

jsquyres:
Sure, I can have a look.  Will probably take me a day or three, however.

config/pmix_setup_hwloc.m4
```diff
@@ -52,72 +50,58 @@ AC_DEFUN([PMIX_SETUP_HWLOC],[
...
-           pmix_have_topology_dup=1],
-          [AC_MSG_RESULT([no])])
+    # NOTE: We have already read PMIx's VERSION file, so we can use
+    # those values
+    pmix_hwloc_min_num_version=PMIX_HWLOC_NUMERIC_MIN_VERSION
+    pmix_hwloc_min_version=PMIX_HWLOC_MIN_VERSION
+    AC_MSG_CHECKING([version at or above v$pmix_hwloc_min_version])
+    AC_PREPROC_IFELSE([AC_LANG_PROGRAM([
+                                        #include <hwloc.h>
+                                        #if (HWLOC_API_VERSION < $pmix_hwloc_min_num_version)
```
jsquyres, line 70: Actually, I just took a quick look: I think the problem is here.

The value that you're reading from `VERSION` (i.e., in `$pmix_hwloc_min_version`) is of the form "1.11.0".
But `HWLOC_API_VERSION` is a hex number, like `0x00020800`.

So you're comparing apples and oranges.

So I think you want to decide whether you want to compare hwloc's API version or package version.

If you want to compare the package version, you can do a trick just like we introduced into PRRTE to compare the PMIx package version number: for each of the A.B.C version numbers you want to compare, concatenate them and make them a big hex number.  Then you can simply compare the 2 hex numbers.

Like this: ...er, the change that was just merged in #1819 looks like it was subsequently removed. ☹️ 

If you look at the diff from #1819 in prte_setup_pmix.m4, new lines 84-89, I convert an A.B.C value to a large hex number that I can then directly compare to another hex number, effectively comparing A.B.C vs. D.E.F to see which one is larger.

Was that removed because it didn't work?

rhc54: No - I'm just trying to make the configure codes consistent. Would rather not have every dependent library with its own method for checking against min version. So I changed `autogen.pl` to generate the numbers (ported from OMPI) and verified that the hex numbers were being properly created. I then use those values in the configure code instead of re-generating them.

If you look in `config/autogen_found_items.m4`, you'll see the results - easy way to ensure that we are comparing apples-to-apples. Looking locally, `PMIX_HWLOC_NUMERIC_MIN_VERSION` is set to `0x00010b00`, which is less than the `0x00010b06` found in HWLOC 1.11.13.

I agree it's an API version - frustrating that HWLOC doesn't provide the actual numeric version. I'm thinking of revising the tests to use `HWLOC_VERSION_MAJOR` and friends as a workaround.

rhc54: Sadly, v1.11 doesn't have the major/minor/release numbers available - so all you can do is check the api version. Not ideal. I suppose we could have one method for checking pre-2.0 that uses the API version, and then another that looks at the real version for 2.0 and above - but that seems like a lot of work for little gain. 🤷‍♂️

rhc54: Think I might have figured out the problem. These failures are all cases where HWLOC is not located in a system location, and therefore must be specified by `--with-hwloc=path`. In the configure code, however, I didn't alter the `CPPFLAGS` to point to where the include file can be found. So I'm thinking that without that alteration, we are probably picking up some old include file that doesn't meet our requirement in the test.

I'll give that a try and see if it fixes things.

PR #3180, opened 2023-10-20
Add a new attribute to specify connection order

rhc54:
Instead of having individual attributes to specify a connection target and others to specify an order for each target, just have an attribute that allows one to specify an ordered list of targets.


rhc54:
Going to commit this as we have seen this nonsensical error before - and it disappears once the attributes have been updated.

PR #3181, opened 2023-10-20
Test - do not merge

rhc54:


rhc54:
@samuelkgutierrez I know we have encountered this before, but I cannot remember the remedy. Adding this attribute causes the shmem subsystem to fail - disable shmem and everything passes just fine, which is why it doesn't show up on my Mac.

Obviously, the attribute itself isn't the issue, and the shmem isn't doing anything special with attrs. Do you recall what caused the problem, and how we got around it?

samuelkgutierrez:
Hi, @rhc54. Oof, I don't recall. I'll update the issue if it comes to me. I'll also dig around past commits, while I'm at it.

samuelkgutierrez:
[This](https://github.com/openpmix/openpmix/pull/3137#issuecomment-1692512786) is the last time it came up.

PR #3190, opened 2023-11-03
hash: Update pmix_hash functions to accept a pmix_keyindex_t*.

samuelkgutierrez:
Update all pmix_hash functions to include an extra parameter: a pmix_keyindex_t *. This provides the ability to have hash lookups across different key/index translation tables, not just pmix_globals.keyindex.

Provide a new pmix_keyindex_t data type that eases the embedding of key/index structures into shared-memory.

Remove all internal pmix_hash APIs of the form pmix_hash_*in_keyindex in favor of the new form provided here.

Note: this work is slated for use in gds/shmem. Specifically, we are aiming to fix a potential issue that may arise when using gds/shmem when key/index values differ between server and client versions.

### DIFF
--- a/src/runtime/pmix_finalize.c
+++ b/src/runtime/pmix_finalize.c
@@ -63,7 +63,6 @@ void pmix_rte_finalize(void)
     int i;
     pmix_notify_caddy_t *cd;
     pmix_iof_req_t *req;
-    pmix_regattr_input_t *p;
 
     if (!pmix_init_called) {
         return;
@@ -147,22 +146,6 @@ void pmix_rte_finalize(void)
     }
     PMIX_LIST_DESTRUCT(&pmix_globals.nspaces);
     PMIX_LIST_DESTRUCT(&pmix_client_globals.groups);
-
-    for (i=0; i < pmix_globals.keyindex.size; i++) {
-        p = (pmix_regattr_input_t*)pmix_pointer_array_get_item(&pmix_globals.keyindex, i);
-        if (NULL != p) {
-            if (NULL != p->name) {
-                free(p->name);
-            }
-            if (NULL != p->string) {
-                free(p->string);
-            }
-            if (NULL != p->description) {
-                PMIx_Argv_free(p->description);
-            }
-            free(p);
-        }
-    }
     PMIX_DESTRUCT(&pmix_globals.keyindex);
     free(pmix_globals.myidval.data.proc);
 
--- a/src/runtime/pmix_init.c
+++ b/src/runtime/pmix_init.c
@@ -19,7 +19,7 @@
  * Copyright (c) 2015      Research Organization for Information Science
  *                         and Technology (RIST). All rights reserved.
  * Copyright (c) 2021-2023 Nanook Consulting.  All rights reserved.
- * Copyright (c) 2022      Triad National Security, LLC. All rights reserved.
+ * Copyright (c) 2022-2023 Triad National Security, LLC. All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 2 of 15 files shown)


src/runtime/pmix_init.c
```diff
@@ -328,8 +327,8 @@ int pmix_rte_init(uint32_t type, pmix_info_t info[], size_t ninfo, pmix_ptl_cbfu
     ret = pmix_hotel_init(&pmix_globals.notifications, pmix_globals.max_events, pmix_globals.evbase,
                           pmix_globals.event_eviction_time, _notification_eviction_cbfunc);
     PMIX_CONSTRUCT(&pmix_globals.nspaces, pmix_list_t);
-    PMIX_CONSTRUCT(&pmix_globals.keyindex, pmix_pointer_array_t);
-    pmix_pointer_array_init(&pmix_globals.keyindex, 1024, INT_MAX, 128);
+    PMIX_CONSTRUCT(&pmix_globals.keyindex, pmix_keyindex_t);
+    pmix_pointer_array_init(pmix_globals.keyindex.table, 1024, INT_MAX, 128);
```
rhc54, line 331: Just a nit - why not just put the array_init call in the constructor?

src/runtime/pmix_finalize.c
```diff
@@ -148,8 +148,8 @@ void pmix_rte_finalize(void)
     PMIX_LIST_DESTRUCT(&pmix_globals.nspaces);
     PMIX_LIST_DESTRUCT(&pmix_client_globals.groups);
 
-    for (i=0; i < pmix_globals.keyindex.size; i++) {
-        p = (pmix_regattr_input_t*)pmix_pointer_array_get_item(&pmix_globals.keyindex, i);
+    for (i=0; i < pmix_globals.keyindex.table->size; i++) {
+        p = (pmix_regattr_input_t*)pmix_pointer_array_get_item(pmix_globals.keyindex.table, i);
```
rhc54, line 152: Might as well just put this in the pmix_keyindex_t destructor?

rhc54 (commented):
Just a couple of suggestions to make the code a tad cleaner

src/runtime/pmix_init.c
samuelkgutierrez: I agree, it's not great. Since we can't include the initial size during construct (and we have to be careful about now we use shared-memory), I left it as a separate step.

src/runtime/pmix_finalize.c
samuelkgutierrez: Makes perfect sense! I've included change this in my latest push.

src/runtime/pmix_init.c
rhc54: Understood - but the initial size truly doesn't matter. You could set it to anything - say 256 - as a lower bound since we know we have a couple of hundred defined attributes anyway. You have no upper bound, and so it will simply realloc when something beyond it's current size gets set - and then realloc again if/when that boundary gets hit.

Even if you leave it at 1024, it is so small compared to the shmem footprint that it is in the noise. So you might as well just put it in the constructor.

samuelkgutierrez: The problem is that we can't realloc using the gds/shmem's TMA.

rhc54: Ouch - in that case, we have a real design problem. Much of the system relies on the ability to realloc as the job scales. Perhaps we need to start by looking at that problem?

samuelkgutierrez: We were aware of this limitation. That's why I'm careful about how I size data structures. Supporting realloc would require a massive change to gds/shmem's memory allocator. To support realloc, we would have to implement a proper memory allocator, not just a simple linear slab allocator.

rhc54: We probably should chat about this by voice instead of here - would have a time?

samuelkgutierrez: Oops. My comment was lost before I resolved this. Let me recreate what was lost:

Offline conversion: this limitation is understood, but we need to have a broader conversion about the current limitations of the gds/shmem memory allocation subsystem and its implications.

samuelkgutierrez: Addressed in #3214.

PR #3191, opened 2023-11-03
gds/shmem: Improve error message in tma_realloc().

samuelkgutierrez:
In tma_realloc(), provide an error message describing the current limitations of the gds/shmem's TMA and a solution.


rhc54:
Thanks - that looks good!

PR #3192, opened 2023-11-05
gds/shmem: Implement first cut of tma_realloc().

samuelkgutierrez:
Implement realloc() for gds/shmem's shared-memory TMA.


rhc54:
Of course - was only holding off as it seemed like WIP and you might not want it merged.

samuelkgutierrez:
My mistake: I can see how that was confusing! Thank you, @rhc54.

PR #3257, opened 2024-01-19
Add a query attribute for number of available slots

rhc54:
Allow someone to query the number of slots available in a given session. Note that this is purely a point-in-time measurement as jobs may be working there way thru the state machine for mapping, and more jobs may be
submitted at any moment.


rhc54:
@hppritcha First half of the query support

Issue #3259, opened 2024-01-23
PMIX_RANK_UNDEF doesn't work with PMIx_Get()


RaymondMichael:
### What version of the PMIx Reference Library are you using? 

4.2.6

### Describe how PMIx was installed 

Running with HPE's PALS launcher.

### Please describe the system on which you are running

Running with 1 host and 2 ranks.

## Details of the problem

In the following program, the PMIx_Get() fails to find the key when the passed proc.rank is PMIX_RANK_UNDEF. When proc.rank is changed to the rank that posted the key, it finds it fine. According to the spec, this should work.

#include <assert.h>
#include <stdio.h>
#include <pmix.h>
#include <string.h>

int
main(int argc, char *argv[])
{
    pmix_value_t *val = NULL;
    pmix_status_t rc = PMIX_SUCCESS;
    pmix_proc_t global_proc, proc;

    rc = PMIx_Init(&global_proc, NULL, 0);
    assert(PMIX_SUCCESS == rc);

    rc = PMIx_Get(&global_proc, PMIX_LOCAL_RANK, NULL, 0, &val);
    assert(PMIX_SUCCESS == rc);

    int local_rank = val->data.uint16;
    PMIX_VALUE_RELEASE(val);

    char key_p[32];
    pmix_value_t value_p;

    sprintf(key_p, "%s-%d", "foo", local_rank);
    value_p.type = PMIX_UINT32;
    value_p.data.uint32 = local_rank + 10;

    rc = PMIx_Put(PMIX_GLOBAL, key_p, &value_p);
    assert(PMIX_SUCCESS == rc);

    rc = PMIx_Commit();
    assert(PMIX_SUCCESS == rc);

    PMIX_PROC_CONSTRUCT(&proc);
    proc.rank = PMIX_RANK_UNDEF;
    strcpy(proc.nspace, global_proc.nspace);

    for (int i = 0; i < 2; i++) {
        char key_g[32];
        pmix_value_t *value_g;
#if 0
        proc.rank = i;
#endif
        sprintf(key_g, "%s-%d", "foo", i);
        rc = PMIx_Get(&proc, key_g, NULL, 0, &value_g);
        if (PMIX_SUCCESS != rc) {
            fprintf(stderr, "PMIx_Get got %d\n", rc);
        } else {
            fprintf(stderr, "Got %d\n", value_g->data.uint32);
        }
    }

    PMIx_Finalize(NULL, 0);

    return 0;
}

[...]

PR #3260, opened 2024-01-24
Fix "get" of key with undefined rank

rhc54:
Users are allowed to request a key using
PMIX_RANK_UNDEF _if_ the key is globally
unique. In order to do this, the gds/hash
component in the client needs to know how
many procs are in the job, and we need to
return the requested data from the server
to the client so it can be found since
the dstore cannot handle undefined ranks.

Thanks to @RaymondMichael for the report.

Fixes https://github.com/openpmix/openpmix/issues/3259

bot:notacherrypick

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -185,6 +185,7 @@ test/simple/simpjctrl
 test/simple/simpio
 test/simple/data-*
 test/simple/simpcoord
+test/simple/legacy
 test/python/run_sched.sh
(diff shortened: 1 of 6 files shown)


RaymondMichael:
Thanks

rhc54:
We're getting ready to drop v4.2.9, so this will be included - waiting for OMPI's final test/review. Should be within a week.

PR #3273, opened 2024-01-31
Protect against empty envar definition for mca_base_param_files

rhc54:
If someone defines the PMIX_MCA_mca_base_param_files or the PMIX_MCA_mca_base_override_param_file envar as NULL (i.e., they specify the envar "name=" with no value), then the envar will be found but have a NULL value. Current code doesn't protect against that contingency, so fix it.


rhc54:
@ggouaillardet I cannot find where you referenced this bug over in OMPI - can you please add the reference here?

PR #3275, opened 2024-02-01
Fix test builds with picky compiler options.

samuelkgutierrez:
Newer compilers don't accept the void cast of asprintf's return value, so address those sites in test code.


rhc54:
Huh - which compiler is that?

rhc54:
FWIW: I'm using clang 15.0.0, which is (I believe) their most recent release, and I'm not seeing it.

samuelkgutierrez:
I felt the same. gcc 13.2.0 is the version that complained about all those sites.

PR #3283, opened 2024-02-07
Improve PMIx attribute lookup efficiency.

samuelkgutierrez:
Break out of the lookup loops when we pass the end of table element occupancy.


rhc54:
Good point - they are left-justified.

PR #3285, opened 2024-02-07
Re-enable gds/shmem by default

samuelkgutierrez:
Revert relevant commits to re-enable use of gds/shmem by default.

### DIFF
--- a/src/mca/gds/shmem/gds_shmem.c
+++ b/src/mca/gds/shmem/gds_shmem.c
@@ -1279,6 +1279,12 @@ assign_module(
             break;
         }
     }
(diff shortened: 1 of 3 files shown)


rhc54:
This is fine. I had hoped that the earlier v5.x series would barf in some manner when the server passed down the dictionary, but they don't - they continue to do the wrong thing. Likewise, if this updated branch in a client finds itself with an earlier v5.x server, then it needs to fallback to the hash component.

I'll add some logic to protect against those scenarios.

samuelkgutierrez:
Thank you, Ralph.

PR #3286, opened 2024-02-08
Rename the gds/shmem component to gds/shmem2

rhc54:
Prevent the v5.0.0 and v5.0.1 releases from attempting to use the revised shmem support. This will force both of them to fall back to hash.


samuelkgutierrez:
Clever, I like it 😎

PR #3294, opened 2024-02-15
Fix the dictionary transfer in shmem2

rhc54:
We can _never_ assume that the indices in our dictionary match the ones in the server's dictionary. Even if the number of entries is the same, there is no guarantee that they are in the same order - and thus, the index of a particular entry can be different, causing errors on our side. Likewise, even if we have more entries than they do, there is no guarantee that their entries are in the same order as ours. So we have to rebuild the dictionary to match what they have - and if we have additional entries, then we add them back at the end of the new dictionary.

(cherry picked from commit 6163f2127e6058815998bc934be63a69cb87383e)

### DIFF
--- a/src/include/pmix_globals.h
+++ b/src/include/pmix_globals.h
@@ -15,7 +15,7 @@
  *                         and Technology (RIST).  All rights reserved.
  * Copyright (c) 2019      Mellanox Technologies, Inc.
  *                         All rights reserved.
- * Copyright (c) 2021-2023 Nanook Consulting.  All rights reserved.
+ * Copyright (c) 2021-2024 Nanook Consulting  All rights reserved.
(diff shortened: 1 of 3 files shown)


rhc54:
@samuelkgutierrez Just an FYI: The new shmem dictionary transfer failed cross-version testing when I tried to port it to v5.0. Apparently, there was a check in there that ignored the dictionary exchange if the client had more keys than the server (or maybe the other way around). Unfortunately, that isn't quite correct - see the above comment as to why.

I fixed it in v5.0 and am forward-porting it here. Not the most efficient algorithm, I grant you - but it does seem to solve all cases that I could try (server has more, client has more, have same entries but different order, have same number but different entries).

PR #3310, opened 2024-03-08
More cleanup of group operations and local client array

rhc54:
Ensure we remove clients from the local client array when deregistered, and that all clients get cleaned up when the nspace is deregistered as some people may not have deregistered clients prior to the nspace.

Do not reuse the "hybrid" field of the tracker to track the group operation as it is used for other purposes and should not be overwritten. Add a new PMIX_GROUP_NONE operation tag for initializing the new tracker field - put it at the end of the enum to avoid causing errors with existing users.

### DIFF
--- a/include/pmix_common.h.in
+++ b/include/pmix_common.h.in
@@ -1740,7 +1740,8 @@ typedef enum {
 
 typedef enum {
     PMIX_GROUP_CONSTRUCT,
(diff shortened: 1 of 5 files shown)


rhc54:
These changes need to go back to PMIx v5.0 release branch.

PR #3315, opened 2024-03-09
Fix 'X may be used uninitialized' warnings.

samuelkgutierrez:


rhc54:
Thanks! Always interesting how different compilers catch different things.

PR #3331, opened 2024-04-04
fixes for pr 3329

hppritcha:
PR 3329 introduced a regression which causes Open MPI's MPI_Intercomm_from_groups method to fail.

Related to https://github.com/open-mpi/ompi/pull/12449


rhc54:
Ouch - thx for catching it!

PR #3338, opened 2024-04-16
3 Bugs in PMIx_Log server path

dsolt:


rhc54:
Hmmm...given that this code gets successfully executed every time OMPI outputs a "show_help" message (which is a _LOT_), I'm a little puzzled by the proposed changes. Let me work thru the plugins to ensure that something else isn't going on here. Can you provide some insight as to where you are seeing these problems, and what you are seeing?

dsolt:
So some of this may be because I was running the debug build.  I'm not sure though.  I can try separating out which of the 3 commits was only needed for debug builds or not.   Also, attached is a simple server / client that we are trying to put together for a use-case document.   The server is EXTREMELY minimal.  It only runs on a single host and only allows for tools to connect to it and only supports PMIx_Log.   It will hang in PMIx_Log without these changes.

[SS.tar.gz](https://github.com/openpmix/openpmix/files/15013173/SS.tar.gz)

dsolt:
Is this a new thing?  I don't remember doing that in JSM, though I admit I haven't looked at that code in a few years now.

rhc54:
It has always been a server requirement in the Standard as well as the library. Sometimes you can get away with avoiding it for a specific function - e.g., returning `PMIX_OPERATION_SUCCEEDED` to indicate that the operation was done atomically and no callback will be issued. However, you were never allowed to execute the callback from within the server "upcall" function.

dsolt:
Ah yeah.  I forgot in JSM we would almost always send a message to our head node and then get a reply back before calling the callback.  I'm not used to this one-node simplicity.

rhc54:
Yeah - FWIW: this is why we have avoided trying to provide server-side code examples. You immediately get into the complexity of thread-shifting, both to allow proper callback execution and because you otherwise are operating the server code inside the PMIx library's progress thread - which means you risk getting into trouble with your own global variables.

Perhaps it would be better to provide pseudo-code and then point to PRRTE for operational code? Just a thought.

PR #3371, opened 2024-07-10
Apply prefix to copied version of the app array

### LINKED ISSUE #3370: PMIx_Spawn prefix handling issue on 5.0.2

5.0.2

RPMs built from source

* Operating system/version: SLES 15 SP5
* Computer hardware: VM
* Network type: VM network

-----------------------------

## Details of the problem

We have a PMIx spawn test program that worked fine with PMIx 4.2.9 but fails on PMIx 5.0.2. I've found two issues:

1. The app.cmd passed in by the client must be dynamically allocated, otherwise there's a invalid free().
2. The PMIX_PREFIX isn't applied to the app.cmd or app.argv passed into the PMIx server.

I think this is related to https://github.com/openpmix/openpmix/pull/3330.

It seems to me that this section of code in `src/client/pmix_client_spawn.c` should be modifying `appsptr[n]` rather than `aptr`:

```
319         // adjust the cmd prefix if required
320         if (NULL != prefix) {
321             // prefix the command
322             pmix_asprintf(&tmp, "%s/%s", prefix, aptr->cmd);
323             free(aptr->cmd);
324             aptr->cmd = tmp;
325         } else if (NULL != defprefix) {
326             // prefix the command
327             pmix_asprintf(&tmp, "%s/%s", defprefix, aptr->cmd);
328             free(aptr->cmd);
329             aptr->cmd = tmp;
330         }
```

rhc54:
When applying a specified prefix, apply the change to the copy of the incoming app array and not the incoming array elements themselves in case they were statically defined.

Thanks to @dgloe-hpe for the report!

Fixes #3370

### DIFF
--- a/.github/actions/mlnx/Dockerfile
+++ b/.github/actions/mlnx/Dockerfile
@@ -1,8 +1,8 @@
-FROM centos:7.6.1810
+FROM almalinux:9
 
 RUN \
-    yum install -y perl perl-Data-Dumper \
-        python3 python3-pip python-virtualenv \
+    dnf install -y perl perl-Data-Dumper \
+        python3 python3-pip zlib \
         automake libtool flex make bzip2 git which rpm-build libevent-devel hwloc hwloc-devel
 
 COPY entrypoint.sh /entrypoint.sh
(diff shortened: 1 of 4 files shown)


rhc54:
@janjust This appears to be the root cause of the MLNX CI failure:
```shell
  0.566 14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"
```
I suspect the problem is that CentOS-7 was declared EOL over the last weekend, and so all things CentOS are now dead. I'm trying to adjust it to use Rocky instead.

PR #3374, opened 2024-07-17
configure: fix broken bashisms resulting in logic failure

eli-schwartz:
Some code which was only valid using GNU bash was included. Detected in Gentoo packaging via:

```
 * QA Notice: Abnormal configure code
 *
 * ./configure: 40936: test: 0: unexpected operator
```

Bash provides the standard `test XXX = YYY` or `[ XXX = YYY ]` utilities. It also provides the ability to spell the equals sign as a double equals. This does nothing whatsoever -- it adds no new functionality to bash, it forbids nothing, it is *literally* an exact alias.

It should never be used under any circumstances. All developers must immediately forget that it exists. Using it is non-portable and does not work in /bin/sh scripts such as configure scripts, and it results in dangerous muscle memory when used in bash scripts because it makes people unthinkingly use the double equals even in /bin/sh scripts. To add insult to injury, it makes scripts take up more disk space (by a whole byte! and sometimes even a few bytes...)

Delete this accidental bashism, and restore the ability to get correct ./configure behavior on systems where /bin/sh is something other than a symlink to GNU bash.

### DIFF
--- a/src/mca/gds/shmem2/configure.m4
+++ b/src/mca/gds/shmem2/configure.m4
@@ -18,7 +18,7 @@ AC_DEFUN([MCA_pmix_gds_shmem2_CONFIG], [
     dnl address space is probably too small for the 'virtual memory hole'
     dnl finding that we do here. Below assumes support for only 32- and 64-bit
     dnl architectures.
-    AS_IF([test $ac_cv_sizeof_void_p -ne 4 && test $oac_have_apple == 0],
+    AS_IF([test $ac_cv_sizeof_void_p -ne 4 && test $oac_have_apple = 0],
           [$1
            pmix_gds_shmem2=yes],
           [$2


rhc54:
I appreciate the contribution, but perhaps we could dial down the rhetoric a tad? GNU bash-isms are not the great evil, but it's fine to clean them up if/when necessary. I suspect you will encounter them quite a bit in various code snippets here-and-there.

eli-schwartz:
It's a bit of a pet peeve of mine and I freely admit to being a bit dramatic about it, but the problem is not GNU bash'isms.

If I saw a mistaken `[[ -n ${VAR} ]]`, which is also a bashism, I would kinda shrug and fix it. The double equals is just dumb, in my highly personal opinion, because it serves no purpose in a bash script either and appears to exist solely to lay traps for the unwary. It feels unfair that it was ever added to bash at all.

Every other bashism I can think of provides useful functionality in a bash script and I am disappointed when I have to use POSIX sh since I can't use the bashisms.

But I am not Chet Ramey so I cannot make the next version of bash issue a diagnostic warning every time `==` is used, saying it is deprecated without replacement and people should stop using it. Oh well, such is life. :P

rhc54:
I'll leave you to your crusade, though I doubt you'll succeed. Let's just please avoid the rant in the future as it has nothing to do with the proposed change. A simple "change `==` to `=` to allow configure to operate under sh as well as bash" would suffice and be more convincing.

PR #3382, opened 2024-07-19
Protect against LTO optimizer

rhc54:
We currently do not support the LTO optimizer
as it is incompatible with our plugin component
architecture. So detect it has been specified
in configure and error out with an explanation.

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -890,6 +890,19 @@ AC_DEFUN([PMIX_SETUP_CORE],[
     CPP_INCLUDES="$(echo $cpp_includes | $SED 's/[[^ \]]* */'"$pmix_cc_iquote"'&/g')"
     CPPFLAGS="$CPP_INCLUDES -I$PMIX_top_srcdir/include $CPPFLAGS"
 
+
+    # We do not currently support the "lto" optimizer as it
+    # aggregates all the headers from our plugins, resulting
+    # in a configuration that generates warnings/errors when
+    # passed through their optimizer phase. We therefore check
+    # for the flag, and if found, output a message explaining
+    # the situation and aborting configure
+    _PMIX_CHECK_LTO_FLAG($CPPFLAGS, CPPFLAGS)
+    _PMIX_CHECK_LTO_FLAG($CFLAGS, CFLAGS)
+    _PMIX_CHECK_LTO_FLAG($LDFLAGS, LDFLAGS)
+    _PMIX_CHECK_LTO_FLAG($LIBS, LIBS)
+
+
     ############################################################################
     # final wrapper compiler config
     ############################################################################
--- a/config/pmix_check_cflags.m4
+++ b/config/pmix_check_cflags.m4
@@ -2,7 +2,7 @@ dnl -*- shell-script -*-
 dnl
 dnl Copyright (c) 2021 IBM Corporation.  All rights reserved.
 dnl
-dnl Copyright (c) 2021      Nanook Consulting.  All rights reserved.
+dnl Copyright (c) 2021-2024 Nanook Consulting  All rights reserved.
 dnl $COPYRIGHT$
 dnl
 dnl Additional copyrights may follow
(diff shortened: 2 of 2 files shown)


config/pmix_check_cflags.m4
```diff
@@ -40,3 +40,14 @@ AC_MSG_CHECKING(if $CC supports ([$1]))
                 AC_MSG_RESULT([yes])
             fi
 ])
+
+AC_DEFUN([_PMIX_CHECK_LTO_FLAG], [
+    chkflg=`echo $1 | grep -- -flto`
+    if test "$chkflg"; then
```
jsquyres, line 46:
```suggestion
    if test -n "$chkflg"; then
```

I can't remember offhand if `test "$foo"` is the same as `test -n "$foo"`.

I'm 100% sure that `-n` means "check for a non-empty string".  I don't remember offhand if having no test operator means "check for a true-ish value" or "check for a non-empty string".

config/pmix.m4
```diff
@@ -890,6 +890,17 @@ AC_DEFUN([PMIX_SETUP_CORE],[
     CPP_INCLUDES="$(echo $cpp_includes | $SED 's/[[^ \]]* */'"$pmix_cc_iquote"'&/g')"
     CPPFLAGS="$CPP_INCLUDES -I$PMIX_top_srcdir/include $CPPFLAGS"
 
+
+    # We do not currently support the "lto" optimizer as it
+    # aggregates all the headers from our plugins, resulting
+    # in a configuration that generates warnings/errors when
+    # passed through their optimizer faze. We therefore check
```
jsquyres, line 897:
```suggestion
    # passed through their optimizer phase. We therefore check
```

config/pmix.m4
```diff
@@ -890,6 +890,17 @@ AC_DEFUN([PMIX_SETUP_CORE],[
...
+    # for the flag, and if found, output a message explaining
+    # the situation and aborting configure
+    _PMIX_CHECK_LTO_FLAG($CPPFLAGS, CPPFLAGS)
+    _PMIX_CHECK_LTO_FLAG($CFLAGS, CFLAGS)
```
jsquyres, line 901: Might want to check $LDFLAGS and $LIBS, too.

I *think* the most appropriate location for `-flto` is in `$LDFLAGS` (it's a linker flag, after all), but people could choose to put it in any weird location (including $CFLAGS, $CPPFLAGS, $LIBS, etc.).

config/pmix_check_cflags.m4
```diff
@@ -40,3 +40,14 @@ AC_MSG_CHECKING(if $CC supports ([$1]))
...
+        AC_MSG_WARN([Configure has detected the presence of the -flto])
+        AC_MSG_WARN([compiler directive in $2. PMIx does not currently])
+        AC_MSG_WARN([support this optimizer as it conflicts with the])
```
jsquyres, line 49:
```suggestion
        AC_MSG_WARN([support this flag as it conflicts with the])
```

config/pmix_check_cflags.m4
rhc54: It is the same, but no harm in adding the `-n` to make it clear.

jsquyres (approved):

config/pmix.m4
eli-schwartz: It's actually "required" to be specified both at compile time and link time.
- If you pass it at link time only, it doesn't find any LTO btyecode to perform LTO on, and it's way too late to convert source code into bytecode.
- If you pass it at compile time only, the compiler *may* (probably does! I know gcc does, but I can't remember what clang does) choose to infer from the presence of bytecode in the object files, that it can do LTO. Object files can double-compile and produce both object code and bytecode, in which case you can choose to disable it at link time (-fno-lto) and it simply ignores the bytecode...

Basically the tl;dr here is that checking in CFLAGS is sufficient but checking in both won't hurt.

PR #3399, opened 2024-08-29
add contrib/construct_event_strings.py to the dist tarball

ggouaillardet:


rhc54:
I have no objection to adding it, but this PR isn't really needed. The output of that script is already included in the tarball, and so the Makefile simply skips running the constructor (as it should).

Still, for completeness, I see no reason not to include it. 🤷‍♂️

ggouaillardet:
i get your point.

An issue was recently open in Open MPI, and it is not clear whether the root cause were some messed up timestamps, or the user did something like `make distclean && autogen.pl --force && configure && make`.

adding this file to the dist tarball could prevent that kind of should-have-not-happened issue.

PR #3424, opened 2024-10-24
Reduce code complexity in the modex operation and fix a bug

rhc54:
Modex operations typically involve a handful of keys from each proc, and we don't really see thousands of procs/node. So let's not spend time creating key maps and adding a lot of complexity to the modex operation. Use the compression capability instead.

The shmem2 component requires that it know when we are done passing modex info to it so that it can mark the involved job segments as "ready". Unfortunately, the current code assumed that all procs in the modex were from the same namespace. While often true, it is not a requirement and does get violated. So provide a "marker" for each involved nspace indicating that all modex info for that nspace has been processed.


rhc54:
@samuelkgutierrez We currently target modex info to the hash component, but I'm hoping to remove that constraint. So would you mind taking a gander at the changes here to ensure I didn't break something in shmem2?

samuelkgutierrez:
Hi, @rhc54. I sure can. I'm unavailable until Nov. 4th, but can take a look then. Will that work?

rhc54:
That would be great - thx!!

samuelkgutierrez (approved):
Hi, @rhc54. The changes look fine to me. The only thing that I would double check is that the use case that caused us to divert modex data to hash isn't broken. I can't recall what that was, but I know something was subtly broken in the modex-->shmem path. Perhaps this commit fixes that.

rhc54:
No, I'm afraid this doesn't address that problem. It has to do with replacing keys with indices - problem is that the non-standard keys being defined by the app are given different local indices. Then when we try to load the modex data, we hit the problem of needing to correct the dictionary - which is now a rather complex operation.

Haven't given up on it - it may be complex, but that doesn't mean unsolvable! Thanks for the review!

samuelkgutierrez:
You bet!

PR #3425, opened 2024-10-25
Reverse the default "squash" setting

rhc54:
Unfortunately, the implementation of the "psquash" framework violated the PMIx design rules - and was not caught at the time it was released. The code _assumes_ that both sides of the communication are using the same psquash component, which is definitely not guaranteed.

I've tried several remedies, but nothing is particularly satisfying. Adding the psquash active component to the connection handshake doesn't solve the backward compatibility issue and just further complicates an already complex operation. Simply defaulting to always using either psquash component breaks backward compatibility as prior releases will continue to dynamically select the active one.

Given the default settings in that framework, the least breakage results from defaulting to using the "squashed" component. Not the ideal solution as that adds code complexity for what appears to be minimal gain, but forcing all prior releases to set an MCA param directing that they use the "native" component is a user-unfriendly solution.

So bite the old tongue and make the "squash" behavior be the only one supported going forward. This doesn't fully solve the problem for prior releases - nothing we can do about it, we screwed up and should never have released the code in its current form.

### DIFF
--- a/src/mca/bfrops/base/Makefile.include
+++ b/src/mca/bfrops/base/Makefile.include
@@ -13,7 +13,7 @@
 # Copyright (c) 2012      Los Alamos National Security, Inc.  All rights reserved.
 # Copyright (c) 2013-2017 Intel, Inc. All rights reserved.
 # Copyright (c) 2016      Cisco Systems, Inc.  All rights reserved.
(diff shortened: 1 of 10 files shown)


rhc54:
Failures to/from master are expected

PR #3451, opened 2024-11-21
.gitignore: add configure~

jsquyres:
The file "configure~" is generated when building OpenPMIx as part of Open MPI.  This causes an annoying git submodule "untracked content" notification.  So let's just add "configure~" to .gitignore so that it's properly ignored by git / git submodules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ autom4te.cache
 aclocal.m4
 compile
 configure
+configure~
 configure.log.*
 make.log.*
 make.install.log.*


rhc54:
Huh - weird. I've never seen that file being created before. 🤷‍♂️ Definitely no harm in ignoring it!

jsquyres:
Sweet -- thanks!

What branches should I cherry-pick this to?

rhc54:
Just the v5.0, I guess - it's the only branch other than master that you are using.

PR #3481, opened 2024-12-23
Add mpi4py CI script

rhc54:


rhc54:
@jsquyres Would you mind giving this a quick glance to ensure I didn't screw something up? Figured we should cover both upstream projects.

.github/workflows/pmix_mpi4py.yaml
```diff
@@ -0,0 +1,185 @@
...
+    - name: Update PATH and set OPAL_PREFIX and LD_LIBRARY_PATH
+      run: |
+        sed -i "\|$RUNNER_TEMP/openmpi/bin|d" $GITHUB_PATH
+        echo OPAL_PREFIX=$RUNNER_TEMP/ompi >> $GITHUB_ENV
+        echo LD_LIBRARY_PATH=$RUNNER_TEMP/ompi/lib >> $GITHUB_ENV
+
+    - name: Test mpi4py (singleton)
+      run:  python test/main.py -v
+      if:   ${{ true }}
+      timeout-minutes: 5
```
jsquyres, line 185: This part is really just testing the relocatability of Open MPI, which you probably don't need.  It's not harmful; it just makes the test take a little longer (I probably shouldn't have included it in the PRTE version, either -- oops!).

jsquyres (approved):

.github/workflows/pmix_mpi4py.yaml
rhc54: Ah, good point - I'll pull it out of both places. Thanks!

PR #3497, opened 2025-01-10
Modify the return from PMIx_Resolve_peers for clarity

rhc54:
We return an error (e.g., "not found") if the nspace or node are not found. However, if they _are_ found but there are no peers from that nspace on the specified node, then return "success" (instead of "not found") with NULL for the array of peers.

Add a docs page that walks thru the "resolve" procedure and explains the various return status codes.


rhc54:
@sonjahapp I believe this might help with the "resolve peers" question - please see what you think.

PR #3519, opened 2025-01-24
Allow arbitrary keys to use data arrays

rhc54:
The shmem2 component incorrectly restricted use of data arrays to PMIx predefined keys. Anyone can pass a data array with a custom key - we just treat it as any other key. The hash component did this correctly, so there is a workaround - make the shmem2 component do it too.


rhc54:
@dgloe-hpe Please give this a try, if/when you can.

dgloe-hpe:
This fixes our problem, thanks!

PR #3522, opened 2025-01-30
Enable fetch of job-level data for another nspace

rhc54:
If a proc (e.g., a tool) wants to fetch a piece of job-level data for a namespace other than their own, then it is likely that the proc won't already have a copy of that data. If they have done a PMIx_Connect or PMIx_Group_construct operation that included the other namespace, then they should have it - but otherwise, they likely don't. This is normally the case for tools, as an example.

Ensure we flow the request up to the server for processing. Given that it is job-level info (i.e., it is a request for a reserved key), then we only want to flow thru the fetch process once. If it is found, then no need to search further in other scopes - if it isn't found in the initial "internal" scope, then it definitely won't be found on some other scope.

### DIFF
--- a/src/client/pmix_client_get.c
+++ b/src/client/pmix_client_get.c
@@ -1103,7 +1103,7 @@ doget:
     /* we didn't find the data in either the server or the internal hash
      * components. If this is a NULL or reserved key, then we do NOT go
      * up to the server unless special circumstances require it */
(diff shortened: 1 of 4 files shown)


rhc54:
@dgloe-hpe I believe this resolves the problem you reported - can you give it a try? I haven't checked to see how this ports to the v5.0 branch, but hopefully cleanly.

PR #3528, opened 2025-02-06
Add attribute to direct app library's GPU support

rhc54:
There apparently are some circumstances when an application can benefit from disabling the internal GPU support in one or more of its libraries. Let's assume that a library might also provide a mechanism by which that support can be defaulted to enabled or disabled.

Add a spawn attribute by which a user/launcher can specify whether to enable of disable such support. Also extend the "info true" utility to accept a broader range of common entries for designating "true" or "false".

Refs https://github.com/openpmix/prrte/pull/2134


rhc54:
@krempel-pt Might be of interest to you folks?

krempel-pt:
> @krempel-pt Might be of interest to you folks?

Thanks for the hint, will look into that.

PR #3529, opened 2025-02-09
Cleanup the "resolve" functions

rhc54:
The resolve functions were being constrained to look at only information already known by the client process - they would do a local "get" for the "local peers" key (or equivalent), but that only returns what the client has been told about.

If the client is asking about other nodes or other nspaces, then it could be that the client hasn't been informed about those nodes/nspaces. So we have to go up to the server (if we are connected) to see if the server has broader knowledge.

However, not all runtimes will inform every server about every job being executed - e.g., if there are no local procs for that job. So if the server sees that the host supports the "query" upcall, then ask the host to respond to the request. If the host doesn't support the upcall, or returns a non-success status (e.g., it might not support that query), then have the server respond to the best of its knowledge.

If the client isn't connected, or the server has no knowledge, then the client will respond to the best of its knowledge.


rhc54:
@sonjahapp Here is the full implementation - see https://github.com/openpmix/prrte/pull/2135 for the host-side query backend.

PR #3539, opened 2025-02-14
pctrl: fix exit code for help command line option

heylinn:
pctrl tool returns wrong exit code 1 for --help command line option.
It happens because of the --targets option check. But when --help is given, --targets is not required.

I tried to add a check for PMIX_CLI_HELP too, but pmix_cmd_line_parse() does not put this option to the results. Thats why I also added mystore() function call.

Please review this fix and give a feedback.
Thank you!

### DIFF
--- a/src/tools/pctrl/pctrl.c
+++ b/src/tools/pctrl/pctrl.c
@@ -142,6 +142,14 @@ int main(int argc, char **argv)
         }
     }
 
+    // check for help command line option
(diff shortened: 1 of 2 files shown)


heylinn:
Yes, your solution is better. It fixes --version exit code too.
Thank you for your reply.

rhc54:
You are welcome to file a revised PR, if you like - or I can do it if you prefer.

heylinn:
I think it will be better if you do it, as an author of the solution.

PR #3555, opened 2025-02-27
Silence error output when running as singleton

rhc54:
When executed as a singleton, we automatically search for available connections, just like a tool would do. However, if they are not found, we should not emit an error message if we were not specifically directed to find one - e.g., a system-level server.

Also fix the query operation to return any partial results it was able to find when operating as
a singleton.

(cherry picked from commit 2a762fff71d0aa57b286f4a3be15bc3cb87421fd)


rhc54:
Included in multi commit PR

PR #3569, opened 2025-03-16
Multiple commits

rhc54:
[Fix segfault on error in shmem2](https://github.com/openpmix/openpmix/commit/6c221c6d63d205b312833160b4b968dc62e7e9aa)

If someone mistakenly provides an incorrect path for the
directory where we will be storing the shmem backing file
(e.g., pass a relative directory and then change the
working directory), then we segfault in shmem2 because
of a stale job object in the tracking system. So ensure
we cleanup the job object when that happens so we can
cleanly fallover to "hash".

Remove a bunch of error logs when this happens since we are
not going to fail, but instead continue running. Add some
debug output to report which file was attempted to be opened.

(cherry picked from commit https://github.com/openpmix/openpmix/commit/d8d8b69da85cc54e450373f61be1066e5b6345a2)

[Add static component pointers](https://github.com/openpmix/openpmix/commit/e0a4b7c58988cb2f80c1647952db6fae04271f66)

With GCC's link time optimization, the compiler ends up seeing two
declarations of MCA component structures.  One in the component
itself, which has potentially a component-specific type, and one
in static-components.h, which is just declared as a base MCA
component struct.  The compiler doesn't like this.

This patch adds a level of indirection used by the static component
lookup code.  static-components.h now declares the existance of a
mca_component_t * and the list is of pointers to pointers.  The
MCA base framework has to do an extra dereference, but this is only
in the initialization path.  We also then have to have the actual
pointer exist, so add a macro PMIX_MCA_BASE_COMPONENT_INIT that we need
to add to every component definition to create the pointer (and
perhaps do more initialization later).

Use the capabilities flag to indicate that this PMIx version
supports the LTO optimizer.

Thanks to Brian Barrett for the design!

(cherry picked from commit https://github.com/openpmix/openpmix/commit/c829e36f0a6bf15a644c753ca3772742ab59f571)

[Rename pstat/linux component](https://github.com/openpmix/openpmix/commit/8b9e035e3b714b73c227166f99b971c2ceaf3d48)

Apparently, the "linux" name causes a problem in
macro expansion, so rename to "plinux"

(cherry picked from commit https://github.com/openpmix/openpmix/commit/f9be0a85047780f24919e5ebc0db63f28b50d886)

[Update NEWS](https://github.com/openpmix/openpmix/commit/1dac73a86facfc036ccce3dd7cb12469e6685b8a)

bot:notacherrypick


rhc54:
Hmmm...not going to be able to bring the LTO compatibility code over here or else it will break PRRTE v3.0 - unless I update it as well. All of which means there would be a change in min PMIx version for PRRTE v3.0 in mid-series. Probably not a good idea.

We'll just have to leave the ignore-LTO-flags behavior in this series and go with the new stuff in PMIx v6, requiring PMIx v6.0 as min version for PRRTE v4 and above.

PR #3588, opened 2025-05-09
Cleanup for show-help system

rhc54:
Add cmd line definitions for use by pmix-info. Remove deprecated function to ensure that older versions of PRRTE don't mistakenly try to build against newer
versions of PMIx as the show-help system is incompatible.


rhc54:
PRRTE builds will fail - this is expected.

PR #3612, opened 2025-05-24
Use pthread_np.h on platforms where it is required

rayegun:
I encountered an error attempting to cross compile for FreeBSD. On some of the BSDs pthread extensions are not found in `pthread.h` but in `pthread.np.h`. For reference here is that error:

```
runtime/pmix_progress_threads.c:176:14: error: call to undeclared function 'pthread_setaffinity_np'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  176 |         rc = pthread_setaffinity_np(trk->engine.t_handle, sizeof(cpu_set_t), &cpuset);
      |              ^
2 errors generated.
make[2]: *** [Makefile:1235: runtime/pmix_progress_threads.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
```

Adding `#include <pthread_np.h>` seems to get compilation working. Note I haven't tested this on an actual FreeBSD machine, however perusing GitHub this seems like a very common pattern to enable pthreads on FreeBSD, and the FreeBSD manpages indicate this is correct.

### DIFF
--- a/src/runtime/pmix_progress_threads.c
+++ b/src/runtime/pmix_progress_threads.c
@@ -19,7 +19,12 @@
 #ifdef HAVE_UNISTD_H
 #    include <unistd.h>
 #endif
+
 #include <pthread.h>
+#if (defined(__FreeBSD__) || defined(__OpenBSD__)) && defined(HAVE_PTHREAD_SETAFFINITY_NP)
+#    include <pthread_np.h>
+#endif
+    
 #include <string.h>
 #include <event.h>
 


rhc54:
Hmmm...I don't know anything about FreeBSD, so I can't speak to the correctness here. I'm afraid this change, however, is incomplete as you'd first need to test for the presence of that header so you don't break other builds (e.g., I don't see that header in my Ubuntu container). If you aren't familiar with Autoconf, I can push an additional change for that purpose.

src/runtime/pmix_progress_threads.c
```diff
@@ -19,7 +19,12 @@
 #ifdef HAVE_UNISTD_H
 #    include <unistd.h>
 #endif
+
 #include <pthread.h>
+#if (defined(__FreeBSD__) || defined(__OpenBSD__)) && defined(HAVE_PTHREAD_SETAFFINITY_NP)
```
rhc54, line 24: I'm a little leery of these OS-specific checks in the code - we generally do them in the Autoconf configure code. For example, in this case we would just check for the existence of the `pthread_np.h` header. I'm not sure why we need to also check for the `HAVE_PTHREAD_SETAFFINITY_NP` definition - is there some reason why we would need that?

Is there an additional required library? It seems a little odd that a function would be defined in a different header, but the actual function would reside in the same `pthread` library.

rayegun:
If you would do the autoconf that would be great! I can wade through it but I've never used autoconf before. I'll try to get to that on Monday if you haven't by then.

src/runtime/pmix_progress_threads.c
rhc54: > About the checking for `HAVE_PTHREAD_SETAFFINITY_NP` that is used to guard the use of the `_np` extensions further below. My assumption would be that we use the same guard when including the header

No, that shouldn't be required. If the header exists, then we include it. The code is simply protected against the case where it doesn't since that variable is defined in the `pthread_np.h` header.

rhc54:
Please check https://github.com/openpmix/openpmix/pull/3615 to see if it works in your scenario - it's the Autoconf way of dealing with the problem. If it does, and with your permission, I'll add you as co-committer to it.

rhc54:
I checked the other PR in a FreeBSD VM and it worked fine. So we'll go with that solution. Please do let me know if it's okay to add your name to it!

PR #3635, opened 2025-06-25
Update to track changes in Standard

rhc54:
Fix a few typos, but mostly complete the change from macros to functions, as per latest changes to Standard


naughtont3:
> Objecting to every call down into the library because of "...without gil". Any idea what is going on?

Looks like the thread shifting into/out  Python interpreter context needs the "Global Interpreter Lock" (GIL). I can poke at it a bit.

PR #3638, opened 2025-06-27
Change default show-load-errors to "none"

rhc54:
We treat a NULL string as boolean "true", which means that the show-load-errors param was defaulting to "all". Change the default configuration option --with-show-load-errors to be "none" so we avoid that behavior.

Refs https://github.com/open-mpi/ompi/issues/13314

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -791,8 +791,8 @@ AC_DEFUN([PMIX_SETUP_CORE],[
                                 mca_base_component_show_load_errors (but
                                 can be overridden at run time by the usual
                                 MCA-variable-setting mechansism).
-                                (default: "all")])])
-
+                                (default: "none")])],
+                                [], [with_show_load_errors=no])
     if test -z "$with_show_load_errors" || \
        test "$with_show_load_errors" = "yes"; then
         with_show_load_errors=all


rhc54:
@hppritcha @jsquyres I haven't seen this before - is this something new? Don't believe the change here could impact finding the MPI library.
```shell
======================================================================
ERROR: testGetLib (test_package.TestMPIABI.testGetLib)
----------------------------------------------------------------------
  File "/home/runner/work/openpmix/openpmix/test/test_package.py", line 119, in testGetLib
    result = mpiabi._get_mpiabi()
  File "/home/runner/work/openpmix/openpmix/build/lib.linux-x86_64-cpython-313/mpi4py/_mpiabi.py", line 258, in _get_mpiabi
    mpiabi = _get_mpiabi_from_libmpi(libmpi)
  File "/home/runner/work/openpmix/openpmix/build/lib.linux-x86_64-cpython-313/mpi4py/_mpiabi.py", line 217, in _get_mpiabi_from_libmpi
    lib = _dlopen_libmpi(libmpi)
  File "/home/runner/work/openpmix/openpmix/build/lib.linux-x86_64-cpython-313/mpi4py/_mpiabi.py", line 210, in _dlopen_libmpi
    raise RuntimeError("\n".join(errors))
RuntimeError: cannot load MPI library
/home/runner/.local/lib: cannot open shared object file: No such file or directory
/opt/hostedtoolcache/Python/3.13.5/x64/lib/libmpi.so: cannot open shared object file: No such file or directory
/opt/hostedtoolcache/Python/3.13.5/x64/lib/libmpi.so.12: cannot open shared object file: No such file or directory
/opt/hostedtoolcache/Python/3.13.5/x64/lib/libmpi.so.40: cannot open shared object file: No such file or directory
libmpi.so: cannot open shared object file: No such file or directory
libmpi.so.12: cannot open shared object file: No such file or directory
libmpi.so.40: cannot open shared object file: No such file or directory
```

rhc54:
Appears to be a new test that was added to mpi4py, but it has a bug in it.

PR #3651, opened 2025-08-01
Don't fail when PMIX_IOF_OUTPUT_TO_FILE directory exists

Matthew-Whitlock:
Currently, setting an output file successfully creates the directory on one process, but the rest then detect an already existing directory. PMIX reports the error and continues, but the output from those processes is lost (even to the standard shared output).


rhc54:
Ouch - thanks for catching that and providing a fix!

Issue #3671, opened 2025-08-27
mca base: MAXPATHLEN not defined on hurd


drew-parsons:
One of the goals of the Debian linux distribution is to build packages for a wide range of cpu architectures, and also, where possible, different operating systems. This includes the [GNU hurd](https://www.gnu.org/software/hurd/) operating system.

A large number of linux packages are able to build successfully on hurd.  But a common point of failure happens when `MAXPATHLEN` is used. In linux it is defined in `<sys/param.h>` provided by libc, e.g. at /usr/include/x86_64-linux-gnu/sys/param.h.  But on hurd it does not get defined.

openpmix uses `MAXPATHLEN` at https://github.com/openpmix/openpmix/blob/47b6634be858ca6e5915eed5fb2c50ceb0b4ece0/src/mca/base/pmix_mca_base_var.c#L303

The problem with hurd can be seen in the debian build of openmpi using its internal copy of pmix, build log at
https://buildd.debian.org/status/fetch.php?pkg=openmpi&arch=hurd-amd64&ver=5.0.8-3&stamp=1756137824&raw=0
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../../../../../../3rd-party/openpmix/src/mca/base -I../../../src/include -I../../../include -iquote/build/reproducible-path/openmpi-5.0.8/debian/build-gfortran/3rd-party/openpmix -iquote/build/reproducible-path/openmpi-5.0.8/3rd-party/openpmix -iquote/build/reproducible-path/openmpi-5.0.8/3rd-party/openpmix/src -iquote/build/reproducible-path/openmpi-5.0.8/debian/build-gfortran/3rd-party/openpmix/include -I/build/reproducible-path/openmpi-5.0.8/3rd-party/openpmix/include -Wdate-time -D_FORTIFY_SOURCE=2 -DNDEBUG -g -O2 -ffile-prefix-map=/build/reproducible-path/openmpi-5.0.8=. -fstack-protector-strong -fcf-protection -O3 -finline-functions -mcx16 -pthread -Werror=implicit-function-declaration -Wformat -Werror=format-security -c ../../../../../../../3rd-party/openpmix/src/mca/base/pmix_mca_base_var.c  -fPIC -DPIC -o .libs/pmix_mca_base_var.o
../../../../../../../3rd-party/openpmix/src/mca/base/pmix_mca_base_var.c: In function 'pmix_mca_base_var_cache_files':
[...]

PR #3672, opened 2025-08-27
define default MAXPATHLEN if not defined by system

drew-parsons:
A definition of `MAXPATHLEN` is not a mandatory part of the POSIX standard, and is not defined for [GNU hurd](https://www.gnu.org/software/hurd/) operating system, where `<sys/param.h>` is not available.

cf. https://www.gnu.org/software/hurd/community/gsoc/project_ideas/maxpath.html
for further context.

But it is difficult to use `getcwd` without `MAXPATHLEN`, so set a default value if it has not been defined in `<sys/param.h>`.

This allows openpmix to build for hurd.

The patch is taken from debian,
https://salsa.debian.org/science-team/pmix/-/blob/e29db6a5b8e523a3b5e0d80922d5958a0d8e0d07/debian/patches/hurd-fix.patch
and has been successfully tested in debian hurd builds, e.g.
https://buildd.debian.org/status/fetch.php?pkg=pmix&arch=hurd-amd64&ver=6.0.0-3&stamp=1756199722&raw=0

Closes: #3671

### DIFF
--- a/src/common/pmix_pfexec.c
+++ b/src/common/pmix_pfexec.c
@@ -88,6 +88,10 @@
 #include "src/common/pmix_pfexec.h"
 #include "src/server/pmix_server_ops.h"
 
+#ifndef MAXPATHLEN /* Hurd */
+#define MAXPATHLEN 1024
+#endif
+
 static pmix_status_t setup_prefork(pmix_pfexec_child_t *child);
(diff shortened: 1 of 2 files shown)


rhc54:
I'll take a gander at the CI error - nothing to do with your PR, so something else has snuck through the backdoor.

PR #3678, opened 2025-09-04
Fix error code on blocking PMIx_Notify_event calls

Matthew-Whitlock:
They currently return the event status being notified, not the status of the event notification function call.


rhc54:
Nice catch!

PR #3686, opened 2025-10-03
Extend listener thread port specification to support ranges

rhc54:
Particularly when the host system is closing ports for security purposes, we can face a situation where we know the range of ports available for the server's listener thread, but not necessarily the specific port to be used. So extend the port specification interface to allow passing of a comma-delimited set of port ranges, and then try those until we find the first one that works.

Note that the MCA param method for passing the port did not require any change as that is always just a string that we parse to extract the port number(s). However, we have extended the value for the PMIX_TCP_IPV4_PORT and PMIX_TCP_IPV6_PORT attributes so that:

* if the value is of type PMIX_INT, then treat the value as a distinct port number to be used

* if the value is of type PMIX_STRING, then treat it as a comma-delimited list of port ranges (e.g., "2,13-25,32")

In other words, those two attributes can now contain either a PMIX_INT or a PMIX_STRING value.

### DIFF
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -1358,13 +1358,16 @@ AC_ARG_ENABLE([ipv6],
 if test "$enable_ipv6" = "yes"; then
     AC_MSG_RESULT([yes])
     pmix_want_ipv6=1
+    pmix_ipv6_enabled="enabled"
 else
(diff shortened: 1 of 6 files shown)


rhc54:
@lipixx Per your request

PR #3711, opened 2025-10-28
Flush namespace sinks' residuals before destroying namespace

Matthew-Whitlock:
I was getting intermittent hangs during PMIx_server_finalize when running MPI jobs with the `--output file=<filename>` option. The problem is that sinks get destroyed when the namespace is destroyed, but any residuals on those sinks keep a pointer to the `pmix_iof_write_event_t` object in those sinks. Flushing the residuals during finalize causes PMIx to add invalid events to the event base, which then hangs when being closed by libevent.

### DIFF
--- a/src/common/pmix_iof.c
+++ b/src/common/pmix_iof.c
@@ -1728,7 +1728,7 @@ void pmix_iof_flush_residuals(void)
     }
 }
 
-void pmix_iof_flush_sink(pmix_iof_sink_t *sink)
+static void flush_sink_residuals(pmix_iof_sink_t *sink)
 {
     pmix_status_t rc = PMIX_SUCCESS;
     pmix_iof_residual_t *res, *next;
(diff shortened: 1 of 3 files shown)


Matthew-Whitlock:
Yeah, it's about the timing. The namespace gets destroyed in `PMIx_server_deregister_nspace` before `PMIx_server_finalize` is called to invoke `pmix_iof_flush_residuals`. When the namespace is destroyed, if any of those residuals have pointers to the destroyed namespace's sinks (which are destroyed by the namespace's destructor), the residual adds an invalid event pointer to the event base (`pmix_iof_sink_t.wev.ev`) when it is later written/flushed.

The changed function signature lets us flush only the specific sinks' residuals, so that any continuing namespaces' outputs don't risk getting mangled by printing unfinished lines.

Matthew-Whitlock:
The safest approach would probably be to put these lines:
```
pmix_iof_flush_sink(sink);
pmix_iof_static_dump_output(sink);
```
Inside of the `pmix_iof_sink_t` class destructor. That way you don't have to worry about making sure you always clean up the residuals before destroying a sink, since the sink handles it itself.

I just didn't want to go verify that it would be safe to access the residuals list and do IO at every existing location where sinks can be destroyed.

Matthew-Whitlock:
> So you are flushing the residuals relating to a specific nspace when that nspace is deregistered. At server finalize, we flush all remaining residuals from the global list. This implies that some namespaces were not deregistered prior to finalizing the server. Is that what you are seeing? Or is the residual list empty?

I just realized what you were asking here. I think the residual list might not be empty even if all namespaces are deregistered, because there could be residuals for the `pmix_client_globals.iof_stdout` and `pmix_client_globals.iof_stderr` sinks.

rhc54:
I went ahead and added it here. Had to add a little protection because clients don't have residuals (only servers do), but this better encapsulates the cleanup.

PR #3716, opened 2025-11-05
Add new PMIX_GROUP_FINAL_MEMBERSHIP_ORDER attribute

rhc54:
The PMIx_Group_construct API itself is not order sensitive in the provided proc array - i.e., we ignore that order when executing the operation. This allows each caller to provide the proc array in arbitrary order - which is advantageous for most libraries.

However, some users may need us to return a specific order of the procs in the final membership. Allow them to specify the order in a new attribute, This can be specified by individual processes or by namespace (with a wildcard rank). If multiple participants provide this attribute, then the values must match - i.e., the desired final membership order must be identical.

### DIFF
--- a/examples/group.c
+++ b/examples/group.c
@@ -83,6 +83,7 @@ int main(int argc, char **argv)
     pmix_query_t query;
     int i;
     bool testquery = false;
+    pmix_data_array_t darray;
     EXAMPLES_HIDE_UNUSED_PARAMS(argc, argv);
 
     for (i=1; i < argc; i++) {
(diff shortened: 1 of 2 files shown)


sonjahapp:
Hi Ralph!

> Occurred to me that if you specify an ordering using an nspace/wildcard combination, then you might want any specific ranks to appear in rank order - true?

I would say yes. For any other rank order, wildcards/nspace cannot be used but rather the explicit order must be provided.

A few (perhaps stupid) questions: Does the process order in the membership array have any influence on how OpenPMIx or PRRTE treat the PMIx Group? Are there any benefits or downsides for PMIx or the host server implementation that could originate from a specific proc order in the membership array? Or is this feature "just" for convenience on PMIx client side - so that the clients would not have to sort the processes by themselves?

I would assume that for MPI this feature is useful for optimization, but not really strictly required, or is it? As long as the group membership array is identical in all procs of the group, MPI should be able to deal with it and sort the procs in whatever order is required.

bosilca:
MPI connect/accept is order agnostic. Moreover, all MPI dynamic processes functionality is order agnostic., as none of it  create an intra communicator (where all processes are part of the same group and the order is important), but an intracomm (where each side of the connect/accept is on their own group with their own original ranking). The only MPI function that need to know the ranking is `MPI_Intercomm_merge` which collapses an intercomm into an intracomm, but at that point we already have the intracommunicator to do MPI-level collectives.

PR #3758, opened 2026-01-05
 Fix systematic race conditions in cleanup sequences

yinliaws:
## Problem
PMIx contains multiple race conditions in cleanup code causing random segfaults during finalization in multi-threaded environments. Issues manifest as:
- Double-free crashes in `keyindex_destruct()`
- Hash table corruption with parameter corruption (`key_size=4195616`)
- List manipulation races in `pmix_release_registered_attrs()`

## Root Cause
Concurrent access to shared data structures during cleanup without proper synchronization, leading to:
- Memory corruption during `pmix_bfrops_base_tma_argv_free()`
- Hash table structure corruption during multi-threaded access
- Race conditions in pointer management during cleanup

## Solution
Applied comprehensive race condition fixes:

### 1. Enhanced keyindex_destruct Protection
- Set pointers to NULL **before** freeing to prevent double-free
- Add mutex protection around cleanup operations
- Validate pointers before cleanup operations

### 2. Complete Hash Table Protection
- Add mutex protection to all 11 hash table access functions
- Validate hash table structure before access
- Prevent parameter corruption during concurrent operations

### 3. List Destruction Protection
- Mark lists as empty before destruction to prevent concurrent access
- Add mutex protection around list operations

## Files Modified
- `src/include/pmix_globals.c` - Enhanced keyindex_destruct protection
- `src/class/pmix_hash_table.c` - Complete hash table race condition protection

## Testing
Systematic debugging approach where each fix eliminated one crash and revealed the next, until all race conditions were resolved. Tested under high concurrency loads.

## Impact
Eliminates random segfaults during PMIx finalization, making PMIx reliable in multi-threaded HPC environments.

### DIFF
--- a/src/class/pmix_hash_table.c
+++ b/src/class/pmix_hash_table.c
@@ -134,9 +134,28 @@ pmix_hash_table_init(pmix_hash_table_t *ht, size_t table_size)
 int /* PMIX_ return code */
 pmix_hash_table_remove_all(pmix_hash_table_t *ht)
 {
-    size_t ii;
-    for (ii = 0; ii < ht->ht_capacity; ii += 1) {
-        pmix_hash_element_t *elt = &ht->ht_table[ii];
+    size_t ii, capacity;
+    pmix_hash_element_t *table;
+    static pthread_mutex_t hash_mutex = PTHREAD_MUTEX_INITIALIZER;
+
+    // RACE CONDITION FIX: Validate inputs first
+    if (NULL == ht) {
+        return PMIX_ERR_BAD_PARAM;
+    }
+
+    pthread_mutex_lock(&hash_mutex);
+
+    // Additional validation after acquiring lock
+    if (NULL == ht->ht_table || ht->ht_capacity == 0) {
+        pthread_mutex_unlock(&hash_mutex);
+        return PMIX_SUCCESS; // Already empty
+    }
+
+    capacity = ht->ht_capacity;
+    table = ht->ht_table;
+
+    for (ii = 0; ii < capacity; ii += 1) {
+        pmix_hash_element_t *elt = &table[ii];
         if (elt->valid && ht->ht_type_methods && ht->ht_type_methods->elt_destructor) {
             ht->ht_type_methods->elt_destructor(elt);
         }
@@ -147,6 +166,8 @@ pmix_hash_table_remove_all(pmix_hash_table_t *ht)
     /* the tests reuse the hash table for different types after removing all */
     /* so we should allow that by forgetting what type it used to be */
     ht->ht_type_methods = NULL;
+    
+    pthread_mutex_unlock(&hash_mutex);
     return PMIX_SUCCESS;
 }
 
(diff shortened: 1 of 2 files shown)


rhc54:
Few comments:

1. rather than using pthread directly, we prefer to use the PMIx abstractions in the src/threads directory. Would you mind converting to those?
2. I see you only dealt with one hash table deconstruction. I'd have to check, but I believe it is used elsewhere as well - e.g., in the actual storage of "put" data. While your specific tests may not have triggered problems, you are changing the hash table code in a way that makes specific assumptions as to how those functions will be called. I believe you'll need to check other usages in the code base to ensure that they _all_ now conform to these assumptions. Otherwise, it would see that we are bound to hit new segfaults when the assumptions are invalid.
3. it bothers me that the code calling the hash table destruct function now has requirements that seem to violate abstraction - e.g., setting the hash table name to NULL and releasing it before calling destruct. That should not be necessary and puts a burden on developers they are almost guaranteed to violate - probably will never realize it exists. Can we not maintain abstraction? I get your using these changes as "markers" for prior destruction - but maybe we need another way to approach it. For example, it is a class and therefore tracks the reference count of each instance - why are we running destructors more than once? Is the real root cause here that the pmix_globals.c code manually deconstructs the table instead of using the class destructor? Perhaps that would be a cleaner fix?
4. you deleted several comments (copies of each other in various locations) that are actually valid. The hash tables are used by the shmem2 component, and the comments are warnings requiring future attention. Probably best to leave them - I'll ask the author to confirm

@samuelkgutierrez Would you mind taking a peek at this? I'd especially appreciate your thoughts on the deleted comments, but also your eyes on the general proposed changes.

samuelkgutierrez:
I would advocate for leaving the comments alone unless there is a compelling reason to remove them. Since the code related to the comments was left untouched, I would personally leave those comments alone.

Overall the changes seem reasonable to me, @rhc54.

rhc54:
The problem here is the lack of a thread shift somewhere - everything in PMIx is required to be funneled through its own internal progress thread that can only execute one operation at a time (basically, it serializes everything). Is this with the head of the respective PMIx/PRRTE master branches? Or some OMPI-distributed thing?

Also, I don't see what this has to do with cleanup, which is what you propose in the PR? Or is that some other symptom of the same scenario?

yinliaws:
Hi @rhc54, Thank you for confirming this and working on the proper fix! This explains why we were seeing the race condition. The missing thread shift in PMIx_IOF_push was allowing direct calls from the main thread while the progress thread was also processing PMIx operations. I appreciate you taking the time to implement the architectural fix. Once the PMIx_IOF_push thread shifting is in place, it should eliminate the race condition we've been debugging. Should I wait for your PMIx fix and close this PR, or would you like me to proceed with a different approach in the meantime? Thanks, Yin

rhc54:
Just for clarity: these are all generated when mpirun is killed by signal?

rhc54:
Confirmed that this appears to be the due to the issues described in https://github.com/open-mpi/ompi/issues/13044, so closing this as unneeded until something else comes up.

PR #3799, opened 2026-02-04
Make thread start/stop marker consistent

rhc54:
Set the thread stopped marker in a common location and ensure that all the thread start/stop functions properly set it.


rhc54:
@hppritcha Please take a look to ensure things are still correct. A review indicated that some of the thread pause/resume/etc functions were not setting the bool, so I am attempting to ensure it always gets properly set by putting the set/unset calls in a common code path.

PR #3832, opened 2026-03-14
Fix segfault in PMIx_Finalize due to NULL cls_destruct_array

xSetech:
This fixes a crash in PMIx_Finalize when MPI programs are run under mpirun/prterun outside of singleton mode by adding NULL guards for cls_construct_array and cls_destruct_array in pmix_obj_run_constructors and pmix_obj_run_destructors..

pmix_object_t_class is statically preinitialized with cls_initialized matching pmix_class_init_epoch, causing pmix_class_initialize() to skip it. However, cls_construct_array and cls_destruct_array remain NULL since they are only populated by pmix_class_initialize().

Objects statically initialized via PMIX_OBJ_STATIC_INIT(pmix_object_t) (e.g., list sentinels within PMIX_LIST_STATIC_INIT) retain obj_class pointing to pmix_object_t_class. When PMIX_DESTRUCT is called on such an object, pmix_obj_run_destructors dereferences the NULL cls_destruct_array, causing a segfault.

### DIFF
--- a/src/class/pmix_object.h
+++ b/src/class/pmix_object.h
@@ -655,6 +655,7 @@ static inline void pmix_obj_run_constructors(pmix_object_t *object)
     assert(NULL != object->obj_class);
 
     cls_construct = object->obj_class->cls_construct_array;
+    if (NULL == cls_construct) return;
     while (NULL != *cls_construct) {
         (*cls_construct)(object);
         cls_construct++;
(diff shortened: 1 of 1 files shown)


xSetech:
MPI_Init in that example succeeds; it's actually a requirement to trigger the bug. I've left it out of the snippet for brevity. The MPI program runs, it just crashes in the finalizer due to the dereference.

I agree the object ought to be constructed in a real program. Even still, so far as I can tell, neither on Linux nor macOS does one get constructed correctly or at all. It does however look like gcc lays out the code differently than clang, and that may be obscuring the null pointer deference in finalization on Linux.

rhc54:
Just FWIW: I cannot reproduce the problem. The library correctly detects singleton operation and sets up for it, then finalizes cleanly. You might try the head of PMIx master or the latest PMIx v6.1.0 release - could be that some earlier issue has been fixed.

Issue #3590, opened 2025-05-09
PMIx 5.0.7 PMIx_server_setup_local_support returning status -50


dgloe-hpe:
### What version of the PMIx Reference Library are you using?
5.0.7

### Describe how PMIx was installed
An RPM built from source

### Please describe the system on which you are running

* Operating system/version: SLES 15 SP6
* Computer hardware: Intel Xeon 6148
* Network type: Slingshot

-----------------------------

## Details of the problem

With PMIx 5.0.7, when we call PMIx_server_setup_local_support, the callback function is getting status -50 (PMIX_ERR_UNPACK_READ_PAST_END_OF_BUFFER). This was working fine in PMIx 5.0.6.

rhc54:
What info keys are you passing to `PMIx_server_setup_application`? I checked and there is no code change in the relevant code areas between the two releases.

rhc54:
Mind you, I can see _how_ what you report can happen - I just cannot explain why it wouldn't be happening in 5.0.6 and then show up in 5.0.7. I posted a patch (https://github.com/openpmix/openpmix/pull/3592) that should remove the possibility. Will have to scratch our heads and try to figure out why the change.

I should note that I cannot reproduce that observation with PRRTE, even though (looking at the code) I _should_ have seen that error. 🤷‍♂

rhc54:
BTW: one way you should be able to turn it off is to add `PMIX_MCA_pnet=^opa` to the environment. If you want to debug it further, add `PMIX_MCA_pnet_base_verbose=100`.

rhc54:
Okay, I tracked down why I cannot reproduce this behavior. The OmniPath component (opa) checks to see if any OmniPath devices are present - it looks for the PCI device class (0x8086) and device ID (0x208) assigned to that NIC. If it doesn't find one in the topology, then it disables itself, thereby preventing you from executing the code that might have returned that error.

Is it possible that you ran the v5.0.7 code on a node that happened to have such a device on it that matches those IDs?

Checking the latest PCI ID data, it looks like this is all stale anyway. Intel (class 0x8086) now has a different ID for their OmniPath 100 series (0xabc0), and Cornelis (class 0x434e) doesn't have any PCI device IDs assigned to it. So I'm just going to delete this component.

rhc54:
Done in the official PMIx v5.0.8 release (https://github.com/openpmix/openpmix/releases/tag/v5.0.8)

dgloe-hpe:
[...]

PR #3852, opened 2026-04-22
Cornelis restore pnet opa

BrendanCunningham:
Three commits in this PR but I'll squash them down to one for the final PR review. Just want to get feedback to start.

Did not update any copyrights even though Cornelis Networks is now the custodian of OPA100 and PSM2.


BrendanCunningham:
Pushed just the revert.

TODO:
- [ ] Commit message has cutline in it. Want review from Ralph on wording/wordsmithing.
- [ ] Double-check that this doesn't reintroduce #3590 .

PR #3867, opened 2026-05-18
pmdl/ompi: use the right type to enumerate myenvars

ggouaillardet:
myenvars is pmix_list_t populated with pmix_mca_base_var_file_value_t objects, so use the right type when enumerating is it setup_fork()

Refs open-mpi/ompi#13303


rhc54:
Thanks!

PR #3908, opened 2026-06-19
Add allocation inheritance datatype and attribute

rhc54:
Introduce a new `pmix_alloc_inheritance_t` (`PMIX_ALLOC_INHERIT`) data
type for expressing what becomes of an allocation when its owning
namespace terminates. Three values are defined:

- `PMIX_ALLOC_INHERIT_NONE` — release the allocation upon termination of
  the owning namespace
- `PMIX_ALLOC_INHERIT_CHILD` — retain until all child namespaces
  (including derived children) terminate
- `PMIX_ALLOC_INHERIT_DEFAULT` — the allocation becomes unreserved  upon
termination of the owning namespace
- `PMIX_ALLOC_INHERIT_CHILD_DEFAULT` —  the allocation becomes unreserved
upon termination of the last derived child namespace

The value is carried in the `pmix_value_t` union as `inheritance` and is
selected via the new `PMIX_ALLOC_INHERITANCE` attribute.

Full bfrops support is added in the base functions (pack, unpack,
std-copy sizing, compare, TMA copy, value load/unload, and print), and
the type is registered in the v61 component — matching how other newly
introduced types are wired into only the most recent wire-format
version. A `PMIx_Alloc_inheritance_string()` converter returns the
trailing word of each value name (`NONE`, `CHILD`, `DEFAULT`) and is used
by the print routine. The dictionary generator learns to map the
`(pmix_alloc_inheritance_t)` annotation to `PMIX_ALLOC_INHERIT` so the
attribute harvests cleanly.

A new `make check` unit test, `bfrops_alloc_inherit`, exercises
pack/unpack (single, multiple, and via `pmix_value_t`), value transfer,
compare, print, and the string converter.

Also includes related allocation attribute updates carried on this
branch: a new `PMIX_SPAWN_TARGET` attribute for mapping spawned
applications onto specific allocations, and a reworked `PMIX_ALLOC_SHARE`
definition that replaces the former `PMIX_ALLOC_RESERVED`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -227,6 +227,7 @@ test/sshot/testcoord
 test/unit/compress
 test/unit/preg
 test/unit/bfrops_regex2
+test/unit/bfrops_alloc_inherit
 test/unit/gds_fallback
 test/unit/pmix_log
 test/unit/run_gds_fallback.pl
--- a/contrib/construct_dictionary.py
+++ b/contrib/construct_dictionary.py
@@ -165,6 +165,8 @@ def harvest_constants(options, path, constants):
                             datatype = "PMIX_UINT16"
                         elif tokens[3] == "(double)":
                             datatype = "PMIX_DOUBLE"
+                        elif tokens[3] == "(pmix_alloc_inheritance_t)":
+                            datatype = "PMIX_ALLOC_INHERIT"
                         else:
                             print("UNKNOWN TOKEN: {tok}".format(tok=tokens[3]))
                             return 1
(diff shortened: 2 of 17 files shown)


rhc54:
@Petter-Programs I'm working on the node assignment to sessions - these definitions control inheritance of the reserved nodes. See what you think.

Petter-Programs:
> @Petter-Programs I'm working on the node assignment to sessions - these definitions control inheritance of the reserved nodes. See what you think.

Looks good to me!

PR #3932, opened 2026-07-05
ptl/base: don't prune a gracefully-finalized proc from collectives

### LINKED ISSUE #3931: PMIx_Fence deadlocks under repeated PMIx_Init/PMIx_Finalize: lost_connection() prunes a gracefully-finalized, reconnecting proc
## Summary

A `PMIx_Fence` over a namespace **deadlocks** when clients repeatedly
`PMIx_Init` → `PMIx_Fence` → `PMIx_Finalize` and then `PMIx_Init` again
(the MPI Sessions usage model). Under timing skew, the processes drift by
whole init/finalize iterations, one process completes and exits, and the
remaining processes hang forever in a fence that the departed process was
still expected to join.

The root cause is in the server-side collective bookkeeping: when a client
disconnects as the result of an **orderly `PMIx_Finalize`**, the server's
`lost_connection()` handler prunes that client from every in-progress
collective it appears in — decrementing `nlocal` and marking the collective
`PMIX_ERR_PARTIAL_SUCCESS` — exactly as it would for a **faulted** client.
But a finalized client in this usage model immediately reconnects
(`PMIx_Init`) and is still required to participate. Pruning it lets a
*next-cycle* fence (already opened by a faster peer) complete short; the
reconnecting client then opens a fresh tracker the others have already
passed, and the namespace's fence sequence permanently desynchronizes.

This surfaced as an intermittent macOS hang in Open MPI's MPI Sessions
support (repeated `MPI_Session_init`/`MPI_Session_finalize`): see
open-mpi/ompi#14125. It reproduces with **no MPI at all** — pure PMIx +
the `test/simple/simptest` reference server.

## Environment

- OpenPMIx `master` @ `6f3a90b8` (reports `7.0.0a1`). Also reproduced on
  `v6.1` @ `v6.1.1rc2` (`cb89026`), the branch Open MPI `main`/`v6.0.x`
  embed.
- macOS 26.5 (arm64, 12 cores); also observed on the GitHub Actions
  `macos-15` runner. Timing-dependent, so heavy CPU load / oversubscription
  makes it far more likely.

## Reproducer (pure PMIx, no MPI)

Client — loops `PMIx_Init` / two wildcard `PMIx_Fence`s / `PMIx_Finalize`,
with a small rank-dependent `usleep` to induce skew:

```c
#include <pmix.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
int main(int argc, char **argv) {
    int iters = (argc > 1) ? atoi(argv[1]) : 200;
    pmix_proc_t myproc, wild;
    unsigned int seed = (unsigned int) getpid();
    FILE *fp = NULL;
    for (int i = 0; i < iters; i++) {
        if (PMIX_SUCCESS != PMIx_Init(&myproc, NULL, 0)) return 1;
        if (NULL == fp) {
            char p[128];
            snprintf(p, sizeof(p), "/tmp/pmt.%s.%d", myproc.nspace, myproc.rank);
            fp = fopen(p, "w"); setvbuf(fp, NULL, _IONBF, 0);
        }
        fprintf(fp, "i%d start\n", i);
        PMIX_PROC_CONSTRUCT(&wild);
        PMIX_LOAD_PROCID(&wild, myproc.nspace, PMIX_RANK_WILDCARD);
        bool t = true, f = false;
        pmix_info_t info;
        PMIX_INFO_LOAD(&info, PMIX_COLLECT_DATA, &t, PMIX_BOOL);
        PMIx_Fence(&wild, 1, &info, 1);        /* modex fence */
        pmix_info_t info2;
        PMIX_INFO_LOAD(&info2, PMIX_COLLECT_DATA, &f, PMIX_BOOL);
        PMIx_Fence(&wild, 1, &info2, 1);       /* barrier fence */
        PMIx_Finalize(NULL, 0);
        usleep((myproc.rank * 137 + (rand_r(&seed) % 400)));
    }
    if (fp) { fprintf(fp, "ALLDONE %d\n", iters); fclose(fp); }
    return 0;
}
```

Run 4 clients under the reference server while the machine is loaded:

```sh
# in one shell: create some load
for i in $(seq 24); do yes > /dev/null & done

# then, repeatedly:
cd test/simple
./simptest -n 4 -e ./client 200
```

Within a run or two it hangs. Inspecting `/tmp/pmt.*` shows the drift, e.g.:

```
pmt.foobar.0: i199 start      <- still stuck in iteration 199's fence
pmt.foobar.1: ALLDONE 200     <- finished all 200 and exited
pmt.foobar.2: i199 start
pmt.foobar.3: ALLDONE 200
```

Two of four completed all 200 iterations while the other two are blocked
in a fence at iteration 199. A client also occasionally sees
`PMIx_Fence` return `-52` = `PMIX_ERR_PARTIAL_SUCCESS`.

**It is specifically the `Init`/`Finalize` cycling that breaks it.** With a
single `PMIx_Init` and a loop of repeated fences (no re-init), the same
harness is 100% reliable across dozens of runs.

## Root cause

`src/mca/ptl/base/ptl_base_sendrecv.c`, `lost_connection()`:

```c
/* it should - adjust the count */
--trk->nlocal;
if (0 < trk->nlocal) {
    rc = PMIX_ERR_PARTIAL_SUCCESS;
} else {
    rc = PMIX_ERR_LOST_CONNECTION;
}
```

For a **wildcard** fence, the tracker's participant list (`trk->pcs`)
contains every proc in the namespace, so a proc that is momentarily
disconnected between its `PMIx_Finalize` (cycle N-1) and its next
`PMIx_Init` (cycle N) matches, by name, a fence tracker that a *faster*
peer already created for cycle N. `lost_connection()` decrements that
tracker's `nlocal` and completes it short — even though the finalized proc
is going to reconnect and is still required to participate.

Notably, the surrounding code in the same function already special-cases
`peer->finalized` — it declines to decrement `nptr->nlocalprocs` and
declines to emit a `PMIX_ERR_LOST_CONNECTION` event for an orderly
finalize — but the collective-pruning loop between them does **not** make
that distinction. The pruning that produces this behavior was introduced in
`389bcf10` ("Provide partial_success error when collectives not complete",
2021-09), i.e., after the earlier re-init work.

## Related prior issues

This is the same family of problem discussed in these (all closed) issues;
the fence-after-reinit hang was the explicit subject of the first one:

- #2202 — "PMIx_fence doesn't work following second PMIx initialization"
  (the direct match; fixed for the MCA-parameter re-registration aspect by
  #2205 "Enable re-init of clients"). It was noted there that removing the
  fence lets init/finalize cycle fine — consistent with what we see here.
- #1558 — "Problem with multiple PMIx_Init / Finalize call sequence"
- #2377 — "Multiple inits and finalizes segfaults"
- #2096 — "More than one client timeout per PMIx_Fence causing error"
- #89 — "possible PMIx server race: PMIx_server_register_client vs fence"

A fix PR follows. It teaches `lost_connection()` to leave a
gracefully-finalized (and not-yet-joined) proc in the collective so its
reconnected incarnation can complete it, while preserving the existing
fault-tolerance behavior for a genuinely faulted client.

jsquyres:
Fixes #3931: a `PMIx_Fence` over a namespace deadlocks under repeated
`PMIx_Init`/`PMIx_Finalize` (the MPI Sessions usage model) because
`lost_connection()` prunes a gracefully-finalized -- and immediately
reconnecting -- proc from an in-progress collective, completing it short
(`PMIX_ERR_PARTIAL_SUCCESS`) and desynchronizing the namespace's fence
sequence. See #3931 for the full analysis and a pure-PMIx reproducer (no
MPI; `test/simple/simptest` + a small client).

### The fix

`lost_connection()` already special-cases `peer->finalized` for the
`nptr->nlocalprocs` decrement and the `PMIX_ERR_LOST_CONNECTION` event,
but the collective-pruning loop between them did not. This teaches that
loop to leave a gracefully-finalized proc that has not yet joined a
tracker in place, so its reconnected incarnation can still complete the
collective. A faulted (non-finalized) proc -- or a finalized proc that had
actually joined the tracker (a non-blocking fence abandoned via Finalize)
-- is still pruned, so fault tolerance for a genuinely dead client is
preserved.

### Validation

- Pure-PMIx reproducer (`simptest` + client, 4 procs, heavy CPU load):
  - before: hangs within a run or two;
  - after: 50/50 runs × 200 cycles complete cleanly, 0 hangs.
- Fault-tolerance regression check: a client that `_exit()`s mid-fence
  (no finalize) still lets the survivors' fence complete
  (`PMIX_ERR_PARTIAL_SUCCESS`) rather than hang.
- End-to-end in Open MPI (this patch applied to the embedded PMIx): the
  MPI Sessions reproducer from open-mpi/ompi#14125 (repeated
  `MPI_Session_init`/`MPI_Session_finalize` under load) goes from hanging
  on ~run 1 to **30/30 runs × 300 cycles (9,000 cycles/rank) clean**.

### Cherry-pick

If this is the right approach, it should also be cherry-picked to **`v6.1`**
-- the OpenPMIx branch that Open MPI's `main` and `v6.0.x` branches embed
(submodule pinned at `v6.1.1rc2` / `cb89026`). Those are the Open MPI
release lines that ship MPI Sessions, so that is where the fix is needed
downstream. (Open MPI `v5.0.x` embeds `v5.0` and predates the Sessions
model, so it is not affected via this path.)

### Note / open question

The server cannot distinguish "gracefully finalized and will reconnect"
from "gracefully finalized for good"; this patch optimizes for the former
(the Sessions model). The theoretical trade-off: a proc that finalizes
*permanently* while still named in peers' ongoing wildcard fence would now
cause that fence to wait rather than complete short. That is not reachable
in well-formed symmetric usage (a proc shouldn't finalize while still
expected in a collective), and completing a whole-namespace fence after a
member has permanently departed was already returning incomplete data --
but flagging it in case a different mechanism is preferred (e.g., a short
grace period before pruning, or fence-generation tracking).

@rhc54 -- would appreciate your review, particularly on that edge case.

### DIFF
--- a/src/mca/ptl/base/ptl_base_sendrecv.c
+++ b/src/mca/ptl/base/ptl_base_sendrecv.c
@@ -8,6 +8,7 @@
  *                         All rights reserved.
  * Copyright (c) 2016      IBM Corporation.  All rights reserved.
  * Copyright (c) 2021-2026 Nanook Consulting  All rights reserved.
+ * Copyright (c) 2026      Jeffrey M. Squyres.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
@@ -102,6 +103,45 @@ static void lost_connection(pmix_peer_t *peer)
                 }
                 if (!flag) {
                     continue;
+                }
+                /* If this peer disconnected as the result of an orderly
+                 * PMIx_Finalize (rather than a fault), do not prune it from a
+                 * *fence* it has not yet joined. Under the repeated
+                 * PMIx_Init/PMIx_Finalize usage model (e.g., the MPI Sessions
+                 * model), the peer will reconnect via a subsequent PMIx_Init
+                 * and is still required to participate in the namespace's
+                 * fences. Because a wildcard fence tracker lists every proc in
+                 * the namespace, a proc that is momentarily disconnected
+                 * between the PMIx_Finalize of one cycle and the PMIx_Init of
+                 * the next matches, by name, a fence tracker that a faster
+                 * peer already opened for the next cycle; pruning it there
+                 * decrements nlocal and lets that fence complete short
+                 * (PMIX_ERR_PARTIAL_SUCCESS), desynchronizing the fence
+                 * sequence and hanging the peers still expecting it. This is
+                 * consistent with the nlocalprocs and event-notification
+                 * handling below, which already special-case peer->finalized.
+                 *
+                 * Scope this strictly to fence (PMIX_FENCENB_CMD). The other
+                 * collective types handled here - connect, disconnect, and
+                 * group construct - establish persistent relationships and are
+                 * not part of the per-cycle init/finalize pattern, so their
+                 * existing fault-tolerance pruning is left unchanged. A
+                 * blocking PMIx_Fence cannot have been joined by a peer that
+                 * subsequently called PMIx_Finalize (the fence must have
+                 * returned first), but a non-blocking PMIx_Fence_nb can be
+                 * abandoned that way, so we still prune when the peer had
+                 * actually joined the tracker. */
+                if (peer->finalized && PMIX_FENCENB_CMD == trk->type) {
+                    bool joined = false;
+                    PMIX_LIST_FOREACH (rinfo, &trk->local_cbs, pmix_server_caddy_t) {
+                        if (PMIX_CHECK_NAMES(&rinfo->peer->info->pname, &peer->info->pname)) {
+                            joined = true;
+                            break;
+                        }
+                    }
+                    if (!joined) {
+                        continue;
+                    }
                 }
                 /* it should - adjust the count */
                 --trk->nlocal;


rhc54:
Yeah, something doesn't sound right here. You shouldn't be able to mix fence operations the way you are claiming. I'm suspicious that this fix will just open another race condition - the real question may be how to avoid mixing the fences in the first place. Might also be more of a usage issue than a PMIx bug. Have to investigate/ponder for a while.

jsquyres:
Thanks Ralph — fair skepticism, let me give you the exact trace so we can decide if it's a band-aid.

The "mixing" is real and reproduces with **no MPI at all** (pure PMIx client under `test/simple/simptest`, in #3931). Mechanism, with two ranks A (fast) and B (slow) each looping `{Init; Fence; Fence; Finalize}`:

1. A finishes cycle N, `Finalize(N)`, reconnects, `Init(N+1)`, calls `Fence(N+1)` → server opens tracker **T**, `pcs={A,B}` (wildcard), `nlocal=2`.
2. B is still in *its* `Finalize(N)`; its socket closes → `lost_connection(B)`.
3. T lists B by name, so B is pruned from **A's next-cycle fence**: `--nlocal`, `PMIX_ERR_PARTIAL_SUCCESS`, T completes with only A.
4. B reconnects, calls `Fence(N+1)` → T is gone → B opens a fresh tracker and waits for A, who has already moved on → drift, then deadlock.

The root enabler is that trackers are keyed by (participants + type) with no generation, so `lost_connection` can't tell that B's disconnect belongs to the *previous* fence generation, not the one A just opened.

On "is this just masking it": the surrounding code in `lost_connection()` already special-cases `peer->finalized` (it declines to decrement `nlocalprocs`, and declines to emit a `LOST_CONNECTION` event) — i.e., it already treats a gracefully-finalized proc as "expected back." The collective-pruning loop between those two was just inconsistent with that intent; this patch aligns it, and I scoped it strictly to `PMIX_FENCENB_CMD` so connect/disconnect/group construct are untouched.

The one behavior change I do want your eyes on: with the patch, a proc that finalizes **permanently** while still named in peers' ongoing wildcard fence will now make that fence *wait* rather than complete short with `PARTIAL_SUCCESS`. In symmetric SPMD that's unreachable, but if you consider legally-asymmetric finalize valid, that's the corner to worry about — and it might argue for the deeper "fence generation" or "don't mix generations" approach you're pointing at instead.

On "usage issue": agreed there's an OMPI-side angle — OMPI does a full `PMIx_Finalize`+`PMIx_Init` per Session cycle, and ref-counting PMIx to keep it up for the process lifetime would avoid the churn entirely. But since repeated init/finalize is intended (that's #2202 / #2205, which Sessions relies on) and pure PMIx currently *hangs* under it rather than erroring, hardening the server seemed worthwhile too. Happy to go whichever way you prefer — including holding this in favor of a generation-based fix if you think the edge case above is a real exposure.

jsquyres:
Human Jeff responding here: I did ask Claude about whether OMPI was doing it wrong / should be using PMIx_Fence in this way.  After thinking about it a bunch, Claude thought that it seemed like a valid use of PMIx_Init / PMIx_Fence / PMIx_Finalize. 🤷‍♂️ 

Either way is fine with me (change something here in PMIx or change something in OMPI -- or both!).  Lemme know what you think.

rhc54:
Yeah, that's why I wondered if this should be a usage issue. We've discussed this in the past, tried to fix it before - and apparently failed to completely do so. The problem is that sessions in this test is completely driving the client library to tear down, and then tries to bring it back. Over and over again. It isn't clear that the library is capable of supporting that kind of behavior.

Trying to chase down every race condition that creates may be more than we can really do. Your fix _might_ resolve this particular test, but what happens when someone runs a different collective? Or just executes some other PMIx function in the pattern?

So it needs some pondering at the global level to decide if this is even a use-case we want to or can support, and if we do, then what will it take to support it in general.

rhc54:
After giving it a lot of thought, there are two issues here:

1. collective tracking in general is idempotent, which means that there is a hidden vulnerability. Imagine a fence operation where one client participates and then immediately dies before the other participants call fence (assume that the runtime doesn't kill the job). Because we only track participation by count and don't track the individual participants, this client gets double counted - once for participating, and again when its connection is unexpectedly lost. This means that the fence (or any collective operation) gets released early for the other participants. Note that the fix proposed here has no impact on this root problem.
2. the repeated tear-down and rebuild of the client library, and its attendant impact on the persistent server, is uncharted territory. Looking at past history, I had "suggested" rather strongly that someone should verify that use-case actually works before relying on it, but that apparently never happened.

The first issue is going to take some significant changes to resolve. For one thing, we'll need to track the actual participants to each collective operation and not just count them. That's a big change in the code. Not saying it shouldn't be done - it needs to be - but just noting it isn't a small change.

The second issue will take a bit of effort to scope out. I think we'll have to leave a final decision on whether or not to support it until we get a better idea of how much change would be required to make it work.

jsquyres:
> After giving it a lot of thought, there are two issues here:

Gotcha.  I dug into the OMPI side a bit with Claude, and it came up with just not calling PMIx_Finalize until an atexit handler.  That *seems* safe, but I can dig into that a bit more to see if there's consequences there.

For your 2 proposals, I think you're probably better to dig into 1) than me.  Want me to do some testing for 2) and see if I can come up with any obvious (or hopefully, non-obvious) gaps?

rhc54:
Let me resolve the first issue before we get into the second one as the changes for the first are going to heavily impact the second.

hppritcha:
the atexit route would be okay to add at some point.  Probably should add a note in the PMIx_init and PMIx_finalize man pages about possible restart limitations.  The [PMIx standard ](https://pmix.org/uploads/2023/05/pmix-standard-v5.0.pdf)doesn't describe such a limitation.

rhc54:
> the atexit route would be okay to add at some point. Probably should add a note in the PMIx_init and PMIx_finalize man pages about possible restart limitations. The [PMIx standard ](https://pmix.org/uploads/2023/05/pmix-standard-v5.0.pdf)doesn't describe such a limitation.

Probably not worth adding a note until we confirm whether or not it's true. The described bug actually has nothing to do with the finalize situation. Claude fixated on a symptom and not the actual problem. I'm working on the problem now.

jsquyres:
This PR is now moot (and not the real fix); replaced by https://github.com/openpmix/openpmix/pull/3934

PR #3933, opened 2026-07-06
mca: fix deprecated variable source warning

francdoc:
This PR is to fix the deprecated MCA variable warning so PMIx reports the right source and the right new variable name.

When the old variable name is set from the environment, PMIx was reporting it as if it came from an MCA params file. It also printed the old variable name as the new variable name.

The case tested was:

- file: `mca_base_component_show_load_errors = 1`
- environment: `PMIX_MCA_mca_component_show_load_errors=0`

Before this change, PMIx reported the source as an MCA variable file and printed:

- `Deprecated variable: mca_component_show_load_errors`
- `New variable:        mca_component_show_load_errors`

After this change, PMIx reports the source as environment/command line and prints:

- `Deprecated variable: mca_component_show_load_errors`
- `New variable:        mca_base_component_show_load_errors`

I also checked the file case. When `mca_component_show_load_errors` is actually in the MCA params file, PMIx still reports the source file and the right new variable name.

Testing:

- `./autogen.pl`
- `../configure --prefix="$PWD/install"`
- `make -j"$(nproc)"`
- `make install`
- checked the environment case with `PMIX_MCA_mca_component_show_load_errors=0`
- checked the file case with `mca_component_show_load_errors = 0` in the MCA params file

Related to open-mpi/ompi#13958.

### DIFF
--- a/.github/workflows/debugger.yaml
+++ b/.github/workflows/debugger.yaml
@@ -45,7 +45,7 @@ jobs:
          # Tweak PRRTE
          mca_params="$HOME/.prte/mca-params.conf"
          mkdir -p "$(dirname "$mca_params")"
-         echo rmaps_default_mapping_policy = :oversubscribe >> "$mca_params"
+         echo mapby = :oversubscribe >> "$mca_params"
     - name: Run direct test
       run: |
          export PATH=$RUNNER_TEMP/prteinstall/bin:$RUNNER_TEMP/pmixinstall/bin:${PATH}
--- a/.github/workflows/group.yaml
+++ b/.github/workflows/group.yaml
@@ -42,7 +42,7 @@ jobs:
          # Tweak PRRTE
          mca_params="$HOME/.prte/mca-params.conf"
          mkdir -p "$(dirname "$mca_params")"
-         echo rmaps_default_mapping_policy = :oversubscribe >> "$mca_params"
+         echo mapby = :oversubscribe >> "$mca_params"
     - name: Run simple group
       run: |
          export PATH=$RUNNER_TEMP/prteinstall/bin:${PATH}
--- a/.github/workflows/pmix_mpi4py.yaml
+++ b/.github/workflows/pmix_mpi4py.yaml
@@ -110,7 +110,7 @@ jobs:
         echo mpi_show_handle_leaks = true >> "$mca_params"
         mca_params="$HOME/.prte/mca-params.conf"
         mkdir -p "$(dirname "$mca_params")"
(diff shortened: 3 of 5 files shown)


rhc54:
Thanks! It still feels like the "original" vs "var" naming is odd, but this does get us to the right answer with minimal fooling around.

rhc54:
Afraid I have NO idea what mpi4py is doing and why this change would cause an error. Could be an upstream change in their test suite - we've had errors appear from that sector before that are due to bugs upstream.

francdoc:
> Thanks! It still feels like the "original" vs "var" naming is odd, but this does get us to the right answer with minimal fooling around.

Thank you, Ralph. I’m glad this helps.

francdoc:
Ralph, I don’t fully understand the mpi4py error either. So I’m reading through it now and will keep looking tomorrow.

rhc54:
I think the issue is here:
```shell
echo rmaps_default_mapping_policy = :oversubscribe >> "$mca_params"
```
Isn't that a deprecated form of the MCA param? If so, it might cause a deprecation warning to be emitted from your revision, and the mpi4py test isn't expecting it. What you could do is modify the CI yml file to use the updated param name and avoid the warning.

francdoc:
Hi Ralph (@rhc54), great. Thank you for the guidance. I will check it today.

francdoc:
Ok it seems that the variable schizo_proxy also made a python test throw an error, I will fix that as well and relaunch the CI with a new push.

rhc54:
@francdoc I pushed a couple of fixes that should resolve the problem. The issue here is that OMPI is pushing at least one MCA param into the environ before fork/exec'ing prte, and that param is deprecated. So your change will always generate a deprecation warning.

Sadly, the mpi4py tests aren't smart enough to deal with that situation - they simply error out if they see any output. So the only way to get past their CI is to silence deprecation warnings by default. The new param I introduced does just that, basically meaning that your change is going to be invisible unless someone specifically _asks_ to see deprecated param warnings.

I don't see any way around the problem as the author is unlikely to make mpi4py more flexible. Note that you will hit the same problem over in OMPI for the OPAL equivalent, so you might have to duplicate this fix over there.

francdoc:
Hi Ralph (@rhc54). Sorry I missed the update earlier. I saw your follow-up fixes and the merge. I’ll look at the OMPI/OPAL PR (https://github.com/open-mpi/ompi/pull/13958) next based on your observations. Thanks again for the guidance and for taking care of the PMIx side.

PR #3937, opened 2026-07-07
Client init/finalize cycling: state-lifecycle spec, teardown fixes, and regression test

rhc54:
## Motivation

A client should be able to cycle `PMIx_Init` → *(work: fences, puts, …)* → `PMIx_Finalize` and then `PMIx_Init` again, any number of times, with each new init presenting a **clean slate** — no state from one cycle leaking into the next. This PR documents how that is intended to work and fixes the client-side defects that currently prevent it, with a regression test to lock it down.

## What's here

**1. Specification** — `docs/how-things-work/init-finalize.rst`
Defines the state lifecycle across the init/finalize boundary, in two halves:
- **Server side:** the split between job-scoped and client-scoped state, and the three teardown triggers — client finalize (recycle the peer in place via destruct/construct, keep it registered), deregister-client (remove + release the peer), and deregister-nspace (drop GDS job data, release all peers, purge all references).
- **Client side:** the symmetry requirements a `PMIx_Finalize` must meet so the next `PMIx_Init` is a clean slate (reset every one-time-init latch, null every freed global, keep the util layer's init/finalize in lockstep), plus the platform caveat for builds without destructor/`-fini` support.

**2. Client-side teardown fixes** (implemented and tested)
- **Singleton flag reset** (`client: reset the singleton flag on every fresh PMIx_Init`): the fresh-init path now clears `pmix_client_globals.singleton`, closing a double-destruct of the server-side IOF lists on a singleton→with-server cycle.
- **TSD key lifecycle** (`runtime: leave no residue or dangling state after finalize`): `pmix_thread_set_main()` was never called, so the TSD key registry was always empty and `pmix_tsd_keys_destruct` was a no-op — every `pmix_net_init` leaked a pthread key slot until `PMIx_Init` failed after `PTHREAD_KEYS_MAX` cycles. Wire `set_main` into `pmix_init_util`, delete keys on teardown, null the freed registry pointer (it was left dangling and aborted the next `realloc`), and reset the `pmix_name_fns` latch.
- **Dangling pointers**: null `pmix_globals.evbase`/`evauxbase` and the cached `shared_thread_tracker` on teardown so nothing in the between-cycles window (or a stray `PMIx_Progress`) touches freed memory.

**3. Regression test** — `test/unit/client_cycle`, wired into `make check`
Cycles `PMIx_Init` → `PMIx_Put`/`Commit` → `PMIx_Fence` → `PMIx_Finalize` 1200 times (a count deliberately above `PTHREAD_KEYS_MAX` on common platforms) and checks each transition. Runs as a singleton, so it needs no server.

## Scope note

The server-side peer-recycle behavior is **specified** here as the intended design; this PR implements the **client-side** half plus the shared spec. The server-side implementation is a natural follow-up.

## Testing
- Clean build under `--enable-debug --enable-devel-check` (zero warnings).
- `make check` in `test/unit`: 9/9 pass, including the new `client_cycle`.
- `test/simple/simptest` (server + clients) still passes.

### DIFF
--- a/docs/how-things-work/index.rst
+++ b/docs/how-things-work/index.rst
@@ -10,6 +10,7 @@ find information on that subject here.
    :maxdepth: 2
 
    session_dirs.rst
+   init-finalize.rst
    distances.rst
    schedulers/index.rst
    sets_groups/index.rst
--- a/docs/how-things-work/init-finalize.rst
+++ b/docs/how-things-work/init-finalize.rst
@@ -1,0 +1,300 @@
+Client Init/Finalize Lifecycle and Server-Side State
+====================================================
+
+This document specifies how the PMIx server manages the per-client and
+per-namespace state associated with a client across the ``PMIx_Init`` /
+``PMIx_Finalize`` boundary. Its purpose is to define the intended
+behavior precisely enough to implement and test it, and in particular to
+guarantee one property:
+
+.. admonition:: The cycling guarantee
+   :class: note
+
+   A client may cycle ``PMIx_Init`` → *(work: fences, gets, puts,
+   events, …)* → ``PMIx_Finalize`` and then ``PMIx_Init`` again, any
+   number of times, reusing the **same namespace and rank**, and each
+   new ``PMIx_Init`` must present a **fresh interaction state** on the
+   server. No state produced in one cycle may leak into the next. The
+   host environment is **not** required to deregister and re-register the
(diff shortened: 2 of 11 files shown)


rhc54:
Consolidating the full init/finalize cycling effort (client + server + tool) into a single self-contained PR, #3938. These commits are included there. Closing this one to avoid interdependent PRs.

PR #3938, opened 2026-07-07
Init/finalize cycling: clean-slate teardown for client, server, and tool

### LINKED ISSUE #3931: PMIx_Fence deadlocks under repeated PMIx_Init/PMIx_Finalize: lost_connection() prunes a gracefully-finalized, reconnecting proc

rhc54:
## Overview

Make the PMIx init/finalize cycle leak-free across **all three roles**. A process must be able to cycle init → work → finalize → init repeatedly — reusing the same namespace/rank — with each new init presenting a clean slate, on the client, the server, and the tool. This is a single self-contained PR covering the whole effort.

The intended lifecycle is captured in a new spec, `docs/how-things-work/init-finalize.rst`, added and extended by the commits below.

Closes #3931

## 1. Client half

A client `PMIx_Finalize` must tear the library all the way down so the next `PMIx_Init` in the same process is a brand-new instance. Fixes a set of cross-cycle residue defects found by cycling a singleton client past the per-process pthread-key limit (it crashed around cycle ~505 on the leaked key slot, and earlier on a stale singleton flag).

- **`docs: specify the client init/finalize state lifecycle`** — the cycling guarantee, the three server teardown triggers, and the state-ownership table.
- **`docs: specify client-side init/finalize teardown`** — the symmetry requirements a clean finalize must uphold.
- **`client: reset the singleton flag on every fresh PMIx_Init`** — a stale singleton flag carried from a no-server cycle into a with-server cycle drove finalize to tear down lists the with-server path never built.
- **`runtime: leave no residue or dangling state after finalize`** — the root-cause fixes: call `pmix_thread_set_main()` so the TSD registry is live (else every cycle leaked a pthread key to exhaustion); delete TSD keys and NULL the registry on teardown; reset the name-fns latch; NULL the event base and the shared progress-thread tracker.
- **`client: tear down singleton IOF lists based on actual construction`** — a singleton's IOF lists must be destructed according to what the singleton path actually built, not a fixed assumption, so a cycle cannot double-destruct or leak them.
- **`examples/dynamic: tolerate a singleton PMIx_Init`** — the dynamic example no longer assumes it was launched with a server, so it runs cleanly as a singleton.
- **`test: add an init/finalize cycling regression test`** — `test/unit/client_cycle`, a singleton cycler wired into `make check`, default count past the key limit.

## 2. Server half

When a client and all its clones finalize, the server must reclaim per-client state rather than leaking the peer until the namespace is deregistered. Fixes orphaned peers accumulating in the `clients` array and `nfinalized` drift that broke the "all local processes finalized" condition.

The retirement mechanism converged over the course of the branch — recycle-in-place, then release, then finally the **tombstone** model, which is what ships here. The intermediate commits are retained so the reasoning is bisectable; the net design is the tombstone.

- **`server: recycle a client peer object on clean finalize`** — first cut: new `pmix_server_peer_finalized()` that recycled the peer in place. *(Superseded below.)*
- **`server: release a finalized peer instead of recycling in place`** — recycling a still-reachable `pmix_peer_t` proved fragile; switched to releasing and reallocating a fresh peer on reconnect. *(Superseded below.)*
- **`server: leave a finalized peer as a tombstone instead of releasing it`** — the shipping design. Both prior forms mutated shared per-namespace/per-rank state (nulled the `clients` slot, cleared/repointed `info->peerid`) from `lost_connection` the instant a cleanly-finalized socket dropped — racing a *different* local process's in-flight spawn/connect/disconnect collective or direct-modex get that resolves ranks through `info->peerid`, an architecture- and timing-sensitive hang of multi-local-process `MPI_Comm_spawn`. A finalized **client** peer is now left in place as an inert, `finalized`-guarded **tombstone** (kept at its slot, `peerid` unchanged), reclaimed at a quiescent point instead: on the rank's next reconnect, or at namespace deregistration. A stranded peer a newer connection already displaced is still freed immediately.
- **`server: free finalized tool peers instead of leaving them stranded`** — the tombstone protects *clients*; `lost_connection` previously retired only pure clients, so finalized **tool** peers leaked (a tool is assigned a fresh namespace each init, so it never reconnects onto the same rank to be reclaimed). A tool is not a job member and is never resolved through `info->peerid`, so it needs no tombstone: `lost_connection` now routes a finalized pure tool to `pmix_server_peer_finalized`, which frees it outright and repoints `peerid` (to a live tool sibling, else `-1`, restoring the state `process_tool_request` uses to recognize a reconnecting tool). A tool the host registered as a client keeps client tombstone semantics and is reclaimed on its tool reconnect via `process_tool_request` (the same reclaim the client connection handler does).
- **`server: fire the "all local processes finalized" callback once per nspace`** — because a tombstone stays counted in `nfinalized` until reclaimed, `nlocalprocs == nfinalized` is already satisfied when the host deregisters the namespace and holds for every rank the teardown walks — firing `pmix_pnet.local_app_finalized` once per rank instead of once for the job. A `local_app_fini_fired` latch on the namespace fires it exactly once. (Single-rank cycling hid this, since "once per rank" and "once" coincide when `nlocalprocs == 1`.)
- **`docs: describe the server-side peer recycle mechanics`** — the peer states and the `proc_cnt`/`nfinalized` invariants (rewritten by the tombstone doc updates below).
- **`test: exercise the server-side peer recycle in make check`** — a persistent `simptest` server forks a `simpcycle` client that loops init → Get/Store/Fence → finalize many times; driven by `run_simpcycle.pl`.

## 3. Tool half

`PMIx_tool_init`/`PMIx_tool_finalize` are a separate entry-point pair from the client's and must satisfy the same symmetry independently. A bare tool init/finalize loop **failed at cycle 1** with `PMIX_ERR_INIT`.

- **`tool: tear the library fully down on PMIx_tool_finalize`** — reset the one-time-init latch + reference-count the finalize (the fix that unbreaks cycling); destruct the `server_globals` lists the tool constructs on every init (`nspaces`, `iof_residuals`, `psets`, `grp_collectives`); free + NULL `tmpdir`/`system_tmpdir` (they leaked and silently pinned the prior cycle's tmpdir); destruct the client-side `peers` array and IOF sinks; drop a duplicate pnet close. (`module_set`, owned by the separate `PMIx_tool_set_server_module` call, is left as a documented follow-up.)
- **`docs: describe tool-side init/finalize teardown`** — a "Tool-Side Teardown" section.
- **`test: add a tool init/finalize cycling test to make check`** — `test/unit/tool_cycle`, the tool analog of `client_cycle`, exercising the tool *library* teardown standalone via `PMIX_TOOL_DO_NOT_CONNECT`.
- **`test: add a server-side tool init/finalize cycling test to make check`** — `test/simple/simptoolcycle` (driven by `run_toolcycle.pl`) covers what `tool_cycle` cannot: the *server's* retirement of a tool peer across cycles. It forks a minimal server that is both the scheduler and the owner of one registered job, then runs two phases, each a fresh re-exec'd child looping init/finalize — a **pure** tool (fresh namespace each cycle; connects as the scheduler so `tool_init` skips its job-info request) and a **client-registered** tool (reuses one namespace/rank). After each phase the parent asserts its `clients` array did not accumulate a peer per cycle.
- **`docs: describe tool finalize and once-per-nspace app-finalized firing`** — updates the lifecycle spec for the pure-tool free path (routed via `lost_connection`) and the once-per-nspace `local_app_finalized` firing.

## Testing

- Zero-warning `--enable-devel-check` build; `make check` fully green across all suites (12/12 in `test/unit`).
- Regression tests in `make check`: `client_cycle`, `run_simpcycle.pl` (server client cycling), `tool_cycle` (tool library), `run_toolcycle.pl` (server-side tool cycling, pure + client-registered).
- Each server-side leak reproduces without its fix and is closed by it: client cycling (25 cycles, server exits 0), pure-tool cycling (was +1 peer/cycle → 0 leak at 50 cycles), client-tool cycling (was +1 peer/cycle → bounded), and `local_app_finalized` firing once per nspace (was once per rank).
- Client singleton: 1200+ cycles. Tool standalone: 1500 cycles.
- `test/simple/simptest` smoke test green.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -210,6 +210,7 @@ test/python/run_sched.sh
 test/python/run_server.sh
 test/simple/pmitest
 test/simple/simpcycle
+test/simple/simptoolcycle
 test/simple/simpfabric
 test/simple/get_put_example
 test/simple/simpvni
@@ -237,6 +238,7 @@ test/unit/client_cycle
 test/unit/tool_cycle
 test/unit/run_gds_fallback.pl
 test/unit/run_simpcycle.pl
+test/unit/run_toolcycle.pl
 test/unit/util_argv
 test/unit/util_basename
 test/unit/util_path
--- a/.gitignore
+++ b/.gitignore
@@ -233,7 +233,9 @@ test/unit/bfrops_alloc_inherit
 test/unit/gds_fallback
 test/unit/pmix_log
 test/unit/collective_status
+test/unit/client_cycle
 test/unit/run_gds_fallback.pl
+test/unit/run_simpcycle.pl
 test/unit/util_argv
 test/unit/util_basename
 test/unit/util_path
--- a/.gitignore
+++ b/.gitignore
@@ -234,6 +234,7 @@ test/unit/gds_fallback
 test/unit/pmix_log
 test/unit/collective_status
 test/unit/client_cycle
+test/unit/tool_cycle
 test/unit/run_gds_fallback.pl
 test/unit/run_simpcycle.pl
 test/unit/util_argv
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -1006,6 +1006,7 @@ AC_DEFUN([PMIX_SETUP_CORE],[
     AC_CONFIG_FILES(pmix_config_prefix[test/run_tests12.pl], [chmod +x test/run_tests12.pl])
     AC_CONFIG_FILES(pmix_config_prefix[test/run_tests13.pl], [chmod +x test/run_tests13.pl])
     AC_CONFIG_FILES(pmix_config_prefix[test/unit/run_gds_fallback.pl], [chmod +x test/unit/run_gds_fallback.pl])
+    AC_CONFIG_FILES(pmix_config_prefix[test/unit/run_simpcycle.pl], [chmod +x test/unit/run_simpcycle.pl])
 #    AC_CONFIG_FILES(pmix_config_prefix[test/run_tests14.pl], [chmod +x test/run_tests14.pl])
 #    AC_CONFIG_FILES(pmix_config_prefix[test/run_tests15.pl], [chmod +x test/run_tests15.pl])
     if test "$WANT_PYTHON_BINDINGS" = "1"; then
--- a/config/pmix.m4
+++ b/config/pmix.m4
@@ -1007,6 +1007,7 @@ AC_DEFUN([PMIX_SETUP_CORE],[
     AC_CONFIG_FILES(pmix_config_prefix[test/run_tests13.pl], [chmod +x test/run_tests13.pl])
     AC_CONFIG_FILES(pmix_config_prefix[test/unit/run_gds_fallback.pl], [chmod +x test/unit/run_gds_fallback.pl])
     AC_CONFIG_FILES(pmix_config_prefix[test/unit/run_simpcycle.pl], [chmod +x test/unit/run_simpcycle.pl])
+    AC_CONFIG_FILES(pmix_config_prefix[test/unit/run_toolcycle.pl], [chmod +x test/unit/run_toolcycle.pl])
 #    AC_CONFIG_FILES(pmix_config_prefix[test/run_tests14.pl], [chmod +x test/run_tests14.pl])
 #    AC_CONFIG_FILES(pmix_config_prefix[test/run_tests15.pl], [chmod +x test/run_tests15.pl])
     if test "$WANT_PYTHON_BINDINGS" = "1"; then
--- a/docs/how-things-work/index.rst
+++ b/docs/how-things-work/index.rst
@@ -10,6 +10,7 @@ find information on that subject here.
    :maxdepth: 2
 
    session_dirs.rst
+   init-finalize.rst
    distances.rst
    schedulers/index.rst
    sets_groups/index.rst
--- a/docs/how-things-work/init-finalize.rst
+++ b/docs/how-things-work/init-finalize.rst
@@ -103,10 +103,12 @@ is normative: "must" denotes a required behavior.
 1. Client finalize (all clones included)
     A client, together with all of its clones, has finalized — the rank's
     ``proc_cnt`` has dropped to zero. The peer object serving that rank is
-    **released** and its client-array slot is nulled, but the rank's
-    **registration** (the ``pmix_rank_info_t`` on ``nptr->ranks``) is
-    deliberately preserved so the client can init again. See `Releasing the
-    peer on finalize`_.
+    left in place as an inert **tombstone** at its existing client-array
+    slot (it is not freed, its slot is not nulled, and ``info->peerid`` is
+    not moved), while the rank's **registration** (the ``pmix_rank_info_t``
+    on ``nptr->ranks``) is preserved so the client can init again. The
+    tombstone is reclaimed later, at the next reconnect for the rank or at
+    namespace deregistration. See `Tombstoning the peer on finalize`_.
 
 2. Host deregisters a client (``PMIx_server_deregister_client``)
     The server removes the peer object from the namespace and releases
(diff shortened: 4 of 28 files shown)


rhc54:
@hppritcha @jsquyres Perhaps one of you could take a look and provide some advice on what mpi4py is crying about now? I have no idea what that test does, and so no way of trying to address a problem.

rhc54:
Claude is churning through trying to make sense of the hang - stay tuned.

rhc54:
Sigh..."So I could not reproduce the mpi4py hang with pure-PMIx examples. It appears specific to OMPI's singleton integration (the prterun it forks) or to a particular mpi4py operation, neither of which the PMIx-only harness has. Reproducing the actual failure would need OMPI built against this PMIx + mpi4py — I did not build that stack unilaterally given its size; say the word and I'll add it to the harness."

Still trying.

rhc54:
Problem appears limited to x86_64, so Claude is going to have to drive it through the CI workflow interface. Will take awhile....

PR #3942, opened 2026-07-08
Update ignores

rhc54:
Ignore an existing build product

Add a golden rule directing contributors to add any build product
introduced by a new source file, component, or generated artifact to
the appropriate .gitignore rather than leaving it untracked or
committing it. Points at the per-directory .gitignore convention and
the git status check after a clean build.


rhc54:
@jsquyres You might want to add a similar golden rule over there

Issue #3936, opened 2026-07-06
Add tests for the dynamic group operations (invite/join/leave, leader failure, abort) via a dockerswarm harness


rhc54:
## Summary

The dynamic, event-driven group operations and their fault paths are
largely untested. The `contrib/dockerswarm` harness added in #3934
exercises group *construction* well (six construction methods across two
nodes, plus a participant-death scenario for construct), but the
invite/join/leave negotiation and the group fault events have little or
no automated coverage. We should build exercisers for the remaining
group operations and wire them into a companion dockerswarm harness
script.

## Current coverage (from #3934)

`contrib/dockerswarm/run-tests.sh` runs, across two nodes:

- construction methods: `group`, `group_bootstrap`, `group_dmodex`,
  `group_lcl_cid`, `asyncgroup`, `multi_nspace_group`;
- participant death during `PMIx_Group_construct` (`group_die`);
- participant death during `PMIx_Connect` (`connect_die`).

`asyncgroup` touches the invite/join happy path, but only as a
"did it complete" check.

## Gaps to cover

The dynamic group protocol is realized through events
(`PMIx_Notify_event`), and none of these paths are exercised
automatically:

- **Invite / join happy path** as a first-class test (not just a side
  effect of `asyncgroup`): leader invites, invitees accept via
  `PMIx_Group_join`, group forms, `PMIX_GROUP_CONSTRUCT_COMPLETE`.
- **Invite declined** — an invitee returns `PMIX_GROUP_INVITE_DECLINED`;
  verify the leader handles the decline and the construct resolves per
  the documented semantics.
- **Invite failed / timeout** — `PMIX_GROUP_INVITE_FAILED`.
- **Leader failure and reselection** — `PMIX_GROUP_LEADER_FAILED`
  delivered to participants; a new leader declared via
  `PMIX_GROUP_LEADER` in the handler; `PMIX_GROUP_LEADER_SELECTED`
  broadcast; and the no-new-leader error case.
- **Construct abort** — `PMIX_GROUP_CONSTRUCT_ABORT` from a participant or
  the leader propagates to all invited processes.
- **Member failure during dynamic construct** —
  `PMIX_GROUP_MEMBER_FAILED`.
- **Destruct with a lost member** — `group_die` covers construct; add the
  destruct analog (a member dies mid-`PMIx_Group_destruct`).
- **`PMIX_GROUP_FT_COLLECTIVE`** — an in-flight group collective adjusts
[...]

PR #3944, opened 2026-07-08
Add tests for dynamic group invite/join and destruct fault paths

rhc54:
Work on #3936, in the stepwise fashion agreed on the issue: first tests for the
already-supported dynamic-group paths, then fill the gaps one at a time.

## 1. Tests for already-supported paths

**Exercisers + companion harness**
- `examples/group_invite.c` — the invite/join happy path as a first-class test.
- `examples/group_destruct_die.c` — the destruct analog of `group_die` (a member
  is lost mid-`PMIx_Group_destruct`; the destruct completes on the survivors).
- `contrib/dockerswarm/run-group-events.sh` — companion harness, separate from
  `run-tests.sh` so the event/fault matrix can grow independently.

**Deterministic make check**
- `test/unit/run_grpinvite.pl` drives `group_invite` under `simptest -n 4`.

## 2. First gap filled: the invite timeout

`PMIx_Group_invite` accepted `PMIX_TIMEOUT` but never armed a timer, so a
non-responding invitee hung the leader forever. This implements it:

- Report each non-responder to the leader via `PMIX_GROUP_INVITE_FAILED` and
  form the group on the members that accepted (reduced membership).
- Restructured the completion flow to make that correct: track invitees by
  identity; do all `PMIx_Notify_event` calls on the caller's thread (they
  deadlock on the progress thread — this also removes a latent deadlock in the
  existing `PROC_TERMINATED` branch); collapse the two-tracker split into one and
  deregister the handler before releasing the tracker.
- Coverage: `examples/group_invite_timeout.c` (+ `run-group-events.sh` scenario)
  and a deterministic `test/unit/run_grptimeout.pl`.

## Testing

Full `test/unit` `make check` is green (15/15), including the invite happy-path
and timeout tests. All exercisers build clean under `--enable-devel-check` and
pass end-to-end under `simptest`; the timeout is stable across repeated runs.

## Still to come (per #3936)

Invite-declined resolution, leader failure/reselection, construct abort,
`PMIX_GROUP_MEMBER_FAILED`, and `PMIX_GROUP_FT_COLLECTIVE` remain; some require
coordination with PRRTE master.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,6 +124,7 @@ examples/group_destruct_die
 examples/group_invite
 examples/group_invite_timeout
 examples/group_invite_decline
+examples/group_invite_abort
 
 include/pmix_version.h
 include/pmix_rename.h
@@ -249,6 +250,7 @@ test/unit/run_grpmember.pl
 test/unit/run_grpinvite.pl
 test/unit/run_grptimeout.pl
 test/unit/run_grpdecline.pl
+test/unit/run_grpabort.pl
 test/unit/util_argv
 test/unit/util_basename
 test/unit/util_path
--- a/.gitignore
+++ b/.gitignore
@@ -122,6 +122,7 @@ examples/abort
 examples/group_leave
 examples/group_destruct_die
 examples/group_invite
+examples/group_invite_timeout
 
 include/pmix_version.h
 include/pmix_rename.h
@@ -245,6 +246,7 @@ test/unit/run_simpcycle.pl
 test/unit/run_toolcycle.pl
 test/unit/run_grpmember.pl
 test/unit/run_grpinvite.pl
+test/unit/run_grptimeout.pl
 test/unit/util_argv
 test/unit/util_basename
 test/unit/util_path
--- a/.gitignore
+++ b/.gitignore
@@ -123,6 +123,7 @@ examples/group_leave
 examples/group_destruct_die
 examples/group_invite
 examples/group_invite_timeout
+examples/group_invite_decline
 
 include/pmix_version.h
 include/pmix_rename.h
@@ -247,6 +248,7 @@ test/unit/run_toolcycle.pl
 test/unit/run_grpmember.pl
 test/unit/run_grpinvite.pl
 test/unit/run_grptimeout.pl
+test/unit/run_grpdecline.pl
 test/unit/util_argv
 test/unit/util_basename
 test/unit/util_path
--- a/.gitignore
+++ b/.gitignore
@@ -244,6 +244,7 @@ test/unit/run_gds_fallback.pl
 test/unit/run_simpcycle.pl
 test/unit/run_toolcycle.pl
 test/unit/run_grpmember.pl
+test/unit/run_grpinvite.pl
 test/unit/util_argv
 test/unit/util_basename
 test/unit/util_path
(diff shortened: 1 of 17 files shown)


rhc54:
Added the **invite-declined** gap (next step after invite-timeout), as two commits per the stepwise plan:

- `client: resolve a group invitation when a member declines` — completes the library so an explicit decline (or a termination) resolves the construct *immediately* instead of only on `PMIX_TIMEOUT`. A new per-member `answered` flag/counter tracks any definitive response separately from the `responded` (accepted) flag; the invitation resolves once every member has answered. The decliner is reported to the leader via `PMIX_GROUP_INVITE_FAILED` and excluded, and the group forms on the accepters — correctly even when no timeout was set (previously that would hang forever on a refusal). The handler also now consults `PMIX_EVENT_AFFECTED_PROC` so a termination is attributed to the departed proc.
- `test: exercise the group invite decline` — new `group_invite_decline` example (leader invites with **no** timeout; the last rank declines), a deterministic `make check` test `run_grpdecline.pl` on a single simptest server, and a matching multi-node dockerswarm scenario (linux + macOS).

Validation: full `test/unit` `make check` passes 17/17 (all four group tests: grpmember, grpinvite, grptimeout, grpdecline). Clean `--enable-devel-check` build.

Next gaps (still to come per #3936): construct abort (client-side), then the PRRTE-coordinated ones — leader failure/reselection, `PMIX_GROUP_MEMBER_FAILED`, and `PMIX_GROUP_FT_COLLECTIVE`.

rhc54:
Added the **construct-abort** gap (client-side, no PRRTE dependency), as two commits per the stepwise plan:

- `client: abort a non-optional group construct when a member fails` — gives the previously defined-but-unused `PMIX_GROUP_CONSTRUCT_ABORT` its first producer. The reduced-membership path (report non-accepters, form the group on whoever accepted) was being taken *unconditionally*; per `PMIX_GROUP_OPTIONAL` (documented default `false` = "return an error if any specified process terminates without having joined"), that path is correct only when the leader opts in. Now, absent `PMIX_GROUP_OPTIONAL`, an invite is all-or-nothing: any decliner/terminated/timed-out invitee aborts the whole construct — every invited participant (including accepters, so they stop waiting for a completion that never comes) receives `PMIX_GROUP_CONSTRUCT_ABORT` and the leader's `PMIx_Group_invite` returns that status. Entirely on the client/event path — no server/host-RM involvement. The decline/timeout exercisers (which test reduced membership) now pass `PMIX_GROUP_OPTIONAL` so they keep exercising it under the corrected default.
- `test: exercise the group invite construct abort` — new `group_invite_abort` example (leader invites the whole job with **no** `PMIX_GROUP_OPTIONAL`; last rank declines), a deterministic `make check` test `run_grpabort.pl`, and a matching multi-node dockerswarm scenario (linux + macOS).

Validation: full `test/unit` `make check` passes **17/17** (all five group tests: grpmember, grpinvite, grptimeout, grpdecline, grpabort). Clean `--enable-devel-check` build. Verified end-to-end under simptest: leader's invite returns `CONSTRUCT_ABORT`, all 4 participants notified, no `CONSTRUCT_COMPLETE`.

Remaining #3936 gaps (PRRTE-coordinated, per the plan): leader failure/reselection, `PMIX_GROUP_MEMBER_FAILED`, and `PMIX_GROUP_FT_COLLECTIVE`. Note the *collective* `PMIx_Group_construct` abort path (as opposed to this invite/join path) needs a server-side cancel in `pmix_server_group.c` + host-RM support, so it belongs with that coordinated work.

PR #4012, opened 2026-07-16
Runtime layer: developer guide, bug fixes, and unit tests

rhc54:
This branch is the result of a review of `src/runtime`. It adds a
developer orientation guide, fixes the concrete defects found during
that review, removes a dead API, and lands unit tests for the parts
that are feasibly testable.

## Documentation
- **Add `src/runtime/AGENTS.md`** (with the customary `CLAUDE.md`
  symlink) orienting contributors to the runtime/bootstrap layer: the
  init/finalize ordering contract and its re-entrancy requirements, the
  process-wide `pmix_globals` object, the non-framework MCA parameter
  registry, the reference-counted progress-thread engine, and the
  `pmix_info` support library. Follows the format of the AGENTS.md files
  already present in sibling `src/` subdirectories.

## Bug fixes
- **Progress-thread CPU-affinity binding.** The range parser tested
  `strtoul`'s end pointer against NULL (it never is), so a bare CPU
  number fell into the range branch, advanced past the string's
  terminator (an out-of-bounds read) and never set the CPU, while an
  explicit range like `3-5` excluded its final endpoint. Test the
  referenced character instead and make the range inclusive. Also tear
  down cleanly on a required-bind failure: the engine thread is already
  running when `pthread_setaffinity_np` fails, so stop it before
  returning the error and have the caller unlink the tracker before
  releasing it (previously it freed a live event base and left a
  dangling list entry).
- **`pmix_info` param/type display.** `pmix_info_do_params` reused its
  outer loop index for the inner `mca_types` search, so more than one
  `type component` pair to `--param` walked off the option array; give
  the inner search its own index. Also drop a leftover debug `printf`
  in `pmix_info_do_type`.
- **`pmix_rte_init` error paths.** Give the `PMIX_NODEID` and
  peer/namespace allocation failures a short label so the startup
  `show_help()` diagnostic names the failing step instead of formatting
  a NULL, and guard the `PMIX_NODE_INFO_ARRAY` handler against a NULL
  `darray`.
- **Harden `pmix_info_make_version_str`** against misuse: a NULL or
  unrecognized scope no longer crashes or reads uninitialized stack, and
  the `greek`/`repo` scopes no longer `strdup(NULL)`.

## Cleanup
- **Remove `pmix_progress_thread_finalize`** — it had no callers, and
  since `pmix_progress_thread_stop` already removes and releases the
  tracker at refcount zero it could never find a live tracker to act on.

## Tests
- **`test/unit/progress_threads`** exercises the name-addressed
  progress-thread engine (reference-counted init/stop, distinct bases
  per name, real event progression, pause/resume, resume-while-active,
  tracker removal at refcount zero, unknown-name handling), using named
  threads so it never disturbs the library's shared progress thread.
- **`test/unit/info_support`** covers `pmix_info_make_version_str` for
  every scope (including the new misuse-hardening paths) and the
  parsable `pmix_info_out`/`pmix_info_out_int` formatting via captured
  stdout.

Both are wired into `make check`. The whole series builds warning-free
under the default `--enable-devel-check`/`-Werror` flags, and each
commit compiles independently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -253,6 +253,8 @@ test/unit/client_cycle
 test/unit/tool_cycle
 test/unit/event_chain
 test/unit/hwloc_datatype
+test/unit/progress_threads
+test/unit/info_support
 test/unit/run_gds_fallback.pl
 test/unit/run_simpcycle.pl
 test/unit/run_toolcycle.pl
--- a/src/runtime/AGENTS.md
+++ b/src/runtime/AGENTS.md
@@ -1,0 +1,302 @@
+# AGENTS.md: The PMIx Runtime / Bootstrap Layer
+
+This document orients AI agents and human contributors working in
+`src/runtime`, the code that brings `libpmix` up and tears it back down.
+It assumes you have already read the top-level
+[`AGENTS.md`](../../AGENTS.md) — the golden rules (prefix conventions,
+`pmix_config.h`-first include order, constant-on-the-left comparisons,
+brace-everything, `#define`-logical-macros-to-0/1, warning-free under
+`--enable-devel-check`), the **thread-safety / progress-thread model and
+the caddy pattern**, the "prefer an MCA parameter over a hard-coded
+constant" rule, the `show_help` regeneration rule, and the
+copyright-header requirement all apply here and are not repeated. This
+file covers what is specific to `src/runtime`: the init/finalize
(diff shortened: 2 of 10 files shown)


rhc54:
Docs built fine - status report is being eaten somewhere.

PR #4036, opened 2026-07-28
Expand the conversions in an output filename pattern

rhc54:
## The problem

`PMIX_IOF_FILE_PATTERN` says the name given in `PMIX_IOF_OUTPUT_TO_FILE` is a
pattern the caller composed, rather than a stem we annotate with the namespace
and rank. We accepted the flag and recorded it, but never implemented what it
promises:

```c
if (NULL == strchr(nptr->iof_flags.file, '%')) {
    pmix_asprintf(&outfile, "%s.out", nptr->iof_flags.file);
} else {
    /* must process the pattern - for now, just take it literally */
    pmix_asprintf(&outfile, "%s.pattern.out", nptr->iof_flags.file);
}
```

A caller asking for per-process filenames of their own design got one nonsense
name — literally `<file>.pattern.out` — shared by every rank in the job.

## What this does

Implements the expansion:

| Conversion | Expands to |
|---|---|
| `%n` | the job's namespace |
| `%r` | the process's rank |
| `%R` | the rank, zero-padded to the width of the job's largest rank |
| `%h` | the hostname of the node writing the file |
| `%%` | a literal `%` |

Anything else introduced by `%` is an error, rather than a filename the caller
did not ask for. The stream suffix (`.out`/`.err`) is still appended, so stdout
and stderr can never land on the same file, and a pattern containing no `%` at
all remains exactly what it was before — a fixed name.

Two design points worth review:

- **The expansion belongs here, not in the launcher that accepted the
  option.** `pmix_iof_setup()` is the only place that knows the rank at the
  moment a file is opened.
- **The directory is now created from the FINAL name.** A pattern may put
  conversions in the directory part (`%h/rank-%R`); taking `pmix_dirname()` of
  the raw pattern creates a directory literally called `%h` and then fails to
  open the file in the one the pattern actually named.

`pmix_iof_check_pattern()` is exported alongside the expander so a launcher can
reject a malformed pattern while the user is still looking at the command line
they typed, rather than at a job that failed to open its output. Both share one
walker, so a pattern the check accepts can never be one the expansion rejects,
or expands differently.

New capability flag `PMIX_CAP_IOF_FILE_PATTERN` lets a launcher tell whether it
may offer the option at all.

## Testing

Exercised through the PRRTE side of this work (openpmix/prrte, `--output
file=NAME:pattern`), which is what motivated it:

- `prterun -np 3 --output "file=/tmp/pat/%h/rank-%R:pattern"` → one file per
  rank under a per-host directory each daemon created for itself;
- `%r`/`%%` and the `.out`/`.err` split verified;
- a bad conversion (`%q`) reported before anything launched;
- unchanged default naming (`<file>.<nspace>.<rank>.out`) confirmed;
- multi-node coverage in PRRTE's `contrib/dockerswarm` suite (168 pass / 0
  fail), including that each daemon expands `%h` and creates the directory its
  own expansion names — which one host cannot show.

### DIFF
--- a/include/pmix_version.h.in
+++ b/include/pmix_version.h.in
@@ -53,6 +53,7 @@
 #define PMIX_CAP_REGEX2          6
 #define PMIX_CAP_GROUP_FT        7
 #define PMIX_CAP_CLI_QUAL_VALUE  8
+#define PMIX_CAP_IOF_FILE_PATTERN 9
 
 
 /*****    DEPRECATED    *****/
--- a/src/common/AGENTS.md
+++ b/src/common/AGENTS.md
@@ -63,6 +63,33 @@ entry point.
 | `pmix_pfexec.c/.h` | (internal) `pmix_pfexec_base_*` | Local fork/exec: when PMIx itself launches processes (singleton, tool spawning children). Collapsed from the former `src/mca/pfexec` framework. |
 | `pmix_strings.c` | `PMIx_*_string` converters | Pure enum→string helpers, no threadshifting. Must stay in sync with the enums in `include/pmix_common.h.in`. |
 
+### Output-file naming lives here (`pmix_iof.c`)
+
+`pmix_iof_setup()` is what opens the per-process output sinks, and it is
+the only place that knows the namespace and rank at the moment a file is
+created — which is why the naming, and the `PMIX_IOF_FILE_PATTERN`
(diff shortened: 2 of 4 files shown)


rhc54:
The consumer this was written for is openpmix/prrte#2566, which adds `--output file=NAME:pattern` on top of it.

PR #4056, opened 2026-07-31
Audit src/hwloc and src/include: fix defects, require C11 atomics

rhc54:
A review of `src/hwloc` and `src/include` (generated files excluded), the
defects it turned up, the tests that now cover them, and the guide updates
that record the reasoning.

### `src/hwloc`

Seven defects, none of which had a test behind it:

- **Stack buffer overflow in the topology print handler.** `print_hwloc_obj`
  declared `char string[1024]` but passed `PMIX_HWLOC_MAX_STRING` (2048) as
  the length to `hwloc_bitmap_snprintf`. A machine with enough PUs for the
  machine-level cpuset to render past 1024 characters overwrote the frame.
- **`pmix_hwloc_get_cpuset_size` ignored its argument and returned
  `SIZE_MAX`.** It measured `hwloc_bitmap_weight()` of a *filled* bitmap;
  a filled bitmap is infinitely set, so hwloc returns `-1`.
  `PMIx_Value_get_size` then wrapped, and the `data_array` walker in
  `bfrops` accumulated `SIZE_MAX` per element.
- **`pmix_hwloc_finalize` never reclaimed the shared-memory segment it
  created.** The cleanup was gated on a flag set only on the *adopt* path
  (the client), but only the *server* sets `shmemfile`/`shmemfd` - so the
  block was unreachable for the one process it applied to.
- **`compute_distances` leaked its `devids` argv on every path**, success
  included, and several error returns also skipped the tracking list.
- **`compute_distances` could not produce a distance *range*.** It measured
  the common ancestor between the device and the single lowest object
  covering the *whole* cpuset, inside the loop over PUs - so nothing varied
  with the loop index and `mindist == maxdist` came back for every device on
  every machine. `pmix_device_distance_t(5)` is explicit that the pair
  exists "to support cases where the process may be bound to more than one
  location, and those locations are at different distances from the device",
  which the implementation could not express. Also, when no PU intersected
  the cpuset the result was `mindist = UINT16_MAX, maxdist = 0` -
  simultaneously "unknown" and "as close as possible", with min > max.
- **Missing NULL screening on public entry points.**
  `PMIx_Parse_cpuset_string` passes its arguments straight through, so a
  NULL string reached `strchr`; `PMIx_Get_cpuset`, `PMIx_Compute_distances`
  and `pmix_hwloc_check_vendor` were similarly exposed. A failed parse also
  left a half-built cpuset on the caller's struct.
- **`strdup(topo->source)` on a host-supplied `PMIX_TOPOLOGY2`** - a host is
  not required to label the topology it hands over. Plus
  `generate_cpuset_string` ignoring `hwloc_bitmap_list_asprintf`'s return,
  the same misread already corrected in its two siblings.

### `src/include`

**C11 atomics are now enforced at configure time.** Two holes:

- `PMIX_SETUP_CC` fell back to reconfiguring for C99 when the compiler could
  not be put in C11 mode. That could never work - the library uses
  `_Atomic`, the C11 atomic convenience types and `<stdatomic.h>`
  unconditionally (`pmix_globals_t` alone declares four bare `atomic_bool`
  fields) - so the build failed on the first translation unit with
  unknown-type errors naming neither C11 nor the compiler.
- The required-atomics block tested a sample: `atomic_store` plus four
  `__atomic_*` builtins, leaving the `_Atomic` keyword, the convenience
  types, `atomic_load`, `atomic_thread_fence`/`memory_order`, and the whole
  `atomic_fetch_*_explicit` family unchecked. All eleven primitives the code
  actually uses are now probed, each fatal and each named in the error, via
  a file-scope `PMIX_REQUIRE_ATOMIC` helper.

Correspondingly, `pmix_stdatomic.h` now includes `<stdatomic.h>`
unconditionally rather than behind `#ifdef HAVE_STDATOMIC_H`.

Also removed: `pmix_hash_string.h` (included twice by `src/util/pmix_hash.c`
but neither macro used; PRRTE carries its own copy and never included ours),
the dead `#undef` cluster / `PMIX_HAVE_THREADS` / self-include in
`pmix_config_bottom.h`, and the unguarded string arguments in the
`pmix_check_*` inline classifiers.

### Tests

`test/unit/hwloc_datatype` goes from 22 to 39 checks. The two that matter
most fail against the previous code: the wide-topology print case
(4096-PU synthetic topology) **segfaults**, and the device-distance range
case reports no range. Both derive their inputs from the topology rather
than from hand-written literals.

New multi-node coverage - `examples/topology.c` driven by
`contrib/dockerswarm/run-topology.sh` - for the three things one process
structurally cannot reach: the topology *handoff* from the local server
(shmem, or XML when no VM hole is available), `PMIX_LOCALITY_STRING` as the
host actually stores it fed to `PMIx_Get_relative_locality`, and a real
cross-node answer. On one host every peer is a node-mate, so a single-host
run cannot distinguish a correct result from one that claims everything is
local; the exerciser fails itself if it finds no off-node peer. The runner
also repeats the suite with `hwloc_hole_kind=none` so the XML fallback is
exercised rather than assumed, and checks that no `hwloc.sm` survives
teardown.

### Testing done

Warning-free build and `make check` 80/80 on macOS/arm64 (clang, hwloc
2.10.0, `--enable-debug --enable-devel-check`), after a full
`autogen.pl` + `configure` cycle for the `.m4` changes. All eleven atomic
probes verified to run, and the generated `configure` verified to
`as_fn_error` on each failure path. `examples/topology.c` verified
end-to-end under `prterun`.

The multi-node `run-topology.sh` suite has **not** been run - no Docker
swarm available here.


rhc54:
### Multi-node suite has now been run

`contrib/dockerswarm/run-topology.sh linux` against the 10-container swarm
(Ubuntu 24.04, hwloc 2.10.0, PRRTE master built against this PMIx) — **6
passed, 0 failed**:

```
=== preflight ===
  PASS topology exerciser present in the shared volume
=== topology + locality across two nodes ===
  PASS all 4 ranks loaded a topology and agreed on locality
=== topology + locality across four nodes ===
  PASS all 8 ranks across 4 nodes agreed on locality
=== same, with the shmem topology path disabled (XML fallback) ===
  PASS no shmem segment: the clients must use the XML string
  PASS all 4 ranks agreed on locality with shmem disabled
=== shmem segment is reclaimed at teardown ===
  PASS no hwloc.sm left behind on node1/node2
```

Per-rank output confirms the assertions are live rather than vacuous — each
rank finds 1 node-mate and 2 off-node peers, and the locality comparison
runs on the host's real stored strings:

```
TOPO rank 0: PASS load_topology  hwloc:2.10.0
TOPO rank 0: PASS locality string is available  SK0:L20:L10:CR0-7:HT0-7:NM0
TOPO rank 0: PASS node-mate shares the node  [peer 2 on node1: rc=PMIX_SUCCESS bits=0x406f ...]
TOPO rank 0: PASS off-node peer is distinguishable  [peer 1 on node2, we are on node1]
TOPO rank 0: PASS job actually spans more than one node  [1 node-mates, 2 off-node]
```

Negative control: the same binary launched with all four ranks on one node
reports `FAIL job actually spans more than one node` on every rank, so the
suite is not green by default.

### One fix this run produced

The XML-fallback case was passing **without ever leaving the shmem path**.
The parameter is registered as `pmix_hwloc_hole_kind` — framework and
component both — and the case was written against `hwloc_hole_kind`, the
name `src/hwloc/AGENTS.md` gave. An unrecognized MCA parameter is accepted
in silence, so the run stayed on shared memory and the case was simply the
default case a second time, reporting PASS.

Pushed as a follow-up commit. The case now starts a persistent DVM with the
parameter, checks both nodes for `hwloc.sm`, and launches the ranks only
once no segment exists — so a parameter that stops taking effect fails the
case instead of quietly reverting it to a duplicate. The guide is corrected
to give the registered names, with the instruction to confirm one against
`pmix_info` rather than reading it off a document.

### Caveat on what the swarm can and cannot show

The containers are unbound — every rank sees all 8 CPUs — so every process
on a node reports the same locality string and only `PMIX_LOCALITY_SHARE_NODE`
is meaningfully exercised. That is the bit the earlier prefix defect broke,
so the regression is covered; finer-grained locality (distinct cores mapping
to distinct bits) still needs a machine where PMIx can actually bind.

PR #4121, opened 2026-08-11
Remove what a deregistration named, and nothing else

rhc54:
`PMIx_server_deregister_resources` matched on the key alone. Its man page, following the Standard, describes narrowing a removal with qualifiers — a `PMIX_NODE_INFO_ARRAY` naming the node that hosts a fabric device, plus the device's name — and passing exactly that removed **every** node entry the server held. Over-removing is the worst of the available behaviors here: nothing in the library can put an entry back, and the host has no way to learn what went missing.

A request element carrying a data array is now read as the qualified request it is, and the two kinds of member in it do different things:

- **Identifiers** (`PMIX_NODEID`, `PMIX_HOSTNAME`) choose which entries the request applies to. An array carrying no identifier applies to every entry with that key — which is how a device that is unique across the system is removed without naming its node.
- **Targets** (anything else) choose the elements to remove from inside those entries. A target matches an element directly, or matches an element that *contains* it — which is what reaches a fabric device, since a node array holds a `PMIX_FABRIC_DEVICE` element carrying its own array of the device's name and id.

An array with identifiers and no target removes the whole entry, as before but narrowed to the node named. Removing the last element an entry described removes the entry too: what would be left names a node and nothing else, which is what an empty registration would have produced, and it would go on seeding every namespace registered afterwards.

The handler also reports what happened — it returned `PMIX_SUCCESS` unconditionally, so a malformed qualifier was indistinguishable from a removal that worked. Both halves of the walk screen what they read: the request comes from a host and the entries were registered by one, so neither array's element types are ours to assume.

Covered by `test/unit/server_setup.c`: two nodes registered under one key with the removal narrowed to each in turn (by hostname and by nodeid), a device removed from a node carrying two, the entry dropped when its last device goes, and the unqualified key-only removal still clearing every entry.

### What this does not change

Non-namespace information is copied into a namespace's data store when that namespace is registered — `hash_cache_job_info`, guarded by the per-namespace `gdata_added` — and nothing re-reads `gdata` afterwards. So a deregistration governs the namespaces registered *after* it and leaves a running job's copy in place. The man page now says so, and `src/server/AGENTS.md` records why closing it is not a small change: the `gds` module struct has no delete-a-key entry point, and one cannot be built the obvious way for `shmem3` without putting a lock on its read path.

Verified: `make` clean under the tree's `--enable-devel-check` configuration; `make check` 21/21 in `test/` and 70/70 + 10/10 + 7/7 in `test/unit/`; docs build clean under `-W`.

### DIFF
--- a/docs/man/man3/PMIx_server_deregister_resources.3.rst
+++ b/docs/man/man3/PMIx_server_deregister_resources.3.rst
@@ -83,16 +83,34 @@ As with all non-blocking PMIx APIs, when a callback is supplied the caller
 DIRECTIVES
 ----------
 
-The ``key`` of each ``info`` element selects the entry to remove. Qualifiers may
-be included to scope the removal, for example:
+The ``key`` of each ``info`` element selects the entries to remove. An element
+whose value is **not** a data array selects by key alone: every registered entry
+carrying that key is removed.
 
-* ``PMIX_NODE_INFO_ARRAY`` (pmix_data_array_t\*) |mdash| scope the removal to a
-  specific node, identified within the array by ``PMIX_NODEID`` or
-  ``PMIX_HOSTNAME``.
-* ``PMIX_FABRIC_DEVICE_NAME`` (char*) |mdash| the name of a fabric device to
-  remove.
-* ``PMIX_DEVICE_ID`` (char*) |mdash| a system-unique device identifier; sufficient
-  on its own to remove the corresponding device.
+An element whose value **is** a ``pmix_data_array_t`` of ``pmix_info_t`` narrows
+the removal. Two kinds of member are recognized within it:
+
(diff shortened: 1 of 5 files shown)


rhc54:
Rebased onto master now that #4119 has landed, and added a second commit: `src/server/AGENTS.md` still carried the warning that `pmix_common.h` and `_register_resources` disagreed about `PMIX_GROUP_ENDPT_DATA`, which #4119 settled. It now describes the shape the attribute carries and where it comes from.

Re-verified after the rebase: clean build, `make check` 21/21 in `test/` and 70/70 + 10/10 + 7/7 in `test/unit/`, docs clean under `-W`.

The remaining half of the original open item — that a deregistration cannot retract what a namespace already holds — is recorded in #4122 rather than fixed here, per the discussion: closing it needs a delete-a-key entry point across every `gds` component, and `shmem3` cannot serve one without putting a lock on its read path.

Issue #4087, opened 2026-08-05
gds/shmem3: modex segment rotation assumes a cumulative payload; delta commits will break it


rhc54:
## Background

`gds/shmem3` now publishes **each modex in its own shared-memory segment** rather than writing every modex into the segment the first one created. A finished segment has been advertised and local clients have it mapped, so it must never be written again; a blob arriving for a modex already flagged `READY_FOR_USE` starts the next generation, and the previous segment is handed off (its backing file is reference counted, so a client that still has it mapped keeps a valid mapping until the last holder lets go).

That change fixed a reproducible server abort — a second collecting fence larger than the first overran a segment sized for the first, and since the allocator behind it is a bump allocator, the `abort()` happened *inside the PMIx server*, taking the daemon and the job with it. It reproduces from four nodes up.

## The dependency this creates

**Dropping generation N is only correct because the modex payload is cumulative.** Today:

- `PMIx_Commit` ships the process's *entire* local store, not a delta — `_commitfn` fetches with `cb->proc = &myid`, `cb->scope = PMIX_LOCAL` and a **NULL key**, which means "all data for this proc" (`src/client/pmix_client.c`). `PMIx_Put` of an existing key overwrites in that store first, so updates are already folded in before anything ships.
- A server contributes each local proc's full set from **its own** store (`cb.scope = PMIX_REMOTE`, NULL key, `src/server/pmix_server_ops.c`). That matters because `commits_pending` means a client that put nothing new before a later fence packs nothing at all — the server covers it.

So generation N+1 is self-contained, and a client only ever reads the newest segment.

The Standard permits this but does not require it. `PMIx_Commit` is specified as making available *all* previously staged key/value pairs, and notes that "depending on the implementation, this may involve transmitting the entire collection of data posted by the process to the local PMIx server."

## What we intend to change

As a performance optimization, commit should stop shipping the whole store when a process adds or updates one value and fences — there is no point circulating a large unchanged set repeatedly. That makes the modex payload a true **delta**.

At that point the current rotation is wrong: generation N+1 would carry only what changed, and dropping generation N would lose everything published earlier and not re-published.

## Options when we get there

1. **Carry forward.** When rotating, copy the previous generation's contents into the new segment before applying the delta. Keeps the read path trivial (one segment, newest wins) and keeps the "a visible segment is never written again" invariant. Costs a copy per fence and sizing has to account for the accumulated set, which partly gives back what the delta saves.
[...]

PR #4158, opened 2026-08-17
Reserve shmem3's addresses instead of hoping for them, and stop a failed modex attach from taking the job data with it

### LINKED ISSUE #4156: gds/shmem3: fence path has no GDS fallback, leaving the client without job data

git master @ `7f5777f38` ("Advertise the versioned framework declaration as a capability"),
`git describe` = `v5.0.10rc1-1009-g7f5777f38`, reports `7.0.0a1`.

Rest of the stack, all built against that PMIx:

* PRRTE master @ `7924a09354` (reports `5.0.0a1`)
* Open MPI main @ `b3983031bb` (reports `6.1.0a1`)

From a git clone:

```shell
shell$ ./configure --prefix=$HOME/.local     # no --enable-debug
shell$ make install
```

PRRTE and Open MPI were configured with `--with-pmix=$HOME/.local`.

* Operating system/version: Rocky Linux 8.10 containers, kernel 6.8.0 x86_64
* Computer hardware: single host, Intel Core i7-8650U, 8 logical CPUs, 16 GB RAM; four
  containers act as compute nodes (`mc-slurmd-1` … `mc-slurmd-4`) under Slurm
* Network type: TCP/Ethernet over the container network; no high-speed fabric

-----------------------------

## Details of the problem

When a client cannot map a gds/shmem3 segment at its required fixed address,
`shmem3_attach()` converts the failure to `PMIX_ERR_TAKE_NEXT_OPTION` and releases the job
tracker, so that the caller can fall back to the next GDS module
(`src/mca/gds/shmem3/gds_shmem3.c:1135-1180`).

The job-data path does that. `src/client/pmix_client.c:1006-1010`:

```c
        if (PMIX_ERR_TAKE_NEXT_OPTION == rc) {
            /* the connect-time GDS module could not deliver our job data to
             * this client; fall back to the next module and re-request */
            rc = fallback_to_next_gds();
        }
```

The fence path does not. `unpack_return()` in `src/client/pmix_client_fence.c:274-277`:

```c
    /* provide an opportunity to store any data (or at least how to access
     * any data) that was included in the fence */
    PMIX_GDS_RECV_MODEX_COMPLETE(rc, pmix_client_globals.myserver, data);
    if (PMIX_UNLIKELY(PMIX_SUCCESS != rc)) {
        PMIX_ERROR_LOG(rc);
        return rc;
    }
```

`PMIX_GDS_RECV_MODEX_COMPLETE` invokes only the peer's assigned module
(`src/mca/gds/gds.h:329-340`), so there is no "next option" to take here: the status is
logged and handed back to the caller. The client is left pointing at shmem3 with its job
tracker released.

**Expected**: a client that cannot attach a shmem3 segment during a fence falls back to the
next GDS module the way it does at init, and the fence completes with the modex stored.

**Actual**: `PMIx_Fence()` returns `PMIX_ERR_TAKE_NEXT_OPTION` to the application, and
job-level lookups for the client's *own* namespace afterwards fail with
`PMIX_ERR_INVALID_NAMESPACE`:

```
[mc-slurmd-2:13213] PMIX ERROR: PMIX_ERR_TAKE_NEXT_OPTION in file client/pmix_client_fence.c at line 276
[rank 3] PMIx_Fence rc=PMIX_ERR_TAKE_NEXT_OPTION
[rank 3] PMIx_Get(PMIX_LOCAL_PEERS) rc=PMIX_ERR_INVALID_NAMESPACE value=(none)
```

Since the fixed-address attach depends on the client's VM layout, this is intermittent in
production: we hit it in 1 of 25 runs of the same multi-node workload.

### Reproducer

The existing `gds_shmem3_force_client_attach_failure` MCA parameter cannot exercise this
path, because it fails *every* attach, including the one at init where the fallback handles
it and switches the client to hash. The shim below fails only attaches after init, which is
where the modex segment is attached.

The job must span at least two nodes. In a single-node job no modex segment is sent to the
client, so no attach happens at fence time and the bug cannot trigger.

`attach_fail_shim.c` — fails `pmix_shmem_segment_attach()` for selected ranks from the Nth
call onwards (calls 1-2 are init, call 3 is the fence):

```c
#define _GNU_SOURCE
#include <dlfcn.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#define PMIX_ERR_NOT_AVAILABLE (-64)

typedef int (*attach_fn_t)(void *shmem, unsigned long addr, int flags, unsigned int layout_id);

static int rank_is_selected(void)
{
    const char *ranks = getenv("ATTACH_FAIL_RANKS");
    const char *my_rank = getenv("PMIX_RANK");
    char list[256];

    if (!ranks || !my_rank) {
        return 0;
    }
    snprintf(list, sizeof(list), "%s", ranks);
    for (char *tok = strtok(list, ","); tok; tok = strtok(NULL, ",")) {
        if (0 == strcmp(tok, my_rank)) {
            return 1;
        }
    }
    return 0;
}

int pmix_shmem_segment_attach(void *shmem, unsigned long addr, int flags, unsigned int layout_id)
{
    static attach_fn_t real = NULL;
    static int calls = 0;
    const char *from_env = getenv("ATTACH_FAIL_FROM");
    int from = from_env ? atoi(from_env) : 3;
    int call;

    if (!real) {
        real = (attach_fn_t) dlsym(RTLD_NEXT, "pmix_shmem_segment_attach");
    }
    call = __atomic_add_fetch(&calls, 1, __ATOMIC_RELAXED);

    if (rank_is_selected() && call >= from) {
        fprintf(stderr, "[shim] rank %s: failing attach call %d\n", getenv("PMIX_RANK"), call);
        return PMIX_ERR_NOT_AVAILABLE;
    }
    return real(shmem, addr, flags, layout_id);
}
```

`pmix_fence_repro.c` — put a key, commit, fence with `PMIX_COLLECT_DATA`, then read back:

```c
#include <pmix.h>
#include <stdio.h>
#include <string.h>

int main(void)
{
    pmix_proc_t myproc, wildcard, peer;
    pmix_value_t put, *val = NULL;
    pmix_info_t info;
    bool collect = true;
    uint32_t nprocs = 1;
    pmix_status_t rc;

    rc = PMIx_Init(&myproc, NULL, 0);
    if (PMIX_SUCCESS != rc) {
        fprintf(stderr, "PMIx_Init failed: %s\n", PMIx_Error_string(rc));
        return 1;
    }

    PMIX_PROC_CONSTRUCT(&wildcard);
    strncpy(wildcard.nspace, myproc.nspace, PMIX_MAX_NSLEN);
    wildcard.rank = PMIX_RANK_WILDCARD;
    if (PMIX_SUCCESS == PMIx_Get(&wildcard, PMIX_JOB_SIZE, NULL, 0, &val) && NULL != val) {
        nprocs = val->data.uint32;
        PMIX_VALUE_RELEASE(val);
    }

    PMIX_VALUE_CONSTRUCT(&put);
    put.type = PMIX_UINT32;
    put.data.uint32 = myproc.rank;
    PMIx_Put(PMIX_GLOBAL, "repro-key", &put);
    PMIx_Commit();

    PMIX_INFO_LOAD(&info, PMIX_COLLECT_DATA, &collect, PMIX_BOOL);
    rc = PMIx_Fence(NULL, 0, &info, 1);
    printf("[rank %u] PMIx_Fence rc=%s\n", myproc.rank, PMIx_Error_string(rc));

    PMIX_PROC_CONSTRUCT(&peer);
    strncpy(peer.nspace, myproc.nspace, PMIX_MAX_NSLEN);
    peer.rank = (myproc.rank + 1) % nprocs;

    val = NULL;
    rc = PMIx_Get(&peer, "repro-key", NULL, 0, &val);
    printf("[rank %u] PMIx_Get(rank %u, repro-key) rc=%s\n",
           myproc.rank, peer.rank, PMIx_Error_string(rc));

    val = NULL;
    rc = PMIx_Get(&wildcard, PMIX_LOCAL_PEERS, NULL, 0, &val);
    printf("[rank %u] PMIx_Get(PMIX_LOCAL_PEERS) rc=%s value=%s\n",
           myproc.rank, PMIx_Error_string(rc),
           (PMIX_SUCCESS == rc && NULL != val) ? val->data.string : "(none)");

    PMIx_Finalize(NULL, 0);
    return 0;
}
```

Build and run — ranks 2 and 3 are the two procs on the second node:

```shell
shell$ gcc -shared -fPIC -o libattachfail.so attach_fail_shim.c -ldl
shell$ gcc -o pmix_fence_repro pmix_fence_repro.c -I$PREFIX/include -L$PREFIX/lib -lpmix
shell$ export LD_PRELOAD=$PWD/libattachfail.so ATTACH_FAIL_RANKS=2,3 ATTACH_FAIL_FROM=3
shell$ prterun -np 4 --host mc-slurmd-1:2,mc-slurmd-2:2 \
         -x LD_PRELOAD -x ATTACH_FAIL_RANKS -x ATTACH_FAIL_FROM ./pmix_fence_repro
```

Output (unaffected ranks first, then the two sabotaged ones):

```
[rank 0] PMIx_Fence rc=PMIX_SUCCESS
[rank 0] PMIx_Get(rank 1, repro-key) rc=PMIX_SUCCESS
[rank 0] PMIx_Get(PMIX_LOCAL_PEERS) rc=PMIX_SUCCESS value=0,1
[rank 1] PMIx_Fence rc=PMIX_SUCCESS
[rank 1] PMIx_Get(rank 2, repro-key) rc=PMIX_SUCCESS
[rank 1] PMIx_Get(PMIX_LOCAL_PEERS) rc=PMIX_SUCCESS value=0,1
[shim] rank 2: failing attach call 3
[shim] rank 3: failing attach call 3
[mc-slurmd-2:13213] PMIX ERROR: PMIX_ERR_TAKE_NEXT_OPTION in file client/pmix_client_fence.c at line 276
[mc-slurmd-2:13212] PMIX ERROR: PMIX_ERR_TAKE_NEXT_OPTION in file client/pmix_client_fence.c at line 276
[rank 3] PMIx_Fence rc=PMIX_ERR_TAKE_NEXT_OPTION
[rank 2] PMIx_Fence rc=PMIX_ERR_TAKE_NEXT_OPTION
[rank 2] PMIx_Get(rank 3, repro-key) rc=PMIX_SUCCESS
[rank 2] PMIx_Get(PMIX_LOCAL_PEERS) rc=PMIX_ERR_INVALID_NAMESPACE value=(none)
[rank 3] PMIx_Get(rank 0, repro-key) rc=PMIX_SUCCESS
[rank 3] PMIx_Get(PMIX_LOCAL_PEERS) rc=PMIX_ERR_INVALID_NAMESPACE value=(none)
```

Per-proc `PMIx_Get` still resolves; it is the job-level data for the client's own namespace
that is no longer reachable.

### How this reaches applications

This was found because it silently corrupts locality in Open MPI, not because anything
aborted:

1. `fence_release()` (`ompi/runtime/ompi_mpi_init.c:395-401`) discards the fence completion
   status, so `MPI_Init` continues as if the modex succeeded.
2. `ompi_proc_complete_init()` (`ompi/proc/proc.c:315`) reads `PMIX_LOCAL_PEERS`, which now
   fails, so the affected procs never build their local-peer list.
3. `MPI_Comm_split_type(MPI_COMM_TYPE_SHARED)` (`ompi/communicator/comm.c:834-835`) includes
   a peer only if `OPAL_PROC_ON_LOCAL_NODE`, so each affected proc ends up alone.

In a 20-process job over four nodes where the five processes on one node hit the fence
error, `MPI_COMM_TYPE_SHARED` produced 8 communicators instead of 4 — three groups of five
plus five singletons — which broke the node accounting of the library running on top. The
natural (unsabotaged) occurrence looks like this:

```
[mc-slurmd-3:11478] PMIX ERROR: PMIX_ERR_TAKE_NEXT_OPTION in file client/pmix_client_fence.c at line 276
[mc-slurmd-3:11479] PMIX ERROR: PMIX_ERR_TAKE_NEXT_OPTION in file client/pmix_client_fence.c at line 276
[mc-slurmd-3:11481] PMIX ERROR: PMIX_ERR_TAKE_NEXT_OPTION in file client/pmix_client_fence.c at line 276
[mc-slurmd-3:11483] PMIX ERROR: PMIX_ERR_TAKE_NEXT_OPTION in file client/pmix_client_fence.c at line 276
[mc-slurmd-3:11485] PMIX ERROR: PMIX_ERR_TAKE_NEXT_OPTION in file client/pmix_client_fence.c at line 276
```

A fallback in the fence path equivalent to `fallback_to_next_gds()` would fix this. Failing
that, leaving the client attached to a module whose job tracker has been released seems
worth avoiding on its own, since the client cannot recover afterwards.

rhc54:
Fixes #4156.

The report is accurate, and two separate things were wrong. Only one of
them is what the title said.

### The fence returned an internal signal to the application

`unpack_return()` has no GDS fallback to take, and cannot cheaply be
given one: `PMIX_GDS_RECV_MODEX_COMPLETE` resolves a single module, and
`PMIX_GDS_FALLBACK_CMD` re-registers job data only, so a completed modex
cannot be re-delivered in another module's format without a new wire
command. `PMIX_ERR_TAKE_NEXT_OPTION` therefore escaped out of
`PMIx_Fence` as its return code. Open MPI does not check that return, so
a job carried on with a modex it had not stored.

### The failed attach also destroyed data that was fine

This is the damaging half. `shmem3_attach()` had a single failure
policy, written for the init case: detach, take the job tracker off the
list, release it. At fence time the segment being attached is the MODEX
one, so that also dropped the job and session segments the client had
been reading successfully since `PMIx_Init` — after which every
job-level lookup for the client's **own** namespace failed with
`PMIX_ERR_INVALID_NAMESPACE`, long after and nowhere near the fence that
caused it. That is how it was noticed: `MPI_Comm_split_type(MPI_COMM_TYPE_SHARED)`
putting a node's five processes into five communicators of one.

A failed modex attach now clears that segment's status and keeps the
tracker, and the fence completes. The client loses the fast path to
remote data and nothing else — the fetch side already gates every modex
read on `PMIX_GDS_SHMEM3_READY_FOR_USE`, so the lookup misses and goes
up to the server like any other, and a later generation can still attach
into the cleared slot.

## Not needing luck in the first place

Losing the shared-memory fast path is a poor outcome even when it is
survivable, so the rest of the branch is about not reaching it.

A client attaches the **job** segment during `PMIx_Init`, when its
address space is nearly empty, and the **modex** segment at a *fence* —
by which time it has loaded an MPI stack, opened components, grown
allocator arenas and registered fabric memory over the same ground. Both
addresses are chosen by the server from its own map and both must be
honored exactly, because the structures in a segment hold absolute
pointers. So the first attach nearly always works and the second is a
gamble; #4156 is that gamble lost, about one run in 25, taking every
rank on a node at once.

So stop asking for the address and hold it instead. Before creating any
of a job's segments the server reserves one contiguous range with the
new `pmix_vmem_reserve()` — a `PROT_NONE` mapping that commits nothing
and costs a VMA. The range travels on every seg blob, and each client
claims it while still inside `PMIx_Init`, the one moment it reliably
can. Every later mapping, including each modex generation, is then
`MAP_FIXED` over ground the process already owns.

Placement changed too. `VMEM_HOLE_BIGGEST` aims at the **midpoint of the
biggest hole**, and hwloc and Open MPI — which share this routine's
ancestry — aim there as well; that convergence is what makes an address
picked in one process likely to be taken in another. `VMEM_HOLE_BIGGEST_OFFSET`
lands a quarter of the way in instead, and the placement is scattered by
a hash of the namespace so concurrent jobs on a node do not agree either.

Four things are load-bearing and are commented as such:

* **Carve by `pmix_shmem_utils_segment_footprint()`, not the requested
  size.** A segment maps a page of header ahead of its data; carving to
  the size asked for overlapped each segment with the next by that page
  and the following `MAP_FIXED` replaced it. It segfaulted the daemon on
  the first fence.
* **Detach restores the reservation** rather than unmapping, or
  releasing a modex generation punches a hole for something else to take.
* **A modex generation holds its slot for as long as it is readable.**
  This began as an alternation between two slots, on the assumption that
  one generation was live at a time; #4087 ended that. A delta
  generation leaves every earlier one mapped on `job->modex_prior`, so
  `arena_alloc_modex()` reads occupancy off the live segments and takes
  the lowest free slot.
* **The session segment stays outside the arena.** A session outlives
  the job that first described it, and `job_destruct()` releases the
  whole arena.

If the arena cannot be reserved, or a modex outgrows its slot, every
segment places itself independently exactly as before — degraded, never
broken.

## Two dead ends, recorded so they are not retried

**Placing at the top of the hole is worse than the midpoint.** It was
the first thing tried. The top of that hole is the underside of the
executable's load address — where ASLR moves processes around most and
where the heap grows — so on aarch64 clients on a second node could not
reserve the range their own server had picked.

**A two-job test cannot show that the scatter works.** ASLR has already
moved each server's load base and the placement is relative to it, so
two jobs land far apart whether or not anything scatters them; the
concurrent-jobs stage added here passes with the scatter compiled out,
which was confirmed by compiling it out. The scatter is for where ASLR
does not do that job — randomization disabled (not rare in HPC) and
non-PIE executables — so the property is pinned down in
`test/unit/util/util_vmem.c` against the placement function directly,
where it is decidable.

## Testing

`examples/modex_attach_fail.c` asserts the three things that must
survive a modex that will not map: the fence succeeds, job-level data
for the client's own namespace is still reachable, and a peer's value
still arrives by some route. `run-gds-tests.sh` drives it over 2, 4 and
8 servers with the failure forced and again without, plus a
concurrent-jobs stage. Two nodes is not a convention there but the whole
test — a single-node job exchanges no remote data, so no modex segment
is ever sent and there is nothing to fail.

The new `gds_shmem3_force_modex_attach_failure` parameter is what makes
that drivable. The existing `force_client_attach_failure` cannot reach
this path: it fails every attach, so the client leaves `PMIx_Init` on
`gds/hash` and never gets as far as a fence on `shmem3` — which is why
the reporter had to `LD_PRELOAD` a shim.

`test/unit/util/util_vmem.c` covers placement and the
reserve/map/restore/release round trip; it skips itself where there is
no `/proc/self/maps`, which is also where `gds/shmem3` is not built.

Both behavioral fixes were checked against negative controls: reverting
them reproduces #4156 verbatim, including the same
`pmix_client_fence.c:276` error line, and reverting the scatter fails
exactly the `different scatter values give different addresses`
assertion.

Verified on macOS (clean build, `make check`) and on Linux through
`contrib/dockerswarm` (warning-free `shmem3` build, gds/client/common/bfrops
suites).

### One thing this does not fix

The 8-server stages of `run-gds-tests.sh` hang intermittently. That is
**pre-existing and unrelated** — measured at 10 of 12 runs on `master`
against 11 of 12 on this branch. It is a deadlock rather than load: it
reproduces on an idle host and stays wedged for ten minutes, with each
stuck rank's main thread in `futex_wait` and its progress thread idle in
`epoll_wait`, blocked in `PMIx_Get` for a remote peer's key whose owner
has already finished. It will be filed separately.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,6 +86,7 @@ examples/debugger
 examples/debuggerd
 examples/datatypes
 examples/delete_key
+examples/modex_attach_fail
 examples/dmodex
 examples/dynamic
 examples/modex_twice
--- a/contrib/dockerswarm/run-gds-tests.sh
+++ b/contrib/dockerswarm/run-gds-tests.sh
@@ -223,6 +223,24 @@ run_across_nodes() {
             bad "$prog $label, $np servers: both fences reported ok but exit was $rc"
             return 1
         fi
+    elif [ "$prog" = modex_attach_fail ]; then
+        # Every rank has to say OK: the fence succeeded, its own
+        # namespace's job data was still reachable, and it still got a
+        # peer's value - by whatever route. Counting the OK lines rather
+        # than trusting the exit status matters here, because the failure
+        # this covers is one a caller that ignores a return code does not
+        # see (openpmix#4156 reached production through exactly that).
+        nok=$(echo "$out" | grep -c 'MODEX-ATTACH-FAIL OK')
+        if [ "$nok" != "$np" ]; then
+            bad "$prog $label, $np servers: only $nok/$np ranks degraded cleanly"
+            echo "$out" | grep -iE 'FAILED|TAKE_NEXT_OPTION|INVALID_NAMESPACE' \
+                | head -3 | sed 's/^/       /'
+            return 1
+        fi
+        if [ "$rc" != 0 ]; then
+            bad "$prog $label, $np servers: every rank was OK but exit was $rc"
+            return 1
+        fi
     elif [ "$prog" = client ]; then
         # client reports per key per peer and returns 0 regardless, so
         # count the confirmations: each of the $np ranks confirms one key
@@ -385,7 +403,8 @@ test_linux() {
         -e PFX="$PMIX_PREFIX" \
         "$IMAGE" bash -euo pipefail -c '
             mkdir -p /opt/prte/tests-gds
-            for p in datatypes dmodex client modex_twice delete_key get_timing fence_timing; do
+            for p in datatypes dmodex client modex_twice delete_key \
+                     modex_attach_fail get_timing fence_timing; do
                 gcc -O0 -g -o /opt/prte/tests-gds/$p /pmix-src/examples/$p.c \
                     -I"$PFX/include" -I/pmix-src/examples \
                     -L"$PFX/lib" -lpmix -Wl,-rpath,"$PFX/lib"
@@ -520,6 +539,34 @@ test_linux() {
         run_across_nodes datatypes "$hosts" "$np" \
             "PMIX_MCA_gds_shmem3_force_client_attach_failure=1" "(fallback)" \
             && ok "$np servers: clients fell back and still read every peer"
+    done
+
+    banner "a modex the client cannot map (openpmix#4156)"
+    # The OTHER attach failure, and the one with no fallback available.
+    # force_client_attach_failure above cannot reach it: that fails every
+    # attach, so the client leaves PMIx_Init on gds/hash and never gets to
+    # a fence on shmem3. This one fails only the modex segment, which is
+    # attached at fence time - and used to take the client's job and
+    # session segments down with it while handing PMIx_Fence's caller
+    # PMIX_ERR_TAKE_NEXT_OPTION.
+    #
+    # Two nodes minimum is not a convention here, it is the whole test: a
+    # single-node job exchanges no remote data, so no modex segment is
+    # ever sent to the client and there is nothing to fail.
+    for geom in $GEOMETRIES; do
+        hosts="${geom%|*}"; np="${geom#*|}"
+        run_across_nodes modex_attach_fail "$hosts" "$np" \
+            "PMIX_MCA_gds_shmem3_force_modex_attach_failure=1" "(no modex map)" \
+            && ok "$np servers: fence succeeded and job data survived"
+    done
+
+    banner "the same program with the modex mapping normally"
+    # The control. If this fails, the program is broken rather than the
+    # degradation path it is meant to be checking.
+    for geom in $GEOMETRIES; do
+        hosts="${geom%|*}"; np="${geom#*|}"
+        run_across_nodes modex_attach_fail "$hosts" "$np" "" "(modex mapped)" \
+            && ok "$np servers: baseline with the modex segment mapped"
     done
 }
 
--- a/contrib/dockerswarm/run-gds-tests.sh
+++ b/contrib/dockerswarm/run-gds-tests.sh
@@ -140,6 +140,7 @@ GEOMETRIES="${GEOMETRIES:-node1:1,node2:1|2 node1:1,node2:1,node3:1,node4:1|4 no
 #   over its checks, so the only usable signal is the absence of its
 #   failure lines together with rank 0's finalize line. Grade it that way
 #   rather than on the exit status, which proves nothing here.
+
 run_across_nodes() {
     local prog="$1" hosts="$2" np="$3" envs="${4:-}" label="${5:-}"
     local out rc nok xargs exports v trace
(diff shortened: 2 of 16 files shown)


rhc54:
Rebased onto #4160, which has to land first.

The `xversion (master)` failure is not something this PR can fix from its own side. The blob keys it adds are rejected by the reader on the **master** side, which refuses any key it does not recognize — so a newer server plus an older client fails `PMIx_Init` outright, and no change here reaches that client. Reproduced locally: this PR's `simptest` against master's `simpclient` gives exactly the CI failure; against master-plus-#4160 it passes, as does the reverse direction.

A `PMIX_GDS_SHMEM3_LAYOUT_ID` bump would not help — that stamp is checked in `pmix_shmem_segment_attach()`, and `unpack_shmem3_seg_blob_and_attach_if_necessary()` parses the blob before it ever attaches, so the client dies before anything reads the stamp. (Renaming the component would work, by making an old peer decline `shmem3` entirely, but that is the remedy for a change in the shared *memory layout*, and nothing here changes what is in the segment.)

#4160 is that one-hunk reader fix, split out so it can go in first. This branch now sits on top of it, so its own copy of the change is gone and it will rebase cleanly once #4160 merges.

Note the 8-server stages of `run-gds-tests.sh` still hang on this stack — that is #4159, which is independent of both of these.

PR #4161, opened 2026-08-17
Name a device by the node it is on, and tell a process which GPUs it may use

rhc54:
Groundwork in PMIx for `--map-by device=gpu` (PRRTE side in openpmix/prrte, opened alongside this; it needs this PR). Motivated by open-mpi/ompi#14169.

The theme is that a device assignment is only worth making if the process can act on it, and three separate things stood between "PMIx knows which GPU" and "the process can use it".

### The device uuid named the wrong machine

`build_device_uuid()` took the hostname from `pmix_globals.hostname` — the host of whatever process was doing the enumeration. That is right only when the topology is the local one, and the caller that matters is exactly the one for which it is not: PRRTE's mapper runs on the head node and enumerates every compute node's topology in turn, so every GPU in a job came back named `gpu://<head node>::cuda0`.

That defeats the reason the uuid travels instead of an ordinal. A process correlates the uuid it was handed against what it computes locally from `PMIX_DEVICE_DISTANCES`; the two agreed on exactly one node. Nothing detects it either — the strings are well-formed and unique per device, just consistently attributed to the wrong machine.

`pmix_hwloc_get_devices()` now takes the hostname as a **required** parameter (a NULL is `PMIX_ERR_BAD_PARAM`, because a default is how this happened). `pmix_hwloc_compute_distances()` is the one caller entitled to pass `pmix_globals.hostname`, since it measures against a cpuset of this machine's PUs.

### PMIx never reported the identity a GPU runtime accepts

PMIx names a device with a uuid of its own construction, which no GPU library has heard of. `pmix_hwloc_device_t` gains `vendor_id`/`vendor`, harvested from the key hwloc's vendor backends record (`NVIDIAUUID`, `AMDUUID`, `LevelZeroUUID`).

Harvested across the whole PCI **function**, not off the OS device that names it: with the CUDA and NVML backends both loaded a GPU exposes `cuda0`, `opencl0d0` and `nvml0` on one function, the enumerator names it `cuda0`, and the uuid is on `nvml0`. Reading only the named device reports "no identity" on precisely the machines that have one.

`NULL` is the ordinary answer — an hwloc built without the vendor backends describes the same GPUs with no way to name any of them — so what to do about it is left to the caller.

### The whole pgpu environment path was dead code

`PMIx_server_setup_fork` asked `ptl`, `pnet`, `gds` and `pmdl` for their contribution to a child's environment and never asked `pgpu`. So `allocate()` harvested the vendor's envars, the host shipped the blob, `setup_local()` cached it — and nothing ever replayed the cache into a process. A component could be selected, do all its work, and have no observable effect. There is no error and no empty result to notice, because the cache is written by one function and read by another that runs zero times.

On top of that, the `amd`/`intel`/`nvd` components disabled themselves in `configure` unless `--enable-test-build` was given. None of them links anything or needs a vendor SDK, and the machine that builds PMIx is routinely not the machine that runs it — detection belongs in `component_open`, which already asks the local topology whether the vendor's hardware is present. They build unconditionally now.

### Per-process environment

`pmix_pgpu_module_t` gains `setup_fork`, called after the namespace-wide envars. Everything `setup_local` caches is per namespace and replayed identically into every child; a GPU assignment is per rank and had nowhere to be set. Framework version → 2.0.0.

The shared work lives in the base as `pmix_pgpu_base_set_visible_devices()`, so a component is three lines. `nvd` sets `CUDA_VISIBLE_DEVICES`, `amd` sets `ROCR_VISIBLE_DEVICES`. `intel` deliberately sets nothing: `ZE_AFFINITY_MASK` takes indices in an ordering PMIx cannot reproduce, and a wrong value in these variables does not fail — it silently shrinks the visible set. For the same reason no index is ever guessed, and none of them touches the vendor's device-ordering variable.

Reading the identity from **this** node's topology is why the work happens at fork time on the daemon: the head node holds one topology for every node that reported identical hardware, so the identities in its copy belong to whichever node reported first.

### Testing

`make check` passes (78 tests). Two unit tests are new or extended:

- `hwloc_devices` enumerates one topology under two node names and checks the assignment moved with the node — and that uuids built from the *hardware's* own identifiers (fabric GUIDs, MACs) do **not** move, since those name the card wherever it is plugged in.
- `pgpu_visible_devices` drives the whole chain against a topology whose GPUs are named by `cuda0` while the identity sits on the sibling `nvml0`, so a value only appears if it was found across the function. It hands the server that topology through the documented `topo_file` hook, which also makes `nvd`'s own NVIDIA check pass on a machine with no GPU, so the module really is selected rather than simulated.

Three topology fixtures are added: the same real machine (from open-mpi/ompi#14169) reduced to 16 PUs and read by an hwloc with and without the vendor backends, plus the existing `test-topo2` whose `AMDUUID` sits on the device that names its function — so all three arrangements are covered rather than whichever one the first fixture happened to use.

### DIFF
--- a/src/hwloc/AGENTS.md
+++ b/src/hwloc/AGENTS.md
@@ -289,6 +289,22 @@ a normal build does exercise it.
   `pmix_hwloc_compute_distances()` is the one caller entitled to pass
   `pmix_globals.hostname`, because it measures against a cpuset of *this*
   machine's PUs and so can only ever be reading the local topology.
+- **A device's vendor identity is usually absent, and that is not a
+  failure.** `pmix_hwloc_device_t.vendor_id` carries the handle a GPU
+  runtime will actually accept — NVIDIA's `GPU-<uuid>` and its AMD and
+  Level Zero equivalents — and hwloc records it only from the matching
+  vendor backend (`NVIDIAUUID`, `AMDUUID`, `LevelZeroUUID`). An hwloc
+  built without them describes the same GPUs with no way to name any of
+  them, so `NULL` is the common answer and callers decide what it means.
+  Two properties worth knowing: the identity is harvested across the
+  whole PCI **function**, not read off the OS device that names it (with
+  CUDA and NVML both loaded the function is named `cuda0` while the uuid
+  sits on `nvml0`, so reading only the named device reports "no identity"
+  on exactly the machines that have one); and *whether* a topology has
+  identities is structural — an absent info attribute changes an object's
+  info count, which `hwloc_topology_diff_build()` calls "too complex" —
+  while the *values* are per-node and must be read from that node's own
+  topology.
 - **Always check `source` first, and match it with `strncasecmp(...,
   "hwloc", 5)`.** Returning `PMIX_ERR_TAKE_NEXT_OPTION` (not an error) for
   a non-hwloc object is the contract. A `NULL` source is legal on *input*
--- a/src/hwloc/AGENTS.md
+++ b/src/hwloc/AGENTS.md
@@ -274,6 +274,21 @@ a normal build does exercise it.
 
 ## Pitfalls specific to this directory
 
+- **A device uuid names the node the device is on, not the node reading
+  the topology.** `pmix_hwloc_get_devices()` takes the hostname as a
+  *required* parameter and `build_device_uuid()` uses it — neither reads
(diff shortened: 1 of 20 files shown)


rhc54:
Chicken/egg problem - PRRTE master cannot build with this change, and cannot update without it.

PR #4182, opened 2026-08-21
Discover interfaces the OS actually has, and their real netmasks

rhc54:
Review of `src/mca/pif` — the framework that enumerates the local node's
network interfaces into `pmix_if_list`, once, during `pmix_init_util()`,
for **every** PMIx process. Four commits: the fixes, a style sweep,
the IPv6 prefix work, and the kernel-index width — the last two both
described below.

### The one worth reading about

`bsdx_ipv4` computed `if_mask` by running `prefix()` over the interface's
own **address** instead of over `ifa_netmask`. A /24 interface therefore
reported whatever the trailing zero bits of its address happened to say —
32 for any address ending in an odd octet. Measured on the machine this
was found on:

```
lo0   addr=127.0.0.1      netmask=255.0.0.0       reported /32, should be /8
en1   addr=192.168.1.155  netmask=255.255.255.0   reported /32, should be /24
```

Nothing crashes on that, which is why it lasted: it quietly turns
`pmix_ifaddrtokindex()` into an exact-address match, so a peer on the
same subnet stops resolving to the local NIC, and `pmix_ifindextomask()`
hands its callers — all of which are outside this tree — a wrong prefix.

### Everything else

| Sev | Component | Defect |
|---|---|---|
| HIGH | `bsdx_ipv4` | `if_mask` from the address, not `ifa_netmask` (above) |
| HIGH | `bsdx_ipv4`, `bsdx_ipv6` | `ifa_addr` dereferenced unchecked. `getifaddrs(3)` on these exact platforms documents it may be NULL; here that is a segfault before the library is up |
| HIGH | `posix_ipv4` | `break` on a failed `SIOCGIFADDR`/`HWADDR`/`MTU` abandoned the whole loop, silently dropping every interface *after* the failing one — surfacing later as "no usable interface", nowhere near the ioctl |
| MED | `posix_ipv4` | unchecked `ifr` malloc; the walk trusted an `ifc_len` the file's own comment says a platform may report larger than the buffer; retry guard was `errno != EINVAL && lastlen != 0`, which swallowed a genuine first-attempt error and retried it to the 10MB cap |
| MED | `linux_ipv6` | loop on `fscanf(...) != EOF`. A partial match neither consumes the input that failed to convert nor reports EOF, so anything short of a full line spins forever appending interfaces built from uninitialised `idx`/`pfxlen`/`scope` |
| MED | both `bsdx` | an unchecked `malloc` of an outer `struct ifaddrs **` before `getifaddrs`, justified by a comment claiming the call segfaults without it. It does not — passing the address of a local pointer is the documented usage — so the allocation was only a NULL for `getifaddrs` to write through |
| MED | `base` | `frameopen` was raised *before* the components opened. If open fails the base never sets the framework's OPEN flag, so `pmix_mca_base_framework_close` takes its not-open branch and our close never runs — stranding the list and leaving the latch up, which would make a later successful open skip discovery entirely |

Two behaviour choices worth flagging: a failed MAC/MTU ioctl now **keeps**
the interface (both are informational, and the `bsdx` components never
populate them at all) rather than discarding it; and per-interface ioctl
chatter moved from `pmix_output(0, …)` to `pmix_output_verbose`.

### The IPv6 prefix, which turned out to be two bugs hiding each other

Split out below because it was initially deferred and should not have
been. `pmix_net_samenetwork()` takes a prefix length and, for `AF_INET6`,
ignored it: unless it was exactly 64 it returned `false`, whatever the
two addresses were. Its own header said as much — *"for IPv6, we only
need to check for /64, there are no other local netmasks"* — which has
not been true for a long time.

That one restriction is why the two IPv6 discovery components could not
agree on `if_mask`:

- `linux_ipv6` reports the true prefix from `/proc`, so its loopback
  entry carries **/128** and was never on the same network as anything —
  including itself. `pmix_ifaddrtokindex("::1")` could not resolve
  loopback at all.
- `bsdx_ipv6` hard-coded **64** for every interface, which kept the
  comparison working by feeding it the only value it understood.

Each bug concealed the other, so neither could be corrected alone. Both
are fixed here. The comparison now checks the whole bytes the prefix
covers and then the leftover bits in the byte it ends inside of, walking
the address a byte at a time — an IPv6 address is in network byte order,
so byte 0 is always the most significant and endianness does not enter
into it. For a prefix of 64 it compares exactly the same bits as before,
so nothing that worked changes; every other prefix previously returned
`false` unconditionally, which is not behaviour anything could have
depended on. `bsdx_ipv6` now converts `ifa_netmask` with a `prefix6()`
helper, the IPv6 twin of the IPv4 `prefix()`.

### The kernel index width

Also split out, and also initially deferred. An OS interface index is a
32-bit unsigned number that Linux issues from a counter which only
increases. PMIx carried it in a `uint16_t` and handed it back through
helpers that did not agree on a width:

| | |
|---|---|
| `pmix_pif_t.if_kernel_index` | `uint16_t` |
| `pmix_ifnametokindex()` | `int16_t` |
| `pmix_ifaddrtokindex()` | `int16_t` |
| `pmix_ifindextokindex()` | `int` |
| `pmix_ifkindextoname()` / `pmix_ifkindextoaddr()` / `pmix_ifmatches()` | `int` parameter |

The disagreement bites well before the field overflows, which is the part
I had missed when I called this exotic. An index of **40000** stores in
the `uint16_t` perfectly well and `pmix_ifindextokindex()` returns it
correctly — but `pmix_ifnametokindex()` returns it as **-25536**, and
every caller screens a kernel index with `if (0 > kindex)`. The same
interface is therefore findable by list index and missing by name at the
same time, and an `if_include` entry naming it is silently skipped. That
is the misidentification, and it needs only `> 32767`, not `> 65535`.

Everything is now one width: the field is `uint32_t`, both `int16_t`
returns become `int` (which is what every consumer already took), and the
unassigned sentinel becomes `UINT32_MAX`.

Widening the field made the compiler flag four comparisons that had been
relying on the old integer promotion, and those were worth more than a
cast — the three functions taking a signed `kindex` are legitimately fed
a `-1` from a failed lookup, and converting that to unsigned yields
`UINT32_MAX`, exactly the sentinel an unfilled entry carries. Each now
rejects a negative index up front.

**Rebuild note:** `pmix_pif_t` lives in the installed
`src/util/pmix_if.h`, and PRRTE's `oob` does walk `pmix_if_list` and
construct `pmix_pif_t` directly, so it needs a rebuild against this. It
needs no source change — its three uses of `if_kernel_index` all go
through an `int` already — and it calls neither of the two functions
whose return type changed. I have not built PRRTE against this branch.

### Testing

New `test/unit/pif_discovery` (23 cases), new coverage of
`pmix_net_samenetwork()` in `test/unit/util/util_net.c` (which had none —
46 cases now), and new kernel-index-width cases in
`test/unit/util/util_if.c` (27 cases). All wired into `make check`.

`pif_discovery` asserts on the *contents* of `pmix_if_list` rather than
on the mechanism that filled it, so the same program is a real test of
whichever pair of components a build selected. It cross-references every
discovered prefix, v4 and v6, against `getifaddrs(3)` — deliberately a
different mechanism from the code under test on **both** platforms.
`util_if` builds its interface list by hand, which is the only way to
reach an index above 32767: no developer's machine has one.

Every claim above was verified by re-breaking the fix and watching the
right test fail:

| restore | fails |
|---|---|
| `if_mask` from the address in `bsdx_ipv4` | `lo0: pif says /32, the OS says /8`; `en1: /32` vs `/24` |
| the hard-coded `/64` in `bsdx_ipv6` | `lo0: pif says /64, the OS says /128` |
| the `/64`-only comparison in `pmix_net_samenetwork()` | `lo0: not on its own network at /128` |
| the `int16_t` return on `pmix_ifnametokindex()` | 3 cases — nametokindex disagrees with indextokindex at 40000 |
| the `uint16_t` `if_kernel_index` field | the 70000 case, on its own |

- macOS (`bsdx_ipv4` + `bsdx_ipv6`): build warning-free, `make check`
  155/155.
- Linux container, `--enable-devel-check` (`posix_ipv4` + `linux_ipv6`):
  build warning-free, `pif_discovery` 23/23, `util_net` 46/46, `util_if`
  27/27. Worth stating because component selection here is a
  **configure-time** decision — those two components get no compiler
  coverage whatsoever on macOS, and vice versa.

The framework and per-component `AGENTS.md` files are updated, including
several statements the code now contradicts — among them "the `malloc`
before `getifaddrs` is intentional … do not simplify it away" and
`linux_ipv6`'s flag-lookup description, which had drifted from the code.

https://claude.ai/code/session_01LWfFcCtTNQHNtVpDxDjajH


rhc54:
Good catch — you were right, and I've done it.

My reason for deferring was weak. I wrote that the honest fix was a
general prefix comparison in `pmix_net_samenetwork()` and then treated it
as out of scope because it lives in `src/util` rather than `src/mca/pif`,
and because `pmix_ifaddrtokindex()` has no in-tree callers so I couldn't
measure a behaviour change against anything. Neither holds up:

- The IPv4 arm of that same function *already* does the general thing via
  `pmix_net_prefix2netmask()`, so this is the file's own idiom, not a new
  design.
- It isn't a behaviour change at all in the risky sense. Every
  `plen != 64` returned `false` unconditionally — nothing can depend on
  that. For `plen == 64` the new code compares bit-for-bit what the old
  one did.
- And the deferral left a live bug in place: on Linux `linux_ipv6`
  records the `/128` the kernel reports for `::1`, so loopback was not on
  the same network as itself and `pmix_ifaddrtokindex("::1")` could never
  resolve it.

What I'd got wrong in the framing is that this was ever one bug. The
`/64`-only comparison and `bsdx_ipv6`'s hard-coded `/64` were concealing
each other: the hard-coded value kept the broken comparison fed with the
only input it understood, which is exactly why "fix `bsdx_ipv6` alone"
looked like a regression and why I stopped. They had to move together.

New commit `9cfac04`. `pmix_net_samenetwork()` had no test coverage at
all, so it has some now (46 cases), and `pif_discovery` gained the IPv6
half of its `getifaddrs(3)` cross-check plus a check that every
discovered address is on its own network — which is the property the
`/64` assumption actually broke. Both halves verified by re-breaking:
restoring the hard-coded 64 fails the netmask cross-check on `lo0`
(`/64` against the OS's `/128`); restoring the `/64`-only comparison
fails the self-match check on the same interface.

macOS `make check` 155/155; Linux container `pif_discovery` 23/23 and
`util_net` 46/46, both builds warning-free. The `docs/todo.rst` entry
that recorded this as deferred is removed, and the `AGENTS.md` files now
explain that the two are a matched pair rather than telling the next
reader it's a known-and-parked issue.

rhc54:
Right again, and it's worse than I described — done in `9037970`.

I'd framed this as a truncation problem needing >65535 interfaces' worth
of churn. The misidentification you suspected is real and needs only
**>32767**, because the widths disagree with each other:

```
pmix_pif_t.if_kernel_index    uint16_t
pmix_ifnametokindex()         int16_t
pmix_ifaddrtokindex()         int16_t
pmix_ifindextokindex()        int
pmix_ifkindextoname/addr/matches()   int parameter
```

An index of 40000 stores in the `uint16_t` fine, and
`pmix_ifindextokindex()` returns it correctly — but
`pmix_ifnametokindex()` returns **-25536**, and every caller screens a
kernel index with `if (0 > kindex)`. So the same interface is findable by
list index and missing by name at the same time, and an `if_include`
entry naming it is silently skipped. No overflow required.

On "any reason not to just do it" — the reason I gave (ABI) was mostly
wrong, and I should have checked instead of asserting:

- The two functions whose **return type** changes have no callers at all,
  in-tree or in PRRTE. Zero risk.
- The **struct field** is the only real ABI surface, and it does matter:
  PRRTE's `oob` walks `pmix_if_list` and does `PMIX_NEW(pmix_pif_t)`
  directly, so it needs a rebuild against this. It needs no source
  change — its three uses of `if_kernel_index` all go through an `int`
  already. I have not built PRRTE against this branch; flagging it so you
  can decide whether you want that sequenced.

Everything is one width now: `uint32_t` field, `int` returns, sentinel
`UINT32_MAX`. Widening the field made the compiler surface four
comparisons relying on the old integer promotion, and those were worth
more than a cast — `pmix_ifkindextoname/addr/matches` take a signed
kindex that callers legitimately feed a `-1` from a failed lookup, and
converting that to unsigned gives `UINT32_MAX`, which is exactly the
sentinel an unfilled entry carries. Each now rejects a negative index up
front.

Tests are in `util_if`, which builds the list by hand since no machine
here has such an index. Both halves verified by re-breaking: the
`int16_t` return alone fails 3 cases (nametokindex vs indextokindex at
40000); the `uint16_t` field alone fails the 70000 case. macOS `make
check` 155/155, Linux container 23/23 + 46/46 + 27/27, both warning-free.

`docs/todo.rst` is now empty of `pif` deferrals — both items you pushed
back on are fixed rather than parked.

PR #4196, opened 2026-08-25
common: add statuses and a process state for a release that kills processes

Petter-Programs:
A resource release that removes nodes still running a job kills the processes on
them. PMIx has no status for that, so a runtime either says nothing — PRRTE
recorded such a process as normally terminated, and the job it belonged to ended
"successfully" — or borrows a status that means something else. The nearest ones
are `PMIX_ERR_JOB_CANCELED`, `PMIX_ERR_PROC_KILLED_BY_CMD` and
`PMIX_PROC_STATE_KILLED_BY_CMD`, all of which say a command ended the job and
none of which can be told apart from an operator cancelling it.

That distinction is the point for an elastic framework, which shrinks by
releasing nodes: without it, the framework cannot tell a loss it inflicted on
itself from one imposed on it, though only the latter should abort the campaign.

PMIx already names the ways a process can be lost: aborted by signal, terminated
without sync, sensor bound exceeded, killed by command. The resource-driven one
is missing, even though the operations that cause it — `PMIX_ALLOC_RELEASE`, and
the DVM size-change events `PMIX_ERR_DVM_MOD` and `PMIX_DVM_IS_READY` — are
already standard.

Adds:

- `PMIX_ERR_JOB_KILLED_BY_RELEASE` (-236), continuing the job-error block
- `PMIX_ERR_PROC_KILLED_BY_RELEASE` (-204), in the process-event block
- `PMIX_PROC_STATE_KILLED_BY_RELEASE` (`PMIX_PROC_STATE_ERROR + 14`), with its
  `PMIx_Proc_state_string()` case

All three are listed in `docs/exceptions.rst`. `pmix_event_strings.c` is
generated from `pmix_common.h.in`, so no generated file is committed; a dry run
of `contrib/construct_event_strings.py` confirms both statuses are harvested.

Consumed by the matching PRRTE change, which reports the processes a DVM node
release kills instead of recording them as normally terminated.

### DIFF
--- a/docs/exceptions.rst
+++ b/docs/exceptions.rst
@@ -1027,6 +1027,8 @@ Constants
 * ``PMIX_ERR_JOB_EXE_NOT_FOUND``: specified executable not found
 * ``PMIX_ERR_JOB_INSUFFICIENT_RESOURCES``: insufficient resources to
   spawn job
+* ``PMIX_ERR_JOB_KILLED_BY_RELEASE``: a resource release removed nodes
+  this job was running on, killing the processes on them
 * ``PMIX_ERR_JOB_SYS_OP_FAILED``: system library operation failed
 * ``PMIX_ERR_JOB_WDIR_NOT_FOUND``: specified working directory not
   found
@@ -1036,6 +1038,10 @@ Constants
 * ``PMIX_ERR_PROC_REQUESTED_ABORT``: process called ``PMIx_Abort``
 * ``PMIX_ERR_PROC_KILLED_BY_CMD``: process was terminated by RTE
   command
+* ``PMIX_ERR_PROC_KILLED_BY_RELEASE``: the node this process ran on was
+  released, killing it
+* ``PMIX_PROC_STATE_KILLED_BY_RELEASE``: the node this process ran on
+  was released, killing it
 * ``PMIX_ERR_PROC_FAILED_TO_START``: process failed to start
 * ``PMIX_ERR_PROC_ABORTED_BY_SIG``: process aborted by signal (e.g.,
   segmentation fault)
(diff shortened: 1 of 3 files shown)


rhc54:
Had to think about this for awhile. We haven't dealt with multi-user DVMs yet, so basically you are dealing with the case where I have a job running, and then issue an allocation request to shrink the DVM - and that request includes at least one node that has procs from one or more of my running jobs. The new codes are intended to distinguish between that scenario and one where I specifically cancel the job itself, leaving the nodes alone.

It is kinda backtracking a bit as we had deprecated a number of proc/job status values years ago as it wasn't clear that PMIx itself needed to carry them. These could also be considered runtime states and status codes - is there a specific reason why then need to be in PMIx?

I'm not necessarily opposed - just looking for some consistency in reasoning.

Petter-Programs:
I mostly agree with your reading, but I wouldn't consider a multi-user DVM to be a precondition for these changes to be useful. We could have a DVM managed by a malleability or fault tolerance library and simultaneously interacted with by a user (the same one running the library) who decides to cancel a job within it. The library may wish to recover from the failure if it was caused by a shrink, but not try to recover a job the user wished to cancel. A similar argument can be made for the other codes, although it requires the user targeting individual processes. For the given library to know anything about these internal states, we need corresponding PMIx functionality.

Ultimately this is a fairly niche case which I'm not super attached to, so just let me know if you'd rather I stick to the preexisting codes in the related PRRTE PR, or narrow it down to just PMIX_ERR_JOB_KILLED_BY_RELEASE, as it could achieve much of the desired distinction on its own.

rhc54:
I'm okay with it - I think the reasoning in the PRRTE PR is sound, and the case you cite here is valid. We may want to review the deprecated values to see if any of those should be promoted to cover these scenarios - when we deprecated them, we didn't have elasticity or malleable libraries managing things, so no justification for their use.

PR #4198, opened 2026-08-26
Let PMIx_Value_get_number read the types that are integers with names

rhc54:
### The problem

`PMIx_Value_get_number()` is the function PMIx offers for reading a number out
of a `pmix_value_t`. It recognized the plain widths plus `PMIX_PID`,
`PMIX_STATUS` and `PMIX_PROC_RANK`, and answered `PMIX_ERR_BAD_PARAM` for
everything else.

But PMIx has a whole family of datatypes that are an integer under a name of
their own — `pmix_scope_t`, `pmix_proc_state_t`, `pmix_alloc_inheritance_t`,
`pmix_device_type_t` and a dozen more — each a fixed-width unsigned integer
with its own type tag and its own member of the value union. None of them was
recognized.

That bites hardest exactly where the standard tells a caller to use the named
type. `PMIX_ALLOC_INHERITANCE` is annotated `(pmix_alloc_inheritance_t)`, and
the documentation's own example loads it as `PMIX_ALLOC_INHERIT`:

```c
PMIX_INFO_LOAD(&info[1], PMIX_ALLOC_INHERITANCE,
               &(pmix_alloc_inheritance_t){PMIX_ALLOC_INHERIT_CHILD_DEFAULT},
               PMIX_ALLOC_INHERIT);
```

A host that then read it back with `PMIx_Value_get_number()` got
`PMIX_ERR_BAD_PARAM` — the reader refused the spelling PMIx had asked for. So
a host had two choices: reach into `value.data` itself, which is how a value
of the wrong type becomes a wrong answer instead of an error, or write the
same special case every other host was writing. PRRTE has one.

### The change

Rather than add an arm for each of these to each of the fifteen `check_*`
functions, `plain_integer_type()` names the plain integer underneath, at the
source end and the destination end alike, and the existing conversions do the
work. The range and sign rules then apply to them exactly as to anything
else — a disposition read into an `int` comes back, a `300` read into one is
refused, a `-1` is refused. Because the value union's members alias, the
source is normalized by copying the value and restamping its type, so nothing
has to know which member held the datum. The widths are the ones
`pmix_bfrops_base_value_load()` stores; the coupling is written down in both
places.

Eighteen types in total: the three already recognized plus `PERSIST`, `SCOPE`,
`DATA_RANGE`, `PROC_STATE`, `ALLOC_DIRECTIVE`, `RESBLOCK_DIRECTIVE`,
`ALLOC_INHERIT`, `JOB_STATE`, `LINK_STATE`, `LOCTYPE`, `DEVTYPE` and the four
storage types.

**The one restriction is the policy that was already there.** `check_rank()`
and `check_status()` refused to unload a structured value into another
structured type. That rule now sits once at the top of the function and covers
the whole family: a scope is not a process state, and an allocation directive
is not a status. Reading any of them as a plain integer, and building any of
them from a plain integer, stays allowed — that is what the function is for.
`PMIX_BOOL` is deliberately outside all of this; a flag is not a count.

### Verification

The existing property test needed only its tables extended — both properties
it states are properties of the new types too — so it now checks them over all
31 source and destination types. Plus a targeted check against the built
library:

```
ALLOC_INHERIT(3) -> UINT8        ok  3        INT(3)   -> ALLOC_INHERIT  ok  byte0=3 byte1=0xab
ALLOC_INHERIT(3) -> INT          ok  3        INT(300) -> ALLOC_INHERIT  REFUSED (range)
ALLOC_INHERIT -> SCOPE       REFUSED (policy) INT(-1)  -> ALLOC_INHERIT  REFUSED (sign)
PID -> PROC_RANK             REFUSED (unchanged)  BOOL -> INT            REFUSED
```

`make check`: 100 pass, 0 fail, 2 skip on macOS (debug build). Man page and
`bfrops/AGENTS.md` updated; docs build warning-free. `PMIX_CAP_GET_NUMBER_ENUMS`
lets a host tell whether it is talking to a library that does this.

### The second commit

`Construct the held-delete list PMIx_Init only ever destructed` is an
independent fix, found by running the suite for the above — it was failing
`client_cycle` and `run_simpcycle.pl` before this branch, and those are the
two failures the 100/0/2 above accounts for.

`pmix_client_held_deletes` is declared `PMIX_LIST_STATIC_INIT` and never
constructed. Two consequences. Its sentinel's `next`/`prev` stay NULL until a
construct links them to itself, so the first `pmix_list_append()` to it stores
through a NULL — and the only caller that appends is
`client_data_delete_handler()`, on precisely the path the list exists to
serve. And `PMIX_DESTRUCT` zeroes an object's magic id after asserting on it,
so `client_teardown()` could run at most once per process; the second
`PMIx_Init` in a process aborted on that assertion in a debug build. The
comment there had it backwards — the list *destructor* does re-link the
sentinel, but the destruct *macro* poisons the object.

One `PMIX_CONSTRUCT` in `PMIx_Init` beside `pending_requests`, which already
had the matching pair.

Happy to split that into its own PR if you would rather review it separately.

### DIFF
--- a/docs/man/man3/PMIx_Value_get_number.3.rst
+++ b/docs/man/man3/PMIx_Value_get_number.3.rst
@@ -70,6 +70,39 @@ the datatype found in the given `val`, taking one of the following actions:
       outside the range of the integer type, then the ``PMIX_ERR_LOST_PRECISION``
       error will be returned
 
+What counts as numeric
+^^^^^^^^^^^^^^^^^^^^^^
+
+Both the plain widths (``PMIX_SIZE``, ``PMIX_INT`` and the sized integer
+types, ``PMIX_FLOAT``, ``PMIX_DOUBLE``) and the PMIx datatypes that *are*
+an integer under a name of their own are numeric for the purposes of this
+function. The latter are:
+
+``PMIX_PID``, ``PMIX_STATUS``, ``PMIX_PROC_RANK``, ``PMIX_PERSIST``,
+``PMIX_SCOPE``, ``PMIX_DATA_RANGE``, ``PMIX_PROC_STATE``,
+``PMIX_ALLOC_DIRECTIVE``, ``PMIX_RESBLOCK_DIRECTIVE``,
+``PMIX_ALLOC_INHERIT``, ``PMIX_JOB_STATE``, ``PMIX_LINK_STATE``,
+``PMIX_LOCTYPE``, ``PMIX_DEVTYPE``, ``PMIX_STOR_MEDIUM``,
+``PMIX_STOR_ACCESS``, ``PMIX_STOR_PERSIST`` and ``PMIX_STOR_ACCESS_TYPE``.
+
+Any of these may appear as the datatype of `val`, as the datatype named by
+the caller, or as both. The range and sign rules above apply to them
+through the integer each one is defined as, so a ``PMIX_ALLOC_INHERIT``
+read into a ``PMIX_INT`` returns its value, and a ``PMIX_INT`` holding
+``300`` read into a ``PMIX_ALLOC_INHERIT`` -- a ``uint8_t`` -- is refused.
+
+There is one restriction. A value of one of these named types is **not**
+unloaded into a *different* named type, however alike the integers
+underneath them may be: a scope is not a process state, and an allocation
+directive is not a status. Asking for one from the other returns
+``PMIX_ERR_BAD_PARAM``. Reading any of them as a plain integer, and
+building any of them from a plain integer, is what this function is for
+and is always allowed.
+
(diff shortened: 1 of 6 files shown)


rhc54:
Force-pushed a layout fix (no behavior change).

The two new statics were defined inline between the forward-declaration block and `PMIx_Value_get_number`, which is the one thing this file never does — every static in it is forward-declared at the top and defined at the bottom, in declaration order. They now follow that.

The diff is purely additive as a result: 101 insertions, 0 deletions, at the declaration block, the function body, and the static-definition tail.

Re-verified on a quiet machine: warning-free build, `make check` 102 total / 100 pass / 2 skip / 0 fail.

PR #4205, opened 2026-08-27
Make a statically initialized list an empty one, not a broken one

rhc54:
`PMIX_LIST_STATIC_INIT` left the sentinel's `next` and `prev` NULL. That is not an empty list, it is an unwalkable one — and every `PMIX_LIST_FOREACH` and `pmix_list_append` in the tree assumes an empty list can be walked and appended to. The assumption holds for a constructed list, so any list touched before its `PMIX_CONSTRUCT` — a framework that never opened, a globals struct belonging to a role that never stood it up — took the process down instead of finding nothing.

### The macro

It now takes the object it initializes, because pointing the sentinel at itself is the only way to spell an empty list at compile time, and there is no other way to reach that address:

```c
pmix_list_t mylist = PMIX_LIST_STATIC_INIT(mylist);

myglobals_t myglobals = {
    .thelist = PMIX_LIST_STATIC_INIT(myglobals.thelist)
};
```

One macro, not two. Every use passes its own name, so a missing argument is a compile error rather than a latent NULL. The three composed initializers — `PMIX_EVENTS_STATIC_INIT`, `PMIX_IOF_SINK_STATIC_INIT`, `PMIX_CLI_RESULT_STATIC_INIT` — thread the name through to the lists they contain. The alternative, teaching `PMIX_LIST_FOREACH` and `pmix_list_append` to tolerate NULL, would put a branch in every walk and every append in the library.

### The second flag drops out

Ten `PMIx_server_*` entry points had been given a `pmix_server_globals.initialized` screen to keep a client or a tool away from state that would kill it. The state no longer kills anybody, and you must call exactly one of `PMIx_Init`, `PMIx_tool_init` or `PMIx_server_init` in the first place — `init_called` is a single latch shared by all three, so a mixed init is undetected and blows up regardless. `pmix_globals.initialized` answers the only question worth asking. The field, both its writes and all ten screens are gone.

### Measured, not argued

All fourteen server entry points called from each role, one per forked child:

| | client | plain tool | launcher |
|---|---|---|---|
| old sentinel | 10 crash | 4 crash | 0 |
| new sentinel | **0** | **0** | **0** |

Re-broken by putting the two pointers back to NULL, leaving the flag collapsed, to confirm the macro is what covers it — the ten and four came straight back.

### Tests

Eight cases across `progress_threads.c`, `server_inventory.c`, `server_registration.c` and `server_setup.c` asserted the `PMIX_ERR_INIT` refusal those screens produced. They now assert survival instead — which is what they were really protecting, and what still fails against a broken sentinel. A client calling a server API is a programming error PMIx does not try to diagnose; what it gets back is not a contract, and the process staying alive is.

`make check` 21/21 and `test/unit` 100 pass / 2 skip, warning-free under `--enable-devel-check`. Docs rebuild clean.

### Notes

- Net **-166 lines** across 36 files.
- `docs/todo.rst`: the "entry points screen the wrong initialization flag" item is retired, with a dated note recording why the small fix beat the large one — nine entry points crashing the same way is one bug in what they all walk, not nine bugs in nine doorways.
- Stacked on #4203, which it carries as its first commit. #4203 can be closed as subsumed, or merged first; either works.


rhc54:
chicken/egg problem

PR #4239, opened 2026-09-10
Track show_help aggregation per job, and drop it when the job does

rhc54:
## The problem

`show_help` duplicate suppression exists to stop *one job's* message storm: the first sighting of a `(filename, topic)` is shown, the rest are counted, and a summary goes out on a five-second timer. The tuple list it keys on is process-global and lives as long as the process — right for a client, wrong for a server.

A PMIx server runs many jobs. A persistent DVM runs them in parallel and one after another for as long as it lives, which can be weeks, and every one of them shared a single set of tuples. So **the first job to trip a diagnostic was the only job ever told about it.** Every later job got either a bare `N more processes have sent help message` counter — and only if the timer happened to fire while somebody was still attached to hear it — or nothing at all.

Measured against a persistent PRRTE DVM, three identical runs of a case that terminates a job because a job it was connected to failed:

```
run 1:  the full diagnostic
run 2:  "3 more processes have sent help message help-prted.txt / connected-term"
run 3:  nothing
```

A fresh DVM per run showed it every time. That is the explanation of why a job died being withheld from the user whose job died.

The existing control cannot reach it either: `PMIX_AGGREGATE_HELP` turns aggregation off for a job that asks, but a job cannot ask to be told about a message that a job which finished last Tuesday already consumed. Confirmed — three runs with `prun --no-aggregate-help` give the same `1, 0, 0`.

## The fix

**Put the job in the key.** A tuple is now `(nspace, filename, topic)` and each job gets its own first sighting of each message.

- The nspace travels with the text as a `PMIX_NSPACE` **directive**, not in the source proc: `plog/stdfd` routes by source, so naming the job a message is *about* there would deliver a daemon's diagnostic into that job's output stream.
- It is compared with `strcmp`, not through `match()`, whose `*` handling would let one job's suppression reach another's.
- A caller naming no job gets its own, which is what every message PMIx raises on its own behalf wants.

**Then let go of it.** `PMIx_server_deregister_nspace()` purges that job's tuples, and both halves matter. Without the flush, whatever the job accumulated is never reported — the timer is the only other thing that would say it, and a job that ends sooner, which is most of them, ends first. Without the free, per-job keying turns a bounded list into one that grows for the life of the server, holding an entry per (job, message) pair for jobs that finished long ago; the reclaim is part of this change rather than a follow-up. The purge runs ahead of the GDS and event teardown in `_deregister_nspace` so a notice raised by the flush still has somewhere to go.

The summary line now names the job, because two jobs running side by side would otherwise print two identical lines nobody can tell apart.

The second commit is separate and small: the one-time notice told users to set MCA parameter `base_help_aggregate` to 0. No such parameter is registered anywhere in this tree — the name is inherited from Open MPI's show_help and `grep` finds it only in the string advertising it — so following the advice did nothing, silently. It now names `PMIX_AGGREGATE_HELP`.

## Testing

- `test/unit/util/util_show_help.c` gains `test_check_dups_per_nspace`: job A and job B each get their own first sighting and their own suppression after it; purging A lets A start over **and leaves B's entry intact** (a purge that swept the list rather than one job would pass the first half and fail the second); purging an unknown job, and NULL, are no-ops. Full PMIx unit suite green.
- End to end against a ten-node PRRTE DVM: the diagnostic now appears on every run rather than only the first. It also now prints **both** terminations in the transitive case — the second was previously swallowed as a duplicate of the first *within a single run*, so this was losing messages inside one job as well as across jobs.
- PRRTE's full multi-node suite: 810 passed, 0 failed, 3 skipped (up from 809/1/3 — the failure was this).

## Companion

Needs the matching PRRTE change, which gives `prte_show_help()` an nspace and converts its call sites: openpmix/prrte#2774. Chicken-and-egg as usual — CI on either side is red until both land.


rhc54:
chicken/egg issue with prrte

PR #4276, opened 2026-09-21
Ignore the config.guess/config.sub backups autoreconf leaves

jsquyres:
Running `autogen.pl` with Autoconf 2.70 or later (seen with Homebrew's autoconf 2.73 and automake 1.18) leaves `config/config.guess~` and `config/config.sub~` in the tree as untracked files.

Why it happens: `autoreconf -ivf` first has Automake copy its `config.guess`/`config.sub` into `config/`. Autoconf 2.70 and later then install their own copies of these files if theirs are newer. They do this through `update_file()` in `Autom4te/FileUtils.pm`, which renames the existing file with a `~` suffix before moving the new one into place.

This PR adds both backups to `.gitignore`, next to the `config.guess`/`config.sub` entries and the other autotools backups the file already ignores (`configure~`, `config.h~`, ...).


jsquyres:
Hey @rhc54 -- could we cherry-pick this to the branches that Open MPI uses as submodules?  Whenever I run "git status" in an Open MPI tree, I always see untracked files in the `3rd-party/openpmix` tree.  We can pick up this .gitignore change the next time we update submodules.

Thanks!

rhc54:
@jsquyres What's this all about?
```shell
[runnervmlun5p][[57215,84],3][btl_tcp_proc.c:369:mca_btl_tcp_proc_create_interface_graph] Unable to find reachable pairing between local and remote interfaces
[runnervmlun5p][[57215,84],3][btl_tcp_proc.c:369:mca_btl_tcp_proc_create_interface_graph] Unable to find reachable pairing between local and remote interfaces
```
I believe I saw this somewhere else and it was identified as a problem in the OMPI btl/tcp code, and that there was a patch being worked on over there to fix it? Or is this something I need to pursue as a possible race condition?

rhc54:
> could we cherry-pick this to the branches that Open MPI uses as submodules? Whenever I run "git status" in an Open MPI tree, I always see untracked files in the `3rd-party/openpmix` tree. We can pick up this .gitignore change the next time we update submodules.

Weird - I never see that on Mac or Linux builds. Is this an artifact of being in your submodule? I have no issue with doing it, but just wonder why you're seeing it.

rhc54:
> I believe I saw this somewhere else and it was identified as a problem in the OMPI btl/tcp code, and that there was a patch being worked on over there to fix it? Or is this something I need to pursue as a possible race condition?

I've confirmed this is on our side - digging into it

jsquyres:
> Weird - I never see that on Mac or Linux builds. Is this an artifact of being in your submodule? I have no issue with doing it, but just wonder why you're seeing it.

I see it even in a plain clone of this repo.  Claude tells me this:

>  What happens during autoreconf -ivf (autogen.pl:1144):
>
>  1. automake --add-missing --copy --force-missing puts Automake 1.18's copies of config.guess and config.sub into config/. Those are dated 2024-07-27 and 2024-05-27.
>  2. Autoconf 2.70 and later then install their own copies of these files from /opt/homebrew/share/autoconf/build-aux/ whenever theirs are newer. Homebrew's autoconf 2.73 ships 2025-07-10 versions, which are newer, so it replaces them.
>  3. That replacement goes through update_file() in Autom4te/FileUtils.pm. When the target file already exists, update_file() renames it to $to~ before moving in the new one. It uses $SIMPLE_BACKUP_SUFFIX, which defaults to ~.

rhc54:
> > Weird - I never see that on Mac or Linux builds. Is this an artifact of being in your submodule? I have no issue with doing it, but just wonder why you're seeing it.
> 
> I see it even in a plain clone of this repo. Claude tells me this:
> 
> >

I can't reproduce it on Mac 26.6 or Ubuntu. I'm using autoconf (GNU Autoconf) 2.73. I don't see these backup files anywhere:
```shell
$ ls -l config/config*
-rwx------@ 1 rhc  staff  50850 Sep 21 09:33 config/config.guess*
-rwx------@ 1 rhc  staff  39943 Sep 21 09:33 config/config.sub*
```
So I wonder if you have some leftover cruft from updating? Did you try a simple `git clean -dfx` on the repo before rebuilding?

jsquyres:
> So I wonder if you have some leftover cruft from updating? Did you try a simple git clean -dfx on the repo before rebuilding?

Yep -- even happens on a clean clone.

I'm guessing it's something in my environment that causes Automake to make those backup copies...?  But even if so, it's something that must be a supported "make a backup" mechanism somewhere.

jsquyres:
Thanks @rhc54 !

rhc54:
Already inbound on v7.0 - here's the backport to v5.0: https://github.com/openpmix/openpmix/pull/4282

